PR #1, opened 2022-03-11
Self-Signed,  Peer and Client Relations

gabrielcocenza:
* Render config templates using jinja2
* self-signed certificates using juju resources
* client relation using elastic search interface
* removed integration-test while snap is not ready
* updated README.md and CONTRIBUTING.md

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -1,104 +1,337 @@
 #!/usr/bin/env python3
+"""Charm code for OpenSearch service."""
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
-#
-# Learn more at: https://juju.is/docs/sdk
-
-"""Charm the service.
-
-Refer to the following post for a quick-start guide that will help you
-develop a new k8s charm using the Operator Framework:
-
-    https://discourse.charmhub.io/t/4208
-"""
 
 import logging
+from pathlib import Path
+from subprocess import CalledProcessError, check_call, check_output
+from typing import Any, Dict, List, Optional
 
-from ops.charm import CharmBase
+from jinja2 import Environment, FileSystemLoader
+from ops.charm import CharmBase, RelationEvent, RelationJoinedEvent
 from ops.framework import StoredState
 from ops.main import main
-from ops.model import ActiveStatus
+from ops.model import BlockedStatus, ModelError, Relation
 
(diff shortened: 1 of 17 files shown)


src/charm.py
```diff
@@ -1,104 +1,334 @@
...
+        """Retrieve IP addresses associated with OpenSearch unit.
+
+        Returns:
+            str: IP address of the unit
+        """
+        return str(self.model.get_binding(PEER).network.bind_address)
+
+    @property
+    def _unit_name(self) -> str:
+        """Get the unit name changing "/" to "-".
```
zeljkoagic, line 81: Why is the name change required? It may be good to clarify it in the docstring.

src/charm.py
```diff
@@ -1,104 +1,334 @@
...
+        for template in CONFIG_MAP.keys():
+            config_path = CONFIG_MAP[template].get("config_path")
+            cmd = CONFIG_MAP[template].get("cmd")
+            chmod = CONFIG_MAP[template].get("chmod")
+            self._write_config(config_path, template, context, chmod)
+            if cmd:
+                try:
+                    check_call(cmd.split())
+                except CalledProcessError as e:
+                    logger.exception(f"Failed to run command {e.cmd}: {e}")
```
zeljkoagic, line 181: Here and elsewhere, cf. the [Juju SDK](https://juju.is/docs/sdk/logging), it is important to not build the logging strings yourself so that the logger can do this for you.

zeljkoagic (changes requested):
Great work, @gabrielcocenza! LGTM, only requesting a minor change wherever formatting is used for logging, to observe the Juju SDK requirement.

src/charm.py
gabrielcocenza: It's required because I use the unit name on the node certificate and private key.
Ex: Unit `openserarch/0` becomes `opensearch-0` and end up generating `opensearch-0.pem` and `opensearch-0-key.pem`.

Using the char `/` confuses the [pathlib](https://docs.python.org/3/library/pathlib.html) to get the `path` to the file.
I will update the docstring to make it clear.

gabrielcocenza:
Thanks for your review @zeljkoagic. I think I've addressed your requests.

zeljkoagic (approved):
Nice work, @gabrielcocenza!

PR #2, opened 2022-03-24
Cross Cluster Replication (CCR)

gabrielcocenza:
- added ccr_leader endpoint as provider using opensearch interface
- added ccr_follower endpoint as requires using opensearch interface
- unit tests for CCR
- instructions on README.md to use CCR

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -6,18 +6,21 @@
 import logging
 from pathlib import Path
 from subprocess import CalledProcessError, check_call, check_output
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
+import requests
 from jinja2 import Environment, FileSystemLoader
 from ops.charm import CharmBase, RelationEvent, RelationJoinedEvent
 from ops.framework import StoredState
 from ops.main import main
 from ops.model import BlockedStatus, ModelError, Relation
+from requests.exceptions import HTTPError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


src/charm.py
```diff
@@ -42,6 +45,59 @@ def __init__(self, *args):
...
+        self.framework.observe(
+            self.on.ccr_leader_relation_joined,
+            self._on_config_changed,
+        )
+        self.framework.observe(
+            self.on.ccr_leader_relation_changed,
+            self._on_config_changed,
+        )
+        self.framework.observe(
+            self.on.ccr_follower_relation_changed,
```
jnsgruk, line 57: What happens on relation-broken? Do we need to make an attempt to clean up the cross-cluster replication gracefully?

gabrielcocenza: Hi Jon. Good observation. I guess that on the leader side there's not much to do if the relation is broken, but on the follower side it sounds a good idea to [stop](https://opensearch.org/docs/latest/replication-plugin/api/#stop-replication) it and than the cluster would automatically turn into standard index.
What do you think?

jnsgruk: I'm no expert on cross-cluster replication, I just noticed we weren't handling that case. I think in general, this charm could definitely benefit from some integration tests that deploy to microk8s, and test what happens when it is scaled up, thne down, and also when configured for CCR, then unrelated, etc. :)

gabrielcocenza:
Closing this PR because the #3 has the logic of CCR integrated with the snap

PR #3, opened 2022-04-20
Cross Cluster Replication (CCR) and Snap integration

gabrielcocenza:
- added ccr_leader endpoint as provider using opensearch-cluster interface
- added ccr_follower endpoint as requires using opensearch-cluster interface
- unit tests for CCR
- instructions on README.md to use CCR
- integration with the opensearch snap that is used as resource until it is published.
- enforce_hostname_verification

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ build/
 *.charm
 .tox/
 .coverage
-.vscode
 __pycache__/
 *.py[cod]
 *.pem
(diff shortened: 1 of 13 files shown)


gabrielcocenza:
The `relation-broken` for the CCR is not part of the scope of this PR, My suggestion for the future implementation of this is to observe `on.ccr_follower_relation_broken` and [delete the replication rule](https://opensearch.org/docs/latest/replication-plugin/auto-follow/#delete-a-replication-rule). The `leader_alias` is `ccr-connection-alias` and the `name` is `all-replication-rule`.

PR #6, opened 2022-10-10
New charm + Feature TLS

Mehdi-Bendriss:
# Issue
This PR addresses [DPE-583](https://warthogs.atlassian.net/browse/DPE-583) with the exception of some items listed on the notes sub-section below.
Namely, this PR addresses:
- Bootstrapping of new OpenSearch VM charm.
- Implementation of TLS for OpenSearch VM from no support to TLS operator charm integration.
Reference to

# Solution
The installation and usual actions to interact with OpenSearch is done through an interface `OpenSearchDistribution`. The implentation we're currently using is "OpenSearchTarball", while the Snap is released, Once released, we'll switch to the "OpenSearchSnap" implementation.
The relation with `tls-certificates-operator` is used to obtain the certificates.

 ## Notes:
- This PR installs OpenSearch from a tarball. 
- This PR does not contain tests, each unit and integration tests will come in a separate PR
- This PR goes with the assumption that the `ulimit -n` has successfully been set to at least `65536`
- This PR assumes that the [SANs issue](https://github.com/canonical/tls-certificates-operator/issues/14) is fixed on the `tls-certificates-operator`.
- I added a local Patch to the `tls-certificates-lib` to support multiple types of SANs to add the OID as a `RegisteredID`, `DNS Names` and `IP Addresses` , while I submit a proper PR on the lib's repo.


lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -0,0 +1,305 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Base class for Opensearch distributions."""
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Those classes should be shared between VM and K8s, some of those are classes are supposed to be extended - i.e each charm should inherit from `OpenSearchBaseCharm` which brings them some shared objects and properties, such as the distribution / config etc... others provide callbacks for things that are proper for each charm VM/K8s, such as the `on_tls_conf_set` in `opensearch_tls`.
The `OpenSearchDistribution` provides a set of methods that need to be implemented by the distribution used in each charm (`OpenSearchSnap` and `OpenSearchTarball` for VMs and another one for K8s)

actions.yaml
```diff
@@ -1,16 +1,52 @@
...
       type: string
-      default: ""
+      description: (Optional) The content of the private key. The content will be auto-generated if not set.
+    password:
+      type: string
+      description: (Optional) The passphrase of the private key. Will NOT be set if omitted.
+  required:
+    - category
+
+#put-client:
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Thanks Shayan. It will be in the future, I will delete it and put in a separate branch

src/opensearch.py
```diff
@@ -0,0 +1,187 @@
...
+class OpenSearchTarball(OpenSearchDistribution):
+    """Snap distribution of opensearch, only overrides properties and logic proper to the snap."""
+
+    def __init__(self, charm, peer_relation: str):
+        super().__init__(charm, peer_relation)
+
+    def install(self):
+        """Download and Un-tar the opensearch distro."""
+        response = requests.get(
+            "https://artifacts.opensearch.org/releases/bundle/opensearch/2.3.0/opensearch-2.3.0-linux-x64.tar.gz"
```
WRFitch, line 126: Is this the best place to get opensearch? Should we be building and storing it ourselves? 

Also, maybe parameterise the opensearch version so it can be easily updated and accessed from elsewhere.

src/opensearch.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def stop(self):
+        """Stop opensearch."""
+        self._run_cmd("ps aux | grep opensearch | awk '{print $2}' | kill -15")
+
+        while self.is_node_up():
+            time.sleep(2)
+
+        """
+        TODO:
```
WRFitch, line 158: Is this information important for a user to know? Also this might be worth making a Jira ticket.

src/charm.py
```diff
@@ -1,104 +1,377 @@
...
+        except OpenSearchInstallError:
+            self.unit.status = BlockedStatus("Could not install OpenSearch.")
+            event.defer()
+
+    def _on_leader_elected(self, _: LeaderElectedEvent):
+        """Handle the install event."""
+        if self.app_peers_data.get("security_index_initialised", None) is not None:
+            return
+
+        self.unit.status = MaintenanceStatus("Configuring admin user if needed...")
```
WRFitch, line 74: maybe move this under the `if` statement?

src/charm.py
```diff
@@ -1,104 +1,377 @@
...
+            try:
+                self.opensearch.restart()
+                del self.unit_peers_data["must_reboot_node"]
+            except OpenSearchStartError:
+                event.defer()
+
+    def _on_update_status(self, _: UpdateStatusEvent):
+        """On update status event.
+
+        We want to periodically (every 6 hours) check if certs are expiring soon (in 24h),
```
WRFitch, line 141: This timing might need adjusting - if it occurs over the weekend, for example, we're out of luck. Maybe up the expiry check to a week in advance?

src/charm.py
```diff
@@ -1,104 +1,377 @@
...
         else:
-            event.set_results({"fortune": "A bug in the code is worth two in the documentation."})
+            # write the admin cert conf on all units, in case there is a leader loss + cert renewal
+            self.opensearch_config.set_admin_tls_conf(current_secrets)
+
+        if should_restart:
+            self.unit_peers_data["must_reboot_node"] = "True"
+
+    def _is_tls_fully_configured(self) -> bool:
+        """Start OpenSearch if TLS fully configured and if already not started."""
```
WRFitch, line 229: nit: this method doesn't start opensearch, it just checks if TLS is ready to go

src/charm.py
```diff
@@ -1,104 +1,377 @@
...
+        return (
+            self.unit.is_leader()
+            and self.app_peers_data.get("security_index_initialised") is None
+            and self.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val) is not None
+        )
+
+    def _initialize_security_index(self, admin_secrets: Dict[str, any]):
+        """Run the security_admin script, it creates and initializes the opendistro_security index.
+
+        IMPORTANT: must only run once per cluster, otherwise the index gets overrode
```
WRFitch, line 309: Is there a way to enforce this?

WRFitch (changes requested):
This is a really good start! I have a lot of feedback, but there's a lot of code here. I understand why this PR is so long, but please keep it smaller in future - I feel like I need to go out for a walk or maybe take a nap after reviewing this.

lib/charms/opensearch/v0/helpers/conf_setter.py
```diff
@@ -0,0 +1,297 @@
...
+"""Utilities for editing yaml config files at any depth level and maintaining comments."""
+import re
+import sys
+import uuid
+from collections.abc import Mapping
+from enum import Enum
+from os.path import exists
+from typing import Dict
+
+from ruamel.yaml import YAML, CommentedSeq
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Ruamel maintains the comment blocks on a YAML file as opposed to PyYaml which loses them after loading the file

lib/charms/opensearch/v0/helpers/conf_setter.py
```diff
@@ -0,0 +1,297 @@
...
+
+        random_id = uuid.uuid4().hex
+        with open(path, "r") as f:
+            lines = f.readlines()
+
+            doc_empty_or_full_comments = True
+            for line in lines:
+                if not line.startswith("#") and line.strip() != "":
+                    doc_empty_or_full_comments = False
+                    break
```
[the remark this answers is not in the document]
Mehdi-Bendriss: So what's happening here is the handling of a document that is exclusively filled with comments, the parser has no way of knowing this is a YAML file, so the comments are lost on load and the returned document is empty. So the idea here is to detect this case and if it's a positive, to assist the parser by adding a dummy key/val that gets deleted later.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def on_tls_conf_set(
+        self, event: CertificateAvailableEvent, scope: Scope, cert_type: CertType, renewal: bool
+    ) -> None:
+        """Called after certificate ready and stored on the corresponding scope databag."""
+        pass
+
+    def on_tls_conf_remove(self):
+        """Called after certificates removed."""
+        pass
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Those are callback methods on the charm itself, they [get called from the OpenSearchTLS lib](https://github.com/canonical/opensearch-operator/blob/init-charm-tls/lib/charms/opensearch/v0/opensearch_tls.py#L121), this is to decouple the logic and concerns of each component, as the lib is not supposed to know the logic following an available certificate for ex

src/opensearch.py
Mehdi-Bendriss: Not for the user, it's a documentation and todo for us mainly. I created the Jira ticket

src/charm.py
Mehdi-Bendriss: Yes, we test whether the unit is the leader unit and whether the `security_index_initialised` flag on the app_data is set or not

lib/charms/opensearch/v0/helpers/conf_setter.py
```diff
@@ -0,0 +1,297 @@
...
+            else:
+                source = dict()
+
+        current_key: str = node_keys.pop(0)
+
+        # handling the insert / update of simple types on sets
+        if current_key.startswith("{"):
+            return self.__get_source_for_set(source, val)
+
+        # handling the insert / update on json objects
```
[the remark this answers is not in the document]
Mehdi-Bendriss: The data structure of a deserialized yaml doc is returned as a dictionary, similarly the input data structure of the serialization process is a dictionary - in this case I used "json" to mean "dict". Of course the end result of the serialization process is a yaml document.

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -0,0 +1,270 @@
...
+        """Request certificate when TLS relation joined."""
+        admin_cert = self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN)
+        if self.charm.unit.is_leader() and admin_cert is None:
+            self._request_certificate(Scope.APP, CertType.APP_ADMIN)
+
+        self._request_certificate(Scope.UNIT, CertType.UNIT_TRANSPORT)
+        self._request_certificate(Scope.UNIT, CertType.UNIT_HTTP)
+
+    def _on_tls_relation_broken(self, _: RelationBrokenEvent) -> None:
+        """We don't need to handle much here."""
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Thanks Will - I missed handling a case serving as a safeguard here: if all certificates (3 per unit) have already been provided -> nothing to handle, we keep the certificates stored to guarantee the continuity of functioning of the node until they hit 1 week before their expiration date. 
BUT - in case only a subset of those requested certs is available, it means that TLS will not work - thus requiring the unit to be put on a BlockedStatus

src/opensearch.py
```diff
@@ -0,0 +1,187 @@
...
+            home="/var/snap/opensearch/current",
+            conf="/var/snap/opensearch/common/config",
+            data="/var/snap/opensearch/common/data",
+            logs="/var/snap/opensearch/common/logs",
+            jdk="/var/snap/opensearch/current/jdk",
+            tmp="/var/snap/opensearch/common/tmp",
+        )
+
+
+class OpenSearchTarball(OpenSearchDistribution):
```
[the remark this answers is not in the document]
Mehdi-Bendriss: The way we should be running OpenSearch is using a snap. Since the snap is not published yet, we needed to have an alternative / temporary way to install and interact with OpenSearch, namely a tarball.

cloudinit-userdata.yaml
```diff
@@ -0,0 +1,14 @@
+cloudinit-userdata: |
```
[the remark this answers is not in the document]
Mehdi-Bendriss: This shouldn't be part of the repo. I added it to .gitignore

src/opensearch.py
Mehdi-Bendriss: This is only a temporary way of installing OpenSearch while we wait for the snap to be published. This class will be replaced by `OpenSearchSnap` once the snap is published.

lib/charms/opensearch/v0/helpers/conf_setter.py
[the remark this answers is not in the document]
WRFitch: OK, that makes more sense. Would you mind renaming it to "dict" or similar?

lib/charms/opensearch/v0/helpers/conf_setter.py
```diff
@@ -0,0 +1,297 @@
...
+
+        self.__deep_update(data, key_path.split(sep), val)
+
+        if inline_array:
+            data = self.__inline_array_format(data, key_path.split(sep), val)
+
+        self.__dump(
+            data,
+            output_type,
+            f"{self.base_path}{config_file}" if output_file is None else output_file,
```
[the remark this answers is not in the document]
Mehdi-Bendriss: The base path can vary from use case to use case, I think it should be supplied by the client, this class should be ignorant of the environment it runs on.

lib/charms/opensearch/v0/helpers/conf_setter.py
```diff
@@ -0,0 +1,297 @@
...
+        sep="/",
+        output_type: OutputType = OutputType.file,
+        output_file: str = None,
+    ) -> Dict[str, any]:
+        """Delete the value of a key (or content of array at index / key) if it exists."""
+        data = self.load(config_file)
+
+        self.__deep_delete(data, key_path.split(sep))
+
+        self.__dump(
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Explanation of the design choice above

lib/charms/opensearch/v0/helpers/networking.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Helpers for networking related operations."""
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Hmm, I checked the helper functions defined there, and I see they're applicable to K8s. Did you have a specific function in mind? 
(The idea of this helper is to outsource the simple network related functions, that are used everywhere, such as getting a unit IP or checking if a host is reachable - for substrate specific things, this shouldn't contain them as you said)

lib/charms/opensearch/v0/helpers/conf_setter.py
```diff
@@ -0,0 +1,297 @@
...
+            f.write(data)
+
+    def __dump(self, data: dict[str, any], output_type: OutputType, target_file: str):
+        """Write the YAML data on the corresponding "output_type" stream."""
+        if output_type in [OutputType.console, OutputType.all]:
+            self.yaml.dump(data, sys.stdout)
+
+        if output_type in [OutputType.file, OutputType.all]:
+            with open(target_file, mode="w") as f:
+                self.yaml.dump(data, f)
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Thank you, I am tracking this on a [Jira ticket](https://warthogs.atlassian.net/browse/DPE-786) - will address it once we're nearing the work on the K8s charm.

lib/charms/opensearch/v0/helpers/conf_setter.py
```diff
@@ -0,0 +1,297 @@
...
+    ) -> Dict[str, any]:
+        """Add or update the value of a key (or content of array at index / key) if it exists."""
+        data = self.load(config_file)
+
+        self.__deep_update(data, key_path.split(sep), val)
+
+        if inline_array:
+            data = self.__inline_array_format(data, key_path.split(sep), val)
+
+        self.__dump(
```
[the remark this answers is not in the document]
WRFitch: After talking this over on MM, I see why you took this approach - avoiding rewrites and restarts is important on PGB because you have to restart the entire application, so my proposed approach suits better, but for opensearch I understand your approach.

lib/charms/opensearch/v0/helpers/conf_setter.py
```diff
@@ -0,0 +1,297 @@
...
+
+            doc_empty_or_full_comments = True
+            for line in lines:
+                if not line.startswith("#") and line.strip() != "":
+                    doc_empty_or_full_comments = False
+                    break
+
+        if doc_empty_or_full_comments:
+            with open(path, "a") as f:
+                f.write(f'\n{random_id}: "{random_id}"')
```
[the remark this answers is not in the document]
Mehdi-Bendriss: You're right, I changed the approach and doing this operation in-memory.

tests/integration/test_charm.py
```diff
@@ -17,40 +17,6 @@
...
-        timeout=1000,
-    )
-    assert ops_test.model.applications[APP_NAME].units[0].workload_status == "active"
-
-    # effectively disable the update status from firing
-    await ops_test.model.set_config({"update-status-hook-interval": "60m"})
-
-
 @pytest.mark.abort_on_fail
 async def test_application_is_up(ops_test: OpsTest):
```
WRFitch, line 21: This test is going to fail without the previous setup steps

WRFitch (approved):

lib/charms/opensearch/v0/helper_databag.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class Scope(BaseStrEnum):
+    """Peer relations scope."""
+
+    APP = "app"
+    UNIT = "unit"
+
+
+class SecretStore:
```
paulomach, line 29: This is very nice - I could see this making up to Ops eventually

lib/charms/opensearch/v0/helper_networking.py
```diff
@@ -0,0 +1,66 @@
...
+    for unit in charm.model.get_relation(peer_relation_name).units:
+        unit_id = unit.name.split("/")[1]
+        unit_ip_map[unit_id] = unit_ip(charm, unit, peer_relation_name)
+
+    return unit_ip_map
+
+
+def is_reachable(ip_address: str, port: int) -> bool:
+    """Attempting a socket connection to a host/port."""
+    s = socket.socket()
```
paulomach, line 59: Don't you need to set the socket family (i.e. `AF_INET`) to this work properly? (or connecting with the tuple does the trick?)

requirements.txt
```diff
@@ -1 +1,6 @@
-ops >= 1.2.0
+bcrypt==4.0.0
+cryptography==38.0.1
+jsonschema==4.16.0
+ops==1.5.2
+requests==2.28.1
+ruamel.yaml==0.17.21
```
paulomach, line 6: Did not found a reference, but are you using this yaml parser on purpose?

src/charm.py
```diff
@@ -1,104 +1,380 @@
...
+        """Handle the install event."""
+        self.unit.status = MaintenanceStatus("Installing OpenSearch...")
+        try:
+            self.opensearch.install()
+        except OpenSearchInstallError:
+            self.unit.status = BlockedStatus("Could not install OpenSearch.")
+            event.defer()
+
+    def _on_leader_elected(self, _: LeaderElectedEvent):
+        """Handle the install event."""
```
paulomach, line 70: fix docstring

src/charm.py
```diff
@@ -1,104 +1,380 @@
...
+        if self.app_peers_data.get("security_index_initialised", None) is not None:
+            return
+
+        if self.app_peers_data.get("admin_user_initialized", None) is None:
```
paulomach, line 74: nit:

```suggestion
        if not self.app_peers_data.get("admin_user_initialized"):
```

src/charm.py
```diff
@@ -1,104 +1,380 @@
...
+        self.unit.status = MaintenanceStatus("Installing OpenSearch...")
+        try:
+            self.opensearch.install()
+        except OpenSearchInstallError:
+            self.unit.status = BlockedStatus("Could not install OpenSearch.")
+            event.defer()
+
+    def _on_leader_elected(self, _: LeaderElectedEvent):
+        """Handle the install event."""
+        if self.app_peers_data.get("security_index_initialised", None) is not None:
```
paulomach, line 71: nit: 
```suggestion
        if self.app_peers_data.get("security_index_initialised"):
```

src/charm.py
```diff
@@ -1,104 +1,380 @@
...
+            admin_secrets = self.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
+            self._initialize_security_index(admin_secrets)
+            self.app_peers_data["security_index_initialised"] = "True"
+
+    def _on_peer_relation_joined(self, _: RelationJoinedEvent):
+        """New node joining the cluster."""
+        current_secrets = self.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
+
+        # In the case of the first unit
+        if current_secrets is None:
```
paulomach, line 121: nit:
```suggestion
        if not current_secrets:
```

src/opensearch.py
```diff
@@ -0,0 +1,195 @@
...
+            )
+        except OpenSearchCmdError:
+            raise OpenSearchStartError()
+
+        if not self.is_started():
+            raise OpenSearchStartError()
+
+    def stop(self):
+        """Stop opensearch."""
+        self._run_cmd("ps aux | grep opensearch | awk '{print $2}' | kill -15")
```
paulomach, line 160: you can just use `pkill` with the same effect
```suggestion
        self._run_cmd("pkill -15 opensearch")
```

paulomach (approved):
Left some questions and comments here and there.
Otherwise, well thought out, stellar work!

lib/charms/opensearch/v0/helper_cluster.py
```diff
@@ -0,0 +1,103 @@
...
+        if nodes_by_roles["cluster_manager"] == 1:
+            return ["cluster_manager", "data"]
+
+        return ["data"]
+
+    @staticmethod
+    def is_cluster_bootstrapped(nodes: List[Node]) -> bool:
+        """Check if cluster is bootstrapped. 2 cm + 1 voting only nodes created."""
+        roles = ClusterTopology.suggest_roles(nodes)
+        return not ("cluster_manager" in roles or "voting_only" in roles)
```
marceloneppel, line 71: Shouldn't you count cluster manager nodes here?

lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -0,0 +1,303 @@
...
+        path = f"{self.base_path}{config_file}"
+
+        if not exists(path):
+            raise FileNotFoundError(f"{path} not found.")
+
+        with open(path, "r") as f:
+            lines = f.read().splitlines()
+
+            random_id = uuid.uuid4().hex
+            lines.append(f"{random_id}: {random_id}")
```
marceloneppel, line 70: Just out of curiosity: why is this random id needed (and removed later)?

src/charm.py
```diff
@@ -1,104 +1,380 @@
...
+        self.framework.observe(self.on.update_status, self._on_update_status)
+
+    def _on_install(self, event: InstallEvent) -> None:
+        """Handle the install event."""
+        self.unit.status = MaintenanceStatus("Installing OpenSearch...")
+        try:
+            self.opensearch.install()
+        except OpenSearchInstallError:
+            self.unit.status = BlockedStatus("Could not install OpenSearch.")
+            event.defer()
```
marceloneppel, line 67: Are you deferring because a re-attempt can succeed without human operator intervention?

src/charm.py
```diff
@@ -1,104 +1,380 @@
...
+
+    def _on_update_status(self, _: UpdateStatusEvent):
+        """On update status event.
+
+        We want to periodically check for 2 things:
+        1- The system requirements are still met
+        2- every 6 hours check if certs are expiring soon (in 7 days),
+            as a safeguard in case relation broken. As there will be data loss
+            without the user noticing in case the cert of the unit transport layer expires.
+            So we want to stop opensearch in that case, since it cannot be recovered from.
```
marceloneppel, line 144: Great check!

lib/charms/opensearch/v0/helper_cluster.py
marceloneppel: I saw that there is a check on `_cleanup_conf_if_bootstrapped()` (charm.py), so maybe change the docstring here?

marceloneppel (approved):
As Paulo commented, stellar work Mehdi! Nice idea about the multiple libraries to better organise each scope of the charm and share with the k8s charm later.

lib/charms/opensearch/v0/helper_networking.py
Mehdi-Bendriss: Thank you Paulo, so it seems that by default, the socket family is set to `AF_INET` indeed, so the tuple connection succeeds

.github/workflows/on_push.yaml
```diff
@@ -0,0 +1,12 @@
+name: Tests
+on:
+  push:
+    branches:
+      - master
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Thanks Marcelo, It will not, great catch!

lib/charms/opensearch/v0/helper_conf_setter.py
Mehdi-Bendriss: This handles the case when a YAML document is only filled with comments. In that specific case, Ruamel (the parser) struggles to understand that this is indeed a YAML document, so it does not load the comments in its internal data structure, so when new data gets written in the file, all the comments are gone. 
This random id is here to give the parser a hint that it is a YAML document, so it preserves all the commented code, then once loaded - this id no longer serves a purpose, so it gets deleted

marceloneppel: Nice, thanks for the explanation!

src/charm.py
Mehdi-Bendriss: Hmm, I thought I would handle the case where the HTTP request to download the binaries fails. Maybe I should attempt a couple of retries on the HTTP download and if it fails again, put it into BlockedStatus? What do you think?

marceloneppel: I think that's a good approach. Most of the data platform charms are using the `tenacity` library to accomplish that, so it can be useful here too.

lib/charms/opensearch/v0/helper_cluster.py
Mehdi-Bendriss: You're right, something is strange in that logic, I rewrote it

marceloneppel (approved):

paulomach (approved):
lgtm

PR #7, opened 2022-10-18
Init charm tls: unit tests

Mehdi-Bendriss:
## Testing

This PR addresses the unit testing part of [DPE-583](https://warthogs.atlassian.net/browse/DPE-583):
Namely, this PR addresses:
- Unit testing of Charm init + TLS


WRFitch (approved):

tests/unit/test_helper_databag.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_put_get(self):
+        """Test putting data of simple type in the secret store."""
+        self.secret_store.put(Scope.UNIT, "unit-key", "unit-val")
+        self.secret_store.put(Scope.APP, "app-key", "app-val")
+
+        self.assertEqual(self.secret_store.get(Scope.UNIT, "unit-key"), "unit-val")
+        self.assertEqual(self.secret_store.get(Scope.APP, "app-key"), "app-val")
+
+    def test_put_get_object(self):
+        """Test putting objects in the secret store."""
```
MiaAltieri, line 39:
```suggestion
        """Test putting and getting objects in/from the secret store."""
```

tests/unit/test_helper_networking.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+        self.assertDictEqual(
+            units_ips(self.charm, PEER), {"0": "1.1.1.1", "1": "2.2.2.2", "2": "3.3.3.3"}
+        )
+
+    def test_is_reachable(self):
+        """Test if host is reachable."""
+        self.assertTrue(is_reachable("google.com", 80))
+        self.assertFalse(is_reachable("googl.comodo", 80))
```
MiaAltieri, line 72: add a check that verifies socket was closed

tests/unit/test_helper_security.py
```diff
@@ -0,0 +1,88 @@
...
+    def test_rfc2253_tls_subject(self):
+        """Test conversion of subject to the rfc2253 format."""
+        subject_1 = "/C=DE/ST=Berlin/L=Berlin/O=Canonical/OU=DataPlatform/CN=localhost"
+        subject_2 = "10.10.10.111"
+
+        self.assertEqual(
+            rfc2253_tls_subject(subject_1),
+            "CN=localhost,OU=DataPlatform,O=Canonical,L=Berlin,ST=Berlin,C=DE",
+        )
+        self.assertEqual(rfc2253_tls_subject(subject_2), f"CN={subject_2}")
```
MiaAltieri, line 88: Seems like you are missing some tests here too

tests/unit/test_opensearch_tls.py
```diff
@@ -0,0 +1,55 @@
...
+
+        for cert_type in [CertType.UNIT_HTTP, CertType.UNIT_TRANSPORT]:
+            self.assertDictEqual(
+                self.charm.tls._get_sans(cert_type),
+                {
+                    "sans_oid": "1.2.3.4.5.5",
+                    "sans_ip": ["1.1.1.1"],
+                    "sans_dns": [self.charm.unit_name, "nebula"],
+                },
+            )
```
MiaAltieri, line 55: seems like youre missing tests here

MiaAltieri (changes requested):
Hey this looks awesome! 

Feel free to make the additional unit tests later in a separate PR.

tests/unit/test_helper_security.py
```diff
@@ -0,0 +1,88 @@
...
+import ops
+from charms.opensearch.v0.helper_security import (
+    cert_expiration_remaining_hours,
+    generate_hashed_password,
+    generate_password,
+    normalized_tls_subject,
+    rfc2253_tls_subject,
+)
+
+ops.testing.SIMULATE_CAN_CONNECT = True
```
paulomach, line 19: since you are using this setup often, consider moving it to `tests/unit/__init__.py`

```python
import ops.testing

ops.testing.SIMULATE_CAN_CONNECT = True
```

paulomach (approved):
looks good, some minor comments

lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -136,7 +144,15 @@ def replace(self, file: str, old_val: str, new_val: any, regex: bool = False) ->
             else:
                 data = data.replace(old_val, new_val)
 
-            f.write(data)
+            if output_type in [OutputType.console, OutputType.all]:
+                print(data)
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Thanks Mia, good catch.

lib/charms/opensearch/v0/helper_networking.py
```diff
@@ -54,11 +54,12 @@ def units_ips(charm: CharmBase, peer_relation_name: str) -> Dict[str, str]:
...
-def is_reachable(ip_address: str, port: int) -> bool:
+def is_reachable(host: str, port: int) -> bool:
     """Attempting a socket connection to a host/port."""
     s = socket.socket()
+    s.settimeout(5)
     try:
-        s.connect((ip_address, port))
+        s.connect((host, port))
         return True
     except Exception:
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Agreed. 
However, The socket connection can throw so many types of exceptions, that are either subclasses of `OSError` or of `Exception` i.e: `socket.error`.
I thought in this specific case, it makes sense to just go generic, since they all mean the same thing in the context of this function, which is the fact the node is not reachable.

tests/unit/test_charm.py
```diff
@@ -4,14 +4,140 @@
...
+        self.assertTrue(isinstance(self.harness.model.unit.status, BlockedStatus))
+
+        # test when TLS relation is broken and cert is expiring soon
+        get_relation.return_value = None
+        is_node_up.return_value = True
+        self.charm.unit_peers_data["certs_exp_checked_at"] = (datetime.now() - timedelta(hours=7)).strftime("%Y-%m-%d %H:%M:%S")
+        self.charm.secrets.put_object(Scope.UNIT, CertType.UNIT_TRANSPORT.val, {"cert": "transport"})
+        cert_expiration_remaining_hours.return_value = 24 * 3
+        self.charm.on.update_status.emit()
+        self.assertTrue(isinstance(self.harness.model.unit.status, BlockedStatus))
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Yes, I will increase the test coverage as I progress :)

tests/unit/test_helper_networking.py
Mehdi-Bendriss: I think it's safe to trust that the native socket api (`socket.close()`) is behaving correctly, or do you see a risk here?

tests/unit/test_helper_security.py
Mehdi-Bendriss: I did not test `to_pkcs8` on purpose, as the TLS integration tests should be the ones validating this is working. If the key is not PKCS8 encoded, TLS configuration won't succeed.

tests/unit/test_helper_security.py
Mehdi-Bendriss: Thanks Paulo, good idea

tests/unit/resources/test.cert
```diff
@@ -0,0 +1,24 @@
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
Mehdi-Bendriss: I deleted those and generated them on the fly

WRFitch (approved):

marcoppenheimer (approved):
I haven't gone through the unit-tests in great detail, it's a fairly massive PR already so didn't want to add swirl to non-prod code, as I imagine they'll be expanded on as time goes on. Left a few general design suggestions.

Loads of really good design choices here, really impressed and a little jealous with how well you've hit the ground running on the code-base, great start!

tests/unit/test_helper_networking.py
MiaAltieri: Sorry not being clear, there is a call which closes the socket, no need to check that the socket is actually closed but might be nice to check that that function is called

lib/charms/opensearch/v0/helper_networking.py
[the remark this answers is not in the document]
MiaAltieri: hmmm, i think it would be better to do: `Except (OSError,socket.error) as e` but also I trust you here!

MiaAltieri (approved):
Looks good to me :)

lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -2,6 +2,7 @@
 # See LICENSE file for licensing details.
 
 """Utilities for editing yaml config files at any depth level and maintaining comments."""
+import logging
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Yes, so the reason I'm using `rueaml.yaml` is because of the fact it allows keeping and writing comments on a yaml doc, whereas PyYaml loses comments on load. OpenSearch has many yaml files with comments serving as docs, so i wanted to keep those comments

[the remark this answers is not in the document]
Mehdi-Bendriss: Good idea for the generic config setter lib!

PR #14, opened 2022-11-17
Implemented horizontal scale-up

Mehdi-Bendriss:
# Issue
This PR addresses [DPE-879](https://warthogs.atlassian.net/browse/DPE-879) and [DPE-881](https://warthogs.atlassian.net/browse/DPE-881) and [DPE-785](https://warthogs.atlassian.net/browse/DPE-785).
Namely, this PR addresses:
- Implementing the scale-up feature on VM charm.
- Integration and unit tests for the scale-up feature on VM charm.
- Removal of the `seed_hosts` property from the config of the cluster manager nodes once cluster bootstrapped.
- minor refactoring on the integration tests.
- Bumped version of OpenSearch to `2.4.0`

# Solution
The scale up is done automatically through OpenSearch's node discovery mechanism which was implemented on a previous PR.
However, we add the deferral of the node registration and OpenSearch start, until all shards (replicas) of the cluster are in a `STARTED` or `UNASSIGNED` state. This to reduce the network latency on the cluster.

 ## Main files to review:
- `lib/charms/opensearch/v0/helper_cluster.py`
- `src/charm.py`
- `tests/integration/ha/helpers.py`
- `tests/integration/ha/test_horizontal_scaling.py`
- `tests/unit/test_helper_cluster.py`

### DIFF
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -1,9 +1,12 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""Utility class for getting cluster configuration info and suggestions."""
+"""Utility class for getting cluster info, configuration info and suggestions."""
 
 from typing import Dict, List
+
+from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
+from tenacity import retry, stop_after_attempt, wait_exponential
 
 # The unique Charmhub library identifier, never change it
 LIBID = "80c3b9eff6df437bb4175b1666b73f91"
@@ -105,3 +108,38 @@ class ClusterTopology:
                 result[role] += 1
 
         return result
+
+
+class ClusterState:
+    """Class for getting cluster state info."""
+
+    @staticmethod
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_exponential(multiplier=1, min=2, max=10),
+        reraise=True,
+    )
+    def shards(opensearch: OpenSearchDistribution, host: str = None) -> List[Dict[str, str]]:
+        """Get all shards of all indexes in the cluster."""
+        return opensearch.request("GET", "/_cat/shards", host=host)
+
+    @staticmethod
+    def busy_shards_by_unit(
+        opensearch: OpenSearchDistribution, host: str = None
+    ) -> Dict[str, List[str]]:
+        """Get the busy shards of the indexes of the cluster."""
+        shards = ClusterState.shards(opensearch, host)
+
+        busy_shards = {}
+        for shard in shards:
+            state = shard.get("state")
+            if state not in ["INITIALIZING", "RELOCATING"]:
+                continue
+
+            unit_name = shard["node"]
+            if unit_name not in busy_shards:
+                busy_shards[unit_name] = []
+
+            busy_shards[unit_name].append(shard["index"])
+
+        return busy_shards
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -127,7 +127,7 @@ class ClusterState:
     def busy_shards_by_unit(
         opensearch: OpenSearchDistribution, host: str = None
     ) -> Dict[str, List[str]]:
-        """Get the busy shards of the indexes of the cluster."""
+        """Get the busy shards of each index in the cluster."""
         shards = ClusterState.shards(opensearch, host)
 
         busy_shards = {}
--- a/lib/charms/opensearch/v0/helper_conf_setter.py
+++ b/lib/charms/opensearch/v0/helper_conf_setter.py
@@ -241,36 +241,36 @@ class YamlConfigSetter:
             return
 
         try:
-            leaf_level = self.__leaf_level(source, node_keys)
+            leaf_container = self.__leaf_container(source, node_keys)
             leaf_key = node_keys.pop(0)
 
-            if leaf_key in leaf_level and (
-                isinstance(leaf_level[leaf_key], Mapping) or not leaf_key.startswith("[")
-            ):
-                del leaf_level[leaf_key]
+            # remove simple type elements and entire collections by key
+            if leaf_key in leaf_container and not leaf_key[0] not in {"{", "["}:
+                del leaf_container[leaf_key]
                 return
 
+            # remove element from set
             if leaf_key.startswith("{"):
-                leaf_level.remove(leaf_key[1:-1])
+                leaf_container.remove(leaf_key[1:-1])
                 return
 
-            # list
-            target_index = self.__target_array_index(leaf_level, leaf_key)
-            del leaf_level[target_index]
+            # remove element from list
+            target_index = self.__target_array_index(leaf_container, leaf_key)
+            del leaf_container[target_index]
         except AttributeError:
             # element not found
             pass
 
-    def __leaf_level(self, current, node_names: List[str]):
+    def __leaf_container(self, current, node_names: List[str]):
         if len(node_names) == 1:
             return current
 
         current_key = node_names.pop(0)
         if current_key.startswith("[") and current_key.endswith("]"):
             target_index = self.__target_array_index(current, current_key)
-            return self.__leaf_level(current[target_index], node_names)
+            return self.__leaf_container(current[target_index], node_names)
 
-        return self.__leaf_level(current[current_key], node_names)
+        return self.__leaf_container(current[current_key], node_names)
 
     @staticmethod
     def __target_array_index(source: list, node_key: str) -> int:
@@ -303,7 +303,7 @@ class YamlConfigSetter:
         """Reformat a multiline YAML array into one with square braces."""
         leaf_k = node_keys[-1]
 
-        leaf_l = self.__leaf_level(data, node_keys)
+        leaf_l = self.__leaf_container(data, node_keys)
         leaf_l[leaf_k] = self.__flow_style()
         leaf_l[leaf_k].extend(val)
 
--- a/lib/charms/opensearch/v0/helper_conf_setter.py
+++ b/lib/charms/opensearch/v0/helper_conf_setter.py
@@ -245,7 +245,7 @@ class YamlConfigSetter:
             leaf_key = node_keys.pop(0)
 
             # remove simple type elements and entire collections by key
-            if leaf_key in leaf_container and not leaf_key[0] not in {"{", "["}:
+            if leaf_key in leaf_container and leaf_key[0] not in {"{", "["}:
                 del leaf_container[leaf_key]
                 return
 
--- a/lib/charms/opensearch/v0/helper_conf_setter.py
+++ b/lib/charms/opensearch/v0/helper_conf_setter.py
@@ -240,22 +240,26 @@ class YamlConfigSetter:
         if source is None:
             return
 
-        leaf_level = self.__leaf_level(source, node_keys)
-        leaf_key = node_keys.pop(0)
+        try:
+            leaf_level = self.__leaf_level(source, node_keys)
+            leaf_key = node_keys.pop(0)
 
-        if leaf_key in leaf_level and (
-            isinstance(leaf_level[leaf_key], Mapping) or not leaf_key.startswith("[")
-        ):
-            del leaf_level[leaf_key]
-            return
+            if leaf_key in leaf_level and (
+                isinstance(leaf_level[leaf_key], Mapping) or not leaf_key.startswith("[")
+            ):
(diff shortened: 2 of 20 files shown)


tests/integration/helpers.py
```diff
@@ -139,3 +151,40 @@ async def http_request(
...
+        return False
+
+    for node_id, node_desc in response["nodes"].items():
+        node_name = node_desc["name"]
+        if node_name not in unit_names_set:
+            return False
+
+        unit_names_set.remove(node_name)
+
+    return len(unit_names_set) == 0
```
MiaAltieri, line 190:
```
shard_members = [ node_desc["name"] for node_desc in response["nodes"].items() ]
return set(unit_names) == set(shard_members)
```

tests/integration/test_charm.py
```diff
@@ -18,18 +18,7 @@
...
-            "update-status-hook-interval": "1m",
-            "cloudinit-userdata": """postruncmd:
-                - [ 'sysctl', '-w', 'vm.max_map_count=262144' ]
-                - [ 'sysctl', '-w', 'fs.file-max=1048576' ]
-                - [ 'sysctl', '-w', 'vm.swappiness=0' ]
-                - [ 'sysctl', '-w', 'net.ipv4.tcp_retries2=5' ]
-        """,
-        }
-    )
+    await ops_test.model.set_config(MODEL_CONFIG)
```
MiaAltieri, line 21: 😎

src/charm.py
```diff
@@ -300,6 +303,20 @@ def _start_opensearch(self) -> bool:
...
+                busy_shards = ClusterState.busy_shards_by_unit(self.opensearch)
+                if busy_shards:
+                    message = WaitingForBusyShards.format(
+                        " - ".join([f"{key}/{','.join(val)}" for key, val in busy_shards.items()])
+                    )
+                    self.unit.status = BlockedStatus(message)
+                    return False
+            except OpenSearchHttpError:
+                pass
+
```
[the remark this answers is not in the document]
Mehdi-Bendriss: hmm, so this indeed checks if other units are busy, to know whether it can be started (becoming "busy" on its own, as the shards syncing will only happen after the opensearch service is started) 
but since this specific unit is not started yet, it can only get this info from already started units, and the safest to query is the leader unit. 
I missed passing the leader unit host to the `busy_shards_by_unit` method.

tests/integration/helpers.py
Mehdi-Bendriss: Nice

src/charm.py
```diff
@@ -300,6 +303,20 @@ def _start_opensearch(self) -> bool:
             self.unit.status = BlockedStatus(" - ".join(missing_sys_reqs))
             return False
 
+        # check if there are shards that are "busy" or "relocating"
+        # defer the start until all the shards are "started"
+        if self.app_peers_data.get("leader_ip"):
```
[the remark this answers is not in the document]
Mehdi-Bendriss: I store the leader ip in the app data bag for convenience, as the leader host is needed to fetch the shards of the cluster

tests/integration/ha/test_horizontal_scaling.py
```diff
@@ -0,0 +1,75 @@
...
+        apps=[TLS_CERTIFICATES_APP_NAME], status="active", timeout=1000
+    )
+
+    # Relate it to OpenSearch to set up TLS.
+    await ops_test.model.relate(APP_NAME, TLS_CERTIFICATES_APP_NAME)
+    await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=5000)
+
+
+@pytest.mark.ha_tests
+@pytest.mark.abort_on_fail
```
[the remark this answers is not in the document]
Mehdi-Bendriss: At this stage, it's not. But, this will be useful for future tests that will be appended below this one, so we don't forget to put it back.

lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -240,22 +240,26 @@ def __deep_delete(self, source, node_keys: List[str]):
...
-
-        # list
-        target_index = self.__target_array_index(leaf_level, leaf_key)
-        del leaf_level[target_index]
+        try:
+            leaf_level = self.__leaf_level(source, node_keys)
+            leaf_key = node_keys.pop(0)
+
+            if leaf_key in leaf_level and (
+                isinstance(leaf_level[leaf_key], Mapping) or not leaf_key.startswith("[")
```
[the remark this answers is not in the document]
Mehdi-Bendriss: `leaf_container[leaf_key]` can refer to something else than a `Dict` here

lib/charms/opensearch/v0/helper_cluster.py
```diff
@@ -105,3 +108,38 @@ def nodes_count_by_role(nodes: List[Node]) -> Dict[str, int]:
...
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_exponential(multiplier=1, min=2, max=10),
+        reraise=True,
+    )
+    def shards(opensearch: OpenSearchDistribution, host: str = None) -> List[Dict[str, str]]:
+        """Get all shards of all indexes in the cluster."""
+        return opensearch.request("GET", "/_cat/shards", host=host)
+
+    @staticmethod
```
marcoppenheimer, line 126: nit.

What was your reasoning behind having these as static methods? they both take class, `busy_shards_by_unit` takes `self`, could we initialise `ClusterState` with `host` arg, and have these as properties?
No real downside, just curious.

lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -240,33 +240,37 @@ def __deep_delete(self, source, node_keys: List[str]):
...
+            # remove element from set
+            if leaf_key.startswith("{"):
+                leaf_container.remove(leaf_key[1:-1])
+                return
+
+            # remove element from list
+            target_index = self.__target_array_index(leaf_container, leaf_key)
+            del leaf_container[target_index]
+        except AttributeError:
+            # element not found
```
marcoppenheimer, line 261: I would `logger.debug` this case just for sanity.

lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -240,33 +240,37 @@ def __deep_delete(self, source, node_keys: List[str]):
...
+            pass
+
+    def __leaf_container(self, current, node_names: List[str]):
```
marcoppenheimer, line 264: I can't remember if this was mentioned before, if so close this.

Python convention for private methods is single `_`, as dunder `__` is reserved for builtins. Probably too much of a faff to change now, but worth mentioning if there isn't an explicit reason to have a double-under here.

src/charm.py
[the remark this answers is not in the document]
marcoppenheimer: I can't fully validate it without running, but I wonder if you might incur a slow-down here.

As `leader_ip` gets set on `leader_elected`, which runs long before `start`, units will attempt to make cluster requests before the leader has in fact `start`ed. You retry, adding time.

Might it be better to have something like (pseudo):
- on start, if leader, start, if successfully started, set `leader_ip`
- on start, if not leader, check if `leader_ip`, start

Did I understand that right?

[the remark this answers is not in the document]
marcoppenheimer: At least that way you save HTTP calls until you're reasonably confident SOMETHING has started.

tests/integration/ha/test_horizontal_scaling.py
```diff
@@ -0,0 +1,75 @@
...
+async def test_horizontal_scale_up(ops_test: OpsTest) -> None:
+    """Tests that new added units to the cluster are discoverable."""
+    # scale up
+    await ops_test.model.applications[APP_NAME].add_unit(count=2)
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME], status="active", timeout=1000, wait_for_exact_units=5
+    )
+    num_units = len(ops_test.model.applications[APP_NAME].units)
+    assert num_units == 5
```
marcoppenheimer, line 65: I'd suggest asserting an `activestatus` or equivalent here, rather than whether or not Juju added 2 more units, which one would hope was the case. 

At least then if the test fails here, we have something to look at (though it's true you validate `Active` indirectly below)

tests/integration/helpers.py
```diff
@@ -139,3 +151,32 @@ async def http_request(
...
+    Args:
+        ops_test: The ops test framework instance.
+        unit_ip: The ip of the unit of the OpenSearch unit.
+        unit_names: The list of unit names in the cluster.
+
+    Returns:
+        Whether The cluster formation is successful.
+    """
+    response = await http_request(ops_test, f"https://{unit_ip}:9200/_nodes", "GET")
+    if "_nodes" not in response or "nodes" not in response:
```
marcoppenheimer, line 174: Why the double-check for `'nodes'` substring?

marcoppenheimer: Ah, type-hint. Could avoid with a `.get('nodes', {})` on L181, but OK if not.

marcoppenheimer (approved):
It's a fairly chonky PR so hard to evaluate everything, but I think your integration tests cover enough cases to say that if they pass, LGTM! :)

I might consider adding an int-test testing 'why' we're scaling. I assume replication? Can we check replication?

lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -240,22 +240,26 @@ def __deep_delete(self, source, node_keys: List[str]):
...
-
-        if leaf_key.startswith("{"):
-            leaf_level.remove(leaf_key[1:-1])
-            return
-
-        # list
-        target_index = self.__target_array_index(leaf_level, leaf_key)
-        del leaf_level[target_index]
+        try:
+            leaf_level = self.__leaf_level(source, node_keys)
```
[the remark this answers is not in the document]
WRFitch: sounds good :)

WRFitch (approved):

lib/charms/opensearch/v0/helper_conf_setter.py
Mehdi-Bendriss: Yeah so I got into the habit of adding `__` as a way to say "class private" and no way to access it (as python name mangle methods prefixed with a double underscore), you can find more details [here](https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name).
However, I agree that it probably is overkill and not needed in this context - will remove those in a next PR

tests/integration/helpers.py
Mehdi-Bendriss: The first key starts with a `_` and the second without, the response of that HTTP call essentially looks like:
```
{
    "_nodes": ...
    "nodes": ....
}
```

src/charm.py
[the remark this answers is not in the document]
Mehdi-Bendriss: That's a very good suggestion, thanks!

lib/charms/opensearch/v0/helper_cluster.py
Mehdi-Bendriss: The reason I I put those as statics, is because they're self-contained helper functions that only depend on the input. The host here can be dynamic (when `None` it's referring to the current unit's IP, and can be set to use the host of an other unit such as the leader

tests/integration/ha/test_horizontal_scaling.py
Mehdi-Bendriss: Do I actually need to check the number of units and status at all? I'm wondering if the following is enough to check for both? 
```
await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=1000, wait_for_exact_units=5)
```

marcoppenheimer: Yup that would work!

src/charm.py
```diff
@@ -300,6 +303,20 @@ def _start_opensearch(self) -> bool:
             self.unit.status = BlockedStatus(" - ".join(missing_sys_reqs))
             return False
 
+        # check if there are shards that are "busy" or "relocating"
+        # defer the start until all the shards are "started"
```
MiaAltieri, line 312:
```suggestion
# OpenSearch automatically finds and adds shards. Meaning  that when a unit starts OpenSearch will automatically add it to the cluster. To prevent overloading the cluster, shards must be added one at a time. To insure this it is necessary to defer starting a shared until all shards have been "started" (i.e. they are not in the "busy" or "relocating" state)
```
^ add this comment and then wrap to 99 char length

MiaAltieri (changes requested):
Looks AWESOME! Really great work here, I would just change one comment.

MiaAltieri (approved):
LLTBTIES (looks like the best thing I've ever seen)

PR #17, opened 2022-11-25
[DPE-785] Implemented safe stopping of OpenSearch

Mehdi-Bendriss:
# Issue
This PR addresses  [DPE-785](https://warthogs.atlassian.net/browse/DPE-785).
Namely, this PR addresses:
- Implementing the safe stopping of the OpenSearch daemon.
- Integration tests for the safe stop.
- Minor refactoring of the base class opensearch distro.

# Solution
In order to safely stop opensearch, the following process must be followed:
- Move all shards from the data-node by shard allocation filtering - this is on a best try scenario as some shards may not be relocated if an existing replica is already in the destination node.
- Wait for the cluster health status to be green - this is on a best try scenario, the cluster might stay yellow if it couldn't move all replicas
- Stop the OpenSearch service.
- Remove the allocation exclusion from the cluster settings
  - in case no unit is online, this removal instruction must be stored in the relation data
  - the next unit that comes online should perform the exclusion removal


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -116,6 +151,12 @@ def unit_id(self) -> int:
         """ID of the current unit."""
         return int(self.unit.name.split("/")[1])
 
+    @property
+    def alternative_host(self) -> str:
+        """Return an alternative host (of another node) in case the current is offline."""
+        all_units_ips = units_ips(self, PEER)
+        return random.choice(list(all_units_ips.values()))
```
zmraul, line 158: Why doing this randomly instead of just picking the lower available unit?

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -126,13 +125,84 @@ def start(self):
...
+
+            if self._charm.alternative_host:
+                # remove the exclusion back
+                self.remove_allocation_exclusions(
+                    self._charm.unit_name, self._charm.alternative_host
+                )
+            else:
+                # no node online, store in the app databag to exclude at a future start
+                self._charm.append_allocation_exclusion_to_remove(self._charm.unit_name)
+        except Exception as e:
```
zmraul, line 156: Can a catch all here could lead to many situations where opensearch becomes bricked? Say with wrong exclusions or with incorrect data on the databag. If so, I would add a blocked status for the charm where this is being handled.

src/charm.py
```diff
@@ -240,6 +257,56 @@ def _on_get_admin_secrets_action(self, event: ActionEvent):
...
+                event.set_results(
+                    {
+                        "message": "Something happened, the OpenSearch service will re-attempt a start."
+                    }
+                )
+                return
+
+            event.set_results({"message": "The OpenSearch service is attempting a restart..."})
+        except OpenSearchStopError as e:
+            logger.error(e)
```
zmraul, line 293: See previous comment. Maybe is worth adding a blocked status as well, with a different message.

src/charm.py
```diff
@@ -240,6 +257,56 @@ def _on_get_admin_secrets_action(self, event: ActionEvent):
...
+
+    def _on_stop_service_action(self, event: ActionEvent):
+        """Stop the OpenSearch service from an action event."""
+        try:
+            self.unit.status = BlockedStatus(ServiceIsStopping)
+            self.opensearch.stop()
+            self.unit.status = BlockedStatus(ServiceStopped)
+            event.set_results({"message": "The OpenSearch service is stopping..."})
+        except OpenSearchStopError:
+            event.set_results(
```
zmraul, line 306: See previous comment. Maybe is worth adding a blocked status as well, with a different message. Also add the error to the log.

zmraul (commented):
Great PR. Some comments.

A note for future thought: `BlockedStatus` is usually for things where the admin needs to do something, but you are using it many times as a waiting status, where things are still being executed. I'm unsure of the best approach for this, but the UX is definitely different from other charms. (Even so, I think you are making great use of status to send information on `juju status` output, in addition to logs)

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: The reasoning behind was to sort of balance the load amongst all nodes, it probably doesn't matter much at this stage of the charm, though

lib/charms/opensearch/v0/opensearch_distro.py
Mehdi-Bendriss: Thanks Raul, good catch!

src/charm.py
Mehdi-Bendriss: (For the comment regarding the use of BlockedStatus instead of Maintenance / Waiting etc.. This will be the part f another PR, as it requires to also change some of the integration tests etc.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -60,6 +65,43 @@ def __init__(self, *args, distro: Type[OpenSearchDistribution] = None):
...
+        """Remove the allocation exclusions from the peer databag if existing."""
+        stored_exclusions = set(
+            self.app_peers_data.get("remove_from_allocation_exclusions", "").split(",")
+        )
+        exclusions_to_keep = ",".join(stored_exclusions - exclusions)
+
+        if self.unit.is_leader():
+            self.app_peers_data["remove_from_allocation_exclusions"] = exclusions_to_keep
+        else:
+            self.unit_peers_data["remove_from_allocation_exclusions"] = exclusions_to_keep
```
marceloneppel, line 99: I don't know enough about OpenSearch, but is ok to have a `set` with multiple exclusions here? I saw that on L79 only one unit name is saved to this field in the unit peers data. Maybe a set with multiple exclusions will not reach this part of the code, so in that case ignore this comment.

src/charm.py
```diff
@@ -240,6 +258,57 @@ def _on_get_admin_secrets_action(self, event: ActionEvent):
...
+
+        event.set_results({"message": "The OpenSearch service is attempting a restart..."})
+
+    def _on_stop_service_action(self, event: ActionEvent):
+        """Stop the OpenSearch service from an action event."""
+        try:
+            self.unit.status = MaintenanceStatus(ServiceIsStopping)
+            self.opensearch.stop()
+            self.unit.status = WaitingStatus(ServiceStopped)
+            event.set_results({"message": "The OpenSearch service is stopping..."})
```
marceloneppel, line 305: Should the message that is sent as result of the event match the waiting status (`The OpenSearch service stopped.`)?

marceloneppel (commented):
As Raúl commented, great PR Mehdi!

tests/integration/helpers.py
```diff
@@ -61,7 +69,19 @@ def get_application_unit_names(ops_test: OpsTest) -> List[str]:
...
+def get_application_unit_ids(ops_test: OpsTest) -> List[int]:
+    """List the unit IDs of an application.
+
+    Args:
+        ops_test: The ops test framework instance
+
+    Returns:
+        list of current unit ids of the application
+    """
+    return [int(unit.name.split("/")[0]) for unit in ops_test.model.applications[APP_NAME].units]
```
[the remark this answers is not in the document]
Mehdi-Bendriss: You're right! good catch!

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: Yes, so you can set 1 single exclusion (like the one with the unit_name), but in order to have multiple exclusions, they must be set as a comma separated string, i.e: "unit1, unit2, unit3..."

marceloneppel (approved):

zmraul (approved):

PR #21, opened 2022-12-06
[DPE-1033] Role management base implementation

Mehdi-Bendriss:
# Issue
This PR addresses  [DPE-1033](https://warthogs.atlassian.net/browse/DPE-1033). Namely, this PR addresses:
- Role management of the OpenSearch units (cluster_manager, data etc..)
- Change of the logic in selecting the nodes that are part of the bootstrapping process of the cluster 

# Solution
- At this point in time, the strategy is to allocate half of the units as `cluster_manager` 
- Allocate `voting_only` if needed a tiebreaker
- Allocate all roles to the remaining units
- At this point in time, the `cluster_manager` nodes are not dedicated cms, but also serve as `data` nodes.
- Any `cluster_manager` node, will be part of the bootstrapping process of the cluster
- The bootstrapping conf gets deleted from the node (if applies) after the opensearch service successfully starts


lib/charms/opensearch/v0/helper_cluster.py
```diff
@@ -29,53 +29,40 @@ def __init__(self, name: str, roles: List[str], ip: str):
...
+        max_managers = planned_units
+        max_voters = planned_units
+        if planned_units % 2 == 0:
+            max_managers -= 1
+            max_voters -= 1
 
-        if nodes_by_roles["cluster_manager"] == 1:
-            return ["cluster_manager", "data"]
+        if max_managers > 3:
+            max_managers = max_managers // 2 + 1
```
marcoppenheimer, line 52: This seemed confusing to me and not intuitive. It's also possible I'm just being slow.

Are you doing something similar to https://github.com/canonical/kafka-operator/blob/main/src/config.py#L121-L122? I'm biased as I wrote it, but it seemed easier to intuit.

If not applicable, a comment here with business logic would be great.

Mehdi-Bendriss: Are you referring to the `//` expression? This is basically the equivalent of `math.ceil(max_managers / 2)`, without the `+1` it would have been a `math.floor`.
And the goal is to have for a cluster of +3 nodes, half the nodes as `cluster_manager` 
Did i manage to clear it up?

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -134,7 +135,7 @@ def set_node(
         if len(cm_ips) > 0:
             self._opensearch.config.put(self.CONFIG_YML, "discovery.seed_hosts", cm_ips)
 
-        if "cluster_manager" in roles and len(cm_ips) < 2:  # cluster NOT bootstrapped yet
+        if "cluster_manager" in roles and contribute_to_bootstrap:  # cluster NOT bootstrapped yet
```
deusebio, line 138: I'd rather being explicit here `contribute_to_bootstrap is True`. I have seen subtle bugs as python can promote unexpected behaviour if the variable is None, empty list, int , etc. I understand that we use typing, but being explicit can also improve readability

deusebio (approved):
Codes look good. I would just be a bit more explicit on "if" statements, as subtle bugs can hide there. Unfortunately if statements do not work with boolean only, but with any type really, with unexpected behaviour sometimes. 

Apart from that, good to go

src/charm.py
```diff
@@ -339,6 +350,10 @@ def _start_opensearch(self, event: EventBase) -> None:
                 self.app_peers_data["remove_from_allocation_exclusions"]
             )
 
+        # cleanup bootstrap conf in the node
+        if self.unit_peers_data.get("bootstrap_contributor"):
```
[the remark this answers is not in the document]
Mehdi-Bendriss: The peer relation databag takes strings and not booleans, therefore `sef.unit_peers_data["bootstrap_contributor"] is True` will always yield `True` (as highlighted in your second comment).
`dict.get("key")` which also implicitely removes the need of having an additional check on the existence of the said key, as it yields `None` by default if the key doesn't exist, making the check fail.

src/charm.py
```diff
@@ -170,6 +163,14 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):
                 if exclusions_to_remove:
                     self.append_allocation_exclusion_to_remove(exclusions_to_remove)
 
+                if data.get("bootstrap_contributor"):
+                    contributor_count = int(
+                        self.app_peers_data.get("bootstrap_contributors_count", 0)
+                    )
+                    self.app_peers_data["bootstrap_contributors_count"] = str(
+                        contributor_count + 1
+                    )
```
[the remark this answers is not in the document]
Mehdi-Bendriss: I thought about it, but I usually prefer arithmetic operations to be as clear and as concise as possible 🙈

lib/charms/opensearch/v0/opensearch_config.py
Mehdi-Bendriss: I think we're pretty safe here, even if a variable is None, empty etc.. a check on that variable will evaluate as `False`.

lib/charms/opensearch/v0/helper_cluster.py
marcoppenheimer: Oh jesus I blanked and thought it was a Java comment. OK it's fine.

marcoppenheimer: Do add a comment though :)

lib/charms/opensearch/v0/opensearch_config.py
marcoppenheimer: IMO one of the nicest things about Python is that you can use these syntactic-sugar shortcuts.
Using `is` is already a shortcut, whereas being explicit would be `contribute_to_bootstrap == True`

As typing is enforced by our linting tooling, those subtle bugs are much less likely. Charm devs use static type checkers pretty ubiquitously (e.g Pylance in VSCode or similar), I think this is OK.

Example peek:

image

lib/charms/opensearch/v0/helper_cluster.py
marcoppenheimer: Or, better yet, add a new var_name for `max_managers`. `max_managers_needed/halved` is a bad name but you get the idea.

lib/charms/opensearch/v0/opensearch_config.py
deusebio: @marcoppenheimer Just a small comment that sometimes can make a difference: "is" is not the same as "==". The two  mean different thing: "is" checks whether two variables point to the same object, "==" checks whether two objects are equal (as specified by the __eq__ magic method). When using bools that works fine (since True, like 1, etc, generally native types, support singletons), but this is generally not true

PR #25, opened 2023-01-18
Add basic client relation and testing infrastructure 

WRFitch:
## Overview
This PR adds a new client relation to the opensearch charm so client applications can easily integrate with it. 

# Context
- some imported libs to ignore:
  - `tests/integration/relations/opensearch_provider/application-charm/lib/charms/data_platform_libs/v0/data_interfaces.py`
    - This one is about 1000 lines, so the diff isn't *that* bad. 

## Release Notes
- Added a charm library that implements a client relation, and associated tests

## Testing
- A test client charm has been added to `tests/integration/relations/`, for more realistic testing. This test is currently in progress.


src/charm.py
```diff
@@ -30,6 +31,20 @@ def __init__(self, *args):
         super().__init__(*args, distro=OpenSearchTarball)  # OpenSearchSnap
 
         self.framework.observe(self.on.install, self._on_install)
+        self.framework.observe(self.on.leader_elected, self._on_leader_elected)
```
Mehdi-Bendriss, line 34: These are already on the `opensearch_base_charm.py` file in the `OpenSearchBaseCharm` class. 
Basically anything common to the VM and K8s charm, should be in that class. Only the things that are proper to the VM charm should be in the `charm.py` file

tests/integration/relations/opensearch_provider/application-charm/lib/charms/data_platform_libs/v0/data_interfaces.py
```diff
@@ -0,0 +1,1080 @@
...
+
+# The unique Charmhub library identifier, never change it
+LIBID = "6c3e6b6680d64e9c89e611d1a15f65be"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
```
deusebio, line 306: which version of ops are you using? this version of the library is only compatible with <=1.5.4 . If you are using ops>=2.0.0, you should upgrade this library

WRFitch: We're on 1.5.4 - I'm happy to upgrade another time, especially as preliminary work for a juju 3 update

tests/integration/relations/opensearch_provider/application-charm/src/charm.py
```diff
@@ -0,0 +1,216 @@
...
+            method,
+            endpoint,
+            port,
+            username,
+            password,
+            host,
+            payload=payload,
+        )
+
+    def request(
```
Mehdi-Bendriss, line 159: Not urgent - But given this is a client app, it should be safe to remove this method and make use of one of the opensearch higher level python clients.

tests/integration/relations/opensearch_provider/application-charm/src/charm.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+
+        # Events related to the second database that is requested
+        # (these events are defined in the database requires charm library).
+        database_name = f'{self.app.name.replace("-", "_")}_second_database'
+        # TODO change this to include only permissions and action groups, and verify that we can
+        # create roles when necessary.
+        roles = {
+            "roles": ["readall"],
+            "permissions": ["TODO find some permissions", ""],
```
Mehdi-Bendriss, line 58: Would be interesting to have different permissions for 2 different indices, and validate the permissions accordingly.

Mehdi-Bendriss (commented):
This is starting to get the client relation to a very good state! left some questions comments here and there

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -0,0 +1,168 @@
...
+class OpenSearchProvider(Object):
+    """Defines functionality for the 'provides' side of the 'opensearch-client' relation.
+
+    Hook events observed:
+        - database-requested
+        - relation-departed
+        - relation-broken
+    """
+
+    def __init__(self, charm: CharmBase, relation_name: str = ClientRelationName) -> None:
```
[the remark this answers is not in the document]
WRFitch: 1. Unfortunately, it does indeed create a circular dependency.
2. nope, copy-paste error from pgb ;)

deusebio (commented):
Looks mostly all right! 

I only have a concern about whether we are handling exceptions correctly and whether we should implement reverting logics in the case we error out at a certain point (beside we should also add some logs to help user debugging and troubleshoot problems)

tests/integration/relations/opensearch_provider/application-charm/src/charm.py
WRFitch: This is the plan re: testing access control :)

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -0,0 +1,168 @@
...
+            # combine agroups and perms into a new role of all perms given.
+            try:
+                self.user_manager.create_role(
+                    role_name=username,
+                    permissions=permissions,
+                    action_groups=action_groups,
+                )
+                roles.append(username)
+            except OpenSearchUserMgmtError:
+                self.unit.status = BlockedStatus("bad relation request - role creation failed.")
```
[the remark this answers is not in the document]
WRFitch: I've overhauled a lot of the exception handling - give it a look and tell me what you think :)

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -0,0 +1,168 @@
...
+
```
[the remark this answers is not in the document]
Mehdi-Bendriss: We should not return anything here. 
@WRFitch - Please also replace  
```
logger.error(err)
logger.error(ClientRelationRoleCreationFailedMessage)
raise OpenSearchUserMgmtError(ClientRelationRoleCreationFailedMessage)
```
By:
```
logger.error(err)
raise
```

tests/integration/relations/opensearch_provider/application-charm/src/charm.py
WRFitch: good idea - this is just a basic implementation, but I'll switch it to a different client in the testing PR

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -0,0 +1,168 @@
...
+        if not self.opensearch.is_node_up() or not self.unit.is_leader():
+            return
+        if self._unit_departing(event.relation):
+            # This unit is being removed, so don't update the relation.
+            self.charm.peers_data.delete(Scope.UNIT, self._depart_flag(event.relation))
+            return
+
+        try:
+            self.user_manager.remove_user(self._relation_username(event.relation))
+            self.user_manager.remove_role(self._relation_username(event.relation))
```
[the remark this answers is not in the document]
WRFitch: If either of these functions fail (I've separated error handling), then the user or role will be added to the app databag as a "lingering user/role", and I've updated the update_status hook to check for these lingering users and roles, and try to remove them. 

Not totally sure on the terminology - I was calling them zombies but that feels unprofessional (despite daemons, ELFs, killing processes, GHIDRA, etc. I like the D&D vibe, but it feels unscientific).

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -0,0 +1,254 @@
...
+                hashed_pwd,
+            )
+            if roles:
+                self.user_manager.patch_user(
+                    username,
+                    [{"op": "replace", "path": "/opendistro_security_roles", "value": roles}],
+                )
+        except OpenSearchUserMgmtError as err:
+            logger.error(err)
+            raise
```
deusebio, line 177: Just to double check...In the case we error out at this point we don't want to revert the role creation in line 154, right?

deusebio (approved):
Good to go! 

Just a double check on the business logic when errorring out at creation. It is also fine to merge as-is as long as UX is what is expected @Mehdi-Bendriss

lib/charms/opensearch/v0/opensearch_relation_provider.py
Mehdi-Bendriss: @deusebio good point! I think there is not immediate need for the cleanup, I think the idea here is to "cleanup later" (periodically on `update_status`) separate "cleanup routine" to simplify and centralize the flow

Mehdi-Bendriss (approved):
LGTM

lib/charms/opensearch/v0/opensearch_relation_provider.py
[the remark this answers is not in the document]
Mehdi-Bendriss: @WRFitch I don't have a string opinion, "orphans" or "ghosts" could also be considered here

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -47,16 +47,15 @@ def create_role(
...
             "PUT",
             f"{ROLE_ENDPOINT}/{role_name}",
             {**(permissions or {}), **(action_groups or {})},
         )
-        logger.debug(put_role_resp)
-        if put_role_resp.get("status") != "OK":
-            raise OpenSearchUserMgmtError(
-                f"creating role {role_name} failed - response: {put_role_resp}"
-            )
+        logger.debug(resp)
```
delgod, line 55: is a password or certificate exposed here?

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -0,0 +1,254 @@
...
+        )
+        self.framework.observe(
+            charm.on[self.relation_name].relation_departed, self._on_relation_departed
+        )
+        self.framework.observe(
+            charm.on[self.relation_name].relation_broken, self._on_relation_broken
+        )
+
+    def _relation_username(self, relation: Relation) -> str:
+        return f"{self.relation_name}_{relation.id}_user"
```
delgod, line 90: I don't think we need postfix `_user` for every user. for example, nobody create accounts like `carlos_user@gmail.com`

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -0,0 +1,254 @@
...
+        network endpoints is unnecessary.
+        """
+        if not self.unit.is_leader():
+            return
+        if not self.opensearch.is_node_up():
+            event.defer()
+            return
+
+        try:
+            extra_user_roles = json.loads(event.extra_user_roles)
```
delgod, line 113: do we have ticket or PR where we rework extra_user_roles approach?

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -0,0 +1,254 @@
...
+            self.charm.peers_data.put(Scope.APP, "lingering_roles", (lingering_roles))
+
+    def clear_lingering_users_and_roles(self) -> None:
+        """Remove users and roles that we failed to delete when removing a relation."""
+        if not self.opensearch.is_node_up() or not self.unit.is_leader():
+            return
+
+        lingering_users = set(
+            self.charm.peers_data.get(Scope.APP, "lingering_users", "").split(",")
+        )
```
delgod, line 217: it is unclear why peer databag magic is needed 😞 
better implementation https://github.com/canonical/mongodb-operator/blob/main/lib/charms/mongodb/v0/mongodb_provider.py#L124-L126

lib/charms/opensearch/v0/opensearch_relation_provider.py
WRFitch: Yep - [DPE-1276](https://warthogs.atlassian.net/browse/DPE-1276), scheduled for this week :)

lib/charms/opensearch/v0/opensearch_users.py
WRFitch: Not by default; passwords and hashes are [hidden resources](https://opensearch.org/docs/latest/security/access-control/api/#reserved-and-hidden-resources), so they're removed from API output, and I can't find a documented option to make them visible. However, I don't think these lines are necessary, so I'm going to remove them anyway.

PR #26, opened 2023-01-18
Setting up for client relation

WRFitch:
Some QoL changes from pgbouncer that may be useful, copyright updates, and a library import to declutter PR #25

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,8 @@ jobs:
 
   unit-test:
     name: Unit tests
+    needs:
+      - lint
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout
@@ -29,7 +31,6 @@ jobs:
   integration-test-lxd-charm:
     name: Integration tests for the charm (lxd)
     needs:
-      - lint
       - unit-test
     runs-on: ubuntu-22.04
     steps:
@@ -51,7 +52,6 @@ jobs:
   integration-test-lxd-tls:
     name: Integration tests for TLS (lxd)
     needs:
-      - lint
       - unit-test
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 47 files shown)


.github/workflows/ci.yaml
```diff
@@ -28,6 +28,9 @@ jobs:
 
   integration-test-lxd-charm:
     name: Integration tests for the charm (lxd)
+    needs:
+      - lint
```
Mehdi-Bendriss, line 32: It could be worth setting this dependency on the `unit-test` job, and only depending on the `unit-test` in the integration tests.

.github/workflows/ci.yaml
```diff
@@ -42,11 +45,14 @@ jobs:
...
           sudo sysctl -w vm.swappiness=0
           sudo sysctl -w net.ipv4.tcp_retries2=5
-          
+
           tox -e charm-integration
 
   integration-test-lxd-tls:
     name: Integration tests for TLS (lxd)
+    needs:
+      - lint
```
Mehdi-Bendriss, line 54: Same as above

Mehdi-Bendriss (approved):

.github/workflows/ci.yaml
```diff
@@ -17,6 +17,8 @@ jobs:
 
   unit-test:
     name: Unit tests
+    needs:
+      - lint
```
deusebio, line 21: SUGGESTION

What about putting unit-test and linting together in the same job but different steps? This way one could do checkout and tox installation only once

deusebio (approved):
All good!

.github/workflows/ci.yaml
WRFitch: Nice idea, but they both take less than a minute and it's easy to tell what's happening at a glance if they're separate jobs - if the setup stage was longer, I'd do this, but I don't think it's necessary

PR #27, opened 2023-01-31
Cleaning up major relation PR

WRFitch:
## Proposal
Some small house-cleaning changes to make PR #25 a little less cluttered.

## Changes
- Removed unnecessary on_pull_request CI that was duplicating test runs
- Set peer relation name to be a constant so it can be used in files that can't inherit from opensearch_base_charm.py
- Moved library unit tests to their own directory under unit/lib


lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -430,3 +430,7 @@ def missing_sys_requirements() -> List[str]:
             missing_requirements.append("net.ipv4.tcp_retries2 should be 5")
 
         return missing_requirements
+
+    def version(self) -> str:
+        """Returns the version number of this opensearch instance."""
+        return self.request("GET", "/").get("version").get("number")
```
Mehdi-Bendriss, line 436: I think you need to handle an `OpenSearchHttpError` here.

I wonder, if we should not actually store the version and the static stuff in the unit peer data bag? basically at the end of node initialization, we could put the version on the unit data. So that we only make an HTTP request to fetch the version. What do you think?

lib/charms/opensearch/v0/constants_charm.py
```diff
@@ -38,3 +38,7 @@
 AdminUserInitProgress = "Configuring admin user..."
 HorizontalScaleUpSuggest = "Horizontal scale up advised: {} shards unassigned."
 WaitingForOtherUnitServiceOps = "Waiting for other units to complete the ops on their service."
+
+# Relation Interfaces
+ClientRelationName = "opensearch-client"
+PeerRelationName = "opensearch-peers"
```
Mehdi-Bendriss, line 44: nit: do we need the "Name" suffix there or would `PeerRelation` be enough?

Mehdi-Bendriss (commented):
Looks good, the only thing to handle is the exception on the version fetching

lib/charms/opensearch/v0/opensearch_distro.py
WRFitch: You're right on the HttpError, I'll add that in.

This is actually stored in the client relation databag when the relation is created, but if we want to reduce requests to opensearch I'm happy to store it in the peer databag as well.

lib/charms/opensearch/v0/constants_charm.py
WRFitch: In my view, the `Name` suffix is necessary because we have `ops` relation objects and our `provider` objects that both represent the _relation_ more accurately - I wanted to make it obvious that this only relates directly to the name.

lib/charms/opensearch/v0/opensearch_distro.py
Mehdi-Bendriss: Hmm, then I think for the time being the relation databag is enough, until we need the version on a broader context (opensearch version upgrades etc..)

lib/charms/opensearch/v0/constants_charm.py
deusebio: **suggestion**: just mentioning, but according to the [pep](https://peps.python.org/pep-0008/#constants), constants in python are usually identified by all upper case and separated with underscore. With camelcase, the names rather look like classes otherwise. 

But I understand this may need quite a refactoring as I see above a number of constants in camel cases. Just mentioning. As a suggestion, up to you whether address this

lib/charms/opensearch/v0/opensearch_distro.py
deusebio: **suggestion**: could this be a property? actually, following this discussion, another option could also be to use the cached_property [decorator](https://docs.python.org/3/library/functools.html#functools.cached_property), where you may optimize things a bit

deusebio (approved):
Just some suggestions

lib/charms/opensearch/v0/constants_charm.py
deusebio: yes, sure. Up to you, as mentioned in the comment it is a "suggestion" (non blocking comments left to the author) and I leave it up to you decide how to proceed on this

lib/charms/opensearch/v0/opensearch_distro.py
WRFitch: Since the charm object is rebuilt every time a hook fires, I don't know how useful the caching in cached_property will be, especially if this function is only called when we set up the client relations so far. Caching this across hooks means something like storing it in StoredState or a unit databag, but this fires infrequently enough that I don't feel it's necessary until we're adding version upgrades. 

I'm probably going to set this to be a normal property - TWIL about `cached_property` though!

Mehdi-Bendriss (approved):

PR #28, opened 2023-02-02
Added user management code for client relation

WRFitch:
## Proposal
Adds a library to manage users in the opensearch API, so we can create and update the users and roles necessary for the client relation. 

## Testing
Integration tests have been run on PR #25, and unit tests have been implemented.


lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+        Output of the request.
+    """
+    resp = opensearch.request("DELETE", f"{USER_ENDPOINT}/{username}/")
+    logger.debug(resp)
+    if resp.get("status") != "OK":
+        raise OpenSearchUserMgmtError()
+    return resp
+
+
+def patch_user(opensearch, user: str, patches: List[Dict[str, any]]) -> Dict[str, Any]:
```
deusebio, line 136: Missing typing

deusebio (commented):
Looks very good. Really appreciated the clean code and the documentation for every function. I just have a couple of suggestion, but do discuss this with @medhi and decide whether you like them or not, depending how you guys are into and comfortable with OOP (I'm obviously quite for it, :P )

1. Given that all functions have the first argument OpenSearchDistribution, this could be a nice spot to have a class that takes the OpenSearchDistribution in the constructor, and then it exposes methods like "create_role, remove_role, ..."

2. I don't know it if  possible at all, but maybe one could return instead of "Dict[str, Any]", some dataclass or something a bit more typed for users and roles. This could then improve the typing (having different types for the values of the dict) and the UX of users (e.g. in the IDE when you code, you already know what are the attributes, whereas using dict, you always have to figure out the name of the keys....)

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+    # check if I have to disable roles before removal
+    if resp.get("status") != "OK":
+        raise OpenSearchUserMgmtError()
+    return resp
+
+
+def create_user(
+    opensearch: OpenSearchDistribution,
+    username: str,
+    roles: List[str],  # TODO add default
```
Mehdi-Bendriss, line 84: Why do we need default roles?

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+    logger.debug(resp)
+    # check if I have to disable roles before removal
+    if resp.get("status") != "OK":
+        raise OpenSearchUserMgmtError()
+    return resp
+
+
+def create_user(
+    opensearch: OpenSearchDistribution,
+    username: str,
```
Mehdi-Bendriss, line 83: nit: `user_name` so that we stay consistent with the 2 words variables / params (i.e: `hashed_pwd`)

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+class OpenSearchUserMgmtError(Exception):
+    """Base exception class for OpenSearch user management errors."""
+
+
+def create_role(
+    opensearch: OpenSearchDistribution,
+    role_name: str,
+    permissions: Dict = {},
```
Mehdi-Bendriss, line 28: Giving a “mutable” object as a default value is dangerous in Python, as the said argument is only evaluated during the `def` phase. Though may not necessarily be the case here, it should still be avoided.

You could rather make it optional the args: `permissions: Optional[Dict[str, str]]`
then later in the code do: `{**(permissions or {}), **...}`

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Returns:
+        Output of the request.
+    """
+    resp = opensearch.request("DELETE", f"{ROLE_ENDPOINT}/{role_name}")
+    logger.debug(resp)
+    # check if I have to disable roles before removal
+    if resp.get("status") != "OK":
+        raise OpenSearchUserMgmtError()
+    return resp
```
Mehdi-Bendriss, line 78: I don't think we need to return the response here, as it is simply an `ACK`, the content might not be exploitable elsewhere. All that probably matters to us is: OK / KO. And Since you're raising an exception above, I think that should be enough.
I think the signature of the function can safely be made to return `None`

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+        OpenSearchUserMgmtError: If the request fails.
+
+    Returns:
+        Output of the request.
+    """
+    resp = opensearch.request("DELETE", f"{ROLE_ENDPOINT}/{role_name}")
+    logger.debug(resp)
+    # check if I have to disable roles before removal
+    if resp.get("status") != "OK":
+        raise OpenSearchUserMgmtError()
```
Mehdi-Bendriss, line 77: I think it would be helpful to propagate the response message in the exception, 
i.e `raise OpenSearchUsrMgmtError(resp["message"])`.

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+        payload["opendistro_security_roles"] = roles
+
+    put_user_resp = opensearch.request(
+        "PUT",
+        f"{USER_ENDPOINT}/{username}",
+        payload,
+    )
+    logger.debug(put_user_resp)
+    if put_user_resp.get("status") != "CREATED":
+        raise OpenSearchUserMgmtError()
```
Mehdi-Bendriss, line 112: Same as previous (propagate the message in the exception)

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+        OpenSearchUserMgmtError: If the request fails.
+
+    Returns:
+        Output of the request.
+    """
+    resp = opensearch.request("DELETE", f"{USER_ENDPOINT}/{username}/")
+    logger.debug(resp)
+    if resp.get("status") != "OK":
+        raise OpenSearchUserMgmtError()
+    return resp
```
Mehdi-Bendriss, line 133: Same as previous, the output does not seem helpful.

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+        Output of the request.
+    """
+    patch_user_resp = opensearch.request(
+        "PATCH",
+        f"{USER_ENDPOINT}/{user}",
+        patches,
+    )
+    logger.debug(patch_user_resp)
+    if patch_user_resp.get("status") != "OK":
+        raise OpenSearchUserMgmtError()
```
Mehdi-Bendriss, line 157: Same as previous comment

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,158 @@
...
+    return patch_user_resp
```
Mehdi-Bendriss, line 158: Same as previous comment

Mehdi-Bendriss (commented):
The logic is clear and well done! The core of the feedback is mainly about the method signatures and exception messages along with the wrapping of the whole UserManagement functions into a class as @deusebio said

lib/charms/opensearch/v0/opensearch_users.py
deusebio: *suggestion* What is the output of opensearch.request? to be honest, I would give my +1 to always return something, even just the input. This gives a bit more freedom and it leaves that up to the user (of the function) to decide what he wants to do. He/She can use it without using the output

```
self.remove_role(my_role)
```

or he/she can pipe that 

```
logger.info(f"Removing user: {self.remove_role(my_role)}")
```

Mehdi-Bendriss: The output of all the user and role management calls are as follows:
```
{
  "status": "OK",
  "message": "role test-role deleted."
}
```
So it's really an `ACK: succeeded / failed`, and since there is already a log above and the message propagated in the exception, I do not see much further value to be exploited from the return of this API. (since those are altering operations `PUT / DELETE` as opposed to a `GET` that would return something we need) 

I personally would opt for:
- Exception thrown -> failed -> catch exception / do whatever with the message (if needed)
- Exception not thrown -> call succeeded

lib/charms/opensearch/v0/opensearch_users.py
WRFitch: Actually, we don't :)

lib/charms/opensearch/v0/opensearch_users.py
WRFitch: In the [code for the actual relation](https://github.com/canonical/opensearch-operator/blob/3d21e42c621be4118d6ac6e9b4b03f610262ac83/lib/charms/opensearch/v0/opensearch_relation_provider.py), we're not actually using the response JSON for anything; as Mehdi says, it's [pretty much just an ACK](https://opensearch.org/docs/latest/security/access-control/api/#create-user), which we're checking in this function, so I think I'll remove it. If we find a use for it later we can add it, but for now I'm happy to remove it.

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __init__(self, opensearch: OpenSearchDistribution):
+        self.opensearch = opensearch
+
+    def create_role(
+        self,
+        role_name: str,
+        permissions: Optional[Dict[str, str]],
+        action_groups: Optional[Dict[str, str]],
+    ) -> Dict[str, Any]:
```
Mehdi-Bendriss, line 36: The return type must be replaced with `None` since it's no longer returning anything
(Same comment on all the other methods of this class)

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -88,21 +89,27 @@ def __init__(self, *args, distro: Type[OpenSearchDistribution] = None):
...
-        self.secrets = SecretsDataStore(self, PEER)
+        self.peers_data = RelationDataStore(self, PeerRelationName)
+        self.secrets = SecretsDataStore(self, PeerRelationName)
         self.tls = OpenSearchTLS(self, TLS_RELATION)
         self.status = Status(self)
         self.service_manager = RollingOpsManager(
             self, relation=SERVICE_MANAGER, callback=self._start_opensearch
         )
 
+        self.user_manager = OpenSearchUserManager(self.opensearch)
```
Mehdi-Bendriss, line 102: Nice!

deusebio (approved):

Mehdi-Bendriss (approved):

PR #29, opened 2023-02-03
Client Relation Tests

WRFitch:
## Proposal
Add more detailed integration tests for the opensearch client relation.

## Testing
Tests include:
- [X] Verifying that an application can actually send and receive data from indices
- [X] Verifying that we're sending the correct connection information, including opensearch version
- ~~[ ] Verifying that access control provided in the relation is applied and respected~~ 
  - This should be tested after access controls are updated to use predefined roles
- [X] Update initial relation testing to check we can create all relations simultaneously
- ~~[ ] Test scaling~~
  - TODO test once scale-down has been implemented
  - ~~[ ] check that nothing explodes when we scale~~
  - ~~[ ] check that we're updating relation endpoints properly~~
- [X] Test that we can remove the relation
  - [X] check that nothing explodes
  - [x] check that we're correctly removing the user and role we created
    - arguably this is already being checked by the application charm check.


tests/integration/relations/opensearch_provider/application-charm/src/charm.py
```diff
@@ -122,9 +118,90 @@ def _on_second_database_endpoints_changed(self, event: DatabaseEndpointsChangedE
...
+    #  Action hooks
+    # ==============
+
+    def _on_single_put_action(self, event: ActionEvent):
+        logger.info(event.params)
+        relation = self.first_database
+        relation_id = event.params["relation-id"]
+        databag = relation.fetch_relation_data()[relation_id]
+        method = "PUT"
+        payload = {"artist": "Vulfpeck", "genre": ["Funk", "Jazz"], "title": "Thrill of the Arts"}
```
deusebio, line 131: **suggestion** maybe it would be best to put the payload as parameter of the action. This way one could:

1. parametrize the call and use this action in multiple context
2. have this in the test itself, which makes it easier to understand the test in the first place: in the test we can supply certain json, run the action, and do assertion based on the input json. Now this is less clear as the `single_put_action` does some data loading which (when I read the tests) I somewhat obscure and I need to go to the action implementation to understand the data I have inserted (and based on which assertions are done)

tests/integration/relations/opensearch_provider/application-charm/src/charm.py
```diff
@@ -122,9 +118,90 @@ def _on_second_database_endpoints_changed(self, event: DatabaseEndpointsChangedE
...
+
+        event.set_results({"results": json.dumps(response)})
+
+    def _on_bulk_put_action(self, event: ActionEvent):
+        logger.info(event.params)
+        relation = self.first_database
+        relation_id = event.params["relation-id"]
+        databag = relation.fetch_relation_data()[relation_id]
+        method = "POST"
+        payload = """{ "index" : { "_index": "albums", "_id" : "2" } }
```
deusebio, line 155: as before

deusebio (approved):
All good! I have a suggestion that, although would not change the high level picture, I believe would make the test implementation more clear and scalable (to other use cases too)

**thought** In kafka we are currently factoring out the application logic into a (possibly a charm) [library](https://github.com/canonical/kafka-operator/blob/main/lib/charms/kafka/v0/client.py) (we have not commited that to charmhub yet, though). Just food for thoughts, but not to be addressed in this PR

tests/integration/relations/opensearch_provider/application-charm/src/charm.py
WRFitch: This was actually the original plan, but I ran into a bug with how JSON is parsed in actions somewhere, and I just couldn't get it through the action without it being mangled, especially when sending a multi-line bulk JSON request. I'll give it another go, since I feel your suggestion would be simpler, but I don't see a problem with having a 1-1 mapping of actions to tests in test charms.

tests/integration/relations/opensearch_provider/helpers.py
```diff
@@ -2,13 +2,55 @@
...
+    Returns:
+        the data that was requested or None
+            if no data in the relation
+
+    Raises:
+        ValueError if it's not possible to get application unit data
+            or if there is no data for the particular relation endpoint
+            and/or alias.
+    """
+    unit_name = f"{application_name}/0"
```
Mehdi-Bendriss, line 39: nit: Do you think it would make sense to pass the unit id in the arguments of the function? so that you're not constrained by unit 0

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -57,27 +63,40 @@ def create_role(
...
             raise OpenSearchUserMgmtError(
                 "role name empty - sending a DELETE request to endpoint root isn't permitted"
             )
 
-        resp = self.opensearch.request("DELETE", f"{ROLE_ENDPOINT}/{role_name}")
+        try:
+            resp = self.opensearch.request("DELETE", f"{ROLE_ENDPOINT}/{role_name}")
+        except OpenSearchHttpError as e:
+            if e.status_code == 404:
+                return {
```
Mehdi-Bendriss, line 87: Why is the returned object needed? for logging?

tests/integration/relations/opensearch_provider/application-charm/src/charm.py
```diff
@@ -183,27 +261,28 @@ def request(
...
+            "method": method.upper(),
+            "url": full_url,
+            "headers": {"Content-Type": "application/json"},
+        }
+
+        if payload:
+            if isinstance(payload, dict):
+                payload = json.dumps(payload)
+            request_kwargs["data"] = payload
+            request_kwargs["headers"]["Accept"] = "application/json"
```
Mehdi-Bendriss, line 277: The `Accept` header is related to the expected content type of the response from the server, whereas `Content-Type` is related to the sort of content the client sends to the server. 
I think this is unnecessary and it's pretty safe to just put it as part of the `headers` dict above

tests/integration/relations/opensearch_provider/test_opensearch_provider.py
```diff
@@ -37,32 +52,153 @@ async def test_database_relation_with_charm_libraries(
...
+    results = json.loads(run_bulk_read_index["results"])
+    logging.error(results)
+    assert results.get("timed_out") is False
+    assert results.get("hits", {}).get("total", {}).get("value") == 3
+
+
+@pytest.mark.client_relation
+async def test_database_version(ops_test: OpsTest):
+    """Check version is accurate."""
+    run_version_query = await run_get_from_index(
```
Mehdi-Bendriss, line 128: nit: I'm not sure the name `get_from_index` is explanatory in the case of non-data related operations, such as retrieving the health or the version etc.. what do you think?

Mehdi-Bendriss (approved):
LGTM, just a few minor comments (incl. accept header)

tests/integration/relations/opensearch_provider/helpers.py
WRFitch: Agreed, I'm now passing the `unit_id` instead

lib/charms/opensearch/v0/opensearch_users.py
WRFitch: It's really for consistency with the rest of the library - If we're sending requests I feel we should return the responses. In this case, returning this fake response allows a caller function to be aware that the role never existed in the first place, so no action is being taken.

tests/integration/relations/opensearch_provider/test_opensearch_provider.py
WRFitch: I've updated it to just be called `get_request` - how's that?

tests/integration/relations/opensearch_provider/application-charm/src/charm.py
WRFitch: I've updated all actions to be a new action called run_request, which runs HTTP requests with a given JSON payload at a set endpoint. Turns out `re.escape()` comes in handy for escaping characters properly

tests/integration/relations/opensearch_provider/test_opensearch_provider.py
Mehdi-Bendriss: I like it !

tests/integration/relations/opensearch_provider/test_opensearch_provider.py
```diff
@@ -37,32 +51,172 @@ async def test_database_relation_with_charm_libraries(
...
+    run_bulk_create_index = await run_request(
+        ops_test,
+        unit_name=ops_test.model.applications[CLIENT_APP_NAME].units[0].name,
+        relation_id=client_relation.id,
+        method="POST",
+        endpoint="/_bulk",
+        payload=re.escape(bulk_payload),
+    )
+    # change assertion to "data written" or something
+    logging.info(json.dumps(run_bulk_create_index["results"]))
```
Mehdi-Bendriss, line 124: Is an assertion missing for `run_bulk_create_index` or should the 2 lines above be deleted?

tests/integration/relations/opensearch_provider/test_opensearch_provider.py
```diff
@@ -37,32 +51,172 @@ async def test_database_relation_with_charm_libraries(
...
+    # was retrieved directly from the database.
+    version = await get_application_relation_data(
+        ops_test, f"{CLIENT_APP_NAME}/0", FIRST_DATABASE_RELATION_NAME, "version"
+    )
+    logging.error(run_version_query)
+    assert version in run_version_query["results"]
+
+
+@pytest.mark.client_relation
+async def test_multiple_relations(ops_test: OpsTest, application_charm):
```
Mehdi-Bendriss, line 161: Nice test

Mehdi-Bendriss (commented):
Should we have a test about verifying requests with a certificate generated from the CA chain of the relation?

.github/workflows/ci.yaml
```diff
@@ -4,8 +4,8 @@ on:
   pull_request:
 
 jobs:
-  lint:
-    name: Lint
+  lint-and-unit-test:
```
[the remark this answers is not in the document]
WRFitch: It was an experiment to try and use fewer runners, but these test runs aren't long at all and we could have more of an impact by optimising our integration tests instead.

.github/workflows/ci.yaml
```diff
@@ -14,45 +14,13 @@ jobs:
...
-    runs-on: ubuntu-22.04
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-      - name: Install dependencies
-        run: python -m pip install tox
       - name: Run tests
         run: tox -e unit
 
-  integration-test-lxd-charm:
```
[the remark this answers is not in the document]
WRFitch: This was also an experiment to lower the footprint of an average test run, especially when we're running enough CI that it's becoming a problem for other developers in the org. I think testing the charm enters a blocked state before applying TLS could be carried out during the TLS tests; considering this charm is supposed to not function without TLS, I think we could refactor thse tests to combine normal charm tests (such as this blocked test) with TLS tests, since they're both part of our expected baseline functionality.

lib/charms/opensearch/v0/opensearch_users.py
WRFitch: So far, this is only for internal use and has nothing to do with the client (requirer) app. Its intended use is to remove roles when the relation is broken, or when we're removing lingering users. If we're trying to remove users that don't exist, I want to return a positive signal that the user doesn't exist, so the caller function can continue with the knowledge that user definitely doesn't exist. However, I forgot to add `else: raise` to this function, so I've added that in

tests/integration/relations/opensearch_provider/test_opensearch_provider.py
WRFitch: They have been removed

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -302,22 +309,22 @@ def request(
...
-                    verify=f"{self.paths.certs}/chain.pem",
-                    headers={"Accept": "application/json", "Content-Type": "application/json"},
-                )
-
+                resp = s.request(**request_kwargs)
                 resp.raise_for_status()
         except requests.exceptions.RequestException as e:
             logger.error(f"Request {method} to {full_url} with payload: {payload} failed. \n{e}")
-            raise OpenSearchHttpError()
+            raise OpenSearchHttpError(status_code=e.response.status_code)
```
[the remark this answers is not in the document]
WRFitch: Response is guaranteed to exist, and status_code is initialised to None in the constructor so this should be available. `raise_for_status()` explicitly states that it exists to raise HTTPError, so I feel this is reliable, but I've added a RequestException handler that just raises a basic OpenSearchHttpError.

deusebio (approved):
good to go!

Mehdi-Bendriss (approved):

PR #32, opened 2023-02-13
Removed lingering user code

WRFitch:
## Proposal
This PR simplifies lingering user management by querying the database directly.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_users.py
+++ b/lib/charms/opensearch/v0/opensearch_users.py
@@ -28,12 +28,23 @@ class OpenSearchUserManager:
     def __init__(self, opensearch: OpenSearchDistribution):
         self.opensearch = opensearch
 
+    def get_roles(self) -> Dict[str, any]:
+        """Gets list of roles.
+
+        Raises:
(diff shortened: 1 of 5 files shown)


lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -22,18 +24,29 @@ class OpenSearchUserMgmtError(Exception):
     """Base exception class for OpenSearch user management errors."""
 
 
-class OpenSearchUserManager:
+class OpenSearchUserManager(Object):
```
Mehdi-Bendriss, line 27: Why are we extending `Object` here?

Mehdi-Bendriss (commented):
Looks good! Just a question about the need to extend `Object`

delgod (approved):

lib/charms/opensearch/v0/opensearch_users.py
Mehdi-Bendriss: I think you already have the charm, so probably `self.model = self.charm.model` and removing the `Object`?

Mehdi-Bendriss (approved):

PR #36, opened 2023-02-20
[DPE-1260] Admin password rotation

Mehdi-Bendriss:
# Issue
This PR implements the `password rotation` feature (spec DA021). Namely, this PR addresses:
- [DPE-1260](https://warthogs.atlassian.net/browse/DPE-1260) implementation of the `get / set-password` actions.
- [DPE-1263](https://warthogs.atlassian.net/browse/DPE-1263) unit tests
- [DPE-1264](https://warthogs.atlassian.net/browse/DPE-1264) integration tests

# Solution
- The roles of the admin user must be explicitly allowed to perform user related operations (i.e changing password)
- Initially the admin password is generated and its hash set on the `internal_users.yml`
- Once OpenSearch is started, in order to change it, the password rotation is done in the following steps: 
    - generate password if not passed
    - generate hash 
    - patch the user using the rest api
    - update the peer rel data

[DPE-1260]: https://warthogs.atlassian.net/browse/DPE-1260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-1263]: https://warthogs.atlassian.net/browse/DPE-1263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-1264]: https://warthogs.atlassian.net/browse/DPE-1264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


actions.yaml
```diff
@@ -20,5 +20,21 @@ set-tls-private-key:
...
+    username:
+      type: string
+      description: The username, the default value 'admin'. Possible values - admin.
+      default: admin
+    password:
+      type: string
+      description: The password will be auto-generated if this option is not set.
+
+get-password:
+  description: Fetch the system user's password and CA chain, which is used by charm. It is for internal charm users and SHOULD NOT be used by applications.
```
WRFitch, line 35: Stating that these credentials are only for internal charm users makes me question the purpose of these actions. Maybe it should be made more explicit that these are for internal charm use, and for juju admins to interface directly with the application. If an application wants  admin access, it should set `extra-user-roles="admin"` in a client relation, rather than accessing this.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -324,16 +327,46 @@ def _on_update_status(self, event: UpdateStatusEvent):
...
 
-    def _on_get_admin_secrets_action(self, event: ActionEvent):
+    def _on_set_password_action(self, event: ActionEvent):
+        """Set new admin password from user input or generate if not passed."""
+        if not self.unit.is_leader():
+            event.fail("The action can be run only on leader unit.")
+            return
+
+        user_name = event.params.get("username")
+        if event.params.get("username") != "admin":
```
WRFitch, line 337: Suggestion: might be worth making an `ADMIN_USERNAME` parameter

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -542,23 +575,39 @@ def _can_service_start(self) -> bool:
...
 
-    def _initialize_admin_user(self):
+    def _initialize_admin_user(self, pwd: Optional[str] = None):
         """Change default password of Admin user."""
-        hashed_pwd, pwd = generate_hashed_password()
-        self.secrets.put(Scope.APP, "admin_password", pwd)
+        is_update = pwd is not None
+        hashed_pwd, pwd = generate_hashed_password(pwd)
+
+        if is_update:
```
WRFitch, line 583: Suggestion: perhaps this could be a separate function named `update_admin_user`? It's no longer being initialised in this case.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -542,23 +575,39 @@ def _can_service_start(self) -> bool:
...
+            resp = self.opensearch.request(
+                "PATCH",
+                "/_plugins/_security/api/internalusers/admin",
+                [{"op": "replace", "path": "/hash", "value": hashed_pwd}],
+            )
+            logger.debug(f"Patch admin user response: {resp}")
+            if resp.get("status") != "OK":
+                raise OpenSearchError(f"{resp}")
+        else:
+            # reserved: False, prevents this resource from being update-protected from:
```
WRFitch, line 594: If we're no longer reserving this user, should we [hide](https://opensearch.org/docs/latest/security/access-control/api/#reserved-and-hidden-resources) it?

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -153,6 +153,19 @@ def set_node(
...
+        # security plugin rest API access
+        self._opensearch.config.put(
+            self.CONFIG_YML,
+            "plugins.security.restapi.roles_enabled",
+            ["all_access", "security_rest_api_access"],
+        )
+        # to use the PUT and PATCH methods of the security rest API
+        self._opensearch.config.put(
+            self.CONFIG_YML,
+            "plugins.security.unsupported.restapi.allow_securityconfig_modification",
```
WRFitch, line 165: As this function expands, it might be worth storing a `default_node_config.yml` file somewhere and updating it all at once, instead of rewriting this file multiple times

tests/integration/actions/test_password_actions.py
```diff
@@ -0,0 +1,129 @@
...
+from tests.integration.tls.test_tls import TLS_CERTIFICATES_APP_NAME
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.actions_tests
+@pytest.mark.abort_on_fail
+@pytest.mark.skip_if_deployed
+async def test_build_and_deploy(ops_test: OpsTest) -> None:
+    """Build and deploy one unit of OpenSearch."""
```
WRFitch, line 30: We're actually deploying two units here

tests/integration/actions/test_password_actions.py
```diff
@@ -0,0 +1,129 @@
...
+    get_leader_unit_ip,
+    http_request,
+    run_action,
+)
+from tests.integration.tls.test_tls import TLS_CERTIFICATES_APP_NAME
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.actions_tests
```
WRFitch, line 26: I feel this could be added into the charm_tests tag

actions.yaml
```diff
@@ -20,5 +20,21 @@ set-tls-private-key:
   required:
     - category
 
-get-admin-secrets:
-  description: Get the admin user password and cert chain used by charm.
+set-password:
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Yes, this is implementing the spec `DA021` :)

actions.yaml
Mehdi-Bendriss: I think this text is related to the spec DA021, and in the spirit of staying consistent with the other charms, we may want to keep it as-is unless / until we update the specs.

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: I do not think so, when we get to the opensearch dashboard, I believe it would be useful to provide the full transparent list of existing users in the cluster.

lib/charms/opensearch/v0/opensearch_config.py
Mehdi-Bendriss: It was decided by Mykola some time ago that we should NOT have templates (that we fill say with jinja or other), but what we could do in a refactoring ticket is to update all those keys at once, by providing a dictionary. So that we write once and that we make it clear all the keys that must be written

WRFitch: ok, that seems like an acceptable compromise. Why are we avoiding templates?

tests/integration/actions/test_password_actions.py
Mehdi-Bendriss: I agree, if we end up with too many actions, we could always externalize those later

lib/charms/opensearch/v0/opensearch_config.py
Mehdi-Bendriss: The reasoning was that it was overkill, and to avoid adding yet another component to manage (i.e jinja)

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: I renamed it to `_put_admin_user`, as what we're essentially doing is in fact not initializing it by updating it in both cases, first case on disk where the initial stage user:pwd is `admin:admin` to `admin:generated_pwd` and the second case using the API to achieve the same

lib/charms/opensearch/v0/opensearch_config.py
WRFitch: ok, thanks for the explanation!

WRFitch (approved):

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -324,16 +327,46 @@ def _on_update_status(self, event: UpdateStatusEvent):
...
+        user_name = event.params.get("username")
+        if event.params.get("username") != "admin":
+            event.fail("Only the 'admin' username is allowed for this action.")
+            return
+
+        password = event.params.get("password") or generate_password()
+        try:
+            self._put_admin_user(password)
+            password = self.secrets.get(Scope.APP, "admin_password")
+            event.set_results({f"{user_name}-password": password})
```
deusebio, line 345: **suggestion**  Uhm, maybe I should remove the "-password". I suppose it is already fairly clear in the UX

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -324,16 +327,46 @@ def _on_update_status(self, event: UpdateStatusEvent):
...
-    def _on_get_admin_secrets_action(self, event: ActionEvent):
+    def _on_set_password_action(self, event: ActionEvent):
+        """Set new admin password from user input or generate if not passed."""
+        if not self.unit.is_leader():
+            event.fail("The action can be run only on leader unit.")
+            return
+
+        user_name = event.params.get("username")
+        if event.params.get("username") != "admin":
+            event.fail("Only the 'admin' username is allowed for this action.")
```
deusebio, line 338: **question** I might miss something here, but what is the point of making an action with a parameter username if I cannot really provide other values than admin?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -324,16 +327,46 @@ def _on_update_status(self, event: UpdateStatusEvent):
...
-        chain = ""
-        admin_secrets = self.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
-        if admin_secrets and admin_secrets.get("chain"):
-            chain = "\n".join(admin_secrets["chain"][::-1])
+        if not self._is_tls_fully_configured():
+            event.fail("admin user or TLS certificates not configured yet.")
+            return
 
-        event.set_results({"password": password if password else "", "chain": chain})
+        admin_password = self.secrets.get(Scope.APP, "admin_password")
```
deusebio, line 359: **suggestion** maybe f"{username}_password" to be parametric, if we plan to support multiple usernames

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -324,16 +327,46 @@ def _on_update_status(self, event: UpdateStatusEvent):
...
+
+        user_name = event.params.get("username")
+        if event.params.get("username") != "admin":
+            event.fail("Only the 'admin' username is allowed for this action.")
+            return
+
+        password = event.params.get("password") or generate_password()
+        try:
+            self._put_admin_user(password)
+            password = self.secrets.get(Scope.APP, "admin_password")
```
deusebio, line 344: **suggestion** I'd use f"{user_name}-password" if we plan to extend this

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -542,23 +575,39 @@ def _can_service_start(self) -> bool:
 
         return True
 
-    def _initialize_admin_user(self):
-        """Change default password of Admin user."""
-        hashed_pwd, pwd = generate_hashed_password()
-        self.secrets.put(Scope.APP, "admin_password", pwd)
+    def _put_admin_user(self, pwd: Optional[str] = None):
```
deusebio, line 578: **suggestion** making this parametric with the username if we plan to extend this?

deusebio (commented):
The code looks really fine! 

I only have one issue about the UX here, in the sense that it is kinda weird to me that we don't allow other username other than "admin". I understand this was according to this [spec](https://docs.google.com/document/d/1x0cy2H1RvSX6o-qlYWIQF1uNN2EqHthrT1N1CEm4FsY/edit) and we want to be generic, but if the action will never be parametric, I would really just skip the parameter in its definition. 
I understand it is important to promote standardisation, but if something does really not apply here, I believe it would be confusing for the user to have a parameter with a fixed value :D

On the other hand, If it could ever be parametric (like people that want to call their admin in other ways, e.g. in kafka we support two users: admin and sync), then we should write the business logic with less "admin" hardcoded, 

Apart from that, I believe it is all good and nice!

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: That is a good point, and it doesn't make much sense in OpenSearch at this stage at least. But the idea is to stay consistent with the UX of the other charms.

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: I also agree, but it seems that all the other charms implement it this way, so i'm following their UX for consistency

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: My understanding is that we do not plan to having more users in OpenSearch (outside the client relation) - and that we'll only have the `admin` user, I'll confirm this and report back

lib/charms/opensearch/v0/opensearch_base_charm.py
deusebio: yes, indeed!

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: This i would prefer to leave for later - when we introduce new users. As it won't be enough to just add the `username`, but also the `backend_roles` and potentially other attributes we don't yet know about.

deusebio (approved):
Good to go!

PR #45, opened 2023-02-27
Nodes config broadcast on scale-down - logic change 

Mehdi-Bendriss:
### Issue
This PR is about changing the broadcast of the nodes' config when a scale-down event happens. So that we no longer rely on the peer relation unit's data.

### Solution
The nodes broadcasting no longer targets an individual unit, but is rather broadcasted to all units
The "change" is calculated based on comparing the passed config with the locally stored one in `opensearch.yml`


WRFitch (approved):

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -720,33 +709,35 @@ def _cleanup_bootstrap_conf_if_applies(self) -> None:
...
+        """Reconfigure the current unit if a new config was computed for it, then restart."""
+        nodes_config = self.peers_data.get_object(Scope.APP, "nodes_config")
+        if not nodes_config:
             return
 
-        node = Node.from_dict(node_conf_update_app)
-        if node.name != self.unit_name:
+        new_node_conf = nodes_config.get(self.unit_name)
+        if not new_node_conf:
             return
```
delgod, line 720: if we don't have a config for the member, maybe it is become removed from the configuration.
should we stop OpenSearch?

delgod (changes requested):

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -475,18 +468,12 @@ def _start_opensearch(self, event: EventBase) -> None:  # noqa: C901
...
-                self.peers_data.get_object(Scope.UNIT, "update-node-config")
-            )
-
         try:
             # Retrieve the nodes of the cluster, needed to configure this node
             nodes = self._get_nodes(False)
 
             # Set the configuration of the node
-            self._set_node_conf(nodes, updated_node_conf.roles if updated_node_conf else None)
+            self._set_node_conf(nodes, conf)
```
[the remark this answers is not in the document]
Mehdi-Bendriss: This was like an "override" - as `_set_node_conf` generates defaults, but I could do the conf fetch in it as well

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: Good point, I think we should restart it in that case, and its conf will be auto-generated

delgod: what if `config exists + no node in config + opensearch is up`

delgod (approved):

PR #50, opened 2023-03-08
Simplify relation users

WRFitch:
## Proposal
Simplify relation users in line with the [spec](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/opensearch_client/v0). Requirer charms now need to send either "admin" or "default" to receive relevant permissions. 

## Context
- These permissions are a first draft, but require focus when reviewing. 
  - [Opensearch Permissions Docs](https://opensearch.org/docs/latest/security/access-control/permissions/)

## Release Notes
- Simplify user creation to two values
- Updated tests. 

## Testing
- Integration & unit tests have been updated to match the new user permissions. 

## TODO 
- Add admin permissions list

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_relation_provider.py
+++ b/lib/charms/opensearch/v0/opensearch_relation_provider.py
@@ -14,18 +14,16 @@ Default security values can be found in the opensearch documentation here:
 https://opensearch.org/docs/latest/security/access-control/index/.
 
 """
-
 import logging
+from copy import deepcopy
 from enum import Enum
+from typing import Dict
 
 from charms.data_platform_libs.v0.data_interfaces import (
     DatabaseProvides,
     DatabaseRequestedEvent,
 )
-from charms.opensearch.v0.constants_charm import (
-    ClientRelationName,
-    PeerRelationName,
-)
+from charms.opensearch.v0.constants_charm import ClientRelationName, PeerRelationName
 from charms.opensearch.v0.helper_databag import Scope
 from charms.opensearch.v0.helper_networking import units_ips
 from charms.opensearch.v0.helper_security import generate_hashed_password
@@ -47,24 +45,16 @@ LIBAPI = 0
 LIBPATCH = 1
 
 
-class ExtraUserRole:
-    def __init__(self, extra_user_roles: str, index: str):
-        self.index = index
-        try:
-            self.permissions = ExtraUserRolePermissions[extra_user_roles.upper()]
-        except TypeError:
-            logger.error("invalid extra_user_roles parameter supplied to relation")
-            raise
-
 class ExtraUserRolePermissions(Enum):
     """An enum of user types and their associated permissions."""
 
-    # Default user has CRUD in a specific index. Update index_patterns to include the index.
+    # Default user has CRUD in a specific index. Update index_patterns to include the index to
+    # which these permissions are applied.
     DEFAULT = {
         "cluster_permissions": ["cluster_monitor"],
         "index_permissions": [
             {
-                "index_patterns": [TODO index name goes here],
+                "index_patterns": [],
                 "dls": "",
                 "fls": [],
                 "masked_fields": [],
@@ -86,11 +76,6 @@ class ExtraUserRolePermissions(Enum):
     # - Removing indices they have created
     # - Node roles
     ADMIN = {}
-
-    def set_index(self, index_name: str):
-        """Set these permissions for a specific index."""
-        for perm_set in self["index_permissions"]:
-            perm_set["index_patterns"] = index_name
 
 
 class OpenSearchProvider(Object):
@@ -154,8 +139,9 @@ class OpenSearchProvider(Object):
 
         username = self._relation_username(event.relation)
         hashed_pwd, pwd = generate_hashed_password()
+        extra_user_roles = event.extra_user_roles if event.extra_user_roles else "default"
         try:
-            self.create_opensearch_users(username, hashed_pwd, event.index, user_roles)
+            self.create_opensearch_users(username, hashed_pwd, event.index, extra_user_roles)
         except OpenSearchUserMgmtError as err:
             logger.error(err)
             self.unit.status = BlockedStatus(str(err))
@@ -168,38 +154,64 @@ class OpenSearchProvider(Object):
         self.database_provides.set_version(rel_id, self.opensearch.version)
 
     def create_opensearch_users(
-        self, username: str, hashed_pwd: str, index: str, extra_user_roles: ExtraUserRolePermissions
+        self,
+        username: str,
+        hashed_pwd: str,
+        index: str,
+        extra_user_roles: str,
     ):
         """Creates necessary opensearch users and permissions for this relation.
 
         Args:
             username: Username to be created
             hashed_pwd: the hash of the password to be assigned to the user
-            access_control: A dict of roles, permissions, and action groups to be applied to the
-                user. A new role will be created to contain the requested permissions and action
-                groups.
+            index: the index to which the users must be granted access
+            extra_user_roles: the level of permissions that the user should be given.
 
         Raises:
             OpenSearchUserMgmtError if user creation fails
         """
-        # TODO this might be setting the values inside the enum?
-        extra_user_roles.set_index(index)
         try:
             # Create a new role for this relation, encapsulating the permissions we care about. We
             # can't create a "default" and an "admin" role once because the permissions need to be
             # set to this relation's specific index.
             self.user_manager.create_role(
                 role_name=username,
-                permissions=extra_user_roles.value,
+                permissions=self.get_extra_user_role_permissions(extra_user_roles, index),
             )
-            self.user_manager.create_user(username, [username], hashed_pwd)
+            roles = [username]
+            self.user_manager.create_user(username, roles, hashed_pwd)
             self.user_manager.patch_user(
                 username,
-                [{"op": "replace", "path": "/opendistro_security_roles", "value": [username]}],
+                [{"op": "replace", "path": "/opendistro_security_roles", "value": roles}],
             )
         except OpenSearchUserMgmtError as err:
             logger.error(err)
             raise
+
+    def get_extra_user_role_permissions(self, extra_user_roles: str, index: str) -> Dict[str, any]:
+        """Get relation role permissions from the extra_user_roles field.
+
+        Args:
+            extra_user_roles: role requested by the requirer unit, provided in relation databag.
+                This needs to be one of "admin" or "default", or it will be set to "default".
+                TODO should this fail and raise an error instead so provider charm authors can
+                guarantee they're getting the perms they expect?
+            index: if these permissions are index-specific, they will be assigned to this index.
+
+        Returns:
+            A dict containing the required permissions for the requested role.
+        """
+        logger.error(ExtraUserRolePermissions._member_names_)
+        if extra_user_roles.upper() not in ExtraUserRolePermissions._member_names_:
+            extra_user_roles = "default"
+
+        permissions = deepcopy(ExtraUserRolePermissions[extra_user_roles.upper()])
+
+        # TODO verify if this needs to be applied to admin role. Probably does.
+        if extra_user_roles == "default":
+            for perm_set in permissions["index_permissions"]:
+                perm_set["index_patterns"] = [index]
 
     def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
         """Check if this relation is being removed, and update the peer databag accordingly."""
--- a/lib/charms/opensearch/v0/opensearch_relation_provider.py
+++ b/lib/charms/opensearch/v0/opensearch_relation_provider.py
@@ -19,6 +19,8 @@ TODO add tls
 
 import json
 import logging
+from dataclasses import dataclass
+from enum import Enum
 from typing import Dict, List
 
 from charms.data_platform_libs.v0.data_interfaces import (
@@ -49,6 +51,52 @@ LIBAPI = 0
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 LIBPATCH = 1
+
+
+class ExtraUserRole:
+    def __init__(self, extra_user_roles: str, index: str):
+        self.index = index
+        try:
+            self.permissions = ExtraUserRolePermissions[extra_user_roles.upper()]
+        except TypeError:
+            logger.error("invalid extra_user_roles parameter supplied to relation")
+            raise
+
+class ExtraUserRolePermissions(Enum):
+    """An enum of user types and their associated permissions."""
+
+    # Default user has CRUD in a specific index. Update index_patterns to include the index.
+    DEFAULT = {
+        "cluster_permissions": ["cluster_monitor"],
+        "index_permissions": [
+            {
+                "index_patterns": [],
+                "dls": "",
+                "fls": [],
+                "masked_fields": [],
+                "allowed_actions": [
+                    "indices_monitor",
+                    "create_index",
+                    "crud",
+                    "data_access",
+                    "indices:data/read/search",
+                    "indices:admin/create",
+                ],
+            }
+        ],
+    }
+
+    # Admin user has control over:
+    # - creating multiple indices
+    # - creating and deleting users and assigning their access controls
+    # - Removing indices they have created
+    # - Node roles
+    ADMIN = {}
+
+    def set_index(self, index_name: str):
+        """Set these permissions for a specific index."""
+        for perm_set in self["index_permissions"]:
+            perm_set["index_patterns"] = index_name
 
 
 class OpenSearchProvider(Object):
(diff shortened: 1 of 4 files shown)


Mehdi-Bendriss:
To summarize the flow.

### What do we need to protected:
1- internal indices (list), mainly for now: `.opendistro_security`, here is a (non?) exhaustive list:
```
[".opendistro_security", ".opendistro-alerting-config", ".opendistro-alerting-alert*", ".opendistro-anomaly-results*", ".opendistro-anomaly-detector*", ".opendistro-anomaly-checkpoints", ".opendistro-anomaly-detection-state"]
``` 
2. The access to the security rest API

---

### To be tested:
1. Try to delete `.opendistro_security` by an `admin client`, if not possible => perfect. If possible => see the section below "1."
2. Try to delete `an internal user` using the rest API, if not possible => perfect. If possible => see the section below "2."

---

### How to protect these?
#### 1. internal indices:
    a. in your permission, you could try to exclude by pattern `.opendistro*` or list them all in case the internal indices may break this pattern
    b. on client relation joined, if index name passed (i.e for non admin users), check the requested index against this list and fail if it's one of the internal ones.

#### 2. Access to security rest api:
On cluster initialization (where we initialize the admin user and security index)
1. create 2 roles: 
  a. `client-role` 
  b. `client-admin-role` which inherits from the `admin` role.
2. in `opensearch.yml` exclude these 2 roles from ever interacting with the security API: (refer to the docs [here](https://opensearch.org/docs/latest/security/access-control/api/#access-control-for-the-api))
       
        # for the client role:
       plugins.security.restapi.endpoints_disabled.client-role.ROLES: ["PUT", "POST", "DELETE", "PATCH"]
       plugins.security.restapi.endpoints_disabled.client-role.INTERNALUSERS: ["PUT", "POST", "DELETE", "PATCH"]      

       # for the client-admin-role
       plugins.security.restapi.endpoints_disabled.client-admin-role.ROLES: ["PUT", "POST", "DELETE", "PATCH"]
       plugins.security.restapi.endpoints_disabled.client-admin-role.INTERNALUSERS: ["PUT", "POST", "DELETE", "PATCH"]
3. For each new client:
  a. when relation joining, create a user mapping with the corresponding role (regular user => `client-role` and admin user => `client-admin-role`)
  b. when relation broken, delete the user mapping 

Note: with this, you probably don't need to set user permissions for client `admin` users.

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -144,54 +192,71 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:
...
+                guarantee they're getting the perms they expect?
+            index: if these permissions are index-specific, they will be assigned to this index.
+
+        Returns:
+            A dict containing the required permissions for the requested role.
+        """
+        if extra_user_roles.upper() not in ExtraUserRolePermissions._member_names_:
+            extra_user_roles = "default"
+
+        permissions = ExtraUserRolePermissions[extra_user_roles.upper()].value
```
delgod, line 252: extra_user_roles should be a comma-separated list.
It is ok to solve in separate PR

delgod (changes requested):

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -144,54 +187,79 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:
...
 
         Args:
             username: Username to be created
             hashed_pwd: the hash of the password to be assigned to the user
-            access_control: A dict of roles, permissions, and action groups to be applied to the
-                user. A new role will be created to contain the requested permissions and action
-                groups.
+            index: the index to which the users must be granted access
+            extra_user_roles: the level of permissions that the user should be given. Can be a
+                comma-separated list of permissions, which should be merged.
```
Mehdi-Bendriss, line 209:
```suggestion
                comma-separated list of roles, which should result in a merged list of permissions.
```

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -120,22 +159,26 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:
...
+            if not (
+                e.response_code == 400
+                and e.response_body.get("error", {}).get("type")
+                == "resource_already_exists_exception"
+            ):
+                raise
 
         username = self._relation_username(event.relation)
         hashed_pwd, pwd = generate_hashed_password()
+        extra_user_roles = event.extra_user_roles if event.extra_user_roles else "default"
```
Mehdi-Bendriss, line 179: Could we lowercase the extra_user_roles as a safeguard? 
```suggestion
        extra_user_roles = event.extra_user_roles.lower() if event.extra_user_roles else "default"
```

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -53,22 +52,62 @@
...
+        "index_permissions": [
+            {
+                "index_patterns": [],
+                "dls": "",
+                "fls": [],
+                "masked_fields": [],
+                "allowed_actions": [
+                    "indices_monitor",
+                    "crud",
+                    "data_access",  # How is this different to crud?
```
Mehdi-Bendriss, line 72: Only `data_access` should stay:
> Combines the `crud` action group with indices:data/*.

For [reference](https://opensearch.org/docs/latest/security/access-control/default-action-groups/)

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -53,22 +52,62 @@
...
+        "cluster_permissions": ["cluster_monitor"],
+        "index_permissions": [
+            {
+                "index_patterns": [],
+                "dls": "",
+                "fls": [],
+                "masked_fields": [],
+                "allowed_actions": [
+                    "indices_monitor",
+                    "crud",
```
Mehdi-Bendriss, line 71: Safe to delete

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -53,22 +52,62 @@
...
+                    "data_access",  # How is this different to crud?
+                    "indices:data/read/search",  # How is this different to crud?
```
Mehdi-Bendriss, line 73: safe to delete

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -53,22 +52,62 @@
...
+                ],
+            }
+        ],
+        "cluster_permissions": [
+            "indices_all",
+            "cluster_all",
+            "crud",
+            "manage",
+            "indices:admin/*",
+            "cluster:admin/*",
```
Mehdi-Bendriss, line 101: Safe to delete, equal to `cluster_all`

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -53,22 +52,62 @@
...
+                    "crud",
+                ],
+            }
+        ],
+        "cluster_permissions": [
+            "indices_all",
+            "cluster_all",
+            "crud",
+            "manage",
+            "indices:admin/*",
```
Mehdi-Bendriss, line 100: Safe to delete, equal to `indices_all`

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -53,22 +52,62 @@
...
+                    "cluster_all",
+                    "indices_all",
+                    "crud",
+                ],
+            }
+        ],
+        "cluster_permissions": [
+            "indices_all",
+            "cluster_all",
+            "crud",
```
Mehdi-Bendriss, line 98: safe to delete, included in `indices_all`

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -144,54 +187,79 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:
...
             raise
 
+    def get_extra_user_role_permissions(self, extra_user_roles: str, index: str) -> Dict[str, any]:
+        """Get relation role permissions from the extra_user_roles field.
+
+        Args:
+            extra_user_roles: role requested by the requirer unit, provided in relation databag.
+                This needs to be one of "admin" or "default", or it will be set to "default".
+                TODO should this fail and raise an error instead so provider charm authors can
+                guarantee they're getting the perms they expect?
```
Mehdi-Bendriss, line 239: I would say that it should fail. @delgod ?

tests/integration/relations/opensearch_provider/application-charm/src/charm.py
```diff
@@ -31,46 +31,38 @@ class ApplicationCharm(CharmBase):
...
 
-        self.framework.observe(
-            self.first_opensearch.on.index_created, self._on_authentication_updated
-        )
-        self.framework.observe(
-            self.first_opensearch.on.authentication_updated, self._on_authentication_updated
-        )
+        # Checking comma-separated permissions. These should still basically have admin
+        # permissions.
+        self.admin_opensearch = OpenSearchRequires(self, "admin", "admin-index", "admin,default")
```
Mehdi-Bendriss, line 46: nice

lib/charms/opensearch/v0/opensearch_relation_provider.py
WRFitch: I've added this in, let me know what you think

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -81,8 +81,8 @@ def create_role(
             f"{ROLE_ENDPOINT}/{role_name}",
             payload={**(permissions or {}), **(action_groups or {})},
         )
-        if resp.get("status") != "OK":
-            raise OpenSearchUserMgmtError(f"creating role {role_name} failed")
+        # if resp.get("status") != "OK":
+        #     raise OpenSearchUserMgmtError(f"creating role {role_name} failed")
```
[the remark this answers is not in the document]
WRFitch: done

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -53,22 +52,67 @@
...
+class ExtraUserRolePermissions(Enum):
+    """An enum of user types and their associated permissions."""
+
+    # Default user has CRUD in a specific index. Update index_patterns to include the index to
+    # which these permissions are applied.
+    DEFAULT = {
+        "cluster_permissions": ["cluster_monitor"],
+        "index_permissions": [
+            {
+                # TODO exclude user index (.opensearch_distro) (and have a set of protected
```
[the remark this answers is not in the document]
WRFitch: It depends if you want to make sure we prevent access to the files you mentioned above, I'm happy either way.

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -120,22 +164,26 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:
         """
         if not self.unit.is_leader():
             return
-        if not self.opensearch.is_node_up():
+        if not self.opensearch.is_node_up() or not event.index:
```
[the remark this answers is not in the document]
WRFitch: If it's not provided in the databag for whatever reason, then it evaluates to false. Maybe it's a bit too defensive, I'll try removing it and running tests

delgod (approved):

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -81,7 +81,10 @@ def create_role(
             f"{ROLE_ENDPOINT}/{role_name}",
             payload={**(permissions or {}), **(action_groups or {})},
         )
-        if resp.get("status") != "OK":
+        logger.error(resp)
+        if resp.get("status") not in ["CREATED", "OK"]:
+            logger.error(f"THIS COMMAND DIDN'T WORK AND HERE'S WHY {resp}")
```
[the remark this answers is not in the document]
WRFitch: whoops, I should make my debug logs a little less obvious

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -144,54 +181,79 @@ def _on_index_requested(self, event: IndexRequestedEvent) -> None:
...
             self.user_manager.create_user(username, roles, hashed_pwd)
-            if roles:
-                self.user_manager.patch_user(
-                    username,
-                    [{"op": "replace", "path": "/opendistro_security_roles", "value": roles}],
-                )
+            self.user_manager.patch_user(
+                username,
+                [{"op": "replace", "path": "/opendistro_security_roles", "value": roles}],
+            )
```
deusebio, line 221: **question** I remember you implemented some rollback logic some time ago, which now seems to have disappeared or was it a different file/feature?

deusebio (commented):
Modifications looks good!

There are a couple of things I wanted to mention though (that may be more general that this specific PR, and therefore could go beyond the scope but should really be I believe):

1. To be consistent with [this](https://github.com/canonical/charm-relation-interfaces/pull/43) also the spec should be slightly changed, to also add the "index" field in the field that are returned by the provider.
2. The "index" should be set on the provider side, possibly [here](https://github.com/canonical/charm-relation-interfaces/pull/43)
3. I believe the OpenSearchProvider/OpenSearchRequirer has not yet been added to the data-platform libs, right? However here we are indeed already merging into main, and the two libs are starting to diverge

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -53,22 +52,57 @@
...
+                ],
+            }
+        ],
+    }
+
+    # Admin user has control over:
+    # - creating multiple indices
+    # - Removing indices they have created
+    ADMIN = {
+        "index_permissions": [
```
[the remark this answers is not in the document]
WRFitch: Nice idea, I'm happy to add this in a future refactor since this PR is holding up some other work

lib/charms/opensearch/v0/opensearch_relation_provider.py
WRFitch: I believe it was in opensearch_users.py, the idea now is to allow create_user functions called with the same username to overwrite that user, so if they're called multiple times we know we have the user we expect

WRFitch: I believe it was in opensearch_users.py, the idea now is to allow create_user functions called with the same username to overwrite that user, so if they're called multiple times we know we have the user we expect

lib/charms/opensearch/v0/opensearch_relation_provider.py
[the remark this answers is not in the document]
Mehdi-Bendriss: Yes, we need to for sure. Okay - so these checks will be in a next PR?

lib/charms/opensearch/v0/opensearch_relation_provider.py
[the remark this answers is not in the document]
Mehdi-Bendriss: Yes - I think my point was about having some predictability and confidence in the code and knowing exactly what might cause what.
We need to have validation at a single point of entry so that we assume everywhere else in the code the inputs are right (later PR perhaps?).

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -81,7 +81,7 @@ def create_role(
             f"{ROLE_ENDPOINT}/{role_name}",
             payload={**(permissions or {}), **(action_groups or {})},
         )
-        if resp.get("status") != "OK":
+        if resp.get("status") not in ["CREATED", "OK"]:
```
Mehdi-Bendriss, line 84: Why 2 potential statuses? one for role create and the other update? if that's the case, I don't think we should handle the update, right?

Mehdi-Bendriss (commented):
LGTM, just an interrogation and should be good to go

lib/charms/opensearch/v0/opensearch_users.py
WRFitch: It was originally intended to handle updates when we call it twice due to an event deferral, but there's no `defer` calls after this function in `_on_index_created`, so it shouldn't be necessary.

lib/charms/opensearch/v0/opensearch_relation_provider.py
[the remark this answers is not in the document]
WRFitch: I'll add them in now, I also wanted to include some input sanitisation anyway

Mehdi-Bendriss (approved):
LGTM, thank you!

deusebio (approved):
Good to go!

delgod (approved):

WRFitch:
NB I've commented out HA tests for now, because I wanted to get this feature in, but I'm working on them now in [this ticket](https://warthogs.atlassian.net/browse/DPE-1569?atlOrigin=eyJpIjoiZmYzZDk0YTI2ZmZjNDgwNDljY2MwM2UzOGRlYWFlZjUiLCJwIjoiaiJ9)

PR #52, opened 2023-03-09
Snap install retry and path updates + lib update

Mehdi-Bendriss:
This PR contains:
- a retry on the installation of the OpenSearch snap.
- change of location of the config from SNAP_COMMON to SNAP_DATA
- update of the data interfaces lib to incorporate new upstream fixes)

## Notes:
The integration tests will pass once [new snap released](https://github.com/canonical/opensearch-snap/pull/21)

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -27,4 +27,4 @@ requires:
 storage:
   opensearch-data:
     type: filesystem
-    location: /var/snap/opensearch/common  # /mnt/opensearch/data
+    location: /var/snap/opensearch  # /mnt/opensearch/data
--- a/src/opensearch.py
+++ b/src/opensearch.py
@@ -38,6 +38,8 @@ class OpenSearchSnap(OpenSearchDistribution):
     """Snap distribution of opensearch, only overrides properties and logic proper to the snap."""
 
     _BASE_SNAP_DIR = "/var/snap/opensearch"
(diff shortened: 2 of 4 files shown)


WRFitch (approved):

metadata.yaml
```diff
@@ -27,4 +27,4 @@ requires:
 storage:
   opensearch-data:
     type: filesystem
-    location: /var/snap/opensearch/common  # /mnt/opensearch/data
+    location: /var/snap/opensearch  # /mnt/opensearch/data
```
deusebio, line 30: this should be updated to comply with the spec but this can be addressed in a separated PR

src/opensearch.py
```diff
@@ -104,7 +109,7 @@ def _build_paths(self) -> Paths:
         """
         return Paths(
             home=f"{self._BASE_SNAP_DIR}/current",
-            conf=f"{self._BASE_SNAP_DIR}/common/config",
+            conf=f"{self._BASE_SNAP_DIR}/current/config",
```
deusebio, line 112: **suggestion** maybe it is worth using two variabel 

```
SNAP_COMMON = f"{self._BASE_SNAP_DIR}/common
SNAP_DATA = f"{self._BASE_SNAP_DIR}/current
```

PR #57, opened 2023-03-28
[DPE-1962] Continuous writing in tests

Mehdi-Bendriss:
### Issue:
This PR implements [DPE-1262](https://warthogs.atlassian.net/browse/DPE-1262) & [DPE-1265](https://warthogs.atlassian.net/browse/DPE-1265), namely this PR implements:

- continuous writes for the integration tests:
    - bulk inserts 
    - single doc inserts
- Added the `opensearch-py` client in tests, for hosts rotation when performing requests 
- Updated `operator_libs_linux` libraries

[DPE-1262]: https://warthogs.atlassian.net/browse/DPE-1262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-1265]: https://warthogs.atlassian.net/browse/DPE-1265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #58, opened 2023-03-28
Horizontal scaling and status related bug fixes

Mehdi-Bendriss:
### Issue:

This PR fixes some issues that arose by running the `continuous writes` routines:
- during scale up, only recompute the roles once the last unit joins 
- make sure the leader unit's roles can be updated
- for the first unit on the cluster, only wait for the `opensearch service` to start, no need to check the host:port
- better health management
- externalized the health management into separate file/class
- updated charm libs

### Notes:
- Do NOT merge before [this PR](https://github.com/canonical/opensearch-operator/pull/57)

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -24,12 +24,11 @@ from charms.opensearch.v0.constants_charm import (
     TLSNotFullyConfigured,
     TLSRelationBrokenError,
     TooManyNodesRemoved,
-    WaitingForSpecificBusyShards,
     WaitingToStart,
 )
 from charms.opensearch.v0.constants_tls import TLS_RELATION, CertType
 from charms.opensearch.v0.helper_charm import Status
-from charms.opensearch.v0.helper_cluster import ClusterState, ClusterTopology, Node
+from charms.opensearch.v0.helper_cluster import ClusterTopology, Node
 from charms.opensearch.v0.helper_databag import (
     RelationDataStore,
     Scope,
@@ -324,7 +323,7 @@ class OpenSearchBaseCharm(CharmBase):
             if not self.alt_hosts:
                 raise OpenSearchHAError(NoNodeUpInCluster)
 
-            health_color = self.health.apply(use_localhost=False)
+            health_color = self.health.apply(wait_for_green_first=True, use_localhost=False)
             if health_color == HealthColors.RED:
                 raise OpenSearchHAError(ClusterHealthRed)
         finally:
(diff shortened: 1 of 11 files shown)


lib/charms/opensearch/v0/opensearch_health.py
```diff
@@ -0,0 +1,149 @@
...
+
+        if status == HealthColors.GREEN:
+            # green: cluster healthy
+            self._charm.status.clear(ClusterHealthRed, app=True)
+            self._charm.status.clear(ClusterHealthYellow, app=True)
+            self._charm.status.clear(WaitingForBusyShards, app=True)
+        elif status == HealthColors.RED:
+            # health RED, some primary shards are unassigned
+            self._charm.app.status = BlockedStatus(ClusterHealthRed)
+            return
```
delgod, line 83: Why we return None if the status is RED and the actual status in all other cases?

lib/charms/opensearch/v0/opensearch_health.py
```diff
@@ -0,0 +1,149 @@
...
+        else:
+            # cluster health is yellow, either a temporary state where shards are moving
+            # around (relocating / initializing) - or a permanent state where some replica
+            # shards are unassigned.
+            self._charm.app.status = (
+                BlockedStatus(ClusterHealthYellow)
+                if status == HealthColors.YELLOW
+                else WaitingStatus(WaitingForBusyShards)
```
delgod, line 91: feel free to ignore: If we have a cascade of ifs, for better readability I prefer to have a cascade for all cases (not inline ifs for some cases).

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -311,8 +324,8 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):
             if not self.alt_hosts:
                 raise OpenSearchHAError(NoNodeUpInCluster)
 
-            health = self._apply_cluster_health(wait_for_green_first=True, use_localhost=False)
-            if health == "red":
+            health_color = self.health.apply(use_localhost=False)
```
delgod, line 327: maybe
```suggestion
            health_color = self.health.apply(wait_for_green_first=True, use_localhost=False)
```

delgod (changes requested):

lib/charms/opensearch/v0/opensearch_health.py
Mehdi-Bendriss: Good catch - thank you, `return` should go

lib/charms/opensearch/v0/opensearch_health.py
```diff
@@ -0,0 +1,149 @@
...
+        app: bool = True,
+    ) -> str:
+        """Fetch cluster health and set it on the app status."""
+        host = self._charm.unit_ip if use_localhost else None
+        status = self._fetch_status(host, wait_for_green_first)
+
+        if app:
+            self.apply_for_app(status)
+        else:
+            self.apply_for_unit(status)
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Good catch, the `app=False` should have been in the `can_service_start`

WRFitch (approved):

delgod (approved):

PR #59, opened 2023-03-29
[DPE-1266] HA test: kill cluster_manager and juju leader nodes 

Mehdi-Bendriss:
### Issue:
This PR implements [DPE-1266](https://warthogs.atlassian.net/browse/DPE-1266), namely this PR implements:
- targeted removal of:
   - elected cluster manager node 
   - elected juju leader node
- externalization of the `app name` for running integration tests against existing running deployments.

[DPE-1266]: https://warthogs.atlassian.net/browse/DPE-1266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -101,7 +101,7 @@ async def create_dummy_indexes(ops_test: OpsTest, unit_ip: str, count: int = 5) 
     stop=stop_after_attempt(15),
 )
 async def delete_dummy_indexes(ops_test: OpsTest, unit_ip: str, count: int = 5) -> None:
-    """Create indexes."""
+    """Delete dummy indexes."""
     for index_id in range(count):
         await http_request(
             ops_test,
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -2,7 +2,7 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 from random import randint
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from charms.opensearch.v0.models import Node
 from pytest_operator.plugin import OpsTest
@@ -10,6 +10,38 @@ from tenacity import retry, stop_after_attempt, wait_fixed, wait_random
 
 from tests.integration.ha.continuous_writes import ContinuousWrites
 from tests.integration.helpers import http_request
+
+
+async def app_name(ops_test: OpsTest) -> Optional[str]:
+    """Returns the name of the cluster running OpenSearch.
+
+    This is important since not all deployments of the OpenSearch charm have the
+    application name "opensearch".
+    Note: if multiple clusters are running OpenSearch this will return the one first found.
+    """
+    status = await ops_test.model.get_status()
+    for app in ops_test.model.applications:
+        if "opensearch" in status["applications"][app]["charm"]:
+            return app
+
+    return None
+
+
+async def get_elected_cm_unit_id(ops_test: OpsTest, unit_ip: str) -> int:
+    """Returns the unit id of the current elected cm node."""
+    # get current elected cm node
+    resp = await http_request(
+        ops_test,
+        "GET",
+        f"https://{unit_ip}:9200/_cluster/state/cluster_manager_node",
+    )
+    cm_node_id = resp.get("cluster_manager_node")
+    if not cm_node_id:
+        return -1
+
+    # get all nodes
+    resp = await http_request(ops_test, "GET", f"https://{unit_ip}:9200/_nodes")
+    return int(resp["nodes"][cm_node_id]["name"].split("-")[1])
 
 
 @retry(
@@ -61,6 +93,20 @@ async def create_dummy_indexes(ops_test: OpsTest, unit_ip: str, count: int = 5) 
                     "index": {"number_of_shards": p_shards, "number_of_replicas": r_shards}
                 }
             },
+        )
+
+
+@retry(
+    wait=wait_fixed(wait=5) + wait_random(0, 5),
+    stop=stop_after_attempt(15),
+)
+async def delete_dummy_indexes(ops_test: OpsTest, unit_ip: str, count: int = 5) -> None:
+    """Create indexes."""
+    for index_id in range(count):
+        await http_request(
+            ops_test,
+            "DELETE",
+            f"https://{unit_ip}:9200/index_{index_id}",
         )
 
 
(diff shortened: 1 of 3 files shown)


delgod (approved):

delgod (approved):

tests/integration/ha/helpers.py
```diff
@@ -64,6 +96,20 @@ async def create_dummy_indexes(ops_test: OpsTest, unit_ip: str, count: int = 5)
         )
 
 
+@retry(
+    wait=wait_fixed(wait=5) + wait_random(0, 5),
+    stop=stop_after_attempt(15),
+)
+async def delete_dummy_indexes(ops_test: OpsTest, unit_ip: str, count: int = 5) -> None:
+    """Create indexes."""
```
MiaAltieri, line 104: nit: update doc string

tests/integration/ha/test_horizontal_scaling.py
```diff
@@ -54,6 +57,11 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
 @pytest.mark.skip_if_deployed
 async def test_build_and_deploy(ops_test: OpsTest) -> None:
     """Build and deploy one unit of OpenSearch."""
+    # it is possible for users to provide their own cluster for HA testing.
+    # Hence, check if there is a pre-existing cluster.
+    if await app_name(ops_test):
+        return
+
     my_charm = await ops_test.build_charm(".")
     await ops_test.model.set_config(MODEL_CONFIG)
```
MiaAltieri, line 66: question: should this model config be applied to a user provided cluster?

tests/integration/ha/test_horizontal_scaling.py
```diff
@@ -257,3 +274,60 @@ async def test_safe_scale_down_roles_reassigning(
...
+    await ops_test.model.wait_for_idle(
+        apps=[app], status="active", timeout=1000, wait_for_exact_units=5, idle_period=60
+    )
+
+    # scale down: remove the juju leader
+    leader_unit_id = await get_leader_unit_id(ops_test)
+
+    await ops_test.model.applications[app].destroy_unit(f"{app}/{leader_unit_id}")
+    await ops_test.model.wait_for_idle(
+        apps=[app], status="active", timeout=1000, wait_for_exact_units=4, idle_period=120
```
MiaAltieri, line 302: question: what is the reasoning behind these number specified - specifically for idle period

tests/integration/ha/test_horizontal_scaling.py
```diff
@@ -257,3 +274,60 @@ async def test_safe_scale_down_roles_reassigning(
...
+    )
+
+    # make sure the duties supposed to be done by the departing leader are done
+    # we expect to have 3 cm-eligible and 1 data-only nodes
+    leader_unit_ip = await get_leader_unit_ip(ops_test)
+    nodes = await all_nodes(ops_test, leader_unit_ip)
+    assert ClusterTopology.nodes_count_by_role(nodes)["cluster_manager"] == 3
+    assert ClusterTopology.nodes_count_by_role(nodes)["data"] == 4
```
MiaAltieri, line 310: does this check that there is 1 data-only node?

tests/integration/ha/test_horizontal_scaling.py
```diff
@@ -257,3 +274,60 @@ async def test_safe_scale_down_roles_reassigning(
...
+
+    # scale-down: remove the current elected CM
+    first_elected_cm_unit_id = await get_elected_cm_unit_id(ops_test, leader_unit_ip)
+    assert first_elected_cm_unit_id != -1
+
+    # scale-down: remove the elected cm
+    await ops_test.model.applications[app].destroy_unit(f"{app}/{first_elected_cm_unit_id}")
+    await ops_test.model.wait_for_idle(
+        apps=[app], status="active", timeout=1000, wait_for_exact_units=3, idle_period=90
+    )
```
MiaAltieri, line 320: nit

```suggestion
    # scale-down: remove the current elected CM
    first_elected_cm_unit_id = await get_elected_cm_unit_id(ops_test, leader_unit_ip)
    assert first_elected_cm_unit_id != -1
    await ops_test.model.applications[app].destroy_unit(f"{app}/{first_elected_cm_unit_id}")
    await ops_test.model.wait_for_idle(
        apps=[app], status="active", timeout=1000, wait_for_exact_units=3, idle_period=90
    )
```

MiaAltieri: also is there a meaning behind `idle_period=90`

MiaAltieri (changes requested):
looks AWESOME - only one change requested - rest or only questions :)

tests/integration/ha/test_horizontal_scaling.py
Mehdi-Bendriss: I think we should assume that a user provided cluster has been pre-configured / tuned by the user, so we shouldn't intrude in the model configuration

tests/integration/ha/test_horizontal_scaling.py
Mehdi-Bendriss: On a scale-up/down event, there might occur a recomputing of the nodes' roles - then the restart of the impacted units. 
When this occurs, depending on the responsiveness of the charm hooks - it could be triggered up to a certain amount of time after being active. 
so to be safe `idle_period` is set, otherwise we could lose a unit (restarting) a few seconds after `wait_for_idle` returns true, making the rest of the test fail.

tests/integration/ha/test_horizontal_scaling.py
Mehdi-Bendriss: That's correct

tests/integration/ha/test_horizontal_scaling.py
```diff
@@ -257,3 +274,60 @@ async def test_safe_scale_down_roles_reassigning(
...
+
+    # check if CM re-election happened
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Good point, so the max it can take is `10sec`, and effectively starting on `destroy_unit`, but the waiting itself is taken care of by the `wait_for_idle` (time for the unit to spin up, setup then idle 90sec), makes sense?

tests/integration/ha/test_horizontal_scaling.py
MiaAltieri: hmm I wonder if there is a constant that could be used for `90` to emphasize this - but if this is a bad suggestion plz ignore

tests/integration/ha/test_horizontal_scaling.py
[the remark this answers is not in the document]
MiaAltieri: true - then I think it would be good to add a comment there and perhaps change `90` to a constant so this is better emphasized

MiaAltieri (approved):

PR #62, opened 2023-03-31
[DPE-1458] HA test: replication read from secondaries (consistency) 

Mehdi-Bendriss:
### Issue:
This PR implement [DPE-1458](https://warthogs.atlassian.net/browse/DPE-1458). Namely, this PR implements:
- integration test for ensuring the data can be indexed from any node and read from any other node
   - index data using the elected cluster manager node's endpoint, read from all other nodes
   - index data on an index with 1 replica shard, and query the replica shard 

### Note:
Do not merge this PR before (https://github.com/canonical/opensearch-operator/pull/63)
Then change base to main

[DPE-1458]: https://warthogs.atlassian.net/browse/DPE-1458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #65, opened 2023-04-04
[DPE-1553] HA test: multi clusters isolation

Mehdi-Bendriss:
### Issue:
This PR implements [DPE-1553](https://warthogs.atlassian.net/browse/DPE-1553) - namely, this PR implements the following: 
- integration test that creates two clusters, check that writes in one cluster NOT replicated to another cluster
- dependencies upgrade

**Notes:** Do NOT merge before [#62](https://github.com/canonical/opensearch-operator/pull/62), then change the base to `main` once done.

[DPE-1553]: https://warthogs.atlassian.net/browse/DPE-1553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/integration/ha/test_ha.py
```diff
@@ -105,7 +111,60 @@ async def test_replication_across_members(
...
+    second_app_docs = await search(ops_test, SECOND_APP_NAME, second_app_leader_ip, index_name)
+
+    # check that the only doc indexed in each cluster is different
+    assert len(current_app_docs) == 1
+    assert len(second_app_docs) == 1
+    assert current_app_docs[0] != second_app_docs[0]
+
+    # cleanup
+    await delete_index(ops_test, app, main_app_leader_unit_ip, index_name)
+    await ops_test.model.remove_application(SECOND_APP_NAME)
```
MiaAltieri, line 167: wait cluster to be fully removed for consistency across runs

MiaAltieri (approved):
Looks great! only one change needed

tests/integration/ha/test_ha.py
Mehdi-Bendriss: That's unfortunately not so straightforward, because we have safeguards on `_storage_detaching` that prevents the last unit from being removed to avoid data loss (by putting the error on an error state). So I think what we need to do here is to put this test at the very end of the test file so that it gets removed with the model and doesn't take unnecessary compute on the CI

delgod (approved):

PR #66, opened 2023-04-04
Test network cut

WRFitch:
## Proposal
Adds a test and relevant helper methods to test how an opensearch cluster responds to network outages. Also fixed an intermittent outage in the client-integration tests.

## Context
- A part of the fix for this network outage is to repeatedly check for network updates in the update-status hook. A simpler way to handle network changes is to use DNS - I've created a ticket for that [here](https://warthogs.atlassian.net/browse/DPE-1651?atlOrigin=eyJpIjoiOTUzMzkxZWZmOTA5NDVjYWI3N2E0YTA0NjY2YzkxZDYiLCJwIjoiaiJ9)


deusebio (approved):
The code looks fine! But it is probably best for some more experience person with tests of network cut to look into this (as I may not be the best person)

tests/integration/helpers.py
```diff
@@ -232,6 +232,20 @@ def get_reachable_units(ops_test: OpsTest, app: str = APP_NAME) -> Dict[int, str
     return result
 
 
+async def unit_hostname(ops_test: OpsTest, unit_name: str) -> str:
```
Mehdi-Bendriss, line 235: Why is this needed?

tests/integration/ha/helpers.py
```diff
@@ -174,6 +186,103 @@ async def assert_continuous_writes_consistency(c_writes: ContinuousWrites) -> No
...
+    Args:
+        model_name: The name of the model
+        hostname: The name of the instance
+        old_ip: old registered IP address
+    """
+    if instance_ip(model_name, hostname) == old_ip:
+        raise Exception("Network not restored, IP address has not changed yet.")
+
+
+def instance_ip(model: str, instance: str) -> str:
```
Mehdi-Bendriss, line 238: Could you use `get_application_unit_ids_ips(ops_test)[unit_id]`? it's broader than `unit_ip` as it gives us a full view and we get to stay consistent with the rest of the tests

last_written_value
Mehdi-Bendriss, line 1: Could you delete this file?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -748,15 +765,24 @@ def _set_node_conf(self, nodes: List[Node]) -> None:
...
-            computed_roles,
-            cm_names,
-            cm_ips,
-            contribute_to_bootstrap,
-        )
+        # Update config only when we have meaningful changes.
+        current_config = self.opensearch_config.load_node()
+        if not (
+            current_config.get("cluster.name") == f"{self.app.name}-{self.model.name}"
+            and current_config.get("node.name") == self.unit_name
```
Mehdi-Bendriss, line 772: These 2 should not change.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -748,15 +765,24 @@ def _set_node_conf(self, nodes: List[Node]) -> None:
...
+            and current_config.get("node.roles") == computed_roles
+            and current_config.get("discovery.seed_hosts") == cm_ips
+            and (current_config.get("cluster.initial_cluster_manager_nodes") == cm_names)
```
Mehdi-Bendriss, line 775: this should not change

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -748,15 +765,24 @@ def _set_node_conf(self, nodes: List[Node]) -> None:
...
-            cm_names,
-            cm_ips,
-            contribute_to_bootstrap,
-        )
+        # Update config only when we have meaningful changes.
+        current_config = self.opensearch_config.load_node()
+        if not (
+            current_config.get("cluster.name") == f"{self.app.name}-{self.model.name}"
+            and current_config.get("node.name") == self.unit_name
+            and current_config.get("node.roles") == computed_roles
```
Mehdi-Bendriss, line 773: This should not change from the unit itself, we have the logic that orchestrates it from the leader, it may otherwise lead to a roles not respecting our specs

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -748,15 +765,24 @@ def _set_node_conf(self, nodes: List[Node]) -> None:
...
+            and current_config.get("discovery.seed_hosts") == cm_ips
```
Mehdi-Bendriss, line 774: we also need `network_host`

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -358,23 +363,35 @@ def _on_update_status(self, event: UpdateStatusEvent):
...
             return
 
-        # if there are exclusions to be removed
+        for relation in self.model.relations.get(ClientRelationName, []):
+            self.opensearch_provider.update_endpoints(relation)
+
+        self.user_manager.remove_users_and_roles()
+
         if self.unit.is_leader():
+            self._compute_and_broadcast_updated_topology(self._get_nodes(False))
```
Mehdi-Bendriss, line 386: Why is this needed?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -748,15 +765,24 @@ def _set_node_conf(self, nodes: List[Node]) -> None:
...
-            self.app.name,
-            self.model.name,
-            self.unit_name,
-            computed_roles,
-            cm_names,
-            cm_ips,
-            contribute_to_bootstrap,
-        )
+        # Update config only when we have meaningful changes.
+        current_config = self.opensearch_config.load_node()
```
Mehdi-Bendriss, line 769: This logic should moved to `_reconfigure_and_restart_unit_if_needed` (refer to previous comment)

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -358,23 +363,35 @@ def _on_update_status(self, event: UpdateStatusEvent):
...
 
+        try:
+            # Retrieve the nodes of the cluster, needed to configure this node
+            nodes = self._get_nodes(False)
+            # Set the configuration of the node
+            self._set_node_conf(nodes)
+        except OpenSearchHttpError:
+            pass
+
+        self._reconfigure_and_restart_unit_if_needed()
```
Mehdi-Bendriss, line 374: This will unfortunately not have any effect on the non leader units, as: 
- `nodes_config` is only set from the leader
- the current logic is roles bound

Please extend the `_reconfigure_and_restart_unit_if_needed` method with a new argument: `conf_from_leader: bool = True`

i.e:
```
def _reconfigure_and_restart_unit_if_needed(self, conf_from_leader: bool = True):
    """Reconfigure the current unit if a new config was computed for it, then restart."""
    if conf_from_leader:
        nodes_config = self.peers_data.get_object(Scope.APP, "nodes_config")
        if not nodes_config:
            return

         new_node_conf = nodes_config.get(self.unit_name)
    else:
         new_node_conf = # recompute the conf

    .....
```

Mehdi-Bendriss: You can restrain this new logic to only checking `network_host`

Mehdi-Bendriss: Actually much easier for now, since we'll implement a config reconciliation later: create a new method in `opensearch_config.py`: `update_host_if_needed(host: str)` that loads the config, checks if passed host is the list of the stored `network.host`, if yes return False otherwise replaces the previous one with the passed one and return True.
If return True, emit a restart unit
(don't forget to add back `["_site_"]` to the liust of hosts)

tests/integration/helpers.py
WRFitch: We're using this hostname to cut network access via LXC in `helpers.cut_network_from_unit`

PR #12, opened 2022-10-26
Increased retry wait in integration tests

Mehdi-Bendriss:
This PR contains: 
- Increase in retry intervals for integration tests that consume HTTP endpoints, to throttle requests
- addition of randomness in the retry waits

PR #70, opened 2023-04-06
[DPE-1555] HA tests: kill db process

Mehdi-Bendriss:
### Proposal
This PR implements [DPE-1555](https://warthogs.atlassian.net/browse/DPE-1555), namely this PR implements:
- an integration test that:
   - find the unit hosting the primary shard of an index
   - kill the opensearch service in that unit 
   - verify writes have continued in the cluster 
   - verify the opensearch service in the said unit has been successfully restart 
   - verify the shards primary reelection happened 
   - verify the unit is back online and as part of the cluster


tests/integration/ha/helpers.py
```diff
@@ -172,3 +173,37 @@ async def assert_continuous_writes_consistency(c_writes: ContinuousWrites) -> No
...
+    """Run kill with signal in specific unit."""
+    # Killing the only instance can be disastrous.
+    if len(ops_test.model.applications[app].units) < 2:
+        await ops_test.model.applications[app].add_unit(count=1)
+        await ops_test.model.wait_for_idle(
+            apps=[app],
+            status="active",
+            timeout=1000,
+            idle_period=IDLE_PERIOD,
+        )
```
delgod, line 200: I would prefer to have this logic in the test itself, not hidden in kill function

tests/integration/ha/test_ha.py
```diff
@@ -117,6 +135,68 @@ async def test_replication_across_members(
...
+        )
+
+    # Kill the opensearch process
+    await send_kill_signal_to_process(
+        ops_test, app, first_unit_with_primary_shard, signal="SIGKILL"
+    )
+
+    # verify new writes are continuing by counting the number of writes before and after 5 seconds
+    writes = await c_writes.count()
+    time.sleep(30)  # the opensearch service is scheduled to restart after 20 sec
```
MiaAltieri, line 169: So the idea here is that writes continue even after the primary was killed. (i.e. can the cluster support writes after the primary was killed. Hence, the amount to sleep here should be "the amount of time for opensearch to elect a new primary" **not** "the amount of time for the service to come back up.

The reasoning here is that if for some reason the old primary never came back up that the cluster would still support writes. I hope that makes sense and that I didn't butcher the explanation 😅

tests/integration/helpers.py
```diff
@@ -351,6 +359,21 @@ async def check_cluster_formation_successful(
     return set(unit_names) == set(registered_nodes)
 
 
+async def is_up(ops_test: OpsTest, unit_ip: str) -> bool:
+    """Return if node up."""
+    try:
+        for attempt in Retrying(
+            stop=stop_after_attempt(15), wait=wait_fixed(wait=5) + wait_random(0, 5)
```
MiaAltieri, line 366: Unless `wait_random(0, 5)` is necessary this is easier to understand and gives reliably timed retries

```suggestion
            stop=stop_after_attempt(15), wait=wait_fixed(wait=5) 
```

MiaAltieri (changes requested):
minor change needed

tests/integration/ha/test_ha.py
```diff
@@ -117,6 +135,68 @@ async def test_replication_across_members(
...
+    units_with_r_shards = [shard.unit_id for shard in shards if not shard.is_prim]
+    assert first_unit_with_primary_shard in units_with_r_shards
+
+    # verify the node with the old primary successfully joined the rest of the fleet
+    assert await check_cluster_formation_successful(
+        ops_test, leader_unit_ip, get_application_unit_names(ops_test, app=app)
+    )
+
+    # continuous writes checks
+    await assert_continuous_writes_consistency(c_writes)
```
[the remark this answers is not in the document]
Mehdi-Bendriss: This essentially checks that: `max_stored_id == count == last_expected_id` (where the latter id is stored in a file). The part where we check about writes not continuing is in the test itself

tests/integration/helpers.py
Mehdi-Bendriss: Though not necessary in this specific test, I would like to stay consistent with the other helpers where a bit of randomness is needed

MiaAltieri: why is the randomness needed elsewhere?

Mehdi-Bendriss: I don't remember exactly the [context](https://github.com/canonical/opensearch-operator/pull/12) that compelled me to do so. Regardless, those are tests' helpers - having randomness is not a negative. Would be worth revisiting in the future to trace back the original reasons and see if they're still relevant.

MiaAltieri (approved):

delgod (approved):

PR #73, opened 2023-04-08
Remove fast-forwards from scaling test

WRFitch:
## Proposal
This test is flaky. I've removed the status fast-forwards in an attempt to stabilise it.


deusebio (approved):
LGTM!

Yes, one has to be careful to use wait_for_idle and fast_forward together to not end up in infinite loop. In Kafka we are still using `fast_forward` but we make sure that the wait_for_idle is shorter than the period of firing of updates status (e.g. [here](https://github.com/canonical/kafka-k8s-operator/blob/main/tests/integration/test_password_rotation.py#L50). This way it speeds up the tests, but tests don't get stuck.

But up to you whether you want to take this comment into account or just skip the fast_forward

Mehdi-Bendriss (approved):

PR #74, opened 2023-04-08
[DPE-1630] HA tests: freeze db process on unit with primary shard

Mehdi-Bendriss:
### Issue:
This PR implements [DPE-1630](https://warthogs.atlassian.net/browse/DPE-1630) - namely this PR implements an integration test that:
- stops the OpenSearch service in a unit hosting the primary shard of the `series_index`
- verifies that the primary shard's reelection successfully happened 
- resume the service in the said unit 
- verifies that the said unit successfully joined back the cluster 
- verifies that the said unit now hosts a replica shard  

[DPE-1630]: https://warthogs.atlassian.net/browse/DPE-1630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/integration/ha/test_ha.py
```diff
@@ -197,6 +197,77 @@ async def test_kill_db_process_node_with_primary_shard(
...
+
+    units_ips = get_application_unit_ids_ips(ops_test, app)
+    leader_unit_ip = await get_leader_unit_ip(ops_test, app=app)
+
+    # find unit hosting the primary shard of the index "series-index"
+    shards = await get_shards_by_index(ops_test, leader_unit_ip, ContinuousWrites.INDEX_NAME)
+    first_unit_with_primary_shard = [shard.unit_id for shard in shards if shard.is_prim][0]
+
+    # Killing the only instance can be disastrous.
+    if len(ops_test.model.applications[app].units) < 2:
```
MiaAltieri, line 215: Seems like you do this throughout the code I think it would be cleaner to add this as a function - or even better as a fixture

tests/integration/ha/test_ha.py
```diff
@@ -197,6 +197,77 @@ async def test_kill_db_process_node_with_primary_shard(
...
+            status="active",
+            timeout=1000,
+            idle_period=IDLE_PERIOD,
+        )
+
+    # Freeze the opensearch process
+    opensearch_pid = await send_kill_signal_to_process(
+        ops_test, app, first_unit_with_primary_shard, signal="SIGSTOP"
+    )
+
```
MiaAltieri, line 228: Might need to sleep for the re-election period here

tests/integration/ha/test_ha.py
```diff
@@ -197,6 +197,77 @@ async def test_kill_db_process_node_with_primary_shard(
...
+        app,
+        first_unit_with_primary_shard,
+        signal="SIGCONT",
+        opensearch_pid=opensearch_pid,
+    )
+
+    # verify that the opensearch service is back running on the unit previously hosting the p_shard
+    assert await is_up(
+        ops_test, units_ips[first_unit_with_primary_shard]
+    ), "OpenSearch service hasn't restarted."
```
MiaAltieri, line 247: Nice assert message 😎

tests/integration/ha/test_ha.py
```diff
@@ -197,6 +197,77 @@ async def test_kill_db_process_node_with_primary_shard(
...
+    opensearch_pid = await send_kill_signal_to_process(
+        ops_test, app, first_unit_with_primary_shard, signal="SIGSTOP"
+    )
+
+    # verify new writes are continuing by counting the number of writes before and after 5 seconds
+    writes = await c_writes.count()
+    time.sleep(5)
+    more_writes = await c_writes.count()
+    assert more_writes > writes, "writes not continuing to DB"
+
```
MiaAltieri, line 234: Check if the step down process was performed and check if a new primary was elected

MiaAltieri (changes requested):
Looks good minor requests.

Let me check about sending the SIGTERM to the unit via juju, a look at your test shows that juju is a bit unhappy and I'm not sure why

tests/integration/ha/test_ha.py
Mehdi-Bendriss: Thanks! I got some feedback in a previous PR to make them obvious in the test as opposed to hiding them in a function, and I agree with it as it should ideally only there be 1 place where unit scaling happens (easier to read / debug)

tests/integration/ha/test_ha.py
Mehdi-Bendriss: there is a sleep a couple of lines below this line

delgod (approved):

tests/integration/ha/test_ha.py
MiaAltieri: Yes - but I think it is good to check that the primary comes up before you send the SIGCONT signal to the old primary

tests/integration/ha/test_ha.py
MiaAltieri: Makes sense! Thanks for sharing this :)

tests/integration/ha/test_ha.py
MiaAltieri: that makes sense - perhaps adding a comment above that line that says:
```
# assert that writes are continuing, sleeping for 5 seconds should be pleanty of time for the reelection to occur since it happens in an order of miliseconds
```
would be useful

MiaAltieri (changes requested):
looks good! just some updates on my comments here

tests/integration/ha/continuous_writes.py
```diff
@@ -102,7 +103,9 @@ async def _create_fully_replicated_index(self):
             # create index with a replica shard on every node
             client.indices.create(
                 index=ContinuousWrites.INDEX_NAME,
-                body={"settings": {"index": {"auto_expand_replicas": "1-all"}}},
+                body={
+                    "settings": {"index": {"number_of_shards": 2, "auto_expand_replicas": "1-all"}}
```
delgod, line 107: let's create only one shard per index. It will make the test more reliable.
```suggestion
                    "settings": {"index": {"number_of_shards": 1, "auto_expand_replicas": "1-all"}}
```

delgod (approved):

tests/integration/ha/continuous_writes.py
delgod: changed mind, please ignore

MiaAltieri (approved):

PR #75, opened 2023-04-12
Test graceful restart

WRFitch:

### DIFF
--- a/tests/integration/ha/test_ha.py
+++ b/tests/integration/ha/test_ha.py
@@ -8,35 +8,29 @@ import time
 
 import pytest
 from pytest_operator.plugin import OpsTest
-from tenacity import RetryError, Retrying, stop_after_delay, wait_fixed
 
 from tests.integration.ha.continuous_writes import ContinuousWrites
-from tests.integration.ha.helpers import (
+from tests.integration.ha.helpers import (  # assert_continuous_writes_consistency,; get_shards_by_index,
     app_name,
-    assert_continuous_writes_consistency,
     get_elected_cm_unit,
-    get_shards_by_index,
+    get_elected_cm_unit_id,
     secondary_up_to_date,
     send_kill_signal_to_process,
 )
-from tests.integration.ha.helpers_data import (
-    create_index,
-    default_doc,
-    delete_index,
-    index_doc,
-    search,
-)
-from tests.integration.ha.test_horizontal_scaling import IDLE_PERIOD
-from tests.integration.helpers import (
+
+# from tests.integration.ha.helpers_data import (
+#     create_index,
+#     default_doc,
+#     delete_index,
+#     index_doc,
+#     search,
+# )
+# from tests.integration.ha.test_horizontal_scaling import IDLE_PERIOD
+from tests.integration.helpers import (  # check_cluster_formation_successful,; get_application_unit_ids,; get_application_unit_ids_ips,; get_application_unit_names,;; is_up,
     APP_NAME,
     MODEL_CONFIG,
     SERIES,
-    check_cluster_formation_successful,
-    get_application_unit_ids,
-    get_application_unit_ids_ips,
-    get_application_unit_names,
     get_leader_unit_ip,
-    is_up,
 )
 from tests.integration.tls.test_tls import TLS_CERTIFICATES_APP_NAME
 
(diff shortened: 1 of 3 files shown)


tests/integration/ha/test_ha.py
```diff
@@ -417,6 +417,111 @@ async def test_freeze_db_process_node_with_elected_cm(
...
+    assert await is_up(
+        ops_test, units_ips[first_elected_cm_unit_id]
+    ), "OpenSearch service hasn't restarted."
+
+    # verify the previously elected CM node successfully joined back the rest of the fleet
+    assert await check_cluster_formation_successful(
+        ops_test, leader_unit_ip, get_application_unit_names(ops_test, app=app)
+    )
+
+
```
MiaAltieri, line 462:
```
await assert_continuous_writes_consistency(ops_test, c_writes, app)
```
verify that no writes were missed to the old primary after it restarted - see [here](https://github.com/canonical/mongodb-operator/blob/292212e1c433e1e076fcdb8dc68c2da88d06a3d2/tests/integration/ha_tests/test_ha.py#L474-L483)

tests/integration/ha/test_ha.py
```diff
@@ -417,6 +417,111 @@ async def test_freeze_db_process_node_with_elected_cm(
...
+    writes = await c_writes.count()
+    time.sleep(5)
+    more_writes = await c_writes.count()
+    assert more_writes > writes, "writes not continuing to DB"
+
+    # verify that the opensearch service is back running on the unit previously elected CM unit
+    assert await is_up(
+        ops_test, units_ips[first_elected_cm_unit_id]
+    ), "OpenSearch service hasn't restarted."
+
```
MiaAltieri, line 456: verify that when the SIGTERM was sent that the old primary notified the cluster of the stepdown, see [here](https://github.com/canonical/mongodb-operator/blob/292212e1c433e1e076fcdb8dc68c2da88d06a3d2/tests/integration/ha_tests/test_ha.py#L463-L473)

I recall Mehdi and i discussing this. @Mehdi-Bendriss perhaps you can add some insight here?

tests/integration/ha/test_ha.py
```diff
@@ -417,6 +417,111 @@ async def test_freeze_db_process_node_with_elected_cm(
...
+    writes = await c_writes.count()
+    time.sleep(5)
+    more_writes = await c_writes.count()
+    assert more_writes > writes, "writes not continuing to DB"
+
+    # verify that the opensearch service is back running on the previous primary shard unit
+    assert await is_up(
+        ops_test, units_ips[first_unit_with_primary_shard]
+    ), "OpenSearch service hasn't restarted."
+
```
MiaAltieri, line 503: verify that when the SIGTERM was sent that the old primary notified the cluster of the stepdown, see [here](https://github.com/canonical/mongodb-operator/blob/292212e1c433e1e076fcdb8dc68c2da88d06a3d2/tests/integration/ha_tests/test_ha.py#L463-L473)

I recall Mehdi and i discussing this. @Mehdi-Bendriss perhaps you can add some insight here?

MiaAltieri (changes requested):
Missing some checks - also might be nice to add some messages to your asserts incase they fail (but I don't have strong feelings here - up to you)

otherwise this looks super clean and awesome, nice work :)

tests/integration/ha/test_ha.py
WRFitch: whoops, missed that. Thanks!

tests/integration/ha/test_ha.py
Mehdi-Bendriss: @MiaAltieri - thank you! @delgod and I discussed this, after a short research it doesn't seem like there is a way to trigger or track the stepdown of a primary shard in a unit. We're tracking it in [another ticket](https://warthogs.atlassian.net/browse/DPE-1701) while we do further research

tests/integration/ha/test_ha.py
```diff
@@ -417,6 +417,113 @@ async def test_freeze_db_process_node_with_elected_cm(
...
+    writes = await c_writes.count()
+    time.sleep(5)
+    more_writes = await c_writes.count()
+    assert more_writes > writes, "writes not continuing to DB"
+
+    # verify that the opensearch service is back running on the unit previously elected CM unit
+    assert await is_up(
+        ops_test, units_ips[first_elected_cm_unit_id]
+    ), "OpenSearch service hasn't restarted."
+
```
Mehdi-Bendriss, line 456: Please add a check that a new cluster manager has been elected, see [here](https://github.com/canonical/opensearch-operator/blob/main/tests/integration/ha/test_ha.py#L249-L253)

delgod (approved):

tests/integration/ha/test_ha.py
```diff
@@ -261,6 +264,7 @@ async def test_kill_db_process_node_with_elected_cm(
     await assert_continuous_writes_consistency(ops_test, c_writes, app)
 
 
+@pytest.mark.skip
```
delgod, line 267: hm, maybe tests skipped by mistake

delgod (changes requested):

tests/integration/ha/test_ha.py
WRFitch: I added these to make sure I'm running the tests I've added, to guarantee they're running ok. These markers will be removed before merge.

Mehdi-Bendriss (approved):

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -226,7 +226,10 @@ def call(
         ) -> requests.Response:
             """Performs an HTTP request."""
             if remaining_retries < 0:
-                if return_failed_resp and error_response:
+                if not error_response:
+                    raise OpenSearchHttpError()
+
+                if return_failed_resp:
                     return error_response
```
Mehdi-Bendriss, line 233: what's the difference between this and how it was previously? could we perhaps revert?

Mehdi-Bendriss (approved):

lib/charms/opensearch/v0/opensearch_distro.py
WRFitch: the `OpenSearchHttpError`raised in line 235 relies on `error_response.text` - if `error_response` doesn't exist, raising that error fails and everything explodes.

MiaAltieri (approved):

delgod (approved):

PR #76, opened 2023-04-12
Update the seed_hosts regularly 

Mehdi-Bendriss:
## Issue
The kill db process HA test raised a bug that occurs for the first unit, which only contains 1 seed_host (itself - since no other node was up in the cluster). If the process on this unit restarts, this node will not be able to join the rest of the cluster as the discovery cannot happen.

## Solution
- we now update all the nodes with newly joined `cluster_manager` eligible nodes addresses
- to avoid having to restart each node upon newly discovered cluster manager eligible:
    - we no longer  rely on `discovery.seed_hosts` as it cannot be changed later without a restart
    - we use `discovery.seed_providers: file`
    - we use `unicast_hosts.txt` for storing: ips + aliases + hostnames ==> hot reloaded

## Notes:
The Snap was updated with this change

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_config.py
+++ b/lib/charms/opensearch/v0/opensearch_config.py
@@ -3,6 +3,7 @@
 
 """Class for Setting configuration in opensearch config files."""
 import logging
+import socket
 from typing import Dict, List
 
 from charms.opensearch.v0.constants_tls import CertType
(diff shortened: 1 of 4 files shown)


WRFitch (approved):

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -171,6 +173,19 @@ def set_node(
...
 
+    def add_seed_hosts(self, cm_ips: List[str]):
+        """Add CM nodes ips / host names to the seed host list of this unit."""
+        cm_ips_hostnames = cm_ips.copy()
+        for ip in cm_ips:
+            name, aliases, addresses = socket.gethostbyaddr(ip)
+            cm_ips_hostnames.extend([name] + aliases + addresses)
+
+        with open(self._opensearch.paths.seed_hosts, "w+") as f:
+            existing = [line.strip() for line in f.readlines()]
```
delgod, line 184: do we need an existing list? I mean existing list can contain nodes that changed IP or nodes that was removed

Mehdi-Bendriss: Hmm, I though it wouldn't hurt - since as long as some of these entries are up-and-running no matter if some are down.
We can remove it and see how it behaves

delgod (approved):

WRFitch (approved):

PR #83, opened 2023-04-25
Reduce execution time of multi clusters tests

Mehdi-Bendriss:
## Issue
This PR reorders the provisioning of resources so that it happens in parallel to reduce the waiting time


welpaolo (approved):

paulomach (approved):
ok

PR #87, opened 2023-05-01
Use Juju 3 on CI 

WRFitch:
## Issue
We need to update to Juju 3, but continue to run tests (at least before release) on juju 2. This PR builds and runs the opensearch charm on Juju 3 in CI, but runs the CI pipeline on juju 2 and 3 on release. To do this, I've also created a setup script for the integration tests, since we also need to set a `JUJU_VERSION` environment variable to guarantee we install the correct libjuju version.


.github/workflows/ci.yaml
```diff
@@ -136,11 +111,12 @@ jobs:
...
           fi
+
       - name: Run integration tests
         run: |
-          # set sysctl values in case the cloudinit-userdata not applied
-          sudo sysctl -w vm.max_map_count=262144 vm.swappiness=0 net.ipv4.tcp_retries2=5
-
+          # set up environment - get juju version and set kernel params
+          echo ${{matrix.runner}}
+          source ./setup.sh
```
delgod, line 119: Do we need to hide two bash commands into the script instead of running them explicitly here?

.github/workflows/ci.yaml
```diff
@@ -58,69 +63,39 @@ jobs:
           - tls-integration
           - client-integration
           - ha-integration
+        runner: ["ubuntu-22.04"]
+        include:
+          - tox-environments: h-scaling-integration
+            runner: "Ubuntu_4C_16G"
```
delgod, line 69: Is `Ubuntu_4C_16G` used only for `h-scaling-integration` or for all the tests?

tox.ini
```diff
@@ -135,28 +131,11 @@ pass_env =
...
     CI
     CI_PACKED_CHARMS
+    JUJU_VERSION
 deps =
-    opensearch-py
-    pytest
-    pytest-asyncio
-    juju==2.9.42.1 # juju 3.0.0 has issues with retrieving action results
-    pytest-operator
+    -r {tox_root}/requirements_juju{env:JUJU_VERSION}.txt
```
delgod, line 136: Do we need a separate file? Why not explicitly specify the juju version here as it was before the change?

.github/workflows/ci.yaml
WRFitch: This is the "big runner" - we only have one, and we're sharing it with other teams, so we're only using it to run the horizontal scaling tests, since they need to run on a more powerful machine.

tox.ini
WRFitch: The libjuju version needs to change with the major Juju version, so for Juju 2 we need to use 2.9.42.2 and for Juju 3 we need to use 3.1.0.1. These version numbers don't line up exactly with libjuju, and we want control over the version regardless, so we need to map Juju 2 to a libjuju version 2, and so on with Juju 3. We also can't have if statements (or any other features of shell scripts, as far as I've found) in tox files, so we can't pick a dependency based on what version of Juju is installed, which is why I'm using this workaround with environment variables. I'm very open to other options if you have them :)

CONTRIBUTING.md
```diff
@@ -32,12 +32,30 @@ source venv/bin/activate
 
 ### Testing
 
+Test environments rely on the `JUJU_VERSION` environment variable to be set to your major juju version before running. To do this, run the following command:
+
+```shell
+source ./setup.sh
+```
```
Mehdi-Bendriss, line 39: I am personally not in favor of this approach. I would prefer we pass the juju version as a tox `posargs` with a default set to `juju 2`, i.e: `tox -e ha-integration -- juju-version=3`

requirements_juju3.txt
```diff
@@ -0,0 +1 @@
+juju>=3
```
Mehdi-Bendriss, line 1: Could we pin the version?

tests/integration/helpers.py
```diff
@@ -33,7 +33,7 @@
 
 MODEL_CONFIG = {
     "logging-config": "<root>=INFO;unit=DEBUG",
-    "update-status-hook-interval": "1m",
+    "update-status-hook-interval": "29s",
```
Mehdi-Bendriss, line 36: I would argue we should rather increase the interval, so that we don't hide charm issues by cleanup logic of update_status. And for the tests where we test the logic of this hook we decrease it as you did

Mehdi-Bendriss (commented):
Good job. I think we need to find a simpler way to manage both versions

CONTRIBUTING.md
WRFitch: I like your option more, and it installs the correct version if we only run the following command: 

`tox -e ha-integration -- 3`

with the configuration: 

```ini
[testenv:ha-integration]
description = Run HA integration tests
pass_env =
    {[testenv]pass_env}
    CI
    CI_PACKED_CHARMS
    JUJU_VERSION
deps =
    -r {tox_root}/requirements_juju{posargs:2}.txt
    -r {tox_root}/requirements_integration.txt
    -r {tox_root}/requirements.txt
commands =
    pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/ha/test_ha.py
```

However, as you can see in the last line of this config, we also have `posargs` in our pytest command so this tox run tries to run tests in a file named `3`, which then fails. I've looked for ways to pass arguments to tox in a normal way, and I'm coming up short - I'll come back to it next week.

tox.ini
```diff
@@ -65,98 +70,27 @@ commands =
...
+[testenv:{j2,j3}-{charm,tls,h-scaling,ha,client}-integration]
+description = Run integration tests
 pass_env =
     {[testenv]pass_env}
     CI
     CI_PACKED_CHARMS
 deps =
+    # These are libjuju versions, not Juju.
+    j2: juju==2.9.42.1
+    j3: juju==3.1.0.1
```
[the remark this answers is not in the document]
WRFitch: I'm not sure how poetry works, is there a way to pass in values to poetry that allows a user to select what packages are installed?

tox.ini
```diff
@@ -5,6 +5,11 @@
 no_package = True
 skip_missing_interpreters = True
 env_list = format, lint, unit
+labels =
+    # Don't run this group in parallel, or with --keep-models because it creates a lot of local VMs.
+    integration = j2-{charm,tls,ha,h-scaling,client}-integration
+    j3-integrat6ion = j3-{charm,tls,ha,h-scaling,client}-integration
+min_version=4
```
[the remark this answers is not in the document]
WRFitch: ah, that seems like weird behaviour. I assumed this would just cause tox to fail if it wasn't the correct version, I'll remove this line in that case.

tox.ini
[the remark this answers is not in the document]
WRFitch: @Mehdi-Bendriss, which UX do you prefer? 

In terms of a single shared tox environment for each libjuju version, I think changing local juju versions isn't frequent right now but it might get in the way when we're planning juju 2 -> 3 upgrade testing. Not disagreeing with your point, but it may be annoying if we have to download it mid-test in the scenario where an integration test uses libjuju 2 to deploy the charm, but libjuju 3 to test the upgraded charm.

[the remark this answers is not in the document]
WRFitch: Since you can call tox environments from one another, an option for automated upgrade testing would be to build the model using libjuju 2 in a `setup-integration` environment, upgrade juju, and then test the model in a second `upgrade-integration` tox environment that uses version 3. It clogs the tox file a bit, but it seems like a relatively clean way to not have to constantly reinstall the same package twice in a test.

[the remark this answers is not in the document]
WRFitch: you add can tox to the [`allowlist-externals` field in the tox config](https://tox.wiki/en/latest/config.html#allowlist_externals), then just call `tox -e another-env` in the `commands` section.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+name: PR CI
```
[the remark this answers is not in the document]
WRFitch: The plan is to remove the juju 2 runs from PRs, since it's what Mehdi uses for development, so these will be different workflows.

delgod (approved):

tox.ini
```diff
@@ -65,85 +69,30 @@ commands =
...
-    pytest-asyncio
-    juju==2.9.42.4  # juju 3.0.0 has issues with retrieving action results
-    pytest-operator
-    -r {tox_root}/requirements.txt
-commands =
-    pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/ha/test_ha.py
-
-[testenv:ha-networking-integration]
-description = Run HA integration tests
+[testenv:{charm, tls, h-scaling, ha, client, ha-storage, ha-networking }-integration]
```
Mehdi-Bendriss, line 72:
```suggestion
[testenv:{charm, tls, h-scaling, ha, client, ha-storage, ha-networking}-integration]
```

tox.ini
```diff
@@ -5,6 +5,10 @@
 no_package = True
 skip_missing_interpreters = True
 env_list = format, lint, unit
+labels =
+    # Don't run this group in parallel, or with --keep-models because it creates a lot of local VMs.
+    integration = {charm,tls,ha,h-scaling,client}-integration
```
Mehdi-Bendriss, line 10:
```suggestion
    integration = {charm,tls,ha,ha-networking,h-scaling,client}-integration
```

PR #88, opened 2023-05-02
Fix client integration bug when adding two relations that access the same index

WRFitch:
## Issue
When adding a second relation with the same index, we fail to create the index and enter a fail/defer/retry loop. This means the user is never created, so the new relation is never accessible. 

## Solution
This PR updates the error response check in `call()` to verify that `error_response is None`. This fixes the issue because we were getting false negatives when evaluating `not error_response` in this function, since a failed response is falsy.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,7 +47,7 @@ jobs:
     needs:
       - lint
       - unit-test
-    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v2
+    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v2.3.9
 
   lite-integration-test:
     strategy:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,10 +54,10 @@ jobs:
       fail-fast: false
       matrix:
         tox-environments:
-          # - charm-integration
-          # - tls-integration
+          - charm-integration
+          - tls-integration
           - client-integration
-          # - ha-integration
+          - ha-integration
     name: ${{ matrix.tox-environments }}
     needs:
       - lint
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
     needs:
       - lint
       - unit-test
-    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v2
+    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v2.3.9
```
WRFitch, line 50: This makes sure we're building charms with charmcraft 2.2, since 2.3 fails to pass the charm_helper lib.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -257,6 +257,9 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):
             # we defer because we want the temporary status to be updated
             event.defer()
```
MiaAltieri, line 259: nit: bump libpatch

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -226,7 +226,9 @@ def call(
         ) -> requests.Response:
```
MiaAltieri, line 226: nit: bump libpatch

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -226,7 +226,9 @@ def call(
...
             """Performs an HTTP request."""
             if remaining_retries < 0:
-                if not error_response:
+                # Can't check `if not error_response`, because Response objects evaluate to False
+                # when the request fails.
```
MiaAltieri, line 230: excellent comment

MiaAltieri (approved):
looks good to me! Minor nits only

Question - is it possible/relevant to add an int test to highlight the bug you ran into?

MiaAltieri (approved):
looks good to me! Minor nits only

Question - is it possible/relevant to add an int test to highlight the bug you ran into?

delgod (approved):

PR #90, opened 2023-05-16
Clearing statuses correctly

WRFitch:
Previously we weren't making use of the `status.clear()` function to clear the unit status after the relation-created hook fires. This PR fixes that, allowing us to preserve status information.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_relation_provider.py
+++ b/lib/charms/opensearch/v0/opensearch_relation_provider.py
@@ -381,12 +381,11 @@ class OpenSearchProvider(Object):
             ips = set([node.ip for node in self.charm._get_nodes(use_localhost=True)])
         except OpenSearchHttpError:
             logger.error("unable to get nodes")
-            # TODO maybe set endpoints to 0?
-            return
+            ips = set()
 
         port = self.opensearch.port
         endpoints = ",".join([f"{ip}:{port}" for ip in ips - omit_endpoints])
         databag_endpoints = relation.data[relation.app].get("endpoints")
 
-        if endpoints and endpoints != databag_endpoints:
+        if endpoints != databag_endpoints:
             self.opensearch_provides.set_endpoints(relation.id, endpoints)
--- a/lib/charms/opensearch/v0/opensearch_relation_provider.py
+++ b/lib/charms/opensearch/v0/opensearch_relation_provider.py
@@ -350,8 +350,9 @@ class OpenSearchProvider(Object):
         if not self.unit.is_leader():
             return
         if self._unit_departing(event.relation):
-            # This unit is being removed, so don't update the relation.
+            # This unit is being removed.
             self.charm.peers_data.delete(Scope.UNIT, self._depart_flag(event.relation))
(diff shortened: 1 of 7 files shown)


lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -343,8 +354,9 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
         if not self.unit.is_leader():
             return
         if self._unit_departing(event.relation):
-            # This unit is being removed, so don't update the relation.
+            # This unit is being removed.
```
Mehdi-Bendriss, line 357: is this supposed to be ran only on the leader?

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -343,8 +354,9 @@ def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
...
             self.charm.peers_data.delete(Scope.UNIT, self._depart_flag(event.relation))
+            self.update_endpoints(event.relation, omit_endpoints={self.charm.unit_ip})
```
Mehdi-Bendriss, line 359: can you explain the logic here?

tests/integration/relations/opensearch_provider/test_opensearch_provider.py
```diff
@@ -187,6 +187,7 @@ async def test_scaling(ops_test: OpsTest):
     """Test that scaling correctly updates endpoints in databag.
 
     scale_application also contains a wait_for_idle check, including checking for active status.
+    Idle_period checks must be greater than 1 minute to guarantee update_status fires correctly.
```
Mehdi-Bendriss, line 190: What happens if the update_status hook runs every 6h?

lib/charms/opensearch/v0/opensearch_relation_provider.py
WRFitch: Update_endpoints only runs on the leader because it updates the app databag. However, I've moved this function to the departing hook, since we can directly omit the `event.departing_unit`.

tests/integration/relations/opensearch_provider/test_opensearch_provider.py
WRFitch: This comment refers specifically to how often the test should check status; since we only fire update_status once every minute, we need to check for slightly longer than that to guarantee that the status check fires, and therefore we're definitely checking the up-to-date status information from the charm.

The departing hooks should now handle the status update correctly, but the update_endpoints function is still called on update_status as a catchall. I could test removing this function call to simulate a really long update_status if you wanted, but I feel that it would be a mistake to ever configure update_status to run this infrequently.

lib/charms/opensearch/v0/opensearch_relation_provider.py
WRFitch: Effectively, we want to update the endpoints but we want to exclude the IP of the departing unit so the client app won't expect that unit to be reachable. I'm now doing this in the departing hook, using `event.departing_unit`'s IP.

delgod (approved):

PR #94, opened 2023-06-07
[DPE-1626] Add support for certificates and CA renewal 

Mehdi-Bendriss:
## Issue
This PR implements [DPE-1626](https://warthogs.atlassian.net/browse/DPE-1626), namely this PR implements:
- Replace the certificates / keys formats to using pkcs12 trustores / keystores
- Allow replacement of CA and regeneration of certificates on all nodes
- Fix bug when TLS relation removed and reestablished  

## Notes:
This PR is pending more info on the following raised [issue](https://github.com/opensearch-project/security/issues/2839) upstream, background being that it seems we can only have the changes reflected after a full cluster restart which is not expected nor ideal. ElasticSearch [supports the rolling restart](https://www.elastic.co/guide/en/elasticsearch/reference/current/update-node-certs-different.html) in such a scenario 

[DPE-1626]: https://warthogs.atlassian.net/browse/DPE-1626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,6 +8,7 @@ labels: bug
 <!-- Thank you for submitting a bug report! All fields are required unless marked optional. -->
 
 ## Steps to reproduce
+<!-- Please enable debug logging by running `juju model-config logging-config="<root>=INFO;unit=DEBUG"` (if possible) -->
 1. 
 
 ## Expected behavior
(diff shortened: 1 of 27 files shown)


phvalguima:
@Mehdi-Bendriss: there was quite some work on this topic from @reneradoi. Do we still need to keep this PR or are we okay to close it?

PR #95, opened 2023-06-12
[DPE-1554] HA tests - scale down preserves data

Mehdi-Bendriss:
## Issue
This PR implements [DPE-1554](https://warthogs.atlassian.net/browse/DPE-1554) -- namely, this PR implements:
- integration test to check if data is preserved after remounting storage with data on new unit
- add external IP on the SANs of the HTTP certificates (to handle the case of public clouds and exposed endpoints) 
- preventing the reset of admin credentials when going from 0 to 1 nodes
- flush content of indices on scale down
- Unit tests fix
- dependencies bump

[DPE-1554]: https://warthogs.atlassian.net/browse/DPE-1554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -238,8 +238,23 @@ def _get_sans(self, cert_type: CertType) -> Dict[str, List[str]]:
...
+        ips = {self.charm.unit_ip}
+        if cert_type == CertType.UNIT_HTTP and self.charm.unit_public_ip:
+            ips.add(self.charm.unit_public_ip)
+
+        for ip in ips.copy():
+            try:
+                name, aliases, addresses = socket.gethostbyaddr(ip)
+                ips.update(addresses)
+
+                dns.add(name)
```
delgod, line 251: fqdn and back resolve hostname can be the same.
```suggestion
                dns.update(name)
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -300,24 +305,35 @@ def _on_peer_relation_departed(self, event: RelationDepartedEvent):
         else:
             event.defer()
 
-    def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):
+    def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # noqa
```
delgod, line 308: please be more specific, like
```suggestion
    def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # noqa: E501
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -464,19 +480,14 @@ def on_tls_conf_set(
...
             return
 
         # Otherwise, we block.
         self.unit.status = BlockedStatus(TLSRelationBrokenError)
 
-        try:
-            self._stop_opensearch()
-        except OpenSearchStopError:
-            event.defer()
-
```
delgod, line 479: do we restart OpenSearch without TLS if the relation is broken?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -170,16 +171,20 @@ def _on_leader_elected(self, event: LeaderElectedEvent):
...
-            self.status.clear(AdminUserInitProgress)
+
+        # User config is currently in a default state, which contains multiple insecure default
+        # users. Purge the user list before initialising the users the charm requires.
+        self._purge_users()
+
+        # this is in case we're coming from 0 to N units, we don't want to use the rest api
+        self._put_admin_user(
+            keep_current_if_exists=self.peers_data.get(Scope.APP, "admin_user_initialized", False)
+        )
```
delgod, line 182: you don't need arg, you can check `admin_user_initialized` inside `_put_admin_user`
```suggestion
        self._put_admin_user()
```

delgod (changes requested):

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -170,16 +171,20 @@ def _on_leader_elected(self, event: LeaderElectedEvent):
...
-            # User config is currently in a default state, which contains multiple insecure default
-            # users. Purge the user list before initialising the users the charm requires.
-            self._purge_users()
-            self._put_admin_user()
-            self.peers_data.put(Scope.APP, "admin_user_initialized", True)
-            self.status.clear(AdminUserInitProgress)
+
+        # User config is currently in a default state, which contains multiple insecure default
+        # users. Purge the user list before initialising the users the charm requires.
+        self._purge_users()
```
WRFitch, line 177: What happens if we delete the leader node in a cluster that's already running?

tox.ini
```diff
@@ -129,6 +129,22 @@ deps =
 commands =
     pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tests_path}/integration/ha/test_ha.py
 
+[testenv:ha-storage-integration]
```
WRFitch, line 132: Should this be added to CI?

Mehdi-Bendriss: Good point, on the CI the storage type is `rootfs` which doesn't allow the detaching / mount etc.. so these tests are automatically skipped in that case. Not having them added to it will prevent us from unnecessarily allocate workers

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: It should not be a problem as long as the security index is initialized, as this file is only loaded once and besides the admin user (which is added after this step) all other users are added through the rest api.
In other words, the only unit(s) where this file is touched (with or without effect) are leader units, if the cluster is already initialized => no harm done. If not initialized => expected

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: We do not. Because if OpenSearch was running, the condition (before this block) `if is_tls_configured` would have been `True`.
As long as the node is well configured and running, we do not break it. New nodes will of course be unable to be configured and will be put on a blocked state.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -876,6 +901,11 @@ def unit_ip(self) -> str:
         """IP address of the current unit."""
         return get_host_ip(self, PeerRelationName)
 
+    @property
+    def unit_public_ip(self) -> str:
```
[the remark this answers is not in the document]
Mehdi-Bendriss: Good point, thank you. I deleted it

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: Good idea, thank you

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -661,10 +670,9 @@ def _purge_users(self):
 
     def _put_admin_user(self, pwd: Optional[str] = None):
         """Change password of Admin user."""
-        is_update = pwd is not None
-        hashed_pwd, pwd = generate_hashed_password(pwd)
-
-        if is_update:
+        # update
+        if pwd is not None:
+            hashed_pwd, pwd = generate_hashed_password(pwd)
```
delgod, line 675: If we don't store the password and hash here, then the`_on_set_password_action` function should store both of them.

delgod (changes requested):

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: We store them on success [here](https://github.com/canonical/opensearch-operator/pull/95/files/8643c1c13d920b3b61c10c9c4ae620cd16adb5ac#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R707-R708)

delgod (approved):

PR #98, opened 2023-06-13
[DPE-1637] HA tests - reuse storage volumes in new cluster

Mehdi-Bendriss:
## Issue
This PR implements [DPE-1637](https://warthogs.atlassian.net/browse/DPE-1637), namely this PR implements:
- integration tests for reuse of storage volumes in a new cluster with the same name

[DPE-1637]: https://warthogs.atlassian.net/browse/DPE-1637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -28,4 +28,4 @@ requires:
 storage:
   opensearch-data:
     type: filesystem
-    location: /var/snap/opensearch  # /mnt/opensearch/data
+    location: /var/snap/opensearch/common  # /mnt/opensearch/data
--- a/tests/integration/ha/test_storage.py
+++ b/tests/integration/ha/test_storage.py
@@ -176,7 +176,7 @@ async def test_storage_reuse_after_scale_down(
 
 @pytest.mark.abort_on_fail
 async def test_storage_reuse_in_new_cluster_after_app_removal(
-        ops_test: OpsTest, c_writes: ContinuousWrites, c_balanced_writes_runner
+    ops_test: OpsTest, c_writes: ContinuousWrites, c_balanced_writes_runner
 ):
     """Check storage is reused and data accessible after removing app and deploying new cluster."""
     app = (await app_name(ops_test)) or APP_NAME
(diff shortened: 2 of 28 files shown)


metadata.yaml
```diff
@@ -28,4 +28,4 @@ requires:
 storage:
   opensearch-data:
     type: filesystem
-    location: /var/snap/opensearch  # /mnt/opensearch/data
+    location: /var/snap/opensearch/common  # /mnt/opensearch/data
```
WRFitch, line 31: Why wouldn't we want the entire snap directory?

tests/integration/ha/test_storage.py
```diff
@@ -172,3 +172,81 @@ async def test_storage_reuse_after_scale_down(
...
+):
+    """Check storage is reused and data accessible after removing app and deploying new cluster."""
+    app = (await app_name(ops_test)) or APP_NAME
+
+    if storage_type(ops_test, app) == "rootfs":
+        pytest.skip(
+            "re-use of storage can only be used on deployments with persistent storage not on rootfs deployments"
+        )
+
+    # scale-down to 1 if multiple units
```
WRFitch, line 189: You use this code a few times. Perhaps it could be extracted to a function?

WRFitch (approved):

delgod (approved):

tests/integration/ha/test_storage.py
Mehdi-Bendriss: thank you, will have an integration tests cleanup a bit later

metadata.yaml
Mehdi-Bendriss: because the only part we care about to be persisted / attached / re-mounted etc.. are the directories that contain data / logs, the rest can just be put back by the charm itself

PR #105, opened 2023-08-10
[DPE-2347] Model plugin support

phvalguima:
Modelling the plugin support in the OpenSearch operator.

This PR implements [DPE-2347](https://warthogs.atlassian.net/browse/DPE-2347) and represents the first work to add plugin management to the operator. The plugin can be broken into 3 parts: (1) a class representing its lifecycle status and configuration; (2) a manager class, responsible to actually run the operational tasks related to installing, configuring, upgrading and removing plugins; and (3) changes in charm-related classes to initialize the plugin manager and to call corresponding steps, e.g. installation, configuration, etc.

This PR implements an abstract class: OpenSearchPlugin, that maps plugin lifecycle status and configuration. That can be seen as follows:
* installation
* configuration
* upgrade
* removal

This PR also implements OpenSearchPluginManager class, responsible for the operations of each step above. Finally, this PR changes the charm main code to call the plugin manager on every config-changed event.

For future works: plugins such as backup (repository-s3) should depend on a separate relation to decide to install/configure/remove the corresponding plugin. In this case, a class that manages that relation in the charm level should also be implemented, responsible for making calls to the OpenSearchPluginManager directly.

[DPE-2347]: https://warthogs.atlassian.net/browse/DPE-2347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


deusebio (commented):
First pass. 

In general I believe you could make the code more readable if you were to use variables names that are more meaningful. I had a bit of a hard time to understand the logic, which could have been easier when variables name help 

I provided some suggestions how to achieve this, but also try to apply this even if it was not specifically mentioned. I believe it will make code more readable and easier to mantain

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,489 @@
...
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+logger = logging.getLogger(__name__)
+
+
+class OpenSearchPlugin:
```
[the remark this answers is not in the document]
phvalguima: I instead adopted: 
```
class OpenSearchPlugin(ABC):
    """Abstract class describing an OpenSearch plugin."""
```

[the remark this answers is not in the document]
deusebio: yes, my bad! that's what I meant! :D

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+            ]
+        )
+
+    def is_installed(self) -> bool:
+        """Returns True if the plugin name is present in the list of installed plugins."""
+        if not self._is_started():
+            return False
+        return self.name in self.distro.list_plugins()
+
+    def install(self, uri: str, batch=True) -> bool:
```
Mehdi-Bendriss, line 231: We don't need a boolean return here, an unsuccessful install should raise an exception

Mehdi-Bendriss (changes requested):

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -305,13 +309,79 @@ def write_file(self, path: str, data: str, override: bool = True):
         with open(path, mode="w") as f:
             f.write(data)
 
+    @property
+    def _plugin(self):
+        """Returns the executable path. Should be overloaded, e.g. by the snap command."""
+        return f"{self.paths.home}/bin/opensearch-plugin"
```
Mehdi-Bendriss, line 315: not needed

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -305,13 +309,79 @@ def write_file(self, path: str, data: str, override: bool = True):
...
+
+    def add_plugin_without_restart(self, plugin: str, batch: bool = False) -> bool:
```
Mehdi-Bendriss, line 317: The plugins related methods should only be as part of the `PluginManager` and not have any say in the restart logic

```suggestion
    def add_plugin(self, plugin: str, batch: bool = False) -> bool:
```

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -305,13 +309,79 @@ def write_file(self, path: str, data: str, override: bool = True):
...
+        """Add a plugin to this node. Restart must be managed in separated."""
+        try:
+            args = "install"
+            if batch:
+                args += " --batch"
+            args += f" {plugin}"
```
Mehdi-Bendriss, line 323: let's just `--batch` for all plugins - given we only offer plugins we trust in conf, and we cannot deal with prompts at the user -> charm level anyway.

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -305,13 +309,79 @@ def write_file(self, path: str, data: str, override: bool = True):
...
+            self._run_cmd(self._plugin, args)
+        except OpenSearchCmdError as e:
+            if "already exists" in e.stderr:
```
Mehdi-Bendriss, line 326:
```suggestion
            if "already exists" in str(e):
```

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -305,13 +309,79 @@ def write_file(self, path: str, data: str, override: bool = True):
...
+                return
+            raise OpenSearchPluginError(f"Failed to install plugin {plugin}")
+
+    def remove_plugin_without_restart(self, plugin):
+        """Remove a plugin without restarting the node."""
+        try:
+            args = f"remove {plugin}"
+            self._run_cmd(self._plugin, args)
```
Mehdi-Bendriss, line 334: user `run_bin` instead

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -305,13 +309,79 @@ def write_file(self, path: str, data: str, override: bool = True):
...
+    def add_to_keystore(self, key: str, value: str, force: bool = False):
+        """Adds a given key to the "opensearch" keystore."""
+        if not value:
+            raise OpenSearchKeystoreError("Missing keystore value")
+        args = "add " if not force else "add --force "
+        args += f"{key}"
+        try:
+            # Add newline to the end of the key, if missing
+            v = value + ("" if value[-1] == "\n" else "\n")
+            self._run_cmd(self._keystore, args, input=v)
```
Mehdi-Bendriss, line 361: Why are we doing this? we should have well formatted keys from the get-go, what we store here should match what is stored in the secrets or peer rel data

lib/charms/opensearch/v0/opensearch_exceptions.py
```diff
@@ -92,3 +92,11 @@ class OpenSearchScaleDownError(OpenSearchError):
 
 class OpenSearchIndexError(OpenSearchError):
     """Exception thrown when an opensearch index is invalid."""
+
+
+class OpenSearchPluginError(OpenSearchError):
```
Mehdi-Bendriss, line 97: Could we put this class in `opensearch_plugins.py`?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,65 @@
...
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+
+OpenSearchPluginsAvailable = {}
```
Mehdi-Bendriss, line 23: I like the idea of having a registry! 
The `Available` part may be a bit confusion (built-in, configured etc..). we could perhaps rename it to: `ConfigExposedPlugins`? or something else that's explicit

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class OpenSearchPluginManager:
```
Mehdi-Bendriss, line 26: Nice

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,65 @@
...
+    @property
+    def plugins(self) -> Dict[str, OpenSearchPlugin]:
+        """Returns dict of installed plugins."""
+        return {
+            key: plugin_data["class"](key, self._charm)
+            for key, plugin_data in OpenSearchPluginsAvailable.items()
+        }
+
+    def plugin_map_config_name_to_class(self) -> Dict[str, OpenSearchPlugin]:
+        """Returns dict of plugins installed either via config or relation.
```
Mehdi-Bendriss, line 41: This method seems to be redundant: `plugins` should be enough. (not sure I understand "installed via relation"?)

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,65 @@
...
+        Relation names will take precedence over config.
+        """
+        return {
+            plugin_data["relation-name"]
+            if plugin_data.get("relation-name", None)
+            else plugin_data["config-name"]: plugin_data["class"](key, self._charm)
+            for key, plugin_data in OpenSearchPluginsAvailable.items()
+        }
+
+    def get_status(self) -> StatusBase:
```
Mehdi-Bendriss, line 56: What value do you see in this method? the status of say the knn plugin is unrelated to the status of the asynchronous search one, I don't think there is value in aggregating the statuses or returning a single plugin's

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+Add the corresponding config-name to the OpenSearchPluginsAvailable.
+
+If a given plugin depends on a relation, e.g. repository-s3, then add relation name
+as well. For example:
+    OpenSearchPluginsAvailable = {
+        ...
+        "opensearch-knn": {
+            "class": OpenSearchPlugin,
+            "config-name": "plugin_opensearch_knn",
+            "relation-name": ""
```
Mehdi-Bendriss, line 58: What would be the use of this option?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+logger = logging.getLogger(__name__)
+
+
+class OpenSearchPlugin(ABC):
```
Mehdi-Bendriss, line 89: Nice concept! 
We should make this class a declarative one, and offload all the operational work to the PluginManager.
We should then probably remove some methods and put them in the PluginManager and add a few methods such as: 
- `config()` => returns a dictionary of the config that must be added for the plugin to work. This will then be the job of the plugin manager to apply it on the `self.charm.opensearch_config`
- `install_source` => where is the plugin going to be picked up from for the plugin manager to locate it for the installation

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def __init__(self, name: str, charm: Object, relname: Optional[str] = None):
+        """Creates the OpenSearchPlugin object.
+
+        The *args enable children classes to pass relations.
+        """
+        self.relname = relname
+        self.charm = charm
+        self.distro = self.charm.opensearch
+        self.CONFIG_YML = self.charm.opensearch_config.CONFIG_YML
```
Mehdi-Bendriss, line 103: Let's avoid interacting directly with the `CONFIG_YML` file. Let's only interact with `self.charm.opensearch_config` methods

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+        self._name = name
+        self._plugin_properties = Properties()
```
Mehdi-Bendriss, line 105:
```suggestion
        self._properties = Properties()
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+        """Returns the current version of the plugin."""
+        with open(os.path.join(f"{self.distro.paths.plugins}", f"{self.PLUGIN_PROPERTIES}")) as f:
+            self._plugin_properties.load(f.read())
+        return self._plugin_properties._properties["version"]
+
+    def _is_started(self) -> bool:
+        return self.distro.is_started()
+
+    def _request(self, *args, **kwargs) -> dict:
+        return self.distro.request(*args, **kwargs)
```
Mehdi-Bendriss, line 118: These are simply proxy methods, they're not needed. 
```suggestion
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def is_related(self) -> bool:
+        """Returns True if a relation is expected and it is set."""
+        if not self.relname:
+            return True
+        return len(self.charm.framework.model.relations[self.relname] or {}) > 0
+
+    def _update_keystore_and_reload(
```
Mehdi-Bendriss, line 126: Let's put anything related to the keystore in its own module and class

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+        except Exception as e:
+            logger.exception(e)
+            raise OpenSearchPluginError("Unknown error during keystore reload")
+        if post["status"] < 200 or post["status"] >= 300:
+            raise OpenSearchPluginError("Error while processing _nodes reload")
+
+    @property
+    def name(self) -> str:
+        """Returns the name of the plugin."""
+        return self._name
```
Mehdi-Bendriss, line 151: This is not needed, let's just return a constant

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+        """Erases relevant data for this plugin. Returns True if restart needed."""
+        if not self._is_started():
+            return False
+        self._update_keystore_and_reload(keystore)
+        return any(
+            [
+                self.distro.config.delete(self.CONFIG_YML, c, v)[c] == v
+                for c, v in opensearch_yml.items()
+            ]
+        )
```
Mehdi-Bendriss, line 167: This is the job of the plugin manager (while solely interacting with a method in `OpenSearchConfig` and not the yaml filte itself)

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def configure(
+        self,
+        opensearch_yml: Dict[str, str],
+        keystore: Dict[str, str] = {},
+    ) -> bool:
+        """Sets the plugin configuration. Returns True if restart needed."""
+        if not self._is_started():
+            return False
+        self._update_keystore_and_reload(keystore)
```
Mehdi-Bendriss, line 217: why is the keystore such a central piece of plugins management?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+            [
+                self.distro.config.put(self.CONFIG_YML, c, v)[c] == v
+                for c, v in opensearch_yml.items()
+            ]
+        )
+
+    def is_installed(self) -> bool:
+        """Returns True if the plugin name is present in the list of installed plugins."""
+        if not self._is_started():
+            return False
```
Mehdi-Bendriss, line 228: The manager should be the one knowing what is installed or not. And the plugin should rather just be declarative and return what it takes for it to be configured. 

(Why are we testing on the start of the opensearch service? if the service is down or restarting this plugin will be marked as not installed)

lib/charms/opensearch/v0/opensearch_plugins.py
Mehdi-Bendriss, line 231: We don't need a boolean return here, an unsuccessful install should raise an exception

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+            return True
+        if self._depends_on:
+            plugins = self.distro.list_plugins()
+            missing = []
+            for dependency in self.depends_on:
+                if dependency not in plugins:
+                    missing.append(dependency)
+                if dependency:
+                    raise OpenSearchPluginError("Missing dependencies")
+        self.distro.add_plugin_without_restart(uri, batch=batch)
```
Mehdi-Bendriss, line 250: This logic should be in the PluginManager

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,277 @@
...
+
+    @property
+    @abstractmethod
+    def depends_on(self) -> List[str]:
+        """Returns a list of plugins it depends on."""
+        pass
+
+    def get_status(self) -> Union[int, StatusBase]:
```
Mehdi-Bendriss, line 258: Not a fan of using `StatusBase` - it is a UI related class of the charm, this shouldn't be used for internal states of plugins etc.. We should create an enum for the plugins' states:
```
class PluginState(BaseStrEnum):
    MISSING = "missing"
    DISABLED = "disabled"
    WAITING_FOR_UPGRADE = "waiting-for-upgrade"
    ACTIVE = "active"


@property
def state() -> PluginState:
    if not self.installed():
        status = PluginState.MISSING
    elif not self.enabled():
        status = PluginState.DISABLED
    elif self.needs_upgrade():
        status = PluginState.WAITING_FOR_UPGRADE
    else:
        status = PluginState.ACTIVE

    return status
```

requirements.txt
```diff
@@ -6,4 +6,6 @@ overrides==7.3.1
 requests==2.31.0
 ruamel.yaml==0.17.31
 tenacity==8.2.2
-urllib3==1.26.15
\ No newline at end of file
+urllib3==1.26.15
+git+https://git.launchpad.net/python-jproperties@debian/main
```
Mehdi-Bendriss, line 10: Safe to remove - let's rely on the pip dependency.

Mehdi-Bendriss (commented):
This is a very good first iteration! 
We should work on removing redundancies or loosening up tight coupling between the components and having a clear workflow `charm -> plugin_manager -> plugin`
Also thinking the exact use of the keystore component

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -191,6 +191,23 @@ def cleanup_bootstrap_conf(self):
...
+        plugins = plugin_manager.plugin_map_config_name_to_class()
+        ret = False
+        for c in configs:
+            if charm.config[c] != plugins[c].is_enabled():
+                # Toggle the plugin
+                if plugins[c].is_enabled():
+                    ret = ret or plugins[c].disable()
+                else:
+                    ret = ret or plugins[c].enable()
+        return ret
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss it is possible to do:
```
juju config plugin_A=true plugin_B=false # change the state of 2x plugins at once
```
So, we cannot assume a single plugin has been changed

lib/charms/opensearch/v0/opensearch_distro.py
phvalguima: @Mehdi-Bendriss this check goes to the input in the subprocess, which must have a "\n" to actually pass the value; whereas the content of ```value``` comes from the S3 relation, which may not have it.
We are using ```input``` so we do not create a temporary file with the keys in plain text while passing them to the keystore. It is more appropriate to be prompted by opensearch-keystore and pass it as an input.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: @Mehdi-Bendriss it is the same pattern as: https://github.com/canonical/postgresql-operator/blob/main/config.yaml#L5

lib/charms/opensearch/v0/opensearch_distro.py
phvalguima: Okay, doing the same thing for ```add_to_keystore``` and the ```--force```

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: @Mehdi-Bendriss overall, you must add a plugin configuration to opensearch.yml or the default keystore (if that given option is a  ```SecureSetting``` class).
In the case of backup, for example, we can see that access- and secret-keys must be set in keystore, as [they are secure settings](https://github.com/opensearch-project/OpenSearch/blob/bb7d23ca3c31140b03899d83648948dbe7229cf2/plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/S3ClientSettings.java#L100-L118)

Also, if we decide to remove a plugin but do not clean the keystore, then opensearch fails to start.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: @Mehdi-Bendriss that is the inverted map, and I'm using both, actually

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: Is there a way to call ```self.distro.config.delete``` without specifying CONFIG_YML? Because every plugin will have a specific list of configs to add/remove

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: We need to do the following changes:
* Move away from self.distro to self.charm.opensearch_config
* Create a proxy method to put and delete a collection of entries in OpenSearchConfig. That method should receive a dict containing the key/values.

lib/charms/opensearch/v0/opensearch_distro.py
phvalguima: * Move the keystore and plugin install/remove/list to the PluginManager: remove from OpenSearchDistribution
* In case we need, we can extend the PluginManager to the k8s example, in the future
* The restart logic will be executed by the charm, as folllows, for example backup plugin:

There will be two files:
src/opensearch_backup.py
```
class OpenSearchCharmBackup(...):

     def __init__(self, charm):
          self.plugin_manager = self.charm.plugin_manager
          self.framework.observe(self.charm.on.s3_credentials_changed, self.s3_credentials_changed)

    def s3_credentials_changed(self, event):
         ...
         if self.plugin_manager.configure("repository-s3", ...):
              # needs a restart
```

In the main charm:
```

class OpenSearchCharm(BaseCharm):

    def __init__(self, ...):
        self.plugin_manager = OpenSearchPluginManager(...)
        self.framework.observe(self.charm.on.config_changed, self.on_config_changed)


    def on_config_changed(self, event):
        ....
        if self.plugin_manager.setUp(self.config): # the plugin manager figures out which to save
            # we need a restart
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def _add_plugin(self, plugin: str) -> bool:
+        """Add a plugin to this node. Restart must be managed in separated."""
+        try:
+            args = f"install --batch {plugin}"
+            self._distro.run_bin("opensearch-plugin", args)
+        except OpenSearchCmdError as e:
+            if "already exists" in str(e):
+                return
+            raise OpenSearchPluginError(f"Failed to install plugin {plugin}: " + str(e))
```
Mehdi-Bendriss, line 196: 1. If we go with the route of having plugin manager running an `_install` per plugin: we could incorporate the code of these methods in the `_install` methods without having them separate

2. Create a couple of more exceptions: `OpenSearchPluginInstall/RemoveError(OpenSearchPluginError)` those are the ones you'll raise below


```suggestion
        """Add a plugin to this node. Restart must be managed in separated."""
        try:
            self._distro.run_bin("opensearch-plugin", f"install --batch {plugin}")
        except OpenSearchCmdError as e:
            if "already exists" in str(e):
                return
            raise OpenSearchPluginInstallError(e)
```

Mehdi-Bendriss (changes requested):
This is a good second pass - thank you. I left some comments

lib/charms/opensearch/v0/opensearch_exceptions.py
phvalguima: We should move it to opensearch_plugin.py and opensearch_keystore.py

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,234 @@
...
+
+        The dict should return:
+        (1) all the configs to remove
+        (2) all the configs to add
+        (3) all the keystore values to be remmoved.
+        """
+        return {"to_remove_opensearch": [], "to_add_opensearch": [], "to_remove_keystore": []}
+
+    @property
+    def status(self) -> PluginState:
```
phvalguima, line 220: We cannot track the status here, as this class is just a spec of who the plugin actually is.

This method should move to plugin manager.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,237 @@
...
+logger = logging.getLogger(__name__)
+
+
+ConfigExposedPlugins = {}
+
+
+class OpenSearchPluginManager:
+    """Manages the currently plugins."""
+
+    CONFIG_YML = "opensearch.yml"
```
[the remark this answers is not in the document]
phvalguima: We use that in the config dict in OpenSearchPlugin

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,236 @@
...
+            # the priority
+            configs[self.CONFIG_YML] = {**configs[self.CONFIG_YML], **c[self.CONFIG_YML]}
+            configs["keystore"] = {**configs["keystore"], **c["keystore"]}
+
+        self._update_keystore_and_reload(configs["keystore"])
+        return self._config.config_put_list(configs[self.CONFIG_YML])
+
+    def disable(self) -> bool:
+        """If enabled, removes plugin configuration or sets it to other values."""
+        configs = {"to_remove_opensearch": [], "to_add_opensearch": [], "to_remove_keystore": []}
```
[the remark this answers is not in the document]
phvalguima: That is the configs that must be either removed or readded to opensearch.yml; or removed from keystore. It is the mirror method from ```self.configure()```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: install method has some pre-checks before actually calling the ```_add_plugin```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,219 @@
...
+        plugins_path: str,
+        os_version: Optional[str] = "",
+        relation_data: Optional[Dict[str, Any]] = {},
+    ):
+        """Creates the OpenSearchPlugin object."""
+        self._plugins_path = plugins_path
+        self._os_version = os_version
+        self._relation_data = relation_data
+        self._properties = Properties()
+        self._depends_on = []
```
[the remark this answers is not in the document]
phvalguima: ```self._properties``` is used later to load the plugin version from ```plugin-properties```. The class needs to be initialized nonetheless.

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,219 @@
...
+                self._properties.load(f.read())
+        except FileNotFoundError:
+            raise OpenSearchPluginError(f"Plugin is missing - not found: {plugin_props_path}")
+        except PermissionError:
+            raise OpenSearchPluginError(
+                f"Plugin incorrectly installed - permission denied: {plugin_props_path}"
+            )
+        except Exception:
+            raise OpenSearchPluginError("Plugin Error - Unknown reason")
+        return self._properties._properties["version"]
```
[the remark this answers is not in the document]
phvalguima: Personally, this is related to the ```def version(self)``` method, which is very internal to the plugin infrastructure. I would rather have it returning exclusively OpenSearchPluginError instead of different excepts.

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def disable(self) -> Dict[str, Any]:
+        """Returns a dict containing different config changes.
+
+        The dict should return:
+        (1) all the configs to remove
+        (2) all the configs to add
+        (3) all the keystore values to be remmoved.
+        """
+        return {"to_remove_opensearch": [], "to_add_opensearch": [], "to_remove_keystore": []}
```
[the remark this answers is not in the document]
phvalguima: I have updated the code to contain an ```OpenSearchPluginConfig``` class that contains both config and secret entries. However, I am not adding any API calls to this class with ```enable_endpoint``` or ```disable_endpoint``` as I think that should be executed on a level closer to the charm and aware of who is the unit leader, for example.

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -191,6 +191,28 @@ def cleanup_bootstrap_conf(self):
...
+
+    def add_plugin(self, plugin_config: Dict[str, str]) -> None:
+        """Adds a plugin configurations into opensearch.yml."""
+        if not plugin_config:
+            return
+        for key, val in plugin_config.items():
+            self._opensearch.config.put(self.CONFIG_YML, key, val)
+
+    def delete_plugin(self, plugin_config: Dict[str, str]) -> None:
+        """Adds a plugin configurations into opensearch.yml."""
```
dmitry-ratushnyy, line 210: Looks like wrong  doc string here

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,227 @@
...
+        # we may need or not a restart
+        return restart
+
+    def _install(self, plugin: OpenSearchPlugin) -> bool:
+        """Installs all the plugins enabled via the config/relation.
+
+        Check if plugin in status: PluginState.MISSING and config/relation is set.
+        Returns True if a restart is needed.
+        """
+        needs_restart = False
```
dmitry-ratushnyy, line 113: nits: do we need this var?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,227 @@
...
+            return PluginState.MISSING
+        if not self._user_requested_to_enable(plugin) or not self._is_enabled(plugin):
+            return PluginState.AVAILABLE
+        if self._needs_upgrade(plugin):
+            return PluginState.WAITING_FOR_UPGRADE
+        return PluginState.ENABLED
+
+    def _is_installed(self, plugin: OpenSearchPlugin) -> bool:
+        """Returns true if plugin is installed."""
+        if plugin.name in self._list_plugins():
```
dmitry-ratushnyy, line 166: nits: looks like this can be simplified a bit

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def _needs_upgrade(self, plugin: OpenSearchPlugin) -> bool:
+        """Returns true if plugin needs upgrade."""
+        current_version = plugin.version
+        num_points = len(self._opensearch.version.split("."))
+        return self._opensearch.version != current_version[:num_points]
+
+    def _is_plugin_relation_set(self, relation_name: str) -> bool:
+        """Returns True if a relation is expected and it is set."""
+        if not relation_name:
```
dmitry-ratushnyy, line 197: just curious why return `True` if no relation name provided

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -0,0 +1,114 @@
...
+        except OpenSearchCmdError as e:
+            raise OpenSearchKeystoreError(str(e))
+
+    def _delete(self, key: str):
+        args = f"remove {key}"
+        try:
+            self._opensearch.run_bin("opensearch-keystore", args)
+        except OpenSearchCmdError as e:
+            if "does not exist in the keystore" in str(e):
+                return
```
dmitry-ratushnyy, line 96: do we want to have some additional debug logging here?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,227 @@
...
+        # Relation is not defined, return config value then
+        config = plugin_data["config"]
+        return {config: self._charm_config[config]}
+
+    def run(self) -> bool:
+        """Runs a check on each plugin: install, execute config changes or remove.
+
+        This method should be called at config-changed event. Returns if needed restart.
+        """
+        restart = False
```
dmitry-ratushnyy, line 88: Do we need this variable?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,227 @@
...
+            return True
+        return len(self._charm.framework.model.relations[relation_name] or {}) > 0
+
+    def _add_plugin(self, name: str) -> None:
+        """Add a plugin to this node. Restart must be managed in separated."""
+        try:
+            self._opensearch.run_bin("opensearch-plugin", f"install --batch {name}")
+        except OpenSearchCmdError as e:
+            if "already exists" in str(e):
+                return
```
dmitry-ratushnyy, line 207: do we want to have some additional debug logging here?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def __init__(self, plugins_path: str, extra_config: Dict[str, Any] = None):
+        """Creates the OpenSearchPlugin object.
+
+        Arguments:
+          plugins_path: str, path to the plugins folder
+          extra_config: dict, contains the data coming from either the config option
+                              or the relation bag
+        """
+        self._plugins_path = plugins_path
```
Mehdi-Bendriss, line 299: since our charms run in linux I think it's safe to do the following: 
```suggestion
        self._plugins_path = f"{plugins_path}/{self.PLUGIN_PROPERTIES}"
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,348 @@
...
+        self._extra_config = extra_config
+        self._properties = Properties()
```
Mehdi-Bendriss, line 301: it doesn't seem to be used in multiple places, let's make this a local variable

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,348 @@
...
+
+        Returns: str, string with the version code for this plugin
+        Raises:
+            FileNotFoundError: if plugin file is not present
+            PermissionError: if plugin file is present, but not set with correct permissions
+        """
+        plugin_props_path = os.path.join(f"{self._plugins_path}", f"{self.PLUGIN_PROPERTIES}")
+        with open(plugin_props_path) as f:
+            self._properties.load(f.read())
+        return self._properties._properties["version"]
```
Mehdi-Bendriss, line 315: Given our charms run in linux, I think it's safe to use `/`:
```suggestion
        properties = Properties()
        with open(self._plugins_path) as f:
            properties.load(f.read())
        return properties._properties["version"]
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,348 @@
...
+
+    @property
+    @abstractmethod
+    def dependencies(self) -> List[str]:
+        """Returns a list of plugin name dependencies."""
+        pass
```
Mehdi-Bendriss, line 321:
```suggestion
    @property
    def dependencies(self) -> Optional[List[str]]:
        """Returns a list of plugin name dependencies."""
        return None
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,348 @@
...
+class PluginState(BaseStrEnum):
+    """Enum for the states possible in plugins' lifecycle."""
+
+    MISSING = "missing"
+    AVAILABLE = "available"
+    ENABLED = "enabled"
+    WAITING_FOR_UPGRADE = "waiting-for-upgrade"
+
+
+class OpenSearchPluginConfig:
```
Mehdi-Bendriss, line 274: see the comment above about potentially enriching `config_entries` and making a separate class for that

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,348 @@
...
+        Format:
+        OpenSearchPluginConfig(
+            self.config_entries = {...},
+            self.secret_entries = {...},
+        )
+        """
+        pass
+
+    @abstractmethod
+    def disable(self) -> Tuple[OpenSearchPluginConfig, OpenSearchPluginConfig]:
```
Mehdi-Bendriss, line 336: if we go with the enriching of `OpenSearchPluginConfig` we won't need this one

Mehdi-Bendriss (commented):
I really like how the code is taking shape! Amazing job, I believe we're close! Thank you!

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,227 @@
...
+
+        This method should be called at config-changed event. Returns if needed restart.
+        """
+        restart = False
+        # Save original configuration to compare later
+        original_config = self._opensearch_config.load_node()
+        for plugin in self.plugins:
+            restart = self._install(plugin) or restart
+            self._configure(plugin)
+            self._disable(plugin)
```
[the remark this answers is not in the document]
phvalguima: For self._configure and self._disable, it grabs the ```opensearch.yml``` before and after running these methods. If there was a change, then it requests a restart. It means fewer reads to ```opensearch.yml```.

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -0,0 +1,114 @@
...
+
+    @property
+    def exists(self) -> bool:
+        """Checks if opensearch-keystore exists."""
+        return os.path.isfile(os.path.join(self._charm.paths.conf, self.KEYSTORE))
+
+    def add(self, entries: Dict[str, str]) -> None:
+        """Adds a given key to the "opensearch" keystore."""
+        if not entries or not self.exists:
+            return  # no key/value to add, no need to request reload of keystore either
```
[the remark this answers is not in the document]
phvalguima: Yes, in practice it should always be there

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,227 @@
...
+            if self._is_plugin_relation_set(plugin_data["relation"]):
+                result = {}
+                relation = self._charm.framework.model.relations[plugin_data["relation"]]
+                for i in range(len(relation)):
+                    result = {**relation[i].data, **result}
+                return result
+            return {}
+        # Relation is not defined, return config value then
+        config = plugin_data["config"]
+        return {config: self._charm_config[config]}
```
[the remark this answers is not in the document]
phvalguima: The ```extra_config``` comes in the case of, e.g. repository-s3, which needs to configure the S3 endpoint in opensearch.yml. For that, it needs access to the relation databag.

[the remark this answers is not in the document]
phvalguima: Your change makes sense, I will add it (but leaving this review opened, so we can discuss if needed).

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,227 @@
...
+                return False
+        return True
+
+    def _needs_upgrade(self, plugin: OpenSearchPlugin) -> bool:
+        """Returns true if plugin needs upgrade."""
+        current_version = plugin.version
+        num_points = len(self._opensearch.version.split("."))
+        return self._opensearch.version != current_version[:num_points]
+
+    def _is_plugin_relation_set(self, relation_name: str) -> bool:
```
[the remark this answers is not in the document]
phvalguima: I have simplified this method, please check it

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,227 @@
...
+        self._opensearch_config.add_plugin(plugin.config().config_entries)
+        self._keystore.add(plugin.config().secret_entries)
+
+    def _disable(self, plugin: OpenSearchPlugin) -> None:
+        """If disabled, removes plugin configuration or sets it to other values."""
+        if self.status(plugin) != PluginState.AVAILABLE:
+            # Only considering "available" status as it represents a plugin that has
+            # been installed but either not yet configured or user explicitly disabled.
+            return
+        config_to_remove, config_to_add = plugin.disable()
```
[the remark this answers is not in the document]
phvalguima: Adopted a different approach, based on our discussions:

* OpenSearchPluginConfig inherits from typing.NamedTuple
* OpenSearchPluginConfig is composed of configs and secrets to be added or removed
* Defined some "actions", in this case: config() and disable() -> each return a OpenSearchPluginConfig object with configurations to add and remove

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -191,6 +191,42 @@ def cleanup_bootstrap_conf(self):
...
+    def add_plugin(self, plugin_config: Dict[str, str]) -> bool:
+        """Adds plugin configuration into opensearch.yml.
+
+        Returns True if all the configurations have been set correctly.
+        """
+        if not plugin_config:
+            return False
+        return all(
+            [
+                self._opensearch.config.put(self.CONFIG_YML, key, val).get(key, None) == val
```
dmitry-ratushnyy, line 211: This expression looks quite complex and hard to understand tbh. 
Maybe it is better to split this into more steps for the sake of readability?

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -191,6 +191,42 @@ def cleanup_bootstrap_conf(self):
...
+    def delete_plugin(self, plugin_config: Dict[str, str]) -> None:
+        """Removes plugin configuration into opensearch.yml.
+
+        Returns True if all the configurations have been removed successfully.
+        """
+        if not plugin_config:
+            return False
+        return all(
+            [
+                config not in self._opensearch.config.delete(self.CONFIG_YML, config).keys()
```
dmitry-ratushnyy, line 225: This expression looks quite complex and hard to understand tbh.
Maybe it is better to split this into more steps for the sake of readability?

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -310,12 +310,15 @@ def write_file(self, path: str, data: str, override: bool = True):
             f.write(data)
 
     @staticmethod
-    def _run_cmd(command: str, args: str = None):
+    def _run_cmd(command: str, args: str = None, stdin: str = None) -> str:
```
dmitry-ratushnyy, line 313: It is not very clear why do we need "stdin" on top of the args

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+        except OpenSearchCmdError as e:
+            raise OpenSearchKeystoreError(str(e))
+
+    def _add(self, key: str, value: str):
+        if not value:
+            raise OpenSearchKeystoreError("Missing keystore value")
+        try:
+            # Add newline to the end of the key, if missing
+            v = value + ("" if value[-1] == "\n" else "\n")
```
dmitry-ratushnyy, line 75: nits: 
How about using "endswith" instead of indexing (readability purpose) ? 
something like "v = value + ("" if value.endswith("\n") else "\n")" inste

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        self._opensearch_config = charm.opensearch_config
+        self._charm_config = self._charm.model.config
+        self._plugins_path = plugins_path
+        self._keystore = charm.opensearch_keystore
+
+    @property
+    def plugins(self) -> List[OpenSearchPlugin]:
+        """Returns List of installed plugins."""
+        return [
+            plugin_data["class"](self._plugins_path, extra_config=self._extra_conf(plugin_data))
```
dmitry-ratushnyy, line 64: This expression looks quite complex and hard to understand tbh.
Maybe it is better to split this into more steps for the sake of readability?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def _install_if_needed(self, plugin: OpenSearchPlugin) -> bool:
+        """Installs all the plugins enabled via the config/relation.
+
+        Check if plugin in status: PluginState.MISSING and config/relation is set.
+        Returns True if a restart is needed.
+        """
+        # installed is needed because we may have a plugin already installed
+        # i.e. we fail the install command but we do not need to do anything else.
+        installed = False
```
dmitry-ratushnyy, line 101: nits:
looks like we can simply return "True"  in line 123 and False in line 124

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,232 @@
...
+        installed_plugins = self._installed_plugins()
+
+        if self.status(plugin) == PluginState.MISSING and self._user_requested_to_enable(plugin):
+            # Check for dependencies
+            missing_deps = [dep for dep in plugin.dependencies if dep not in installed_plugins]
```
dmitry-ratushnyy, line 106: nits: this looks like filter call
`missing_deps = list(filter(lambda dep: dep not in installed_plugins, plugin.dependencies))`

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -158,12 +158,12 @@ def is_node_up(self) -> bool:
         except (OpenSearchHttpError, Exception):
             return False
 
-    def run_bin(self, bin_script_name: str, args: str = None):
+    def run_bin(self, bin_script_name: str, args: str = None, stdin: str = None) -> str:
```
dmitry-ratushnyy, line 161: It is not very clear why do we need "stdin" on top of the args

lib/charms/opensearch/v0/opensearch_distro.py
phvalguima: Improved the docstring, check it out

lib/charms/opensearch/v0/opensearch_distro.py
phvalguima: Improved docstring

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,342 @@
...
+
+    An action can be: config() or disable().
+    Each returns a namedtuple containing the configurations and secrets to be added and
+    removed in that operation.
+    """
+
+    config_entries_to_add: Dict[str, str] = {}
+    config_entries_to_del: Dict[str, str] = {}
+    secret_entries_to_add: Dict[str, str] = {}
+    secret_entries_to_del: Dict[str, str] = {}
```
Mehdi-Bendriss, line 271: We should perhaps use a regular class here
```suggestion
class OpenSearchPluginConfig:
    """Represent the configuration of a plugin to be applied when configuring or disabling it."""
    
    def __init__(
        config_entries_to_add: Optional[Dict[str, str]] = None,
        config_entries_to_del: Optional[Dict[str, str]] = None,
        secret_entries_to_add: Optional[Dict[str, str]] = None,
        secret_entries_to_del: Optional[Dict[str, str]] = None,
    )
        self.config_entries_to_add = config_entries_to_add or {}
        self.config_entries_to_del = config_entries_to_del or {}
        self.secret_entries_to_add = secret_entries_to_add or {}
        self.secret_entries_to_del = secret_entries_to_del or {}
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,342 @@
...
+
+    PLUGIN_PROPERTIES = "plugin-descriptor.properties"
+
+    def __init__(self, plugins_path: str, extra_config: Dict[str, Any] = None):
+        """Creates the OpenSearchPlugin object.
+
+        Arguments:
+          plugins_path: str, path to the plugins folder
+          extra_config: dict, contains the data coming from either the config option
+                              or the relation bag
```
Mehdi-Bendriss, line 285:
```suggestion
          extra_config: dict, contains config entries coming from optional relation data
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,342 @@
...
+            FileNotFoundError: if plugin file is not present
+            PermissionError: if plugin file is present, but not set with correct permissions
+        """
+        properties = Properties()
+        with open(self._plugins_path) as f:
+            properties.load(f.read())
+        return properties._properties["version"]
+
+    @property
+    @abstractmethod
```
Mehdi-Bendriss, line 305:
```suggestion
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -0,0 +1,342 @@
...
+            config_entries_to_add = {...},
+            config_entries_to_del = {...},
+            secret_entries_to_add = {...},
+            secret_entries_to_del = {...},
+        )
+        """
+        pass
+
+    @property
+    @abstractmethod
```
Mehdi-Bendriss, line 339:
```suggestion
    @abstractproperty
```

tox.ini
```diff
@@ -62,6 +62,7 @@ description = Run unit tests
 pass_env = 
     LIBJUJU_VERSION_SPECIFIER
 deps =
+    juju{env:LIBJUJU_VERSION_SPECIFIER:==3.1.2.0}
```
Mehdi-Bendriss, line 65: I believe this should be removed after merging from main

Mehdi-Bendriss (commented):
Thank you so much! This is the final pass on this PR :) I left minor comments.

As a side note, I created this ticket for a future PR to add support for plugins enabled through the REST API instead of the `opensearch.yml`

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -338,9 +348,10 @@ def _run_cmd(command: str, args: str = None):
 
             if output.returncode != 0:
                 logger.error(f"{command}:\n Stderr: {output.stderr}\n Stdout: {output.stdout}")
-                raise OpenSearchCmdError()
-        except (TimeoutError, subprocess.TimeoutExpired):
-            raise OpenSearchCmdError()
+                raise OpenSearchCmdError(output.stderr)
```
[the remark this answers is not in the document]
phvalguima: Removing that indentation will trigger this error in all cases, that is not what we want.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -0,0 +1,258 @@
...
+        for key, val in stored_plugin_conf.items():
+            if are_configs_enabled.get(key) != val:
+                return False
+        return True
+
+    def _needs_upgrade(self, plugin: OpenSearchPlugin) -> bool:
+        """Returns true if plugin needs upgrade."""
+        current_version = plugin.version
+        num_points = len(self._opensearch.version.split("."))
+        return self._opensearch.version != current_version[:num_points]
```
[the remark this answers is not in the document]
phvalguima: Plugin version format is sometimes in: x.y.z.A, for example: "2.9.0" may have plugins with version="2.9.0.0". We need this code to correct that

Mehdi-Bendriss (approved):
Thank you!

dmitry-ratushnyy (approved):

PR #113, opened 2023-08-30
[DPE-2421] Change replication factor of new built-in indices

Mehdi-Bendriss:
## Issue
OpenSearch 2.9.0 introduced a "bug" or at least strange behavior where some indices have fixed numbers of replicas, this prevents the cluster from ever being green in a 1 node setup.

This PR implements [DPE-2421](https://warthogs.atlassian.net/browse/DPE-2421) - namely, this PR implements:
- reconfiguration of the replication factor of those indices

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_fixes.py
+++ b/lib/charms/opensearch/v0/opensearch_fixes.py
@@ -1,0 +1,46 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Base class for the OpenSearch Fixes of bugs introduced by upstream."""
+from charms.opensearch.v0.opensearch_exceptions import OpenSearchHttpError
+
+# The unique Charmhub library identifier, never change it
+LIBID = "3bdf0a053a53493abefe8265dac85419"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+
+class OpenSearchFixes:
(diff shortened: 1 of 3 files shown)


lib/charms/opensearch/v0/opensearch_fixes.py
```diff
@@ -0,0 +1,46 @@
...
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+
+class OpenSearchFixes:
+    """This class implements various fixes of bugs introduced in OpenSearch upstream."""
+
+    def __init__(self, charm):
+        self._charm = charm
```
deusebio, line 22: **suggestion** I would try to avoid doing dependency injection with the charm itself, but rather I would create a dependency tree with components. I see that below you only use the `self._charm.opensearch` object, so maybe here you could just provide this component as input there in place of charms. 

The reason why I'm saying this, is that using charm as an input creates a circular dependency (in python we don't see it as it is not compiled but if for instance you were to use typing to annotate the charm attribute, that would not be possible), since Charm definition needs OpenSearchFixes, that needs (since it is its argument) the OpenSearch charm (otherwise, how it could know that exists an `opensearch` attribute)

deusebio (approved):
LGTM

Just a suggestion attached

juditnovak (approved):
WONDERFUL, green pipelines again!!!!!!!!!!!!!!!!!!!!!! :-) :-) :-)

lib/charms/opensearch/v0/opensearch_fixes.py
Mehdi-Bendriss: Good point, thank you! I will revisit all the injections I'm doing across all the components

Issue #119, opened 2023-09-11
OpenSearch Snap depends on snapd 2.60+


phvalguima:
Commit: https://github.com/canonical/opensearch-snap/commit/e9964ef7b8c41df8bdd193bc126b3ff3a2a21065

In opensearch-snap added a dependency to snapd 2.60, otherwise installing opensearch snap fails with permission denied to ```/dev/shm/performance-analyzer```.
However, not all deployments will start at this version and should be refreshed before installation can proceed.

We should add something like:
```

    def __init__(self, charm, peer_relation: str):
        super().__init__(charm, peer_relation)

        for attempt in Retrying(stop=stop_after_attempt(5), wait=wait_fixed(wait=5)):
            with attempt:
                cache = snap.SnapCache()
                self._opensearch = cache["opensearch"]
+              self._snapd = cache["snapd"]


    @override
    def install(self):
        """Install opensearch from the snapcraft store."""
        if self._opensearch.present:
            return

        try:
+          self._snapd.ensure(snap.SnapState.Latest, channel="latest/stable")
            self._opensearch.ensure(snap.SnapState.Latest, channel="edge")
            self._opensearch.connect("process-control")
+          self._opensearch.connect("shmem-perf-analyzer")
        except SnapError as e:
            logger.error(f"Failed to install opensearch. \n{e}")
            raise OpenSearchInstallError()
```

PR #120, opened 2023-09-11
[DPE-2589] OpenSearchSnap refreshes snapd before installation

phvalguima:
Commit: https://github.com/canonical/opensearch-snap/commit/e9964ef7b8c41df8bdd193bc126b3ff3a2a21065

In opensearch-snap added a dependency to snapd 2.60, otherwise installing opensearch snap fails with permission denied to /dev/shm/performance-analyzer.
However, not all deployments will start at this version and should be refreshed before installation can proceed.

Resolves issue: https://github.com/canonical/opensearch-operator/issues/119

### DIFF
--- a/src/opensearch.py
+++ b/src/opensearch.py
@@ -65,7 +65,6 @@ class OpenSearchSnap(OpenSearchDistribution):
             self._snapd.ensure(snap.SnapState.Latest, channel="latest/stable")
             self._opensearch.ensure(snap.SnapState.Latest, channel="edge")
             self._opensearch.connect("process-control")
-            self._opensearch.connect("shmem-perf-analyzer")
         except SnapError as e:
             logger.error(f"Failed to install opensearch. \n{e}")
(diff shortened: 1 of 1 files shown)


src/opensearch.py
```diff
@@ -61,8 +62,10 @@ def install(self):
             return
 
         try:
+            self._snapd.ensure(snap.SnapState.Latest, channel="latest/stable")
             self._opensearch.ensure(snap.SnapState.Latest, channel="edge")
             self._opensearch.connect("process-control")
+            self._opensearch.connect("shmem-perf-analyzer")
```
Mehdi-Bendriss, line 68: this can be removed, this interface is `auto-connect`able by default

Mehdi-Bendriss (approved):

src/opensearch.py
phvalguima: Removed in a new commit.

PR #129, opened 2023-10-01
Update integration tests

phvalguima:


phvalguima:
Some tests are failing b/c OpenSearch returns "503" instead of a string: https://pastebin.ubuntu.com/p/9JPccby7yQ/
Also, seems snapd is not always responding in time: https://pastebin.ubuntu.com/p/VPBj6s95hm/

Fixing the first with an ```isinstance``` check and returning "HealthColor.UNKNOWN" if not a string; the second problem, added debug options on snapd.

PR #130, opened 2023-10-03
Try fix integration tests

phvalguima:


phvalguima:
Some tests are failing b/c OpenSearch returns "503" instead of a string: https://pastebin.ubuntu.com/p/9JPccby7yQ/
Also, seems snapd is not always responding in time: https://pastebin.ubuntu.com/p/VPBj6s95hm/

Fixing the first with an isinstance check and returning "HealthColor.UNKNOWN" if not a string; the second problem, added debug options on snapd.

phvalguima:
`run_script` needs a longer timeout. `h-scaling-integration` fails with: https://pastebin.canonical.com/p/XdFK9CZMC2/
At least 60s

lib/charms/opensearch/v0/opensearch_health.py
```diff
@@ -118,37 +119,38 @@ def apply_for_unit(self, status: str, host: Optional[str] = None):
...
+                    shards_by_state = ClusterState.shards_by_state(
+                        self._opensearch, host=host, alt_hosts=self._charm.alt_hosts
+                    )
+                    busy_shards = shards_by_state.get("INITIALIZING", 0) + shards_by_state.get(
+                        "RELOCATING", 0
+                    )
+                    return HealthColors.YELLOW_TEMP if busy_shards > 0 else HealthColors.YELLOW
+
+        except RetryError:
+            return HealthColors.UNKNOWN
```
phvalguima, line 156: Proposed fix for: https://github.com/canonical/opensearch-operator/pull/130#issuecomment-1744451875

phvalguima:
The error caused by snapd time out: https://pastebin.ubuntu.com/p/rrpMpRXWVY/
Is not really resolved. However, the new method `wait_for_all_units_active` will simply ignore the "error" state, as it does with any other non-active status. Therefore, it gives the unit a fair chance to retry.

We should consider moving to `wait_for_all_units_active` everywhere, and implementing an equivalent method to wait for application, using it whenever we also want the App level to be `active`.

tests/integration/ha/test_ha.py
```diff
@@ -680,11 +680,11 @@ async def test_multi_clusters_db_isolation(
     await ops_test.model.wait_for_idle(
         apps=[app],
         status="active",
-        timeout=1000,
+        timeout=3600,
```
juditnovak, line 683: Silly small thing but don't we wanna have a constant for this? :-)
(Feel free to ignore :-) )

Mehdi-Bendriss:
Thanks! I took some items from this PR and the remaining of the work has been centralized in https://github.com/canonical/opensearch-operator/pull/134

PR #125, opened 2023-09-21
[DPE-2327] add knn support

phvalguima:
Modelling the K-NN plugin support.

This PR implements [DPE-2327](https://warthogs.atlassian.net/browse/DPE-2327) with K-NN support. It adds a new config option: plugin_opensearch_knn that allows to enable and disable the K-NN plugin.

This PR is based on https://github.com/canonical/opensearch-operator/pull/105

[DPE-2327]: https://warthogs.atlassian.net/browse/DPE-2327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #132, opened 2023-10-05
[DPE-2385] Extends ML Plugins to cover MLCommons

phvalguima:
Modelling the ML Commons plugin support.

This PR implements [DPE-2385](https://warthogs.atlassian.net/browse/DPE-2385) with ML Commons. It adds a new config option: plugin_opensearch_mlcommons that allows to enable and disable the ml commons plugin.

It also implements integration test covering this new option [DPE-2383](https://warthogs.atlassian.net/browse/DPE-2383).

This PR is based on https://github.com/canonical/opensearch-operator/pull/105 and KNN PR: https://github.com/canonical/opensearch-operator/pull/125

[DPE-2385]: https://warthogs.atlassian.net/browse/DPE-2385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-2383]: https://warthogs.atlassian.net/browse/DPE-2383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -350,7 +350,78 @@ def disable(self) -> OpenSearchPluginConfig:
             config_entries_to_add={"knn.plugin.enabled": False},
         )
 
+    @property
+    def dependencies(self) -> Optional[List[str]]:
+        """Returns a list of plugin name dependencies."""
+        return ["opensearch-ml"]
```
Mehdi-Bendriss, line 356: Could you confirm that this is needed? a very brief research suggested they're separate

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -350,7 +350,78 @@ def disable(self) -> OpenSearchPluginConfig:
...
+
+
+class OpenSearchPluginMLCommons(OpenSearchPlugin):
+    """Implements the opensearch-ml-commons plugin."""
+
+    def config(self) -> OpenSearchPluginConfig:
+        """Returns a plugin config object to be applied for enabling the current plugin."""
+        return OpenSearchPluginConfig(
+            config_entries_to_add={
+                "plugins.ml_commons.only_run_on_ml_node": True,
```
Mehdi-Bendriss, line 371: all these seem to be populating the default values of the plugin, it should be safe to remove

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -350,7 +350,78 @@ def disable(self) -> OpenSearchPluginConfig:
...
+                "plugins.ml_commons.native_memory_threshold": 90,
+                "plugins.ml_commons.allow_custom_deployment_plan": False,
+                "plugins.ml_commons.model_auto_redeploy.lifetime_retry_times": 3,
+                "plugins.ml_commons.model_auto_redeploy_success_ratio": 0.8,
+                "plugins.ml_commons.enable_inhouse_python_model": False,
+                # Enablement settings
+                "plugins.ml_commons.remote_inference.enabled": True,
+                "plugins.ml_commons.model_access_control_enabled": False,
+                "plugins.ml_commons.connector_access_control_enabled": False,
+                "plugins.ml_commons.model_auto_redeploy.enable": False,
```
Mehdi-Bendriss, line 391: Since these are set to False which is their default values, safe to delete

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -350,7 +350,78 @@ def disable(self) -> OpenSearchPluginConfig:
...
+        )
+
+    def disable(self) -> OpenSearchPluginConfig:
+        """Returns a plugin config object to be applied for disabling the current plugin."""
+        return OpenSearchPluginConfig(
+            config_entries_to_add={
+                "plugins.ml_commons.remote_inference.enabled": False,
+                "plugins.ml_commons.model_access_control_enabled": False,
+                "plugins.ml_commons.connector_access_control_enabled": False,
+                "plugins.ml_commons.model_auto_redeploy.enable": False,
```
Mehdi-Bendriss, line 402: we can delete these, if they were initially set to `True` we could have just list them as part of `config_entries_to_del` to set them back to their default value (False)

tests/integration/plugins/helpers.py
```diff
@@ -170,3 +170,74 @@ async def create_index_and_bulk_insert(
...
+
+    resp = await http_request(ops_test, "POST", endpoint, payload=node_ids, app=app)
+    return resp
+
+
+@retry(
+    wait=wait_fixed(wait=5),
+    stop=stop_after_attempt(5),
+)
+async def mlcommons_predict_model(
```
Mehdi-Bendriss, line 231:
```suggestion
async def mlcommons_model_predict(
```

tests/integration/plugins/helpers.py
```diff
@@ -170,3 +170,74 @@ async def create_index_and_bulk_insert(
...
+    ops_test: OpsTest,
+    app: str,
+    unit_ip: str,
+    model_id: str,
+    prediction_type: str = "text_embedding",
+    text_docs: Dict[str, Any] = {},
+) -> str:
+    """Waits to load model and returns model_id or None."""
```
Mehdi-Bendriss, line 239: Copy / past error :)

tests/integration/plugins/test_plugins.py
```diff
@@ -236,3 +240,74 @@ async def test_knn_training_search(ops_test: OpsTest) -> None:
         except RetryError:
             # The search should fail if knn_enabled is false
             assert not knn_enabled
+
+
+@pytest.mark.abort_on_fail
+async def test_mlcommons_upload_and_predict_model(ops_test: OpsTest) -> None:
```
Mehdi-Bendriss, line 246:
```suggestion
async def test_mlcommons_model_upload_and_prediction(ops_test: OpsTest) -> None:
```

tests/integration/plugins/test_plugins.py
```diff
@@ -236,3 +240,74 @@ async def test_knn_training_search(ops_test: OpsTest) -> None:
...
+    """Uploads and runs the model."""
+
+    async def __wait_model_task(task_id: str):
```
Mehdi-Bendriss, line 249: There seems to be something wrong in this function - should the asserts be under the `for`? what does this function do?

tests/integration/plugins/test_plugins.py
```diff
@@ -236,3 +240,74 @@ async def test_knn_training_search(ops_test: OpsTest) -> None:
...
+            leader_unit_ip,
+            model_config={
+                "name": "huggingface/sentence-transformers/all-MiniLM-L12-v2",
+                "version": "1.0.1",
+                "model_format": "TORCH_SCRIPT",
+            },
+        )
+    ).get("task_id", None)
+
+    assert task_id is not None
```
Mehdi-Bendriss, line 272: Could you add an explanatory message in all your asserts? (including the ones after this block on code)
```suggestion
    assert task_id is not None, "The task_id is None"
```

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: Actually, confirmed as you've said, that is not needed. The `train` API is actually different

.github/workflows/integration.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
             runner: "Ubuntu_4C_16G"
     name: ${{ matrix.tox-environments }}
     runs-on: ${{ matrix.runner }}
-    timeout-minutes: 360
+    timeout-minutes: 60
```
Mehdi-Bendriss, line 42: This might be too short depending on the test: `h-scaling, ha-integration and ha-networking-integration` often exceed that , I would prefer we keep as is for the time being.

Mehdi-Bendriss (approved):

tests/integration/plugins/helpers.py
```diff
@@ -82,9 +82,12 @@ def generate_bulk_training_data(
     result_list = []
     for i in range(docs_count):
         result += json.dumps({"index": {"_index": index_name, "_id": i}}) + "\n"
+        # Breaks the single-dimension (ndim*docs_count samples) data vector above into
+        # a matrix with (ndim, docs_count), where each vector has len() == ndim
         result_list.append([float(data[j]) for j in range(i * dimensions, (i + 1) * dimensions)])
         inter = {vector_name: result_list[i]}
         if has_result:
+            # Naming "result", 1xdocs_count dimension vector, as "price"
             inter["price"] = float(responses[i])
```
dmitry-ratushnyy, line 91: It is not in the PR, but
just curious why there is a cast to float right before call to json.dums?

tests/integration/plugins/helpers.py
```diff
@@ -170,3 +173,78 @@ async def create_index_and_bulk_insert(
...
+    stop=stop_after_attempt(7),
+)
+async def mlcommons_register_model(
+    ops_test: OpsTest,
+    app: str,
+    unit_ip: str,
+    model_config: Dict[str, Any],
+) -> Optional[List[Dict[str, Any]]]:
+    """Uploads models."""
+    endpoint = f"https://{unit_ip}:9200/_plugins/_ml/models/_register"
```
dmitry-ratushnyy, line 189: nits: port number is asking to be in some constant

tests/integration/plugins/helpers.py
```diff
@@ -170,3 +173,78 @@ async def create_index_and_bulk_insert(
...
+    stop=stop_after_attempt(7),
+)
+async def mlcommons_wait_task_model(
+    ops_test: OpsTest,
+    app: str,
+    unit_ip: str,
+    task_id: str,
+) -> Optional[str]:
+    """Waits to load model and returns model_id or None."""
+    endpoint = f"https://{unit_ip}:9200/_plugins/_ml/tasks/{task_id}"
```
dmitry-ratushnyy, line 207: maybe we can have a function or a property what will generate a base endpoint and will be shared between the functions something like

@property 
def endpoint_base_url():
return f"https://{unit_ip}:{PORT}/_plugins/_ml/"

tests/integration/plugins/helpers.py
phvalguima: These should be float values. I am casting it to float to ensure I have "0" and not "0.0"

dmitry-ratushnyy: so you need floats everywhere, except 0?

dmitry-ratushnyy (approved):

Mehdi-Bendriss (approved):

PR #134, opened 2023-10-11
[DPE-2658] - Integration tests stabilization

Mehdi-Bendriss:
## Issue
Some integration tests such as the client integration seem to consistently fail, this PR implements [DPE-2658](https://warthogs.atlassian.net/browse/DPE-2658), namely this PR implements:
- Tests side:
    - alternative wait for idle on all tests with more granularity
    - breakdown of `ha-tests` into 2 files including `ha-multi-clusters-integration`
- charms side:
    - cluster health handling of erroneous response payload
    - upgrade of charm libs - with regards to snaps
    - removal of snapd refresh that causes the install to hang sometimes
    - Handling of plugin operation error `on_config_changed`
    - Introduction custom event for forcing re-trigger of a deferral:
         - removal of deferral strategy using `_status_update`
         - replace it with `DeferTriggerEvent.emit()`
    - Avoiding setting the status of a unit or an application UNLESS the current status is not equal to it, to avoid updating the idle flag of the model
    - Retry of lock acquiring in scale-down 
    - Upgrade of urllib to version 2

[DPE-2658]: https://warthogs.atlassian.net/browse/DPE-2658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


juditnovak (approved):
WOOOOOOOOW

phvalguima:
Hi @Mehdi-Bendriss, the client-integration ran for a long time, but it seems it faced a timeout error in one of its tests: https://pastebin.ubuntu.com/p/QWSzFcgqPS/

Right before, there was also two other lines in the same unit:
```
2023-10-16 18:22:24 INFO unit.opensearch/1.juju-log server.go:316 opensearch-client:5: Health: {'error': {'root_cause': [{'type': 'security_exception', 'reason': 'OpenSearch Security not initialized for cluster:monitor/health'}], 'type': 'security_exception', 'reason': 'OpenSearch Security not initialized for cluster:monitor/health'}, 'status': 503}
2023-10-16 18:22:24 ERROR unit.opensearch/1.juju-log server.go:316 opensearch-client:5: 'int' object has no attribute 'lower'
```
Full juju status of that failure: https://pastebin.ubuntu.com/p/kyv7mCv8TQ/

Now, it should not be a problem the amount of workloads you have there. So, the erroring unit is the reason for the test to fail.

I had a proposal in this PR: https://github.com/canonical/opensearch-operator/pull/130#issuecomment-1744451875

Mehdi-Bendriss:
Thanks @phvalguima ! the `health` case you described should be handled [here](https://github.com/canonical/opensearch-operator/pull/134/files#diff-29386b01645a55e8571c903122c13345e969c7b15d67aa82169e1d95dee82e43R144-R149)

delgod (approved):

PR #137, opened 2023-10-16
[DPE-2345][DPE-2451] restore action

phvalguima:
Extends https://github.com/canonical/opensearch-operator/pull/135 and adds restore support.

This PR will roll the cluster's indices back to the snapshot and override any conflicting index.

It tests two scenarios:
1) Snapshot / restore in the same cluster
2) Snapshot / redeploy cluster / restore

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_backups.py
+++ b/lib/charms/opensearch/v0/opensearch_backups.py
@@ -46,7 +46,7 @@ class OpenSearchBaseCharm(CharmBase):
 
 import json
 import logging
-from typing import Any, Dict, Set, Tuple
+from typing import Any, Dict, List, Set, Tuple
 
 import requests
 from charms.data_platform_libs.v0.s3 import S3Requirer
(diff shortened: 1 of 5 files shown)


Mehdi-Bendriss (commented):
Will review this PR once feedback is addressed on the create / list backups PR. As they're closely intertwined.

phvalguima:
Following up https://github.com/canonical/opensearch-operator/pull/135 and addressing @delgod last few comments there:
* plugin_opensearch_knn: why switching from false to true
* to have a similar interface with postgresql charm and change the `_fetch_snapshot` naming convention
* remove restrictions of backup / restore only on leader unit
* to discover the HTTP/HTTPS protocol from the `endpoint` provided
* `region` should not be mandatory in the charm
* get the `storage_class` information from s3-integrator and configure it, if available
* Change:
```
-                BlockedStatus(f"Disabling backup not possible: {snapshot_status}")
+                MaintanceStatus(f"Disabling backup postponed till backup in progress: {snapshot_status}")
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -143,14 +147,15 @@ def __init__(self, charm: Object):
...
 
     @property
     def _plugin_status(self):
         return self.charm.plugin_manager.get_plugin_status(OpenSearchBackupPlugin)
 
     def _on_list_backups_action(self, event: ActionEvent) -> None:
         """Returns the list of available backups to the user."""
-        if not self._check_action_can_run(event):
+        if not self._can_unit_perform_backup(event):
             event.fail("Failed: backup service is not configured yet")
```
delgod, line 159: still UX should be improved. e.g. id should be a time of backup start, it should be a table with in-progress/failed/successfull status, etc
```
$ juju run postgresql/3 list-backups
...
backups: |-
  backup-id             | backup-type  | backup-status
  ----------------------------------------------------
  2023-10-19T11:36:23Z  | physical     | finished
```

delgod (changes requested):

juditnovak (approved):
LGTM (on my VERY limited level of understanding)

juditnovak (approved):
LGTM

PR #138, opened 2023-10-17
Fixed error handling in client integration

Mehdi-Bendriss:
## Issue
This PR fixes bugs causing intermittent failures on CI:
- some missing exception handling on the client relations - namely non caught http errors
- referring to a non existing variable on the `e.status_code`
- usage of status.set from the user relations, to avoid overriding the model last active since status
- increase timeout to 60s (from 25s) on `_run_cmd` 
- added retry on `_run_cmd`


juditnovak (approved):
Seems wonderful and quite nasty 2 track down these hidden corners...

PR #141, opened 2023-10-19
Fixed jdk path

Mehdi-Bendriss:
## Issue
The new version of the snap relies on the jdk of the snap environment instead of the one bundled by OpenSearch.
This PR replaces the path of the JDK by the correct one


juditnovak (approved):
LGTM

delgod (approved):

PR #143, opened 2023-11-14
Removed juju 2 from CI pipelines

Mehdi-Bendriss:
## Issue
This PR removes Juju 2 support on CI and have all tests ran on juju 3.


tests/integration/ha/test_large_deployments.py
```diff
@@ -147,7 +147,12 @@ async def test_set_roles_manually(
     await wait_until(
         ops_test,
         apps=[app],
-        apps_statuses=["active"],
+        apps_full_statuses={
+            app: {
+                "blocked": [PClusterWrongNodesCountForQuorum],
+                "active": [],
+            },
+        },
```
juditnovak, line 155: What is this one about...?

juditnovak (approved):

delgod (approved):

tests/integration/ha/test_large_deployments.py
Mehdi-Bendriss: This is to fix an Integration test for large deployments where the unit or app can be put in a blocked state because a quorum cannot be reached. Previously, only the case where the unit was put in a blocked state was handled - this is now adding the second case

PR #145, opened 2023-11-24
[DPE-2935] Integrating Prometheus Exporter

juditnovak:


tests/integration/helpers_deployments.py
```diff
@@ -321,37 +322,35 @@ async def wait_until(  # noqa: C901
...
+    logger.info("\n\n\n")
+    logger.info(
+        subprocess.check_output(
+            f"juju status --model {ops_test.model.info.name}", shell=True
+        ).decode("utf-8")
+    )
+
+    start = datetime.now()
+    while start + timedelta(seconds=timeout) > datetime.now():
+        sleep(10)
```
juditnovak, line 334: I have no explanation for this.

With `Retry()` this loop breaks. I've spent AGES on this thing.
Works on `main`, breaks here. No exception. Same tenacity version.
NOT related to the change on line 26.
I don't understand. After one loop the execution simply stops with `Retry()` 
Works perfectly with the "simulation" of the tenacity functionality.
If any mis-behavior on charm side: this dead simple execution flow surely should let it escalate. (So far nothing observed, I got green pipelines again.)

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -96,35 +112,28 @@ def run(self) -> bool:
                     self._install_if_needed(plugin),
                     self._configure_if_needed(plugin),
                     self._disable_if_needed(plugin),
+                    self._remove_if_needed(plugin),
```
juditnovak, line 115: Since `disable()` above  does nothing else but deleting configuration... Ideally it should come after the plugin is removed.

However since we're removing the plugin
 1. when there's no more COS relation (i.e. no conusmer)
 2. right before we are actually removing the plugin
This should be safe enough (to what I understand).

In a worst case scenario we may get complaints in the logs.

I actually tried to go for a "post"-cleanup procedure... But with the current plugin lifecycle and status determination logic, it was virtually impossible to fit it into the scheme.

juditnovak: Just for clarity: Of course not the `cleanup()` function itself. But to be able to reach that stage of the execution ;-)
(As in this case there are no grounds to get into a `DISABLED` state...)

phvalguima:
Hi @juditnovak, I have found a way to manage the removal of a given plugin, without neither removing the method `_remove_plugin_if_needed` from the `run()` method in plugin_manager, nor having to extend `RelationBrokenEvent` into plugin_manager.

juditnovak:
@phvalguima Sure, changes from your diff applied as a separate commit

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -354,3 +356,23 @@ def disable(self) -> OpenSearchPluginConfig:
...
+        return "prometheus-exporter"
+
+    def config(self) -> OpenSearchPluginConfig:
+        """Returns a plugin config object to be applied for enabling the current plugin."""
+        return OpenSearchPluginConfig(
+            config_entries_to_add={
+                "prometheus.metric_name.prefix": "opensearch_",
+                "prometheus.indices": False,
+                "prometheus.cluster.settings": False,
+                "prometheus.nodes.filter": "_all",
```
delgod, line 376: COS should collect metrics from each host individually, so only local metrics should be included.
```suggestion
                "prometheus.nodes.filter": "_local",
```

lib/charms/opensearch/v0/opensearch_cos.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2023 Canonical Ltd.
```
delgod, line 1: I am not sure it functionality needed

delgod (changes requested):

lib/charms/opensearch/v0/opensearch_plugins.py
juditnovak: Moved to snap: https://github.com/canonical/opensearch-snap/pull/47

lib/charms/opensearch/v0/opensearch_cos.py
juditnovak: Removed.

delgod (approved):

phvalguima (approved):
LGTM

PR #148, opened 2023-11-29
[POC] Checking main pipeine status

juditnovak:


juditnovak:
Attempt successful, it's `main` is good

PR #151, opened 2023-11-30
[DPE-3068[ New user for COS monitoring

juditnovak:
[OpenSearch Prometheus Exporter](https://github.com/Aiven-Open/prometheus-exporter-plugin-for-opensearch) requires auhentication when security pluign is switched on. More info: https://github.com/Aiven-Open/prometheus-exporter-plugin-for-opensearch/issues/60

Thus we're adding this new user.


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -501,30 +505,31 @@ def _on_set_password_action(self, event: ActionEvent):
...
             return
 
         password = event.params.get("password") or generate_password()
         try:
+            label = self.secrets.password_key(user_name)
             self._put_admin_user(password)
-            password = self.secrets.get(Scope.APP, f"{user_name}-password")
-            event.set_results({f"{user_name}-password": password})
+            password = self.secrets.get(Scope.APP, label)
+            event.set_results({label: password})
```
delgod, line 517: please don't broke UX here

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -501,30 +505,31 @@ def _on_set_password_action(self, event: ActionEvent):
...
         user_name = event.params.get("username")
-        if user_name != "admin":
-            event.fail("Only the 'admin' username is allowed for this action.")
+        if user_name not in ["admin", COSUser]:
+            event.fail(f"Only the 'admin' and {COSUser} username is allowed for this action.")
             return
 
         password = event.params.get("password") or generate_password()
         try:
+            label = self.secrets.password_key(user_name)
```
delgod, line 514: I think one secret should store passwords for all internal users

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1041,6 +1066,19 @@ def _check_certs_expiration(self, event: UpdateStatusEvent) -> None:
...
         )
 
+    def _scrape_config(self) -> List[Dict]:
+        """Generates the scrape config as needed."""
+        pwd = self.secrets.get(Scope.APP, self.secrets.password_key(COSUser))
+        return [
+            {
+                "metrics_path": "/_prometheus/metrics",
+                "static_configs": [{"targets": [f"{self.unit_ip}:{COSPort}"]}],
+                "tls_config": {"insecure_skip_verify": True},
```
delgod, line 1076: is it possible to send a CA instead of skip CA check?

delgod (changes requested):

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Hi @delgod, that is used between OpenSearch and cos-agent, both running in the same node. In any case, we will have to work on it in two sides: (1) to check the COS relation, if we receive we can define a certificate via relation; and (2) add that cert to the JVM truststore.

I am putting an Epic to revisit our certificate / passwords story in OpenSearch, I'd advise we consider double-checking COS as part of that Epic, wdyt?
I've created an issue: DPE-3141, to look into that more specifically in the coming weeks.

juditnovak: We have changed this to:

```
"tls_config": {"ca": <Opensearch CA CERT>}
```

delgod (dismissed):

phvalguima (dismissed):
LGTM

phvalguima (approved):

delgod (approved):

PR #153, opened 2023-12-08
[DPE-3136][DPE-3137] Migrate to poetry dp workflows and secrets

phvalguima:
This PR is built on top of backup as it will expand backup testing to cover AWS and GCP, using the secrets management mechanisms in data-platform-workflows.

It should be seen as a follow-up from https://github.com/canonical/opensearch-operator/pull/135 and https://github.com/canonical/opensearch-operator/pull/137

Following changes will also be added:
1) Add test coverage for AWS+GCP
2) List backup action should print `backup-type` as either `full` or `partial` instead of `physical`
3) Remove workaround: https://github.com/canonical/opensearch-operator/blob/main/requirements.txt#L18


phvalguima:
Closing this PR in favor of #157.

PR #158, opened 2024-01-09
[DPE-3243][DPE-3246] Finishes the backup with fully async restore

phvalguima:
[DPE-3243](https://warthogs.atlassian.net/browse/DPE-3243): list-backups now contains only relevant information, in this case the backup-id and its status; the PR removes the "backup-type" column, as it has no meaning in OpenSearch

[DPE-3246](https://warthogs.atlassian.net/browse/DPE-3246): backups will be async and restores will be synchronous. Users can follow the restore in the juju status. In any case, the indices that are being restored will be closed, so the user needs to wait.

[DPE-3243]: https://warthogs.atlassian.net/browse/DPE-3243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-3246]: https://warthogs.atlassian.net/browse/DPE-3246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DPE-3192]: https://warthogs.atlassian.net/browse/DPE-3192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -72,6 +72,7 @@ HorizontalScaleUpSuggest = "Horizontal scale up advised: {} shards unassigned."
 WaitingForOtherUnitServiceOps = "Waiting for other units to complete the ops on their service."
 NewIndexRequested = "new index {index} requested"
 RestoreStarting = "Restore action started."
+PluginConfigStart = "Plugin configuration started."
 
 
 # Relation Interfaces
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -57,7 +57,6 @@ PClusterWrongRolesProvided = "Cannot start cluster with current set of roles."
 PClusterWrongNodesCountForQuorum = (
     "Even number of members in quorum if current unit started. Add or remove 1 unit."
 )
-RestoreActionFailed = "Restore action failed, please check the logs."
 
 # Wait status
 RequestUnitServiceOps = "Requesting lock on operation: {}"
@@ -71,7 +70,7 @@ TLSNewCertsRequested = "Requesting new TLS certificates..."
 HorizontalScaleUpSuggest = "Horizontal scale up advised: {} shards unassigned."
 WaitingForOtherUnitServiceOps = "Waiting for other units to complete the ops on their service."
 NewIndexRequested = "new index {index} requested"
-RestoreStarting = "Restore action started."
+RestoreInProgress = "Restore in progress..."
 PluginConfigStart = "Plugin configuration started."
 
 
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -57,6 +57,7 @@ PClusterWrongRolesProvided = "Cannot start cluster with current set of roles."
 PClusterWrongNodesCountForQuorum = (
     "Even number of members in quorum if current unit started. Add or remove 1 unit."
 )
+RestoreActionFailed = "Restore action failed, please check the logs."
 
 # Wait status
 RequestUnitServiceOps = "Requesting lock on operation: {}"
@@ -70,6 +71,7 @@ TLSNewCertsRequested = "Requesting new TLS certificates..."
 HorizontalScaleUpSuggest = "Horizontal scale up advised: {} shards unassigned."
 WaitingForOtherUnitServiceOps = "Waiting for other units to complete the ops on their service."
 NewIndexRequested = "new index {index} requested"
+RestoreStarting = "Restore action started."
 
 
 # Relation Interfaces
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -281,11 +281,14 @@ class ClusterState:
         opensearch: OpenSearchDistribution,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
+        expand_wildcards: str = None,
     ) -> List[Dict[str, str]]:
         """Get all shards of all indexes in the cluster."""
-        idx = opensearch.request("GET", "/_cat/indices", host=host, alt_hosts=alt_hosts)
+        endpoint = "/_cat/indices"
+        if expand_wildcards:
+            endpoint = f"{endpoint}?expand_wildcards={expand_wildcards}"
         idx = {}
-        for index in opensearch.request("GET", "/_cat/indices", host=host, alt_hosts=alt_hosts):
+        for index in opensearch.request("GET", endpoint, host=host, alt_hosts=alt_hosts):
             idx[index["index"]] = {"health": index["health"], "status": index["status"]}
         return idx
 
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -32,12 +32,7 @@ class IndexStateEnum(BaseStrEnum):
     CLOSED = "closed"
 
 
-class IndexHealthEnum(BaseStrEnum):
-    """Enum for index health."""
-
-    GREEN = "green"
-    YELLOW = "yellow"
-    RED = "red"
+INDEX_WILDCARD = "all"
 
 
 class ClusterTopology:
@@ -284,9 +279,7 @@ class ClusterState:
         expand_wildcards: str = None,
     ) -> List[Dict[str, str]]:
         """Get all shards of all indexes in the cluster."""
-        endpoint = "/_cat/indices"
-        if expand_wildcards:
-            endpoint = f"{endpoint}?expand_wildcards={expand_wildcards}"
+        endpoint = f"/_cat/indices?expand_wildcards={INDEX_WILDCARD}"
         idx = {}
         for index in opensearch.request("GET", endpoint, host=host, alt_hosts=alt_hosts):
             idx[index["index"]] = {"health": index["health"], "status": index["status"]}
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -6,6 +6,7 @@ import logging
 from random import choice
 from typing import Dict, List, Optional
 
+from charms.opensearch.v0.helper_enums import BaseStrEnum
 from charms.opensearch.v0.models import Node
 from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
 from tenacity import retry, stop_after_attempt, wait_exponential
@@ -22,6 +23,21 @@ LIBPATCH = 1
 
 
 logger = logging.getLogger(__name__)
+
+
+class IndexStateEnum(BaseStrEnum):
+    """Enum for index states."""
+
+    OPEN = "open"
+    CLOSED = "closed"
+
+
+class IndexHealthEnum(BaseStrEnum):
+    """Enum for index health."""
+
+    GREEN = "green"
+    YELLOW = "yellow"
+    RED = "red"
 
 
 class ClusterTopology:
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -32,9 +32,6 @@ class IndexStateEnum(BaseStrEnum):
     CLOSED = "closed"
 
 
-INDEX_WILDCARD = "all"
-
-
 class ClusterTopology:
     """Class for creating the best possible configuration for a Node."""
 
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -276,10 +276,9 @@ class ClusterState:
         opensearch: OpenSearchDistribution,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
-        expand_wildcards: str = None,
     ) -> List[Dict[str, str]]:
         """Get all shards of all indexes in the cluster."""
-        endpoint = f"/_cat/indices?expand_wildcards={INDEX_WILDCARD}"
+        endpoint = "/_cat/indices?expand_wildcards=all"
         idx = {}
         for index in opensearch.request("GET", endpoint, host=host, alt_hosts=alt_hosts):
             idx[index["index"]] = {"health": index["health"], "status": index["status"]}
--- a/lib/charms/opensearch/v0/opensearch_backups.py
+++ b/lib/charms/opensearch/v0/opensearch_backups.py
@@ -123,6 +123,14 @@ class OpenSearchRestoreCheckError(OpenSearchRestoreError):
     """Exception thrown when restore status check errors."""
 
 
+class OpenSearchRestoreFailedClosingIdxError(OpenSearchRestoreError):
+    """Exception thrown when restore fails to close indices."""
+
+
+class OpenSearchRestoreMismatchBetweenIdxAndSnapshotError(OpenSearchRestoreError):
+    """Exception thrown when restore fails to close indices."""
+
+
 class BackupServiceState(BaseStrEnum):
     """Enum for the states possible once plugin is enabled."""
 
@@ -166,6 +174,15 @@ class OpenSearchBackup(Object):
         self.framework.observe(
             self.charm.on.check_restore_status_action, self._on_check_restore_status_action
         )
+
+    def _on_update_status(self, _):
+        """Checks if the backup and/or restores are finished."""
+        status = self._check_repo_status()
+        if status == BackupServiceState.SUCCESS:
+            # There is a repo setup, check if we have any backup/restore registered in the
+            # peer relation's app databag
+            logger.info(f"Checking if backup in progress: {self.is_backup_in_progress()}")
+            logger.info(f"Checking if restore in progress: {self._check_if_restore_finished()}")
 
     @property
     def _plugin_status(self):
@@ -216,22 +233,16 @@ class OpenSearchBackup(Object):
         else:
             event.fail("Failed: invalid output format, must be either json or table")
 
-    def _restore_and_try_close_indices_if_needed(
-        self, backup_id: int, wait_for_completion: bool = False
-    ) -> Tuple[Dict[str, Any], Set[str]]:
-        """Restores and processes any exception related to running indices.
+    def _close_indices_if_needed(self, backup_indices: List[str]) -> Set[str]:
+        """Closes indices that will be restored.
 
-        Closing option is preferable, as the restore action as a whole may not succeed.
-        This method returns the final output and list of closed indices. The calling method
-        should also check the output for any other error messages, as a more generic error
-        (i.e. non-closing index error) will pass by as much as a successful restore.
+        Returns a set of indices that were closed or raises an exception:
+        - OpenSearchRestoreFailedClosingIdxError if any of the indices could not be closed.
         """
         closed_idx = set()
         try:
             for attempt in Retrying(stop=stop_after_attempt(5), wait=wait_fixed(wait=30)):
                 with attempt:
-                    # First, for overlapping indices
-                    backup_indices = self._list_backups()[backup_id]["indices"]
                     to_close = [
                         index
                         for index, state in ClusterState.indices(self.charm.opensearch).items()
@@ -246,73 +257,45 @@ class OpenSearchBackup(Object):
                         # Try closing the index
                         o = self._request("POST", f"{index}/_close")
                         logger.debug(
-                            f"_restore_and_try_close_indices_if_needed: request closing {index} returned {o}"
+                            f"_close_indices_if_needed: request closing {index} returned {o}"
                         )
+                        assert o.get("acknowledged", False)
+                        assert index in o.get("indices", {})
                         closed_idx.add(index)
-                    output = self._request(
-                        "POST",
-                        f"_snapshot/{S3_REPOSITORY}/{backup_id}/_restore?wait_for_completion={str(wait_for_completion).lower()}",
-                        payload={
-                            "indices": ",".join(
-                                [
-                                    f"-{idx}"
-                                    for idx in INDICES_TO_EXCLUDE_AT_RESTORE & set(backup_indices)
-                                ]
-                            ),
-                            "ignore_unavailable": False,
-                            "include_global_state": False,
-                            "include_aliases": True,
-                            "partial": False,
-                        },
-                    )
-                    logger.debug(
-                        f"_restore_and_try_close_indices_if_needed: restore call returned {output}"
-                    )
-                    if (
-                        self.get_service_status(output)
-                        == BackupServiceState.SNAPSHOT_RESTORE_ERROR_INDEX_NOT_CLOSED
-                    ):
-                        logger.debug(
-                            f"_restore_and_try_close_indices_if_needed retrying, output is: {output}"
-                        )
-                        # The .opensearch-sap-log-types-config, for example does not show
-                        # in the _cat/indices but does block a restore operation.
-                        to_close = output["error"]["reason"].split("[")[2].split("]")[0]
-                        o = self._request("POST", f"{index}/_close")
-                        logger.debug(
-                            f"_restore_and_try_close_indices_if_needed: request closing {index} returned {o}"
-                        )
-                        closed_idx.add(index)
-                        # We still have non closed indices, try again
-                        raise Exception()
-                    # Finally, if status returns success, ensure all indices have been updated
-                    # otherwise, raise an assert error and retry
-                    if wait_for_completion:
-                        shards = output.get("snapshot", {}).get("shards", {})
-                        assert shards.get("successful", -1) == shards.get("total", 0)
-                        break
-                    elif output.get("accepted", False):
-                        rel = self.model.get_relation(PeerRelationName)
-                        if rel and closed_idx:
-                            rel.data[self.charm.app]["restore_in_progress"] = ",".join(closed_idx)
-                            break
-                        else:
-                            raise OpenSearchRestoreSingleUnitNotAsyncError()
-                    raise OpenSearchRestoreError()
         except RetryError:
-            logger.error("_restore_and_try_close_indices_if_needed: fail all retries")
-            if not self.model.get_relation(PeerRelationName):
-                raise OpenSearchRestoreSingleUnitNotAsyncError()
-        return output, closed_idx
+            raise OpenSearchRestoreFailedClosingIdxError(
+                "_close_indices_if_needed: fail all retries"
+            )
+        return closed_idx
 
-    def _on_update_status(self, _):
-        """Checks if the backup and/or restores are finished."""
-        status = self._check_repo_status()
-        if status == BackupServiceState.SUCCESS:
-            # There is a repo setup, check if we have any backup/restore registered in the
-            # peer relation's app databag
-            logger.info(f"Checking if backup in progress: {self.is_backup_in_progress()}")
-            logger.info(f"Checking if restore in progress: {self._check_if_restore_finished()}")
+    def _issue_restore_request(
+        self, backup_id: int, backup_indices: List[str], wait_for_completion: bool = False
+    ) -> Dict[str, Any]:
+        """Runs the restore and processes the response according to wait_for_completion flag."""
+        # backup_indices = self._list_backups()[backup_id]["indices"]
+        output = self._request(
+            "POST",
+            f"_snapshot/{S3_REPOSITORY}/{backup_id}/_restore?wait_for_completion={str(wait_for_completion).lower()}",
+            payload={
+                "indices": ",".join(
+                    [f"-{idx}" for idx in INDICES_TO_EXCLUDE_AT_RESTORE & set(backup_indices)]
+                ),
+                "ignore_unavailable": False,
+                "include_global_state": False,
+                "include_aliases": True,
+                "partial": False,
+            },
+        )
+        logger.debug(f"_issue_restore_request: restore call returned {output}")
+        if (
+            self.get_service_status(output)
+            == BackupServiceState.SNAPSHOT_RESTORE_ERROR_INDEX_NOT_CLOSED
+        ):
+            to_close = output["error"]["reason"].split("[")[2].split("]")[0]
+            raise OpenSearchRestoreFailedClosingIdxError(
+                f"_issue_restore_request: fails to close {to_close}"
+            )
+        return output
 
     def _check_if_restore_finished(self) -> bool:  # noqa: C901
         rel = self.model.get_relation(PeerRelationName)
@@ -354,8 +337,8 @@ class OpenSearchBackup(Object):
                 closed_idx.discard(idx)
 
         if indices_not_found:
-            raise OpenSearchRestoreError(
-                f"Failed to find indices {indices_not_found} in /recovery"
+            raise OpenSearchRestoreMismatchBetweenIdxAndSnapshotError(
+                f"Failed to find indices {indices_not_found} in /_recovery"
             )
         if (
             not closed_idx
@@ -378,6 +361,18 @@ class OpenSearchBackup(Object):
             event.fail(f"Failed: {e}")
             return
 
+    def _backup_is_available_for_restore(self, backup_id: int) -> bool:
+        """Checks if the backup_id exists and is ready for a restore."""
+        backups = self._list_backups()
+        try:
+            return (
+                backup_id in backups.keys()
+                and self.get_snapshot_status(backups[backup_id]["state"])
+                != BackupServiceState.SUCCESS
+            )
+        except OpenSearchListBackupError:
+            return False
+
     def _on_restore_backup_action(self, event: ActionEvent) -> None:  # noqa: C901
         """Restores a backup to the current cluster."""
         if not self._can_unit_perform_backup(event):
@@ -387,40 +382,46 @@ class OpenSearchBackup(Object):
         if not self._check_if_restore_finished():
             event.fail("Failed: previous restore is still in progress")
             return
+        # Now, validate the backup is working
         backup_id = str(event.params.get("backup-id"))
+        if self._backup_is_available_for_restore(backup_id):
+            event.fail(f"Failed: no backup-id {backup_id}")
+            return
 
         # Restore will try to close indices if there is a matching name.
         # The goal is to leave the cluster in a running state, even if the restore fails.
         # In case of failure, then restore action must return a list of closed indices
         closed_idx = set()
         try:
-            backups = self._list_backups()
-            if backup_id not in backups.keys():
-                event.fail(f"Failed: no backup-id {backup_id}, options available: {backups}")
-                return
-            backup_id_state = self.get_snapshot_status(backups[backup_id]["state"])
-            if backup_id_state != BackupServiceState.SUCCESS:
-                event.fail(
-                    f"Failed: no backup-id {backup_id} not successful, state is: {backup_id_state}"
-                )
-                return
-            output, closed_idx = self._restore_and_try_close_indices_if_needed(
+            closed_idx = self._close_indices_if_needed()
+            output = self._issue_restore_request(
                 backup_id, event.params.get("wait-for-completion", False)
             )
             logger.debug(f"Restore action: received response: {output}")
             logger.info(f"Restore action succeeded for backup_id {backup_id}")
-        except OpenSearchRestoreSingleUnitNotAsyncError:
-            event.fail(
-                "Failed: single unit restore with wait_for_completion=false is not supported!\n"
-                "closed indices: {closed_idx}"
-            )
+        except OpenSearchRestoreFailedClosingIdxError as e:
+            event.fail(f"Failed: {e}")
             return
-        except OpenSearchListBackupError as e:
-            event.fail(f"Failed: {e}, closed indices: {closed_idx}")
-            return
+
+        # Post execution checks
+        # Was the call successful?
         if self.get_service_status(output) != BackupServiceState.SUCCESS:
             status = str(self.get_service_status(output))
             event.fail(f"Failed with: {status}, closed indices: {closed_idx}")
+            return
+        # Finally, if status returns success, ensure all indices have been updated
+        # otherwise, raise an assert error and retry
+        if event.params.get("wait-for-completion", False):
+            # We are waiting for restore to complete, then we need to check the status
+            shards = output.get("snapshot", {}).get("shards", {})
+            assert shards.get("successful", -1) == shards.get("total", 0)
+        elif output.get("accepted", False):
+            rel = self.model.get_relation(PeerRelationName)
+            if rel and closed_idx:
+                rel.data[self.charm.app]["restore_in_progress"] = ",".join(closed_idx)
+            else:
+                event.fail("Failed: no relation available to store the restore status")
+                return
         event.set_results({"state": "successful restore!", "closed-indices": str(closed_idx)})
 
     def _on_create_backup_action(self, event: ActionEvent) -> None:  # noqa: C901
--- a/lib/charms/opensearch/v0/opensearch_backups.py
+++ b/lib/charms/opensearch/v0/opensearch_backups.py
@@ -50,7 +50,7 @@ import math
 from typing import Any, Dict, List, Set, Tuple
 
 from charms.data_platform_libs.v0.s3 import S3Requirer
-from charms.opensearch.v0.constants_charm import RestoreActionFailed, RestoreStarting
+from charms.opensearch.v0.constants_charm import RestoreInProgress
 from charms.opensearch.v0.helper_cluster import ClusterState, IndexStateEnum
 from charms.opensearch.v0.helper_enums import BaseStrEnum
 from charms.opensearch.v0.opensearch_exceptions import (
@@ -65,6 +65,7 @@ from charms.opensearch.v0.opensearch_plugins import (
 from ops.charm import ActionEvent
 from ops.framework import EventBase, Object
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
+from tenacity import RetryError, Retrying, stop_after_attempt, wait_fixed
 
 # The unique Charmhub library identifier, never change it
 LIBID = "d301deee4d2c4c1b8e30cd3df8034be2"
@@ -256,8 +257,11 @@ class OpenSearchBackup(Object):
             ):
                 indices_to_close.add(index)
 
-        if not self._close_indices(indices_to_close):
-            raise OpenSearchRestoreIndexClosingError()
+        try:
+            if not self._close_indices(indices_to_close):
+                raise OpenSearchRestoreIndexClosingError()
+        except OpenSearchError as e:
+            raise OpenSearchRestoreIndexClosingError(e)
         return indices_to_close
 
     def _restore(self, backup_id: int) -> Dict[str, Any]:
@@ -280,8 +284,10 @@ class OpenSearchBackup(Object):
         ):
             to_close = output["error"]["reason"].split("[")[2].split("]")[0]
             raise OpenSearchRestoreIndexClosingError(f"_restore: fails to close {to_close}")
+
         if "snapshot" not in output or "shards" not in output.get("snapshot"):
             raise OpenSearchRestoreCheckError(f"_restore: unexpected response {output}")
+
         return output["snapshot"]
 
     def _is_restore_complete(self) -> bool:
@@ -304,7 +310,7 @@ class OpenSearchBackup(Object):
             return (
                 backup_id in backups.keys()
                 and self.get_snapshot_status(backups[backup_id]["state"])
-                != BackupServiceState.SUCCESS
+                == BackupServiceState.SUCCESS
             )
         except OpenSearchListBackupError:
             return False
@@ -319,11 +325,11 @@ class OpenSearchBackup(Object):
             return
         # Now, validate the backup is working
         backup_id = str(event.params.get("backup-id"))
-        if self._is_backup_available_for_restore(backup_id):
+        if not self._is_backup_available_for_restore(backup_id):
             event.fail(f"Failed: no backup-id {backup_id}")
             return
 
-        self.charm.status.set(MaintenanceStatus(RestoreStarting))
+        self.charm.status.set(MaintenanceStatus(RestoreInProgress))
 
         # Restore will try to close indices if there is a matching name.
         # The goal is to leave the cluster in a running state, even if the restore fails.
@@ -340,24 +346,22 @@ class OpenSearchBackup(Object):
             OpenSearchRestoreCheckError,
         ) as e:
             event.fail(f"Failed: {e}")
-            self.charm.status.set(BlockedStatus(RestoreActionFailed))
             return
 
         # Post execution checks
         # Was the call successful?
         state = self.get_service_status(output)
         if state != BackupServiceState.SUCCESS:
-            self.charm.status.set(BlockedStatus(RestoreActionFailed))
+            event.fail(f"Restore failed with {state}")
             return
 
         shards = output.get("shards", {})
-        if not shards or shards.get("successful", -1) != shards.get("total", 0):
-            self.charm.status.set(BlockedStatus(RestoreActionFailed))
+        if shards.get("successful", -1) != shards.get("total", 0):
             event.fail("Failed to restore all the shards")
             return
 
         msg = "Restore is complete" if self._is_restore_complete() else "Restore in progress..."
-        self.charm.status.clear(RestoreStarting)
+        self.charm.status.clear(RestoreInProgress)
         event.set_results(
             {"backup-id": backup_id, "status": msg, "closed-indices": str(closed_idx)}
         )
@@ -365,16 +369,11 @@ class OpenSearchBackup(Object):
     def _on_create_backup_action(self, event: ActionEvent) -> None:  # noqa: C901
         """Creates a backup from the current cluster."""
         if not self._can_unit_perform_backup(event):
-            event.fail("Failed: backup service is not configured yet")
+            event.fail("Failed: backup service is not configured or busy")
             return
 
         new_backup_id = None
         try:
-            # Check if any backup is not running already, or RetryError happens
-            if self.is_backup_in_progress():
-                event.fail("Backup still in progress: aborting this request...")
-                return
-
             # Increment by 1 the latest snapshot_id (set to 0 if no snapshot was previously made)
             new_backup_id = int(max(self._list_backups().keys() or [0])) + 1
             logger.debug(
@@ -412,13 +411,15 @@ class OpenSearchBackup(Object):
         """
         # First, validate the plugin is present and correctly configured.
         if self._plugin_status != PluginState.ENABLED:
-            event.fail(f"Failed: plugin is not ready yet, current status is {self._plugin_status}")
+            logger.warning(
+                f"Failed: plugin is not ready yet, current status is {self._plugin_status}"
+            )
             return False
 
         # Then, check the repo status
         status = self._check_repo_status()
         if status != BackupServiceState.SUCCESS:
-            event.fail(f"Failed: repo status is {status}")
+            logger.warning(f"Failed: repo status is {status}")
             return False
         return not self.is_backup_in_progress()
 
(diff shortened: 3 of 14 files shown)


lib/charms/opensearch/v0/opensearch_secrets.py
```diff
@@ -41,21 +41,63 @@
...
+            return
+        # There is a callback function to update the secret
+        old_val = secret.get_content()
+        new_val = secret.peek_content()
+        if new_val == old_val:
+            # Nothing to do, these dictionaries have the same content
+            return
+        if cb(old_val, new_val):
+            # Remove the old value
+            secret.get_content(refresh=True)
```
juditnovak, line 89: NOTE: https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/data_interfaces.py#L530-L542

If we want cross-versions backward compatibility, we may need 2 handle errors here

lib/charms/opensearch/v0/opensearch_secrets.py
```diff
@@ -41,21 +41,77 @@
...
+                "ERROR either URI or label should be used for getting an owned secret but not both",
+                "ERROR secret owner cannot use --refresh",
+            ]
+            if isinstance(err, model.ModelError) and not any(
+                msg in str(err) for msg in known_model_errors
+            ):
+                raise
+            # Due to: ValueError: Secret owner cannot use refresh=True
+            self._secret_content = self.meta.get_content()
+
```
juditnovak, line 110: Is there a reason to have a new, separate function for the `secret-changed` event specifically, instead of integrating it with the existing `_update_juju_secrets()` function?

Also, I think secrets manipulation should happen via the interface functions of this class, not private methods.

Also, what is this `cb` supposed to refer to...?

I have the gutfeeling that this should rather be merged with the existing functionality rahter than implemented as brand new -- outside of the existing caching method.

phvalguima: Hi @juditnovak, as we've discussed, this function is added outside of any other in `opensearch_secrets.py` because I wanted to quickly fix our charm to support 3.1.7 and then, we move to a more permanent solution using our common library.

Callback (or `cb`) is used to distinguish between secrets. With the changes landed in 3.1.7, now some secrets (e.g. passwords) will be able to read both old and new values and execute the change on their database, whereas other secrets (e.g. private keys/certs) can be refreshed right away. The callback is for that distinction:
* private keys / certs: `cb=None`, they just execute the refresh
* passwords: `cb=change_my_password(old_val, new_val)`, they take the longer tour, where we execute the password and, once we are sure it worked, secret-changed proceeds and refresh to the latest secret value

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -272,20 +280,69 @@ def _restore_and_try_close_indices_if_needed(
...
+        """Checks the status of indices that have been recovered from snapshot."""
+        try:
+            if self._check_if_restore_finished():
+                event.set_results({"state": "successful restore!"})
+                return
+            event.set_results({"state": "restore in progress..."})
+            return
+        except Exception as e:
+            event.fail(f"Failed: {e}")
+            return
```
Mehdi-Bendriss, line 335:
```suggestion
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -272,20 +280,69 @@ def _restore_and_try_close_indices_if_needed(
...
+        return False
+
+    def _on_check_restore_status_action(self, event: ActionEvent) -> None:
+        """Checks the status of indices that have been recovered from snapshot."""
+        try:
+            if self._check_if_restore_finished():
+                event.set_results({"state": "successful restore!"})
+                return
+            event.set_results({"state": "restore in progress..."})
+            return
```
Mehdi-Bendriss, line 332:
```suggestion
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -424,7 +483,9 @@ def _on_s3_credentials_changed(self, event: EventBase) -> None:
 
         try:
             plugin = self.charm.plugin_manager.get_plugin(OpenSearchBackupPlugin)
-            self.charm.plugin_manager._configure_if_needed(plugin)
+            if self.charm.plugin_manager.status(plugin) == PluginState.ENABLED:
+                self.charm.plugin_manager.disable(plugin)
```
Mehdi-Bendriss, line 487: Could you explain why we're disabling ?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -209,6 +209,20 @@ def _install_if_needed(self, plugin: OpenSearchPlugin) -> bool:
 
         return self._install_plugin(plugin)
 
+    def configure(self, plugin: OpenSearchPlugin) -> bool:
```
Mehdi-Bendriss, line 212: Could you explain why?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -387,6 +386,17 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
         # acquire lock to ensure only 1 unit removed at a time
         self.ops_lock.acquire()
 
+        def _get_cluster_color():
```
Mehdi-Bendriss, line 389: Let's take this part with regards to scale-down out of this PR and mere our efforts on this (the issue we talked about last week)

Mehdi-Bendriss (commented):
Thanks Pedro! I have a few comments

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -127,6 +127,10 @@ def stop(self):
         while self.is_started() and (datetime.now() - start).seconds < 60:
             time.sleep(3)
 
+    def get_service_status(self):
```
Mehdi-Bendriss, line 130: Do we need this? we already have:
- `is_started`
- `is_failed`
- `is_node_up`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -272,20 +280,69 @@ def _restore_and_try_close_indices_if_needed(
...
             logger.error("_restore_and_try_close_indices_if_needed: fail all retries")
         return output, closed_idx
 
+    def _check_if_restore_finished(self) -> bool:
+        try:
+            output = self._request(
+                "GET",
+                "_cat/recovery?format=json",
+            )
+        except Exception as e:
```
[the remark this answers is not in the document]
phvalguima: That is not complete, right? we may have other errors here, which makes it easy to forget when writing code:
```
        Raises:
            ValueError if method or endpoint are missing
            OpenSearchHttpError if hosts are unreachable
            requests.HTTPError if connection to opensearch fails
```

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Because it contains configuration from the other plugin, mainly different keys in `OpenSearchKeystore`.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: `_configure_if_needed` does: (1) check IF we have a plugin installed but not configured and then does its work; and (2) internal to OpenSearchPlugin and its Manager.
Think about the case of moving from a GCP cloud to AWS. In this case, we *do* have an already configured plugin, but we want to reconfigure it nonetheless. That is why we need a method that removes/adds configuration disregard what is the current plugin status.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -272,20 +280,69 @@ def _restore_and_try_close_indices_if_needed(
...
+            indices_status = json.loads(output)
+        elif isinstance(output, list):
+            indices_status = output
+        elif isinstance(output, dict):
+            indices_status = [output]
+        else:
+            raise OpenSearchRestoreCheckError(
+                f"_cat/recovery returned {type(output)} instead of str, list or dict"
+            )
+        if len([index["stage"] for index in indices_status if index["type"] == "snapshot"]) == 0:
```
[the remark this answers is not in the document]
phvalguima: Changed this part entirely to simplify the logic here.

lib/charms/opensearch/v0/opensearch_backups.py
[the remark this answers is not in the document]
Mehdi-Bendriss: It should be as `OpenSearchHttpError` already encompasses `requests.HTTPError`, and no need to catch the ValueError here

juditnovak (approved):
I have little knowledge on backups so far, LGTM :sweat_smile:

juditnovak (changes requested):
Bah, sorry, I'm an iddiot for approving a second ago.

PLS-PLS-PLS could we add proper amount of reliable integration testing for *such* an amount of chaneges in the backup logic...?

tests/integration/ha/test_backups.py
```diff
@@ -118,6 +119,25 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
...
+                ops_test, leader_id, "list-backups", params={"output": "json"}
+            )
+            logger.info(f"list-backups output: {action}")
+            # Expected format:
+            # namespace(status='completed', response={'return-code': 0, 'backups': '{"1": ...}'})
+            backups = json.loads(action.response["backups"])
+            logger.info(backups)
+            assert action.status == "completed"  # The actual action status
+            assert len(backups) > 0  # The number of backups
+            for _, backup in backups.items():
```
juditnovak, line 136:
```
for backup in backups.values()
```

phvalguima:
Hi @juditnovak, I've brought the integration tests from PR#159 to this one + added more unit tests. Currently, this is the coverage state [in here](https://github.com/canonical/opensearch-operator/actions/runs/7555511387/job/20570657688?pr=158):
```
unit: commands[1]> coverage report
Name                                                           Stmts   Miss  Cover
----------------------------------------------------------------------------------
lib/charms/opensearch/v0/constants_charm.py                       50      0   100%
lib/charms/opensearch/v0/constants_secrets.py                      6      0   100%
lib/charms/opensearch/v0/constants_tls.py                         16      0   100%
lib/charms/opensearch/v0/helper_charm.py                          36      0   100%
lib/charms/opensearch/v0/helper_cluster.py                       155     26    83%
lib/charms/opensearch/v0/helper_conf_setter.py                   186     30    84%
lib/charms/opensearch/v0/helper_enums.py                          11      0   100%
lib/charms/opensearch/v0/helper_networking.py                     52      2    96%
lib/charms/opensearch/v0/helper_security.py                       30      0   100%
lib/charms/opensearch/v0/models.py                               116     11    91%
lib/charms/opensearch/v0/opensearch_backups.py                   427    189    56%
lib/charms/opensearch/v0/opensearch_base_charm.py                509    296    42%
lib/charms/opensearch/v0/opensearch_config.py                     94     20    79%
lib/charms/opensearch/v0/opensearch_distro.py                    218     80    63%
lib/charms/opensearch/v0/opensearch_exceptions.py                 33      1    97%
lib/charms/opensearch/v0/opensearch_fixes.py                      18      8    56%
lib/charms/opensearch/v0/opensearch_health.py                     80     53    34%
lib/charms/opensearch/v0/opensearch_internal_data.py             114      9    92%
lib/charms/opensearch/v0/opensearch_keystore.py                   85     30    65%
lib/charms/opensearch/v0/opensearch_locking.py                    48     28    42%
lib/charms/opensearch/v0/opensearch_nodes_exclusions.py           78     49    37%
lib/charms/opensearch/v0/opensearch_peer_clusters.py             150     79    47%
lib/charms/opensearch/v0/opensearch_plugin_manager.py            176     43    76%
lib/charms/opensearch/v0/opensearch_plugins.py                    64      2    97%
lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py       7      7     0%
lib/charms/opensearch/v0/opensearch_relation_provider.py         152     27    82%
lib/charms/opensearch/v0/opensearch_secrets.py                   171     25    85%
lib/charms/opensearch/v0/opensearch_tls.py                       142     39    73%
lib/charms/opensearch/v0/opensearch_users.py                     115     54    53%
----------------------------------------------------------------------------------
TOTAL                                                           3339   1108    67%
```

tests/integration/ha/test_backups.py
```diff
@@ -118,6 +119,25 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
...
+    for attempt in Retrying(stop=stop_after_attempt(8), wait=wait_fixed(15)):
+        with attempt:
+            action = await run_action(
+                ops_test, leader_id, "list-backups", params={"output": "json"}
+            )
+            logger.info(f"list-backups output: {action}")
+            # Expected format:
+            # namespace(status='completed', response={'return-code': 0, 'backups': '{"1": ...}'})
+            backups = json.loads(action.response["backups"])
+            logger.info(backups)
```
juditnovak, line 133: Should we log this?

juditnovak: (If yes -- perhaps a bit more verbose? :-) )

tests/integration/ha/test_backups.py
```diff
@@ -232,19 +250,7 @@ async def test_backup_cluster(
...
-
-    # Expected format:
-    # namespace(status='completed', response={'return-code': 0, 'backups': '{"1": ...}'})
-    backups = json.loads(list_backups.response["backups"])
-    assert list_backups.status == "completed"
-    assert len(backups.keys()) == int(action.response["backup-id"])
-    assert backups[action.response["backup-id"]]["state"] == "SUCCESS"
-
-    # continuous writes checks
-    await assert_continuous_writes_consistency(ops_test, c_writes, app)
```
juditnovak, line 247: Why don't we need to check for this anymore?

tests/integration/ha/test_backups.py
```diff
@@ -310,40 +316,14 @@ async def test_restore_cluster_after_app_destroyed(ops_test: OpsTest) -> None:
...
-            TEST_BACKUP_INDEX,
-            query={"query": {"term": {"_id": doc_id}}},
-            preference="_only_local",
-        )
-        # Validate the index and document are present
-        assert len(docs) == 1
-        assert docs[0]["_source"] == default_doc(TEST_BACKUP_INDEX, doc_id)
+    # This is the same check as the previous restore action.
+    # Call the method again
+    await test_restore_cluster(ops_test)
```
juditnovak, line 321: It's a bit awkward one test calling the other...

I'm trying to decide if that's a good trick or not...

Don't we wanna have a helper function with the body of this test, and that would be called both in an individual test, and in here?

tests/integration/ha/test_backups.py
```diff
@@ -310,40 +316,14 @@ async def test_restore_cluster_after_app_destroyed(ops_test: OpsTest) -> None:
...
-            ops_test,
-            app,
-            u_ip,
-            TEST_BACKUP_INDEX,
-            query={"query": {"term": {"_id": doc_id}}},
-            preference="_only_local",
-        )
-        # Validate the index and document are present
-        assert len(docs) == 1
-        assert docs[0]["_source"] == default_doc(TEST_BACKUP_INDEX, doc_id)
```
juditnovak, line 336: Sry I may be getting lost... Where is this index check thing gone and if it's gone completely, why don't we need it anymore?

tests/integration/ha/test_backups.py
```diff
@@ -365,49 +345,6 @@ async def test_remove_and_readd_s3_relation(ops_test: OpsTest) -> None:
...
```
juditnovak, line 413: Here again I find this check that I don't see anywhere migrated...

phvalguima: @juditnovak it is not gone, but this check is done in the _backup and _restore tests already.

tests/integration/ha/test_backups.py
phvalguima: @juditnovak it is not gone, but this check is done in the _backup and _restore tests already.

tests/integration/ha/test_backups.py
phvalguima: Yes, I am adding one, so we can also keep the c_writes.

tests/integration/ha/test_backups.py
phvalguima: I do, but if I reuse in the disaster recovery scenarios, then this test won't really work b/c effectively the writes will break for a certain time.

phvalguima: I am returning these checks, but taking another route to reuse certain tests.

tests/integration/ha/test_backups.py
phvalguima: Now, we have ContinuousWrites in the Backup test (01). Then, for restore, DR (disaster recovery) and relation removed/readded scenarios, we start the ContinuousWrites *after* the restoration of the cluster.

That is because our goal is to validate the cluster is functional post-restore.

juditnovak (dismissed):
Thx very-very much 4 all those tests!!!!!

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -235,105 +258,193 @@ def _restore_and_try_close_indices_if_needed(
...
+                f"Failed to find indices {indices_not_found} in /_recovery"
+            )
+        if (
+            not closed_idx
+            and self.charm.unit.is_leader()
+            and "restore_in_progress" in rel.data[self.charm.app]
+        ):
+            del rel.data[self.charm.app]["restore_in_progress"]
+            return True
+        return False
```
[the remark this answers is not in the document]
Mehdi-Bendriss: And perhaps add the `self.charm.peers_data.delete(Scope.APP, "restore_in_progress")` on the caller side when _leader unit.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -235,105 +258,193 @@ def _restore_and_try_close_indices_if_needed(
...
+
+    def _on_check_restore_status_action(self, event: ActionEvent) -> None:
+        """Checks the status of indices that have been recovered from snapshot."""
+        try:
+            if self._check_if_restore_finished():
+                event.set_results({"state": "successful restore!"})
```
Mehdi-Bendriss, line 357: before this you should probably add `self.charm.peers_data.delete(Scope.APP, "restore_in_progress")`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -235,105 +258,193 @@ def _restore_and_try_close_indices_if_needed(
...
+            return False
 
     def _on_restore_backup_action(self, event: ActionEvent) -> None:  # noqa: C901
         """Restores a backup to the current cluster."""
         if not self._can_unit_perform_backup(event):
             event.fail("Failed: backup service is not configured yet")
             return
 
+        if not self._check_if_restore_finished():
+            event.fail("Failed: previous restore is still in progress")
```
Mehdi-Bendriss, line 382: same comment here

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -235,105 +258,193 @@ def _restore_and_try_close_indices_if_needed(
...
+            return
+        # Now, validate the backup is working
         backup_id = str(event.params.get("backup-id"))
+        if self._backup_is_available_for_restore(backup_id):
```
Mehdi-Bendriss, line 386: please method names should be consistent `_is_....`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -235,105 +258,193 @@ def _restore_and_try_close_indices_if_needed(
...
         if self.get_service_status(output) != BackupServiceState.SUCCESS:
             status = str(self.get_service_status(output))
             event.fail(f"Failed with: {status}, closed indices: {closed_idx}")
+            return
+        # Finally, if status returns success, ensure all indices have been updated
+        # otherwise, raise an assert error and retry
+        if event.params.get("wait-for-completion", False):
+            # We are waiting for restore to complete, then we need to check the status
+            shards = output.get("snapshot", {}).get("shards", {})
+            assert shards.get("successful", -1) == shards.get("total", 0)
```
Mehdi-Bendriss, line 416: same as previously - please avoid using asserts in the code.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -235,105 +258,193 @@ def _restore_and_try_close_indices_if_needed(
...
+        elif output.get("accepted", False):
+            rel = self.model.get_relation(PeerRelationName)
```
Mehdi-Bendriss, line 418: please use `self.charm.peers_data....` - this will avoid having to do these checks etc...

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -345,15 +456,23 @@ def _on_create_backup_action(self, event: ActionEvent) -> None:
             )
 
             logger.info(f"Backup request submitted with backup-id {new_backup_id}")
+            logger.info(f"Backup completed with backup-id {new_backup_id}")
+
+            if wait_for_completion and self.is_backup_in_progress():
+                event.fail(f"Timed out while waiting for: backup id {new_backup_id}")
```
Mehdi-Bendriss, line 462:
```suggestion
                event.fail(f"Timed out while waiting for: backup id {new_backup_id} - Please check the status again in the future.")
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -345,15 +456,23 @@ def _on_create_backup_action(self, event: ActionEvent) -> None:
...
             OpenSearchListBackupError,
         ) as e:
             event.fail(f"Failed with exception: {e}")
             return
+        if wait_for_completion:
+            event.set_results({"backup-id": new_backup_id, "status": "Backup completed."})
+            return
+
+        rel = self.model.get_relation(PeerRelationName)
+        rel.data[self.charm.app]["backup_in_progress"] = str(new_backup_id)
```
Mehdi-Bendriss, line 475: same comment with regards to `peers_data`.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -345,15 +456,23 @@ def _on_create_backup_action(self, event: ActionEvent) -> None:
...
         except (
-            RetryError,
-            ValueError,
             OpenSearchHttpError,
-            requests.HTTPError,
             OpenSearchListBackupError,
         ) as e:
             event.fail(f"Failed with exception: {e}")
             return
+        if wait_for_completion:
```
Mehdi-Bendriss, line 470: this can be put below the previous [condition](https://github.com/canonical/opensearch-operator/pull/158/files#diff-9a3ba6c99d30f70486eef55dba2362fbc0480e6752eb8bb98db1c2e5b24242b4R461)

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -389,10 +512,42 @@ def _list_backups(self) -> Dict[int, str]:
...
-        """Returns True if backup is in progress, False otherwise."""
-        return self._query_backup_status(backup_id) in [
+        """Returns True if backup is in progress, False otherwise.
+
+        First, check if we have a marker of a running backup. If yes, then we check for both the
+        saved backup_id in peer relation AND backup_id, if provided. If neither are in progress,
+        return False and clean up the marker from the relation.
+        """
+        rel = self.model.get_relation(PeerRelationName)
+        if rel:
```
Mehdi-Bendriss, line 522: same remark..

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -389,10 +512,42 @@ def _list_backups(self) -> Dict[int, str]:
...
+            # If relation is present, then check the peer relation data
+            peer_relation_backup_id = rel.data[self.charm.app].get("backup_in_progress", None)
+            if peer_relation_backup_id and self._query_backup_status(backup_id) in [
+                BackupServiceState.SNAPSHOT_IN_PROGRESS
+            ]:
+                # We have a backup in progress
+                return True
+            else:
```
Mehdi-Bendriss, line 530: `else` not needed after a `return`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -389,10 +512,42 @@ def _list_backups(self) -> Dict[int, str]:
...
+                # Clean up status
+                if (
+                    self.charm.unit.is_leader()
+                    and "backup_in_progress" in rel.data[self.charm.app]
+                ):
+                    del rel.data[self.charm.app]["backup_in_progress"]
```
Mehdi-Bendriss, line 536: use peers_data

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -389,10 +512,42 @@ def _list_backups(self) -> Dict[int, str]:
...
         ]
+        if backup_id_in_progress:
+            # We have a backup in progress
+            if (
+                peer_relation_backup_id
+                and peer_relation_backup_id != backup_id_in_progress
+                and self.charm.unit.is_leader()
+            ):
+                # We need to update our peer relation as the IDs were different
+                rel.data[self.charm.app]["backup_in_progress"] = str(backup_id_in_progress)
```
Mehdi-Bendriss, line 548: this method shouldn't have to do this, given its name indicates that its sole focus is to check whether the backup is in progress, consider moving this to the caller side, or slightly rename the method

lib/charms/opensearch/v0/opensearch_backups.py
Mehdi-Bendriss: the other plugin == S3 plugin?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
Mehdi-Bendriss: It looks like `configure` and `disable` are simply proxy methods for `_apply_config` - it is safe to refer to `apply_config(...)` directly instead of having these 2 new methods

tests/integration/ha/continuous_writes.py
```diff
@@ -133,12 +134,19 @@ async def max_stored_id(self) -> int:
...
         """Create index with 1 p_shard and an r_shard on each node."""
         client = await self._client()
+        units = await get_application_unit_ips(self._ops_test, app=self._app)
+
         try:
             # create index with a replica shard on every node
             client.indices.create(
                 index=ContinuousWrites.INDEX_NAME,
                 body={
-                    "settings": {"index": {"number_of_shards": 2, "auto_expand_replicas": "1-all"}}
```
Mehdi-Bendriss, line 141: let's not change this please. Multiple tests rely on the health of the cluster as per this configuration.

tests/integration/ha/continuous_writes.py
```diff
@@ -39,13 +39,14 @@ class ContinuousWrites:
     LAST_WRITTEN_VAL_PATH = "last_written_value"
     CERT_PATH = "/tmp/ca_chain.cert"
 
-    def __init__(self, ops_test: OpsTest, app: str):
+    def __init__(self, ops_test: OpsTest, app: str, initial_count: int = 0):
```
Mehdi-Bendriss, line 42: This shouldn't be needed. As the continuous writes gets deleted at the end of each test, and we start again anew from 0 on new tests.

Mehdi-Bendriss (commented):
Thank you Pedro, I have some comments

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -145,28 +167,35 @@ def __init__(self, charm: Object):
...
         )
+        self.framework.observe(self.charm.on.update_status, self._on_update_status)
         self.framework.observe(self.charm.on.create_backup_action, self._on_create_backup_action)
         self.framework.observe(self.charm.on.list_backups_action, self._on_list_backups_action)
         self.framework.observe(self.charm.on.restore_action, self._on_restore_backup_action)
+        self.framework.observe(
+            self.charm.on.check_restore_status_action, self._on_check_restore_status_action
+        )
+
+    def _on_update_status(self, _):
```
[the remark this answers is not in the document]
phvalguima: Okay, as we've discussed, also removed the list of "closed_idx" that we were saving in the peer relation.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -145,28 +167,35 @@ def __init__(self, charm: Object):
...
     @property
     def _plugin_status(self):
         return self.charm.plugin_manager.get_plugin_status(OpenSearchBackupPlugin)
 
     def _format_backup_list(self, backup_list: List[Tuple[Any]]) -> str:
         """Formats provided list of backups as a table."""
-        backups = [
-            "{:<10s} | {:<12s} | {:s}".format(" backup-id ", "backup-type", "backup-status")
-        ]
+        backups = ["{:<10s} | {:s}".format(" backup-id ", "backup-status")]
```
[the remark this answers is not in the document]
phvalguima: Called it output instead

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -235,105 +258,193 @@ def _restore_and_try_close_indices_if_needed(
...
+                f"_issue_restore_request: fails to close {to_close}"
+            )
+        return output
+
+    def _check_if_restore_finished(self) -> bool:  # noqa: C901
+        rel = self.model.get_relation(PeerRelationName)
+        if not rel:
+            # Running on a single unit, wait_for_completion=false not supported
+            return True
+        if not rel.data[self.charm.app].get("restore_in_progress"):
```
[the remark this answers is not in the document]
phvalguima: No need anymore

lib/charms/opensearch/v0/opensearch_backups.py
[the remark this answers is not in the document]
phvalguima: This is going away

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: I did not get it. This is here just to check at the end if we should say "backup completed", because we waited for it; or is this an async test and we inform "backup is running"

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -235,105 +258,193 @@ def _restore_and_try_close_indices_if_needed(
...
+        logger.debug(f"_issue_restore_request: restore call returned {output}")
+        if (
+            self.get_service_status(output)
+            == BackupServiceState.SNAPSHOT_RESTORE_ERROR_INDEX_NOT_CLOSED
+        ):
+            to_close = output["error"]["reason"].split("[")[2].split("]")[0]
+            raise OpenSearchRestoreFailedClosingIdxError(
+                f"_issue_restore_request: fails to close {to_close}"
+            )
+        return output
```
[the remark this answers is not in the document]
phvalguima: I moved this error to be under `except RetryError`, if that is what you originally meant.

tests/integration/ha/continuous_writes.py
phvalguima: B/c GH runners are small, I need to run microceph + OpenSearch in 2x units only. Therefore, I need here to be flexible.

phvalguima: @Mehdi-Bendriss I will roll this change back as we have more powerful GH runners now.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,135 +210,223 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+            return
+        # Finally, if status returns success, ensure all indices have been updated
+        # otherwise, raise an assert error and retry
+        if event.params.get("wait-for-completion", False):
+            # We are waiting for restore to complete, then we need to check the status
+            shards = output.get("snapshot", {}).get("shards", {})
+            if shards.get("successful", -1) != shards.get("total", 0):
+                event.fail("Failed: restore failed to complete")
+        elif output.get("accepted", False):
+            if not closed_idx:
```
Mehdi-Bendriss, line 403:
```suggestion
                event.fail("Failed: restore failed to complete")
                return
                
        if output.get("accepted", False) and not closed_idx:
```

lib/charms/opensearch/v0/opensearch_backups.py
Mehdi-Bendriss: Line 443 you can do:
```
if wait_for_completion:
    if self.is_backup_in_progress():
        event.fail(
            f"Timed out while waiting for: backup id {new_backup_id} - Please check the status again in the future."
        )
    else:
        event.set_results({"backup-id": new_backup_id, "status": "Backup completed."})
else:
    event.set_results({"backup-id": new_backup_id, "status": "Backup is running."})

     
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -389,10 +493,15 @@ def _list_backups(self) -> Dict[int, str]:
...
-        """Returns True if backup is in progress, False otherwise."""
-        return self._query_backup_status(backup_id) in [
-            BackupServiceState.SNAPSHOT_IN_PROGRESS,
-        ]
+        """Returns True if backup is in progress, False otherwise.
+
+        We filter the _query_backup_status() and seek for the following states:
+        - SNAPSHOT_IN_PROGRESS
+        """
+        if self._query_backup_status() in [BackupServiceState.SNAPSHOT_IN_PROGRESS]:
```
Mehdi-Bendriss, line 501: This is a proxy method, we should perhaps rename this method `_backup_status` and put the code from `_query_backup_status` here and test on the status directly instead of a boolean.

tests/integration/ha/test_backups.py
```diff
@@ -114,44 +119,201 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
...
+    )
+
     await c_writes.clear()
     logger.info("\n\n\n\nThe writes have been cleared.\n\n\n\n")
 
 
+@pytest.fixture(scope="session")
+def microceph():
+    """Starts microceph radosgw."""
+    import subprocess
```
Mehdi-Bendriss: please imports at the top of the files.

tests/integration/ha/test_backups.py
```diff
@@ -114,44 +119,201 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
...
+    if "microceph" not in subprocess.check_output(["sudo", "snap", "list"]).decode():
+        import os
+
+        import requests
+
+        uceph = "/tmp/microceph.sh"
+
+        with open(uceph, "w") as f:
+            resp = requests.get(
+                "https://raw.githubusercontent.com/canonical/microceph-action/main/microceph.sh"
```
Mehdi-Bendriss, line 143: please download this file and add it to the tests resources if needed

tests/integration/ha/test_backups.py
```diff
@@ -114,44 +119,201 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
...
+                "secretkey",
+                "-b",
+                "data-charms-testing",
+                "-z",
+                "5G",
+            ]
+        )
+    # Now, return the configuration
+    ip = subprocess.check_output(["hostname", "-I"]).decode().split()[0]
+    return {"url": f"http://{ip}", "access-key": "accesskey", "secret-key": "secretkey"}
```
Mehdi-Bendriss, line 168: can you explain what accesskey / secretkey hardcoded values are?

tests/integration/ha/helpers_data.py
```diff
@@ -247,3 +247,22 @@ async def search(
         with attempt:  # Raises RetryError if failed after "retries"
             resp = await http_request(ops_test, "GET", endpoint, payload=query, app=app)
             return resp["hits"]["hits"]
+
+
+async def index_count(
```
Mehdi-Bendriss, line 252:
```suggestion
async def index_docs_count(
```

tests/integration/ha/test_backups.py
```diff
@@ -114,44 +119,201 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
...
+
+
+async def _wait_backup_finish(ops_test, leader_id):
```
Mehdi-Bendriss, line 171: tests should start with `test_` they should also be decorated with `@pytest.mark.abort_on_fail` - please refer to the rest of the integration tests for reference.

tests/integration/ha/test_backups.py
```diff
@@ -114,44 +119,201 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
...
+            backups = json.loads(action.response["backups"])
+            logger.debug(f"Backups recovered: {backups}")
+            assert action.status == "completed"  # The actual action status
+            assert len(backups) > 0  # The number of backups
+            for backup in backups.values():
+                logger.debug(f"Backup is: {backup}")
+                assert backup["state"] == "SUCCESS"  # The backup status
+
+
+async def _assert_test_cont_writes_works(ops_test: OpsTest) -> ContinuousWrites:
```
Mehdi-Bendriss, line 190: same comment as above as to how should tests be defined. Tests should not return objects.

tests/integration/ha/test_backups.py
```diff
@@ -114,44 +119,201 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
...
+
+    Given we are restoring an index, we need to make sure ContinuousWrites restart at
+    the tip of that index instead of doc_id = 0.
+
+    Closes the writer at the end.
+    """
+    app = (await app_name(ops_test)) or APP_NAME
+    leader_unit_ip = await get_leader_unit_ip(ops_test, app=app)
+
+    initial_count = await index_count(ops_test, app, leader_unit_ip, ContinuousWrites.INDEX_NAME)
```
Mehdi-Bendriss, line 201: this should be `== 0` - as continuous writes are cleared after each test.

tests/integration/ha/test_backups.py
```diff
@@ -114,44 +119,201 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
...
+    except Exception as e:
+        await writer.stop()
+        raise e
+
+    time.sleep(5)
+    result = await writer.stop()
+    assert result.count > initial_count
+
+
+async def _backup_cluster(ops_test: OpsTest) -> None:
```
Mehdi-Bendriss, line 240: same comment

Mehdi-Bendriss (changes requested):
Thanks Pedro, I left some comments, the tests may need to be reviewed to stay aligned and consistent with the rest of the integration tests.

tests/integration/ha/test_backups.py
phvalguima: @Mehdi-Bendriss this is temporary and will go away with: https://github.com/canonical/opensearch-operator/pull/159

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,135 +210,223 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+        """Requests indices to be closed and return their status."""
+        resp = self._request(
+            "POST",
+            f"{','.join(indices)}/_close",
+            payload={
+                "ignore_unavailable": "true",
+            },
+        )
+        if not resp or not resp.get("acknowledged", False):
+            return {"acknowledged": False}
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss yes, it changes the "shards_ack" field:
```
$ curl -XPOST -k -u admin:PWD https://10.206.26.149:9200/series_index,test_backup_index,.plugins-ml-config/_close
{"acknowledged":true,"shards_acknowledged":true,"indices":{"series_index":{"closed":true},"test_backup_index":{"closed":true},".plugins-ml-config":{"closed":true}}}
$ curl -XPOST -k -u admin:PWD https://10.206.26.149:9200/series_index,test_backup_index,.plugins-ml-config/_close
{"acknowledged":true,"shards_acknowledged":false,"indices":{}}
```

I will make it a binary method.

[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss I think that `_close_indices` should return True if we acknowledged. Tbh, the risk of having a missing index is not really high, because that method is only called: `_close_indices_if_needed` and we do a check between backup indices, available indices currently, indices that are already closed and indices that should be excluded of a restore.

(I also renamed _close_indices to _call_close_index_api, as the names were getting confusing :)

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,135 +210,223 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+            "POST",
+            f"_snapshot/{S3_REPOSITORY}/{backup_id}/_restore?wait_for_completion={str(wait_for_completion).lower()}",
+            payload={
+                "indices": ",".join(
+                    [f"-{idx}" for idx in INDICES_TO_EXCLUDE_AT_RESTORE & set(backup_indices)]
+                ),
+                "ignore_unavailable": False,
+                "include_global_state": False,
+                "include_aliases": True,
+                "partial": False,
```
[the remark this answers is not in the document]
phvalguima: I kept the "partial", because I want it to be more explicit for a future traveller:
```
        output = self._request(
            "POST",
            f"_snapshot/{S3_REPOSITORY}/{backup_id}/_restore?wait_for_completion={str(wait_for_completion).lower()}",
            payload={
                "indices": ",".join(
                    [f"-{idx}" for idx in INDICES_TO_EXCLUDE_AT_RESTORE & set(backup_indices)]
                ),
                "partial": False,  # It is the default value, but we want to avoid partial restores
            },
        )
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,135 +210,223 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+        """
+        try:
+            indices_status = self._request(
+                "GET",
+                "/_recovery?human",
+            )
+        except OpenSearchHttpError as e:
+            raise OpenSearchRestoreCheckError(f"Failed GET to /_recovery with : {e}")
+        if not isinstance(indices_status, dict):
+            indices_status = json.loads(indices_status)
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss `str`. It will always have a json format, but my worry here is later change of APIs, either requests in python, for example.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,135 +210,223 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+        closed_indices_state = {
+            index: IndexStateEnum.CLOSED
+            if state and state.get("closed") == "true"
+            else IndexStateEnum.OPEN
+            for index, state in resp.get("indices", {}).items()
+        }
+        return {"acknowledged": True, "indices": closed_indices_state}
+
+    def _close_indices_if_needed(self, backup_id: int) -> Set[str]:
+        """Closes indices that will be restored.
```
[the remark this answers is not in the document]
phvalguima: I agree these methods need a renaming. But  I prefer to keep this method as its actual go: closes indices if it is really needed; and we have a method that actually does the API call.

Otherwise I need to bring the close Retry logic up to the restore action, which is already big enough.

tests/integration/ha/test_backups.py
phvalguima: @Mehdi-Bendriss these are default values of the script. This logic will eventually go away with this PR: https://github.com/canonical/opensearch-operator/pull/159

tests/integration/ha/test_backups.py
phvalguima: Erh, not sure I get it, that is an aux. method that pretty much every test in `test_backups.py` calls at some point... It is not intended to be a test itself.
I will move it to `ha/helpers.py`

tests/integration/ha/test_backups.py
phvalguima: As mentioned above, I am restoring the backup, it will contain a non-empty `series_index`. That is just to take an initial point of reference, then doing some other logic; finally checking if the count has increased.

tests/integration/ha/test_backups.py
phvalguima: As discussed: I will add a TODO on each of the lines for microceph

phvalguima:
@Mehdi-Bendriss can you give another review?

I had to rebase this change, following a new commit to main.
Right before, [this was the result of the CI](https://github.com/canonical/opensearch-operator/actions/runs/7669539274/job/20903967063). The rebase affects `tests/conftest.py`.

lib/charms/opensearch/v0/helper_cluster.py
```diff
@@ -24,6 +25,16 @@
...
 
 
+class IndexStateEnum(BaseStrEnum):
+    """Enum for index states."""
+
+    OPEN = "open"
+    CLOSED = "closed"
+
+
+INDEX_WILDCARD = "all"
```
Mehdi-Bendriss, line 35: please delete

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,116 +198,171 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
 
-        Closing option is preferable, as the restore action as a whole may not succeed.
-        This method returns the final output and list of closed indices. The calling method
-        should also check the output for any other error messages, as a more generic error
-        (i.e. non-closing index error) will pass by as much as a successful restore.
+        Returns a set of indices that were closed or raises an exception:
+        - OpenSearchRestoreIndexClosingError if any of the indices could not be closed.
+
+        Raises:
+            OpenSearchHttpError
```
Mehdi-Bendriss, line 246: please catch this exception in this method, do not propagate it

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,116 +198,171 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+        for index, state in ClusterState.indices(self.charm.opensearch).items():
+            if (
+                index in backup_indices
+                and state["status"] != IndexStateEnum.CLOSED
+                and index not in INDICES_TO_EXCLUDE_AT_RESTORE
+            ):
+                indices_to_close.add(index)
+
+        if not self._close_indices(indices_to_close):
+            raise OpenSearchRestoreIndexClosingError()
```
Mehdi-Bendriss, line 260:
```suggestion
        try:
            if not self._close_indices(indices_to_close):
                raise OpenSearchRestoreIndexClosingError()
        except OpenSearchHttpError as e:
            raise OpenSearchRestoreIndexClosingError(e)
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,116 +198,171 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
-                    assert shard_status["total"] == shard_status["successful"]
-        except RetryError:
-            logger.error("_restore_and_try_close_indices_if_needed: fail all retries")
-        return output, closed_idx
-
-    def _on_restore_backup_action(self, event: ActionEvent) -> None:  # noqa: C901
+            return (
+                backup_id in backups.keys()
+                and self.get_snapshot_status(backups[backup_id]["state"])
+                != BackupServiceState.SUCCESS
```
Mehdi-Bendriss, line 307: should this be `==`?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,116 +198,171 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+        logger.debug(f"_restore: restore call returned {output}")
+        if (
+            self.get_service_status(output)
+            == BackupServiceState.SNAPSHOT_RESTORE_ERROR_INDEX_NOT_CLOSED
+        ):
+            to_close = output["error"]["reason"].split("[")[2].split("]")[0]
+            raise OpenSearchRestoreIndexClosingError(f"_restore: fails to close {to_close}")
+        if "snapshot" not in output or "shards" not in output.get("snapshot"):
+            raise OpenSearchRestoreCheckError(f"_restore: unexpected response {output}")
+        return output["snapshot"]
```
Mehdi-Bendriss, line 285:
```suggestion
            raise OpenSearchRestoreIndexClosingError(f"_restore: fails to close {to_close}")
            
        if "snapshot" not in output or "shards" not in output.get("snapshot"):
            raise OpenSearchRestoreCheckError(f"_restore: unexpected response {output}")
        
        return output["snapshot"]
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,116 +198,171 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
         if not self._can_unit_perform_backup(event):
             event.fail("Failed: backup service is not configured yet")
             return
-
+        if not self._is_restore_complete():
+            event.fail("Failed: previous restore is still in progress")
+            return
+        # Now, validate the backup is working
         backup_id = str(event.params.get("backup-id"))
+        if self._is_backup_available_for_restore(backup_id):
```
Mehdi-Bendriss, line 322: negation missing? 
```suggestion
        if not self._is_backup_available_for_restore(backup_id):
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,116 +198,171 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
-            status = str(self.get_service_status(output))
-            event.fail(f"Failed with: {status}, closed indices: {closed_idx}")
-        event.set_results({"state": "successful restore!", "closed-indices": str(closed_idx)})
 
-    def _on_create_backup_action(self, event: ActionEvent) -> None:
+        # Post execution checks
+        # Was the call successful?
+        state = self.get_service_status(output)
+        if state != BackupServiceState.SUCCESS:
+            self.charm.status.set(BlockedStatus(RestoreActionFailed))
```
Mehdi-Bendriss, line 350: the status should not be altered on a failure, `event.fail` should be used here.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,116 +198,171 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+            return
+
+        shards = output.get("shards", {})
+        if not shards or shards.get("successful", -1) != shards.get("total", 0):
```
Mehdi-Bendriss, line 354:
```suggestion
        if shards.get("successful", -1) != shards.get("total", 0):
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -205,116 +198,171 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
+            self.charm.status.set(BlockedStatus(RestoreActionFailed))
```
Mehdi-Bendriss, line 355: same comment as above - `event.fail` is enough

lib/charms/opensearch/v0/constants_charm.py
```diff
@@ -70,6 +71,8 @@
 HorizontalScaleUpSuggest = "Horizontal scale up advised: {} shards unassigned."
 WaitingForOtherUnitServiceOps = "Waiting for other units to complete the ops on their service."
 NewIndexRequested = "new index {index} requested"
+RestoreStarting = "Restore action started."
```
Mehdi-Bendriss, line 74:
```suggestion
RestoreInProgress = "Restore in progress..."
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -388,11 +433,16 @@ def _list_backups(self) -> Dict[int, str]:
...
+
+        We filter the _query_backup_status() and seek for the following states:
+        - SNAPSHOT_IN_PROGRESS
+        """
+        if self._query_backup_status() in [BackupServiceState.SNAPSHOT_IN_PROGRESS]:
+            # We have a backup in progress
+            return True
+        return False
 
     def _query_backup_status(self, backup_id=None) -> BackupServiceState:
```
Mehdi-Bendriss, line 447: Please add a retry here - and catch the OpenSearchHttpError - and return the right state if such exception occurs

tests/integration/ha/helpers.py
```diff
@@ -437,3 +440,74 @@ async def print_logs(ops_test: OpsTest, app: str, unit_id: int, msg: str) -> str
...
+            )
+            # Expected format:
+            # namespace(status='completed', response={'return-code': 0, 'backups': '{"1": ...}'})
+            backups = json.loads(action.response["backups"])
+            logger.debug(f"Backups recovered: {backups}")
+            if action.status == "completed" and len(backups) > 0:
+                logger.debug(f"list-backups output: {action}")
+                return
+            else:
+                raise Exception("Backup not finished yet")
```
Mehdi-Bendriss, line 460:
```suggestion
            
            raise Exception("Backup not finished yet")
```

tests/integration/ha/helpers.py
```diff
@@ -437,3 +440,74 @@ async def print_logs(ops_test: OpsTest, app: str, unit_id: int, msg: str) -> str
...
+                ops_test,
+                "GET",
+                f"https://{unit_ip}:9200/_recovery?human",
+            )
+            for info in indices_status.values():
+                # Now, check the status of each shard
+                for shard in info["shards"]:
+                    if shard["type"] == "SNAPSHOT" and shard["stage"] != "DONE":
+                        raise Exception()
+            return
```
Mehdi-Bendriss, line 477:
```suggestion
```

Mehdi-Bendriss (commented):
Thanks Pedro. It looks good! I have a few minor points to clarify. 

(In another PR - I would like us to reduce the number of exceptions we expose in the backup module, we could probably get away with simply `OpenSearchBackupError, OpenSearcchRestoreError`

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: The `_query_backup_status` is in charge of returning a state. This method returns a boolean. So, what I will do is apply your ask to the `_query_backup_status`

Mehdi-Bendriss (approved):

juditnovak (approved):
LGTM

PR #160, opened 2024-01-12
**OLD** [DPE-3306] Support repository-gcs

phvalguima:
Extends original backup to also support repository-gcs plugin. The `opensearch-backup` library is extended to choose which plugin should be enabled based on the endpoint provided. If the endpoint string has `googleapis.com`, then repository-gcs is installed; whereas the remaining will use repository-s3.


lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -160,9 +161,30 @@ def __init__(self, charm: Object):
             self.charm.on.check_restore_status_action, self._on_check_restore_status_action
         )
 
+    @property
+    def type(self) -> str:
```
Mehdi-Bendriss, line 165: we should probably make this and all the following properties protected (name starts with `_`)

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -160,9 +161,30 @@ def __init__(self, charm: Object):
...
+            if "googleapis.com" in self.s3_client.get_s3_connection_info().get("endpoint")
+            else "s3"
+        )
+
+    @property
+    def _plugin_class(self):
+        """Returns the plugin class based on the type."""
+        if self.type == "s3":
+            return OpenSearchS3BackupPlugin
+        return OpenSearchGCSBackupPlugin
```
Mehdi-Bendriss, line 178: nit: 
```suggestion
        return OpenSearchS3BackupPlugin if self.type == "s3" else OpenSearchGCSBackupPlugin
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -142,17 +148,33 @@ def run(self) -> bool:
...
-                    self._configure_if_needed(plugin),
-                    self._disable_if_needed(plugin),
-                    self._remove_if_needed(plugin),
-                    restart_needed,
-                ]
-            )
+            # These are independent plugins.
+            # Any plugin that errors, if that is an OpenSearchPluginError, then
+            # it is an "expected" error, such as missing additional config; should
+            # not influence the execution of other plugins.
```
Mehdi-Bendriss, line 158: very nice!

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -275,16 +307,48 @@ def _on_update_status(self, event):
...
+###########################################################################################
+# These are the basic types of errors:
+# * OpenSearchPluginError: base class for all plugin errors in the plugins code
+# * OpenSearchPluginConfigError: base class for all plugin errors related to config options
 class OpenSearchPluginError(OpenSearchError):
     """Exception thrown when an opensearch plugin is invalid."""
 
+    def __init__(self, msg, only_log=False):
+        super().__init__(msg)
+        self._only_log = only_log
```
Mehdi-Bendriss, line 319: Could you explain this? what is the alternate course of action when a plugin errors?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -275,16 +307,48 @@ def _on_update_status(self, event):
...
+
+    @property
+    def only_log(self) -> bool:
+        """Returns the only_log property."""
+        return self._only_log
+
+
+class OpenSearchPluginConfigError(OpenSearchPluginError):
```
Mehdi-Bendriss, line 327: Though theoretically nice, we do not need to go to this level of granularity. This can be solved and conveyed directly through the error messages in the raised exceptions

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -458,6 +528,12 @@ def config(self) -> OpenSearchPluginConfig:
             secret_entries_to_del = [...]|{...},
         )
         """
+        if not self._extra_config.get("access-key") and not self._extra_config.get("secret-key"):
+            raise OpenSearchPluginMissingConfigError(
+                "Missing AWS access-key and secret-key configuration",
+                affected_configs=["access-key", "secret-key"],
+                only_log=True,
```
Mehdi-Bendriss, line 535: This is not needed, the message `Missing AWS access-key and secret-key configuration` should already be enough.

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -492,3 +568,84 @@ def disable(self) -> OpenSearchPluginConfig:
...
+        """
+        if not self._extra_config.get("service-account") and not self._extra_config.get(
+            "secret-key"
+        ):
+            raise OpenSearchPluginMissingConfigError(
+                "Missing GCS service-account or secret-key configuration",
+                affected_configs=["service-account", "secret-key"],
+                only_log=True,
+            )
+        import base64
```
Mehdi-Bendriss, line 603: please imports at the top of the file

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -492,3 +568,84 @@ def disable(self) -> OpenSearchPluginConfig:
...
+
+        data, conf = (
+            (self._extra_config["service-account"], "service-account")
+            if self._extra_config.get("service-account")
```
Mehdi-Bendriss, line 607: why do we need the `service-account` if we can do simply with `secret-key` as indicated in the rest of the condition? I don't understand exactly what's happening here 
My understanding was that for the `GCS` plugin, we should exclusively use `service-account` ?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -458,6 +528,12 @@ def config(self) -> OpenSearchPluginConfig:
             secret_entries_to_del = [...]|{...},
         )
         """
+        if not self._extra_config.get("access-key") and not self._extra_config.get("secret-key"):
```
Mehdi-Bendriss, line 531: shouldn't this be an `or` ?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -275,16 +307,48 @@ def _on_update_status(self, event):
...
+
 
 class OpenSearchPluginRelationClusterNotReadyError(OpenSearchPluginError):
     """Exception thrown when making API calls and cluster is not yet ready.
 
     This exception in specific should be handled by classes managing the relations of plugins.
     """
 
+    def __init__(self, only_log=True):
+        super().__init__("The cluster is not ready yet.", only_log)
```
Mehdi-Bendriss, line 350: related to the point above

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -492,3 +568,84 @@ def disable(self) -> OpenSearchPluginConfig:
...
+
+        data, conf = (
+            (self._extra_config["service-account"], "service-account")
+            if self._extra_config.get("service-account")
+            else (self._extra_config.get("secret-key"), "secret-key")
+        )
+        try:
+            account = " ".join(base64.b64decode(data).decode().rstrip().splitlines())
+        except Exception as e:
+            # This error means that GCS is wrongly configured.
```
Mehdi-Bendriss, line 613: what kind of exact error are we expecting here, seems like `data` can never be null, so a base64 encoding will always succeed

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -492,3 +568,84 @@ def disable(self) -> OpenSearchPluginConfig:
...
+                str(e), affected_configs=[conf], only_log=True
+            )
+        return OpenSearchPluginConfig(
+            # Try to remove the previous values
+            secret_entries_to_del=[
+                "gcs.client.default.credentials_file",
+            ],
+            secret_entries_to_add={
+                # Remove any entries with None value
+                k: v
```
Mehdi-Bendriss, line 628:
```suggestion
                key: val
```

Mehdi-Bendriss (commented):
Great work! I have a few comments

juditnovak (commented):
Sry, am I correct that this PR isn't "up" anymore, as it's been split to other PRs instead?

juditnovak (changes requested):
The level of testing is FAR too minimal on this PR.

Could we have PLS corresponding solid integration tests...? I'm really not an expert on the topic, but more than 400 lines of changes (backups, plugins, secrets, etc.) just doesn't feel sufficient to be covered by 2 unittest in our charms world...

Could you pls-pls-pls make sure that all new features would be tested, and --assuming we were checking code coverage-- it would certainly not drop in percentage after this merge? Thx very much (on behalf of the next developer, who surely won't need to face any potential bugs :-) )

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: There are some issues that I do not want to raise, like for example if repository-gcs is not completely configured, whereas repository-s3 is. In these plugins, I raise an exception with _only_log=True.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: Actually, I am making distinctions later down the road between the types of errors. I want to then treat: `repository-gcs` with a `OpenSearchPluginMissingConfigError` is not something I should be alarmed, for example.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: This error should be logged only. I noticed in the past few CIs we have been throwing exceptions for this error instead. This is a transitory error that happens at certain stages of the cluster.
It is safe to only log it.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: yes, on the long run, we should use service-account. However, I do not have support for service-account yet in s3-integrator, so I am passing service account via secret-key for testing.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: not if base64 is not power of 2 length OR not encoded at all.

PR #161, opened 2024-01-15
Add secret fix for 3.1.7

phvalguima:
This PR extends opensearch_secrets.py to react to new secret-changed behavior in Juju 3.1.7+, as described in [LP#2037120](https://bugs.launchpad.net/juju/+bug/2037120).

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -76,7 +76,6 @@ jobs:
         with:
           provider: lxd
           juju-channel: ${{ inputs.juju-snap-channel }}
-          bootstrap-options: "--agent-version=3.1.6"
 
       - name: Download packed charm(s)
         uses: actions/download-artifact@v3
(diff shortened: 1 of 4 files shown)


Mehdi-Bendriss (approved):

welpaolo (approved):

juditnovak:
Guys, I know I'm late to the party, but this isn't the right way to fix the issue.

The correct place is inside the `_get_juju_secret()` function.
No, we don't need to worry that a secret's value may change over the execution of an event handler. Local caching makes sure that the value of the secret remains static throughout the run of the event handler.

@phvalguima Can I pls ask to remove the `refresh=True` line from `on_secret_changed()` and add it to `_get_juju_secret()` instead in upcoming PR https://github.com/canonical/opensearch-operator/pull/160/files ? IT's a small change, and the upcoming PR has code that's strongly related.

Issue #164, opened 2024-01-24
`_start_opensearch` does not catch 


phvalguima:
There seems to be some race condition happening at `_post_start_init`, where some of the CI tests fail with `OpenSearchUserMgmtError`. That is caused in the stack at the end.

Side note: `_post_start_init` should be renamed to something else, as it runs as the very first line at `_start_opensearch`.

Stack trace: https://pastebin.ubuntu.com/p/WRMY3FnZCc/
```
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-opensearch-4/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 276, in call
    response = s.request(**request_kwargs)
  File "/var/lib/juju/agents/unit-opensearch-4/charm/venv/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
  File "/var/lib/juju/agents/unit-opensearch-4/charm/venv/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
  File "/var/lib/juju/agents/unit-opensearch-4/charm/venv/requests/adapters.py", line 532, in send
    raise ReadTimeout(e, request=request)
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='10.206.26.175', port=9200): Read timed out. (read timeout=5)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-opensearch-4/charm/lib/charms/opensearch/v0/opensearch_users.py", line 137, in get_users
    return self.opensearch.request("GET", f"{USER_ENDPOINT}/")
  File "/var/lib/juju/agents/unit-opensearch-4/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 297, in request
    resp = call(retries, resp_status_code)
  File "/var/lib/juju/agents/unit-opensearch-4/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 285, in call
    return call(
  File "/var/lib/juju/agents/unit-opensearch-4/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 238, in call
    raise OpenSearchHttpError()
charms.opensearch.v0.opensearch_exceptions.OpenSearchHttpError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-opensearch-4/charm/./src/charm.py", line 94, in <module>
[...]

PR #165, opened 2024-01-24
[DPE-3352] Catch OpenSearchUserMgmtError at start cycle

phvalguima:
Catch OpenSearchUserMgmtError thrown by _post_start_init.

Bug-Fixes: #164


Mehdi-Bendriss (changes requested):
Thanks! As discussed, this would only mask / hide the issue. We should rather fix the root cause of why is the request timing out - when the previous requests made to the same endpoint succeed.

Issue #154, opened 2023-12-10
Race condition due to mix of start/restart logic


phvalguima:
When running the `OpenSearchPluginManager`, it is failing to correctly restart all the nodes post-install / -configuration. The only node that reliably gets restarted is the leader unit. I am occasionally getting clusters that have not accounted for the newly installed `repository-s3` plugin (i.e. they did not restart after install). I can see the locks being required and then released, but journalctl does not show a restart at the same corresponding time.

The first thing is that we are using the `acquire_lock` event with 2x different types of callbacks: `_start_opensearch` and `_restart_opensearch`. Looking at RollingOps code, I can see it sets [the callback_override function at `acquire_lock.emit`](https://github.com/canonical/charm-rolling-ops/blob/7745ceee696007003485b6912ce01aae87ab0bed/lib/charms/rolling_ops/v0/rollingops.py#L388) and then, [it uses the method that was defined `at constructor time`](https://github.com/canonical/charm-rolling-ops/blob/7745ceee696007003485b6912ce01aae87ab0bed/lib/charms/rolling_ops/v0/rollingops.py#L401). That means we are always using `_start_opensearch`.

The first step is to get the charm to use a single callback. The `_restart_opensearch` is de facto choice, as it optionally stops the running opensearch service.

However, `_restart_opensearch` checks first if we have a `starting` flag set in the peer relation. That is important at the moment of the very first start, but after it can cause some deadlocks (e.g. the following `_start_opensearch` may set that flag and prematurely exit, before removing the flag, then no other _restart_opensearch calls will be able stop the service and no new configuration will be picked up).

The second step is to find a more suitable check to decide if we should stop the service.

The next issue: we are over-complexifying the `_restart_opensearch` by calling the `_start_opensearch` in it. The `_start_opensearch` [has multiple exits with a `defer`](https://github.com/canonical/opensearch-operator/blob/550b205ab99ec3e1b8416c95e736e0bd90ddb9a2/lib/charms/opensearch/v0/opensearch_base_charm.py#L601-L625). That effectively means we are deferring the RunWithLockEvent, which means potentially new restarts.

---------------------------------------------------------------------------------------------

I think there is a mix of tasks in this section of the code.

The start and restart should be distinct steps. Indeed, OpenSearch start can be very complex, if we consider large clusters; where we need a combination of `on_start`, `peer_joined` and `config_changed` events to trigger different information.

[...]

PR #177, opened 2024-02-06
[DPE-3573][DPE-3579] Fix #164 - check _cluster/settings instead of opensearch.yml for plugins

### LINKED ISSUE #164: `_start_opensearch` does not catch

phvalguima:
Adds a GET /_cluster/settings, which returns the configuration of the cluster at the moment, including defaults. This way, plugins can check if they should add their configs and request a restart during the start cycle of the charm.

The `_is_enabled` is also extended to use the cluster settings instead of checking the configuration files directly. 

Moved `OpenSearchPluginConfig` to extend `pydantic.BaseModel`. It ensures that configuration is translated to `Dict[str, str]`, in special the boolean value, that must have a format of: `true` or `false`.

Also closes [DPE-3579](https://warthogs.atlassian.net/browse/DPE-3579): move the plugin settings status to use the charm's `status.{set,clear}` APIs.

[DPE-3579]: https://warthogs.atlassian.net/browse/DPE-3579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -58,6 +58,23 @@ class ClusterTopology:
         return base_roles + ["cluster_manager"]
 
     @staticmethod
+    def get_cluster_settings(
+        opensearch: OpenSearchDistribution,
+        host: Optional[str] = None,
+        alt_hosts: Optional[List[str]] = None,
+        include_defaults: bool = False,
+    ) -> Dict[str, any]:
+        """Get the cluster settings."""
+        settings = opensearch.request(
+            "GET",
+            f"/_cluster/settings?flat_settings=true&include_defaults={str(include_defaults).lower()}",
+            host=host,
+            alt_hosts=alt_hosts,
+        )
+
+        return dict(settings["defaults"] | settings["persistent"] | settings["transient"])
+
+    @staticmethod
     def recompute_nodes_conf(app_name: str, nodes: List[Node]) -> Dict[str, Node]:
         """Recompute the configuration of all the nodes (cluster set to auto-generate roles)."""
         # in case the cluster nodes' roles were previously "manually generated" - we need
--- a/lib/charms/opensearch/v0/opensearch_plugin_manager.py
+++ b/lib/charms/opensearch/v0/opensearch_plugin_manager.py
@@ -10,9 +10,12 @@ This class is instantiated at the operator level and is called at every relevant
 config-changed, upgrade, s3-credentials-changed, etc.
 """
 
+import copy
+import functools
 import logging
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
+from charms.opensearch.v0.helper_cluster import ClusterTopology
 from charms.opensearch.v0.opensearch_exceptions import (
     OpenSearchCmdError,
     OpenSearchNotFullyReadyError,
@@ -78,6 +81,11 @@ class OpenSearchPluginManager:
         self._keystore = OpenSearchKeystore(self._charm)
         self._event_scope = OpenSearchPluginEventScope.DEFAULT
 
+    @functools.cached_property
+    def cluster_config(self):
+        """Returns the cluster configuration."""
+        return ClusterTopology.get_cluster_settings(self._charm.opensearch, include_defaults=True)
+
     def set_event_scope(self, event_scope: OpenSearchPluginEventScope) -> None:
         """Sets the event scope of the plugin manager.
 
@@ -129,25 +137,31 @@ class OpenSearchPluginManager:
             }
         return {**self._charm_config, "opensearch-version": self._opensearch.version}
 
+    def check_plugin_manager_ready(self) -> bool:
+        """Checks if the plugin manager is ready to run."""
+        return (
+            self._charm.opensearch.is_node_up()
+            and len(self._charm._get_nodes(True)) == self._charm.app.planned_units()
+            and self._charm.health.apply()
+            in [
+                HealthColors.GREEN,
+                HealthColors.YELLOW,
+            ]
+        )
+
     def run(self) -> bool:
         """Runs a check on each plugin: install, execute config changes or remove.
 
         This method should be called at config-changed event. Returns if needed restart.
         """
-        if not self._charm.opensearch.is_started() or self._charm.health.apply() not in [
-            HealthColors.GREEN,
-            HealthColors.YELLOW,
-        ]:
-            # If the health is not green, then raise a cluster-not-ready error
-            # The classes above should then defer their own events in waiting.
-            # Defer is important as next steps to configure plugins will involve
-            # calls to the APIs of the cluster.
-            logger.info("Cluster not ready, wait for the next event...")
+        if not self.check_plugin_manager_ready():
             raise OpenSearchNotFullyReadyError()
 
         err_msgs = []
         restart_needed = False
         for plugin in self.plugins:
+            logger.info(f"Checking plugin {plugin.name}...")
+            logger.debug(f"Status: {self.status(plugin)}")
             # These are independent plugins.
             # Any plugin that errors, if that is an OpenSearchPluginError, then
             # it is an "expected" error, such as missing additional config; should
@@ -172,8 +186,8 @@ class OpenSearchPluginManager:
                 # This is a more serious issue, as we are missing some input from
                 # the user. The charm should block.
                 err_msgs.append(str(e))
-            except OpenSearchPluginError as e:
-                logger.error(f"Plugin {plugin.name} failed: {str(e)}")
+            logger.debug(f"Finished Plugin {plugin.name} status: {self.status(plugin)}")
+        logger.info(f"Plugin check finished, restart needed: {restart_needed}")
 
         if err_msgs:
             raise OpenSearchPluginError("\n".join(err_msgs))
@@ -255,6 +269,39 @@ class OpenSearchPluginManager:
         except KeyError as e:
             raise OpenSearchPluginMissingConfigError(plugin.name, configs=[f"{e}"])
 
+    def _compute_settings(
+        self, config: OpenSearchPluginConfig
+    ) -> Tuple[Dict[str, str], Dict[str, str]]:
+        """Returns the current and the new configuration."""
+        current_settings = self.cluster_config
+        # We use current_settings and new_conf and check for any differences
+        # therefore, we need to make a deepcopy here before editing new_conf
+        new_conf = copy.deepcopy(current_settings)
+        for key_to_del in config.config_entries_to_del:
+            if key_to_del in new_conf:
+                del new_conf[key_to_del]
+        new_conf |= config.config_entries_to_add
+
+        logger.debug(
+            "Difference between current and new configuration: \n"
+            + str(
+                {
+                    "in current but not in new_conf": {
+                        k: v for k, v in current_settings.items() if k not in new_conf.keys()
+                    },
+                    "in new_conf but not in current": {
+                        k: v for k, v in new_conf.items() if k not in current_settings.keys()
+                    },
+                    "in both but different values": {
+                        k: v
+                        for k, v in new_conf.items()
+                        if k in current_settings.keys() and current_settings[k] != v
+                    },
+                }
+            )
+        )
+        return current_settings, new_conf
+
     def apply_config(self, config: OpenSearchPluginConfig) -> bool:
         """Runs the configuration changes as passed via OpenSearchPluginConfig.
 
@@ -262,27 +309,25 @@ class OpenSearchPluginManager:
         1) Remove the entries to be deleted
         2) Add entries, if available
 
-        For example:
-            KNN needs to:
-            1) Remove from configuration: {"knn.plugin.enabled": "True"}
-            2) Add to configuration: {"knn.plugin.enabled": "False"}
-
         Returns True if a configuration change was performed.
         """
         self._keystore.delete(config.secret_entries_to_del)
         self._keystore.add(config.secret_entries_to_add)
-        # Add and remove configuration if applies
-        if config.config_entries_to_del:
-            self._opensearch_config.delete_plugin(config.config_entries_to_del)
-
-        if config.config_entries_to_add:
-            self._opensearch_config.add_plugin(config.config_entries_to_add)
-
         if config.secret_entries_to_del or config.secret_entries_to_add:
             self._keystore.reload_keystore()
 
-        # Return True if some configuration entries changed
-        return True if config.config_entries_to_add or config.config_entries_to_del else False
+        current_settings, new_conf = self._compute_settings(config)
+        if current_settings == new_conf:
+            # Nothing to do here
+            logger.info("apply_config: nothing to do, return")
+            return False
+
+        # Update the configuration
+        if config.config_entries_to_del:
+            self._opensearch_config.delete_plugin(config.config_entries_to_del)
+        if config.config_entries_to_add:
+            self._opensearch_config.add_plugin(config.config_entries_to_add)
+        return True
 
     def status(self, plugin: OpenSearchPlugin) -> PluginState:
         """Returns the status for a given plugin."""
@@ -320,32 +365,28 @@ class OpenSearchPluginManager:
     def _is_enabled(self, plugin: OpenSearchPlugin) -> bool:
         """Returns true if plugin is enabled.
 
-        The main question to answer is if we would remove the configuration from
-        opensearch.yaml and/or add some other configuration instead.
-        If yes, then we know that the service is enabled.
+        The main question to answer is if we would have it in the configuration
+        from cluster settings. If yes, then we know that the service is enabled.
 
-        The disable() method is used, as a given entry will be removed from opensearch.yml.
-
-        If disable() returns a list, then check if the keys in the stored_plugin_conf
-        matches the list values; otherwise, if a dict, then check if the keys and values match.
-        If they match in any of the cases above, then return True.
+        Check if the configuration from enable() is present or not.
         """
         try:
-            plugin_conf = plugin.disable().config_entries_to_del
-            keystore_conf = plugin.disable().secret_entries_to_del
-            stored_plugin_conf = self._opensearch_config.get_plugin(plugin_conf)
-
-            if any(k not in self._keystore.list() for k in keystore_conf):
+            current_settings, new_conf = self._compute_settings(plugin.config())
+            if current_settings != new_conf:
                 return False
 
-            # Using sets to guarantee matches; stored_plugin_conf will be a dict
-            if isinstance(plugin_conf, list):
-                return set(plugin_conf) == set(stored_plugin_conf)
-            else:  # plugin_conf is a dict
-                return plugin_conf == stored_plugin_conf
+            # Now, focus on the keystore part
+            keys_available = self._keystore.list()
+            keys_to_add = plugin.config().secret_entries_to_add
+            if any(k not in keys_available for k in keys_to_add):
+                return False
+            keys_to_del = plugin.config().secret_entries_to_del
+            if any(k in keys_available for k in keys_to_del):
+                return False
         except (KeyError, OpenSearchPluginError) as e:
             logger.warning(f"_is_enabled: error with {e}")
             return False
+        return True
 
     def _needs_upgrade(self, plugin: OpenSearchPlugin) -> bool:
         """Returns true if plugin needs upgrade."""
--- a/lib/charms/opensearch/v0/opensearch_relation_provider.py
+++ b/lib/charms/opensearch/v0/opensearch_relation_provider.py
@@ -328,18 +328,18 @@ class OpenSearchProvider(Object):
         """
         if not self.unit.is_leader():
             # We're updating app databags in this function, so it can't be called on follower
-            # units.
+            # units. This is not checked in `set_tls_ca`, in data-interfaces.
             return
-        if not ca_chain:
-            try:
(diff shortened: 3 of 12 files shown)


phvalguima:
To clarify: start logic for me is all the logic directly involved in getting the service up and running for the very first time.

Regarding the questions:
> why do we need to go through the restart method for anything start related? it seems we are complicating the logic of the restart by making it a single point of entry. (please refer to PR-175 for ideas)

There are some reasons:
1) The charm-rolling-ops never checks directly for the callback method in the event itself. It [sets the callback_override in the relation](https://github.com/canonical/charm-rolling-ops/blob/4bae5c031cb7a8d5fd3819ace1f1496c87c0aae4/lib/charms/rolling_ops/v0/rollingops.py#L388) and then, in a next RunWithLockEvent, it [searches in the relation databag for the override](https://github.com/canonical/charm-rolling-ops/blob/4bae5c031cb7a8d5fd3819ace1f1496c87c0aae4/lib/charms/rolling_ops/v0/rollingops.py#L401-L403). So, if we switch methods, we may have:
* Hook #1, Event #1: Unit runs: acquire_lock.emit(callback_override=_start_opensearch)
* Hook #1, Event #1: Unit finishes Event #1
...
* Hook #1, Event #X: Unit runs: acquire_lock.emit(callback_override=_restart_opensearch) <--- we have overriden the value in the relation databag
...
* Hook #N, Event #1: Unit received "RunWithLockEvent" and recovers the callback: _restart_opensearch is called
**There is a potential race condition here**
2) The start logic is still being executed several hooks later, for each unit. That increases the chance of overlapping _start_opensearch and _restart_opensearch

> I'm not sure I understand what we're trying to solve in the changes in the _on_start hook? I re-read the previous logic, and it seems it was correct as it was prior to the new changes

For some reasons: #154 - I find the start logic in the charm has a lot going on. It needs information that only comes later, like certificates. That means we have some "mixing", where certain events of the lifecycle have to happen even before the actual start. That means: **we need a way to defer**.
The first step: the `_on_start` gets more of the logic, which also cleans up the `_start_opensearch` and allows us to have one single `callback_override`: `_restart_opensearch`. That first simplification *moves some of the defer up to the _on_start*. Then, I still have to deal, as you said, with delayed connections. You can see I added tenacity and I have several retrials in the start/stop. Now, these are synchronous.
I cannot make the tests work 100% with simply removing the extra "restart" caused by plugin-manager originally - [it randomly fails by never receiving the lock](https://github.com/canonical/opensearch-operator/actions/runs/7812769145/job/21310816939).

-------------------------------------------------------------------------------------------------------

Regarding your points:

1. plugins: repurposed the PR back to focus only in the plugin

2. experiment with sync start: yes, that was my first trial. But then I noticed it did not resolve the problem - our starting cycle has several stages and we need some way to defer the start until we are ready. That is where I started to work in the RollingOpsManager instead. I am trying to have a retriable locking system.

3. As mentioned above, I do not want to accidentally mix _start_opensearch and _restart_opensearch. Given a start is just a "restart without stopping", I see no reason to merge both methods.

4. Ack, done in: charm-rolling-ops#12. That was my 1st iteration on the problem. I will prepare a PR using OpenSearch for the testing. Re:PR#175, if you check the rolling-ops change, I think we should extend "WorkloadLockManager" and overload "can_node_be_safe_stopped", where it would check the output of "wait_for_shards_relocation" to see if it is safe to stop that node.

phvalguima:
The backup error in: https://github.com/canonical/opensearch-operator/actions/runs/7951972946/job/21713806527, is a status cleanup addressed in #186

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -332,15 +332,16 @@ def update_certs(self, relation_id, ca_chain=None):
...
             try:
-                ca_chain = self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val).get(
-                    "chain"
-                )
+                if ca_chain := self.charm.secrets.get_object(
+                    Scope.APP, CertType.APP_ADMIN.val
+                ).get("chain"):
+                    self.opensearch_provides.set_tls_ca(relation_id, ca_chain)
+                else:
+                    logger.warning("unable to get ca_chain")
```
juditnovak, line 340: If I get it right, the `if ... else ...` can't raise `AttributeError`.

Why not to move it after the `try ... except`, based on whether `ca_chain` is defeined or not?

juditnovak (dismissed):

lib/charms/opensearch/v0/opensearch_relation_provider.py
phvalguima: Hi @juditnovak the `AttributeError` is happening because the original `ca_chain` is set to `None` at early stages.

phvalguima:
The  `h-scaling-integration` failed with a time out:
```
2024-02-19T19:14:36.5332986Z [32mINFO    [0m tests.integration.helpers_deployments:helpers_deployments.py:271 	App: opensearch - Units - conditions unmet.
2024-02-19T19:14:36.5334811Z [32mINFO    [0m tests.integration.helpers_deployments:helpers_deployments.py:272 
2024-02-19T19:14:36.5335427Z 	app: opensearch active -- message: None
2024-02-19T19:14:36.5335956Z 		opensearch-3  -- (10.171.13.159) -- [idle (since: 18:44:39)] active: 
2024-02-19T19:14:36.5336557Z 		opensearch-5  -- (10.171.13.194) -- [idle (since: 18:44:39)] active: 
2024-02-19T19:14:36.5337257Z 		opensearch-6  -- (10.171.13.245) -- [idle (since: 18:44:37)] waiting: Awaiting service operation
2024-02-19T19:14:36.5337962Z 		opensearch-7  -- (10.171.13.183) -- [idle (since: 18:44:39)] active: 
2024-02-19T19:14:36.5338550Z 		opensearch-8* -- (10.171.13.236) -- [idle (since: 18:44:48)] active: 
2024-02-19T19:14:36.5338885Z 
2024-02-19T19:14:36.5339362Z [1m[31mERROR   [0m tests.integration.helpers_deployments:helpers_deployments.py:346 wait_until -- Timed out!
```

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Exactly.

```
           try:
                ca_chain := self.charm.secrets.get_object(
                    Scope.APP, CertType.APP_ADMIN.val
                ).get("chain")
            except AttributeError:
                # cert doesn't exist - presumably we don't yet have a TLS relation.
                return
                
          if ca_chain:
               self.opensearch_provides.set_tls_ca(relation_id, ca_chain)
           else:
                logger.warning("unable to get ca_chain")
```
Generally we should have as little in the `try ... except` as possible

lib/charms/opensearch/v0/helper_cluster.py
```diff
@@ -57,6 +57,27 @@ def suggest_roles(nodes: List[Node], planned_units: int) -> List[str]:
...
+        """Get the cluster settings."""
+        settings = opensearch.request(
+            "GET",
+            f"/_cluster/settings?flat_settings=true&include_defaults={str(include_defaults).lower()}",
+            host=host,
+            alt_hosts=alt_hosts,
+        )
+
+        result = {}
+        for el in ["defaults", "transient", "persistent"]:
```
Mehdi-Bendriss, line 76: is this needed? the output seems to be exactly those 3 keys

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -253,34 +278,68 @@ def _disable_if_needed(self, plugin: OpenSearchPlugin) -> bool:
         except KeyError as e:
             raise OpenSearchPluginMissingConfigError(e)
 
+    def _compute_settings(
+        self, config: OpenSearchPluginConfig
+    ) -> Tuple[Dict[str, str], Dict[str, str]]:
+        """Returns the current and the new configuration."""
+        current_settings = ClusterTopology.get_cluster_settings(
```
Mehdi-Bendriss, line 285: Perhaps this can be retrieved before everything in the `run()` method and passed around in all methods that rely on it (the response payload is very big), we could probably get away with downloading/loading it once and the diffs should still be well computed

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -253,34 +278,68 @@ def _disable_if_needed(self, plugin: OpenSearchPlugin) -> bool:
...
+            self._charm.opensearch,
+            include_defaults=True,
+        )
+        to_remove = dict(
+            zip(config.config_entries_to_del, [None] * len(config.config_entries_to_del))
```
Mehdi-Bendriss, line 290: Does this mean that the deletion of a config entry is equivalent to setting it as `None`?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -253,34 +278,68 @@ def _disable_if_needed(self, plugin: OpenSearchPlugin) -> bool:
...
+        )
+        new_conf = {
+            **current_settings,
+            **to_remove,
+            **config.config_entries_to_add,
+        }
```
Mehdi-Bendriss, line 296: wouldn't this result everytime in the same dict per plugin?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -311,32 +370,28 @@ def _user_requested_to_enable(self, plugin: OpenSearchPlugin) -> bool:
...
+            if current_settings != new_conf:
                 return False
 
-            # Using sets to guarantee matches; stored_plugin_conf will be a dict
-            if isinstance(plugin_conf, list):
-                return set(plugin_conf) == set(stored_plugin_conf)
-            else:  # plugin_conf is a dict
-                return plugin_conf == stored_plugin_conf
+            # Now, focus in on the keystore part
+            keys_available = self._keystore.list()
```
Mehdi-Bendriss, line 384: Does adding/remove a key warrant a restart?

Mehdi-Bendriss (commented):
Thanks Pedro - I left some questions and comments

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: @Mehdi-Bendriss the last always takes precedence. That is how it can look:
```
>>> current_settings = {1:2, 3:4}
>>> to_remove = {3: None}
>>> to_add = {3: 5}
>>> print({**current_settings, **to_remove})
{1: 2, 3: None}
>>> print({**current_settings, **to_remove, **to_add})
{1: 2, 3: 5}
>>> print(dict(current_settings | to_remove | to_add))
{1: 2, 3: 5}
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -129,25 +132,40 @@ def _extra_conf(self, plugin_data: Dict[str, Any]) -> Optional[Dict[str, Any]]:
             }
         return {**self._charm_config, "opensearch-version": self._opensearch.version}
 
-    def run(self) -> bool:
-        """Runs a check on each plugin: install, execute config changes or remove.
+    def check_plugin_manager_ready(self) -> bool:
+        """Checks if the plugin manager is ready to run."""
+        if not (self._charm.opensearch.is_started() and self._charm.opensearch.is_node_up()):
+            raise OpenSearchNotFullyReadyError()
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss we want to run it after `_start_opensearch` has finished, however the `config-changed` happens right after the `start` event and the start event is almost always being deferred. This check ensures we can have:
1. Should I request a restart `True | False`; or a third option (via exception): we are not ready yet to decide.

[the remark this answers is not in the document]
phvalguima: Actually, in the `check_plugin_manager_ready`, you're right. Added it.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: @Mehdi-Bendriss yes, +1. That was my plan for the /cluster/setting, just like we were discussing to do with the `opensearch-plugin list`.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: @Mehdi-Bendriss no, only writing to the config files, why? This is about deciding the current plugin status and if we have the keys written to the keystore

lib/charms/opensearch/v0/helper_cluster.py
phvalguima: I am not sure this is always the case, tbh. There is nothing in the docs that gives me that guarantee

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: That logic is to decide if a plugin is enabled, i.e. (1) it exists; and (2) the config options that enable it are present. That also includes the keys in the keystore. This method does not decide if we need a restart or not.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: I am moving this task to DPE-3603.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
Mehdi-Bendriss: I think this is my point. If we're always adding `to_add` after `to_remove` - why add `to_remove` at all?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
Mehdi-Bendriss: I'm probably missing something, but why not simply use the keys of `to_remove` to drop them from the dict directly? why go with `key: None`?

lib/charms/opensearch/v0/helper_cluster.py
phvalguima: Okay, I will set to return only the "transient" value, as this is what the docs in ES states it overrides everything else: https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-update-settings.html

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: No reason other than me trying to put everything on a single-liner, lol.
I will delete the keys from the cluster setting

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: Not if they are trying to manipulate different keys

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: I can, but then I need a deepcopy of current_settings, as I use it later to build the difference between old and new configuration dicts.

lib/charms/opensearch/v0/helper_cluster.py
phvalguima: Hey @Mehdi-Bendriss, no luck :(

After this change, I can see our CI failed, and there was a reboot early on:
```
DEBUG unit.opensearch/1.juju-log server.go:325 opensearch-peers:0: Difference between current and new configuration: 
{'in current but not in new_conf': {}, 'in new_conf but not in current': {'knn.plugin.enabled': 'true'}, 'in both but different values': {}}
DEBUG unit.opensearch/1.juju-log server.go:325 opensearch-peers:0: Status: installed
```

Readding the original for (on a simplified structure, but still the same basic logic), resolves the issue: https://pastebin.ubuntu.com/p/8R3JJmVgV3/

Mehdi-Bendriss (approved):

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -255,34 +269,65 @@ def _disable_if_needed(self, plugin: OpenSearchPlugin) -> bool:
...
+            if key_to_del in new_conf:
+                del new_conf[key_to_del]
+        new_conf |= config.config_entries_to_add
+
+        logger.debug(
+            "Difference between current and new configuration: \n"
+            + str(
+                {
+                    "in current but not in new_conf": {
+                        k: v for k, v in current_settings.items() if k not in new_conf.keys()
```
dmitry-ratushnyy, line 290: nits: you can use set for this 😄

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -320,32 +365,28 @@ def _user_requested_to_enable(self, plugin: OpenSearchPlugin) -> bool:
...
+        Check if the configuration from enable() is present or not.
         """
         try:
-            plugin_conf = plugin.disable().config_entries_to_del
-            keystore_conf = plugin.disable().secret_entries_to_del
-            stored_plugin_conf = self._opensearch_config.get_plugin(plugin_conf)
-
-            if any(k not in self._keystore.list() for k in keystore_conf):
+            current_settings, new_conf = self._compute_settings(plugin.config())
+            if current_settings != new_conf:
```
dmitry-ratushnyy, line 375: is this check to find that the dict contains the same elements or ?

dmitry-ratushnyy (approved):

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: I think it is worthy to log the values as well.

PR #181, opened 2024-02-06
[DPE-3660] Add backups testing infra

phvalguima:
Re adds the backup testing infra + AWS testing for backup, as described in [DPE-3660](https://warthogs.atlassian.net/browse/DPE-3660), on top of the new CI testing infra.

It is broken in two types of tests:
1. Per cloud test: backup, restore, relation removal/readd, disaster recovery
2. All clouds: test configuration changes and backup / restore, test expected API call responses

The tests will use different groups, so they can run in parallel.

[DPE-3660]: https://warthogs.atlassian.net/browse/DPE-3660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -9,6 +9,7 @@ import time
 from typing import Dict, List, Optional
 
 from charms.opensearch.v0.models import Node
+from charms.opensearch.v0.opensearch_backups import S3_REPOSITORY
 from pytest_operator.plugin import OpsTest
 from tenacity import (
     RetryError,
@@ -23,6 +24,7 @@ from ..helpers import (
     get_application_unit_ids,
     get_application_unit_ids_hostnames,
     get_application_unit_ids_ips,
+    get_leader_unit_ip,
     http_request,
     juju_version_major,
     run_action,
@@ -456,10 +458,11 @@ async def print_logs(ops_test: OpsTest, app: str, unit_id: int, msg: str) -> str
     return msg
 
 
-async def wait_for_backup(ops_test, leader_id):
+async def wait_for_backup_system_to_settle(ops_test, leader_id, unit_ip):
     """Waits the backup to finish and move to the finished state or throws a RetryException."""
     for attempt in Retrying(stop=stop_after_attempt(8), wait=wait_fixed(15)):
         with attempt:
+            # First, check if current backups are finished
             action = await run_action(
                 ops_test, leader_id, "list-backups", params={"output": "json"}
             )
@@ -467,17 +470,12 @@ async def wait_for_backup(ops_test, leader_id):
             # namespace(status='completed', response={'return-code': 0, 'backups': '{"1": ...}'})
             backups = json.loads(action.response["backups"])
             logger.debug(f"Backups recovered: {backups}")
-            if action.status == "completed" and len(backups) > 0:
+            if action.status != "completed" or len(backups) == 0:
+                raise Exception("Failed to retrieve backup list or list is empty")
+            else:
                 logger.debug(f"list-backups output: {action}")
-                return
 
-            raise Exception("Backup not finished yet")
-
-
-async def wait_for_restore(ops_test, unit_ip):
-    """Waits the backup to finish and move to the finished state or throws a RetryException."""
-    for attempt in Retrying(stop=stop_after_attempt(8), wait=wait_fixed(15)):
-        with attempt:
+            # Now, check if we have finished the restore
             indices_status = await http_request(
                 ops_test,
                 "GET",
@@ -487,7 +485,20 @@ async def wait_for_restore(ops_test, unit_ip):
                 # Now, check the status of each shard
                 for shard in info["shards"]:
                     if shard["type"] == "SNAPSHOT" and shard["stage"] != "DONE":
-                        raise Exception()
+                        raise Exception(f"Recovery failed for shard {shard}")
+
+            raise Exception("Backup system not settled yet")
+
+
+async def delete_backup(ops_test: OpsTest, backup_id: int) -> None:
+    """Deletes a backup."""
+    # Now, check if we have finished the restore
+    unit_ip = await get_leader_unit_ip(ops_test)
+    await http_request(
+        ops_test,
+        "DELETE",
+        f"https://{unit_ip}:9200/_snapshot/{S3_REPOSITORY}/{backup_id}",
+    )
 
 
 async def start_and_check_continuous_writes(ops_test: OpsTest, unit_ip: str, app: str) -> bool:
@@ -509,18 +520,25 @@ async def start_and_check_continuous_writes(ops_test: OpsTest, unit_ip: str, app
     return result.count > initial_count
 
 
-async def backup_cluster(ops_test: OpsTest, leader_id: int) -> str:
+async def create_backup(ops_test: OpsTest, leader_id: int, unit_ip: str) -> bool:
     """Runs the backup of the cluster."""
     action = await run_action(ops_test, leader_id, "create-backup")
     logger.debug(f"create-backup output: {action}")
 
-    await wait_for_backup(ops_test, leader_id)
-    return int(action.response["backup-id"])
+    await wait_for_backup_system_to_settle(ops_test, leader_id, unit_ip)
+    return action.status == "completed"
 
 
-async def restore_cluster(ops_test: OpsTest, backup_id: int, unit_ip: str, leader_id: int) -> bool:
+async def restore(ops_test: OpsTest, backup_id: int, unit_ip: str, leader_id: int) -> bool:
     action = await run_action(ops_test, leader_id, "restore", params={"backup-id": backup_id})
     logger.debug(f"restore output: {action}")
 
-    await wait_for_restore(ops_test, unit_ip)
+    await wait_for_backup_system_to_settle(ops_test, leader_id, unit_ip)
     return action.status == "completed"
+
+
+async def list_backups(ops_test: OpsTest, leader_id: int) -> Dict[str, str]:
+    action = await run_action(ops_test, leader_id, "list-backups", params={"output": "json"})
+    assert action.status == "completed"
+    backups = action.response["backups"]
+    return backups
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -534,7 +534,11 @@ async def create_backup(ops_test: OpsTest, leader_id: int, unit_ip: str) -> int:
 
 
 async def restore(ops_test: OpsTest, backup_id: int, unit_ip: str, leader_id: int) -> bool:
-    action = await run_action(ops_test, leader_id, "restore", params={"backup-id": backup_id})
+    """Restores a backup."""
+    id = backup_id
+    if not isinstance(backup_id, int):
+        id = int(backup_id)
+    action = await run_action(ops_test, leader_id, "restore", params={"backup-id": id})
     logger.debug(f"restore output: {action}")
 
     await wait_for_backup_system_to_settle(ops_test, leader_id, unit_ip)
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -514,8 +514,9 @@ async def start_and_check_continuous_writes(ops_test: OpsTest, unit_ip: str, app
     writer = ContinuousWrites(ops_test, app, initial_count=initial_count)
     await writer.start()
     time.sleep(10)
-    result = await writer.stop()
-    return result.count > initial_count
+    await assert_continuous_writes_consistency(ops_test, writer, app)
+    # Clear the writer manually, as we are not using the conftest c_writes_runner to do so
+    await writer.clear()
 
 
 async def create_backup(ops_test: OpsTest, leader_id: int, unit_ip: str) -> str:
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -544,19 +544,14 @@ async def list_backups(ops_test: OpsTest, leader_id: int) -> Dict[str, str]:
 
 
 async def assert_cwrites_backup_consistency(
-    ops_test: OpsTest, app: str, leader_id: int, unit_ip: str, backup_id: str, loss: float = 0.25
+    ops_test: OpsTest, app: str, leader_id: int, unit_ip: str, backup_id: str
 ) -> None:
-    """Ensures that continuous writes index has at least the value below.
-
-    assert new_count >= <current-doc-count> * (1 - loss) documents.
-    """
+    """Ensures that continuous writes index has at least the value below."""
     original_count = await index_docs_count(ops_test, app, unit_ip, ContinuousWrites.INDEX_NAME)
     # As stated on: https://discuss.elastic.co/t/how-to-parse-snapshot-dat-file/218888,
     # the only way to discover the documents in a backup is to recover it and check
     # on opensearch.
     # The logic below will run over each backup id, restore it and ensure continuous writes
-    # index loss is within the "loss" parameter.
     assert await restore(ops_test, backup_id, unit_ip, leader_id)
     new_count = await index_docs_count(ops_test, app, unit_ip, ContinuousWrites.INDEX_NAME)
-    # We expect that new_count has a loss of documents and the numbers are different.
-    assert new_count >= int(original_count * (1 - loss)) and new_count != original_count
+    assert new_count > 0 and new_count != original_count
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -566,5 +566,9 @@ async def assert_cwrites_backup_consistency(
     # index loss is within the "loss" parameter.
     assert await restore(ops_test, backup_id, unit_ip, leader_id)
     new_count = await index_docs_count(ops_test, app, unit_ip, ContinuousWrites.INDEX_NAME)
+    logger.info(
+        f"Testing restore for {ContinuousWrites.INDEX_NAME} - "
+        f"original count pre-restore: {original_count}, and now, new count: {new_count}"
+    )
     # We expect that new_count has a loss of documents and the numbers are different.
     assert new_count >= int(original_count * (1 - loss)) and new_count < original_count
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -553,12 +553,6 @@ async def assert_cwrites_backup_consistency(
     original_count = await index_docs_count(ops_test, app, unit_ip, ContinuousWrites.INDEX_NAME)
     logger.info(f"cwrites current count is: {original_count}")
 
-    # Now, remove the index to ensure we are not counting the same documents twice
-    resp = await http_request(
-        ops_test, "DELETE", f"https://{unit_ip}:9200/{ContinuousWrites.INDEX_NAME}"
-    )
-    logger.info(f"Index deletion response: {resp}")
-
     # As stated on: https://discuss.elastic.co/t/how-to-parse-snapshot-dat-file/218888,
     # the only way to discover the documents in a backup is to recover it and check
     # on opensearch.
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -474,7 +474,7 @@ async def wait_for_backup(ops_test, leader_id):
             raise Exception("Backup not finished yet")
 
 
-async def wait_restore_finish(ops_test, unit_ip):
+async def wait_for_restore(ops_test, unit_ip):
     """Waits the backup to finish and move to the finished state or throws a RetryException."""
     for attempt in Retrying(stop=stop_after_attempt(8), wait=wait_fixed(15)):
         with attempt:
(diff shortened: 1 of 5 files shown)


tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+@pytest.mark.parametrize(
+    "cloud_name",
+    [
+        (pytest.param("microceph", marks=pytest.mark.group("microceph"))),
+        (pytest.param("aws", marks=pytest.mark.group("aws"))),
+    ],
+)
+@pytest.mark.abort_on_fail
+async def test_restore_cluster_after_app_destroyed(
+    ops_test: OpsTest, cloud_configs, cloud_credentials, cloud_name
```
Mehdi-Bendriss, line 290: missing type hints

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+) -> None:
+    """Deletes the entire OpenSearch cluster and redeploys from scratch.
+
+    Restores the backup and then checks if the same TEST_BACKUP_INDEX is there.
+    """
+    global global_backup_id
+
+    app = (await app_name(ops_test)) or APP_NAME
+
```
Mehdi-Bendriss, line 299: we need to get and store the count of documents of the target backup here

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    await _configure_s3(ops_test, config, cloud_credentials[cloud_name], app)
+
+    logger.info("Restoring backup")
+    assert await restore_cluster(
+        ops_test,
+        global_backup_id,  # backup_id
+        unit_ip,
+        leader_id,
+    )
+    assert await start_and_check_continuous_writes(ops_test, unit_ip, app)
```
Mehdi-Bendriss, line 278: is this still relevant?

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+
+    logger.info("Creating backup")
+    backup_id = await backup_cluster(
+        ops_test,
+        leader_id,
+    )
+    assert backup_id > 0
+    if cloud_name not in backups_by_cloud:
+        backups_by_cloud[cloud_name] = []
+    backups_by_cloud[cloud_name].append(backup_id)
```
Mehdi-Bendriss, line 342: this can be deleted I think

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+        leader_id,
+    )
+
+    logger.info("Creating backup")
+    backup_id = await backup_cluster(
+        ops_test,
+        leader_id,
+    )
+    assert backup_id > 0
+    if cloud_name not in backups_by_cloud:
```
Mehdi-Bendriss, line 340: we need count the restored documents and compare them with the original count of the original cluster

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    await ops_test.model.applications[app].destroy_relation(
+        "s3-credentials", f"{S3_INTEGRATOR}:s3-credentials"
+    )
+    await ops_test.model.wait_for_idle(
+        apps=[app],
+        status="active",
+        timeout=1400,
+        idle_period=IDLE_PERIOD,
+    )
+
```
Mehdi-Bendriss, line 376: can we:
1. call `list-backups` and verify output
2. attempt to create a new backup and verify output and behavior

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    backups_by_cloud[cloud_name].append(backup_id)
+
+    for id in [global_backup_id, backup_id]:
+        logger.info(f"Restoring backup-id: {id}")
+        assert await restore_cluster(
+            ops_test,
+            id,  # backup_id of the 1st backup and then the latest backup
+            unit_ip,
+            leader_id,
+        )
```
Mehdi-Bendriss, line 406: missing previous checks we mentioned in other comments

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+
+    my_charm = await ops_test.build_charm(".")
+    await ops_test.model.set_config(MODEL_CONFIG)
+    # Deploy TLS Certificates operator.
+    config = {"ca-common-name": "CN_CA"}
+    await ops_test.model.deploy(TLS_CERTIFICATES_APP_NAME, channel="stable", config=config),
+
+    s3_charm = S3_INTEGRATOR
+    # Convert to integer as environ always returns string
+    app_num_units = 3
```
Mehdi-Bendriss, line 438: same comment as above

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    )
+    await ops_test.model.wait_for_idle(
+        apps=[S3_INTEGRATOR],
+        status="active",
+        timeout=TIMEOUT,
+    )
+    await ops_test.model.wait_for_idle(
+        apps=[app],
+        timeout=TIMEOUT,
+    )
```
Mehdi-Bendriss, line 505: these 2 can be merged in one by using `wait_until`..

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+
+    # Not using _configure_s3 as this method will cause opensearch to block
+    await ops_test.model.applications[S3_INTEGRATOR].set_config(config)
+    await run_action(
+        ops_test,
+        0,
+        "sync-s3-credentials",
+        params=credentials,
+        app=S3_INTEGRATOR,
+    )
```
Mehdi-Bendriss, line 496: do we expect a charm status change here? like blocked. If yes, we should test on it.

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+        ops_test, "GET", f"https://{unit_ip}:9200/_snapshot/{S3_REPO_NAME}", json_resp=True
+    )
+    logger.info(f"Response: {resp}")
+    assert resp["status"] == 404
+    assert "repository_missing_exception" in resp["error"]["type"]
+
+
+@pytest.mark.group("all")
+@pytest.mark.abort_on_fail
+async def test_repo_is_misconfigured(ops_test: OpsTest) -> None:
```
Mehdi-Bendriss, line 472:
```suggestion
async def test_wrong_s3_credentials(ops_test: OpsTest) -> None:
```

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    assert "repository_exception" in resp["error"]["type"]
+    assert "Could not determine repository generation from root blobs" in resp["error"]["reason"]
+
+
+@pytest.mark.group("all")
+@pytest.mark.abort_on_fail
+async def test_backup_and_switch_configs(
+    ops_test: OpsTest,
+    cloud_configs,
+    cloud_credentials,
```
Mehdi-Bendriss, line 521: missing type hints.

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    )
+    logger.info(f"Response: {resp}")
+    assert resp["status"] == 500
+    assert "repository_exception" in resp["error"]["type"]
+    assert "Could not determine repository generation from root blobs" in resp["error"]["reason"]
+
+
+@pytest.mark.group("all")
+@pytest.mark.abort_on_fail
+async def test_backup_and_switch_configs(
```
Mehdi-Bendriss, line 518: i'm not entirely sure I understand what we're testing here?

Mehdi-Bendriss (changes requested):
Thanks Pedro. Good first pass. I left some comments. In general - we need to ensure that we're covering on our testing the data aspect and charm's actions more thoroughly. Please refer to the mongodb charm for ideas.

tests/integration/ha/test_backups.py
phvalguima: @Mehdi-Bendriss this tests config changed, as: I change `s3-integrator` to point to: microceph, then aws and an incorrect config; and see how it behaves

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+)
+from .helpers_data import index_docs_count
+
+logger = logging.getLogger(__name__)
+
+
+S3_REPO_NAME = "s3-repository"
+
+
+backups_by_cloud = {}
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss I have one test case which iterates over all the clouds (the config-changed one). Then, I need to know which backups correspond to which clouds.

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+        }
+    return results
+
+
+@pytest.fixture(scope="session", autouse=True)
+def clean_backups_from_buckets(github_secrets, cloud_configs, cloud_credentials):
+    """Teardown to clean up created backups from clouds."""
+    yield
+
+    creds = cloud_credentials.copy()
```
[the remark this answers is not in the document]
phvalguima: Indeed, that was there because originally GCS, it needed to change the credentials.

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    global global_backup_id
+
+    app = (await app_name(ops_test)) or APP_NAME
+    leader_id = await get_leader_unit_id(ops_test)
+    unit_ip = await get_leader_unit_ip(ops_test)
+    config = cloud_configs[cloud_name]
+
+    logger.info(f"Syncing credentials for {cloud_name}")
+    await _configure_s3(ops_test, config, cloud_credentials[cloud_name], app)
+
```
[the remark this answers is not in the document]
phvalguima: I added to the `ha/helpers` the `list_backups` method, is that what you mean?

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    assert global_backup_id > 0
+    if cloud_name not in backups_by_cloud:
+        backups_by_cloud[cloud_name] = []
+    backups_by_cloud[cloud_name].append(global_backup_id)
+
+    # Comparing the number of docs without stopping c_writes
+    initial_count = await index_docs_count(ops_test, app, unit_ip, ContinuousWrites.INDEX_NAME)
+    time.sleep(5)
+    count = await index_docs_count(ops_test, app, unit_ip, ContinuousWrites.INDEX_NAME)
+    assert count > initial_count
```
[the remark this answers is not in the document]
phvalguima: I removed this chunk of code from the original `test_backup_and_list`

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss: now each test that restores, will actually restore against all backups available and then check. 

Also, the sequence of tests takes at least 3x backups of the entire cluster, which means we do test multiple times backing up a cluster post restore.

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    config = cloud_configs[cloud_name]
+
+    logger.info(f"Syncing credentials for {cloud_name}")
+    await _configure_s3(ops_test, config, cloud_credentials[cloud_name], app)
+
+    logger.info("Creating backup")
+    global_backup_id = await backup_cluster(
+        ops_test,
+        leader_id,
+    )
```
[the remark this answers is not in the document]
phvalguima: I added a check to the `create_backup`, and it returns a bool. Is that what you mean here?

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    logger.info(f"Syncing credentials for {cloud_name}")
+    await _configure_s3(ops_test, config, cloud_credentials[cloud_name], app)
+
+    logger.info("Restoring backup")
+    assert await restore_cluster(
+        ops_test,
+        global_backup_id,  # backup_id
+        unit_ip,
+        leader_id,
+    )
```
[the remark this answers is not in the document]
phvalguima: I modified the tests, so for every restore I check if the new index has the same number of documents as the original index. There are also multiple backup / restore / backup of the same index (the ContinuousWrite one) in the entire test.

tests/integration/ha/test_backups.py
phvalguima: it is grabbed from the opensearch APIs.

tests/integration/ha/test_backups.py
phvalguima: Yes, it proves the cluster can receive writes post-restore

tests/integration/ha/test_backups.py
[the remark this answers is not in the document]
Mehdi-Bendriss: I mean that given we are about to test backups, we should probably know what our environment has already in the context of backups. 
One way to do this is to invoke the `list-backups` charm action which should yield an empty list, so we should assert on that

tests/integration/ha/test_backups.py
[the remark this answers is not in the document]
Mehdi-Bendriss: the `create-backup` backup charm action returns a status alongside the backup-id, as shown [here](https://github.com/canonical/opensearch-operator/blob/main/lib/charms/opensearch/v0/opensearch_backups.py#L406) - I meant to say that we should test to ensure that the `action.status == "Backup is running."`

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,568 @@
...
+    global global_backup_id
+
+    unit_ip = await get_leader_unit_ip(ops_test)
+    app = (await app_name(ops_test)) or APP_NAME
+    leader_id = await get_leader_unit_id(ops_test)
+    config = cloud_configs[cloud_name]
+
+    logger.info(f"Syncing credentials for {cloud_name}")
+    await _configure_s3(ops_test, config, cloud_credentials[cloud_name], app)
+
```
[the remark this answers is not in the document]
phvalguima: I merged the tests "test_backup_and_list" and "test_restore" in a single one. Now, we add data and then restore and check.

tests/integration/ha/test_backups.py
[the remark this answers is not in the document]
phvalguima: Added a new method that runs this check: ```assert_cwrites_backup_consistency```

tests/integration/ha/test_backups.py
phvalguima: Added a global variable for that: `cwrites_backup_doc_count`. This way, when we do the DR test, we run against each of the backups we have knowledge of and double check with the original # of docs present inthis global variable.

We need a global variable b/c backups are executed on separated tests and then we restore each on the DR test.

tests/integration/ha/helpers.py
```diff
@@ -512,7 +512,6 @@ async def start_and_check_continuous_writes(ops_test: OpsTest, unit_ip: str, app
 async def backup_cluster(ops_test: OpsTest, leader_id: int) -> int:
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss it does not work, Juju will translate a `"1"` into a `1` (int) before calling the action. So, we need to actually have a string for Juju not to translate it. Here is one example (after setting "backup-id as string in actions.yaml):

```
$ juju run opensearch/leader restore backup-id="1"

Operation 37 failed to schedule any tasks:
validation failed: (root).backup-id : must be of type string, given 1

$ juju run opensearch/leader restore backup-id=1

Operation 39 failed to schedule any tasks:
validation failed: (root).backup-id : must be of type string, given 1
```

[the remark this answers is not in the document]
Mehdi-Bendriss: I was under the impression that the new backup ID format from [your other PR](https://github.com/canonical/opensearch-operator/pull/200) is a date string: `YYYY-MM-DDTHH:MM:SSZ`..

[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss that PR is built on top of this one. We need to merge this first.

tests/integration/ha/helpers.py
```diff
@@ -504,24 +514,62 @@ async def start_and_check_continuous_writes(ops_test: OpsTest, unit_ip: str, app
     )
     writer = ContinuousWrites(ops_test, app, initial_count=initial_count)
     await writer.start()
-    time.sleep(10)
-    result = await writer.stop()
-    return result.count > initial_count
+    time.sleep(60)
+    # Ensure we have writes happening and the index is consistent at the end
+    await assert_continuous_writes_increasing(writer)
+    await assert_continuous_writes_consistency(ops_test, writer, app)
```
MiaAltieri, line 520: just FYI if this check fails then writes won't be cleared. (i.e. if a customer runs this test on their deployed cluster then they will have a bunch of "junk/test" writes. Adding a fixure to tests that clears writes after tests can be useful

tests/integration/ha/helpers.py
```diff
@@ -504,24 +514,62 @@ async def start_and_check_continuous_writes(ops_test: OpsTest, unit_ip: str, app
...
 
-async def restore_cluster(ops_test: OpsTest, backup_id: int, unit_ip: str, leader_id: int) -> bool:
-    action = await run_action(ops_test, leader_id, "restore", params={"backup-id": backup_id})
+async def restore(ops_test: OpsTest, backup_id: int, unit_ip: str, leader_id: int) -> bool:
+    """Restores a backup."""
+    id = backup_id
+    if not isinstance(backup_id, int):
+        id = int(backup_id)
+    action = await run_action(ops_test, leader_id, "restore", params={"backup-id": id})
     logger.debug(f"restore output: {action}")
```
MiaAltieri, line 542: also it is necessary to check that the writes are actually restored

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,570 @@
...
+    return results
+
+
+@pytest.fixture(scope="session", autouse=True)
+def remove_backups(
+    # ops_test: OpsTest,
+    cloud_configs: Dict[str, Dict[str, str]],
+    cloud_credentials: Dict[str, Dict[str, str]],
+):
+    """Remove previously created backups from the cloud-corresponding bucket."""
```
MiaAltieri, line 119: I believe you can set up your cloud to autoclear every day

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,570 @@
...
+    await assert_continuous_writes_increasing(c_writes)
+    await assert_continuous_writes_consistency(ops_test, c_writes, app)
+    await assert_cwrites_backup_consistency(ops_test, app, leader_id, unit_ip, backup_id)
+    global cwrites_backup_doc_count
+    cwrites_backup_doc_count[backup_id] = await index_docs_count(
+        ops_test,
+        app,
+        unit_ip,
+        ContinuousWrites.INDEX_NAME,
+    )
```
MiaAltieri, line 254: this test is called `...and_restore` - does this test actually check the restore?

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,570 @@
...
+        assert await restore(ops_test, backup_id, unit_ip, leader_id)
+        count = await index_docs_count(ops_test, app, unit_ip, ContinuousWrites.INDEX_NAME)
+
+        # Ensure we have the same doc count as we had on the original cluster
+        assert count == cwrites_backup_doc_count[backup_id]
+
+        # restart the continuous writes and check the cluster is still accessible post restore
+        assert await start_and_check_continuous_writes(ops_test, unit_ip, app)
+
+    # Now, try a backup & restore with continuous writes
```
MiaAltieri, line 400: where is the restore?

tests/integration/ha/test_backups.py
```diff
@@ -0,0 +1,570 @@
...
+
+        assert (
+            backup_id := await create_backup(
+                ops_test,
+                leader_id,
+                unit_ip=unit_ip,
+            )
+        ) != ""
+        # continuous writes checks
+        await assert_continuous_writes_increasing(writer)
```
MiaAltieri, line 566: where is the restore?

MiaAltieri (commented):
looks great! And very clean - just a few questions

tests/integration/ha/test_backups.py
phvalguima: Maybe, but these are shared across our teams.

tests/integration/ha/test_backups.py
phvalguima: @MiaAltieri yes, the `assert_cwrites_backup_consistency` restores the indices in the DB and then compares with the current values we have.

tests/integration/ha/helpers.py
phvalguima: These are helpers, they do not run `assert`.

tests/integration/ha/test_backups.py
phvalguima: Indeed, there is a method to do so, but its name was not really helpful! I renamed it to `assert_restore_indices_and_compare_consistency` so one can clearly see this check

tests/integration/ha/test_backups.py
phvalguima: The name of the method is pretty misleading thou :)
Therefore, renamed it to `assert_restore_indices_and_compare_consistency`

MiaAltieri: you are amazing! Thank you for doing that 🙇

MiaAltieri (approved):
great work!! <3 looks awesome

Mehdi-Bendriss (approved):

PR #186, opened 2024-02-16
[DPE-3578] clear backup status

phvalguima:
Move the status messages in `opensearch_backup` module to comply with `charm.status.{set,clear}`. All text has been moved to `constant_charms.py`.

This PR also simplifies the exception structure, where the plugins should log their own errors if that is not supposed to come back to the plugin manager and the charm itself. The exception `OpenSearchPluginClusterNotReadyError` has been abandoned in favor of the default exception for the same type of error in `constant_exceptions`.

Tests have been updated accordingly.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_backups.py
+++ b/lib/charms/opensearch/v0/opensearch_backups.py
@@ -50,7 +50,15 @@ import math
 from typing import Any, Dict, List, Set, Tuple
 
 from charms.data_platform_libs.v0.s3 import S3Requirer
-from charms.opensearch.v0.constants_charm import PluginConfigError, RestoreInProgress
+from charms.opensearch.v0.constants_charm import (
+    BackupConfigureStart,
+    BackupDeferRelBrokenAsInProgress,
+    BackupInDisabling,
+    BackupSetupFailed,
+    BackupSetupStart,
+    PluginConfigError,
+    RestoreInProgress,
+)
 from charms.opensearch.v0.helper_cluster import ClusterState, IndexStateEnum
 from charms.opensearch.v0.helper_enums import BaseStrEnum
 from charms.opensearch.v0.opensearch_exceptions import (
@@ -65,7 +73,7 @@ from charms.opensearch.v0.opensearch_plugins import (
 )
 from ops.charm import ActionEvent
 from ops.framework import EventBase, Object
-from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
+from ops.model import BlockedStatus, MaintenanceStatus, WaitingStatus
 from tenacity import RetryError, Retrying, stop_after_attempt, wait_fixed
 
 # The unique Charmhub library identifier, never change it
@@ -482,7 +490,7 @@ class OpenSearchBackup(Object):
         if self.s3_client.get_s3_connection_info().get("tls-ca-chain"):
             raise NotImplementedError
 
-        self.charm.status.set(WaitingStatus("Setting up backup service"))
+        self.charm.status.set(MaintenanceStatus(BackupSetupStart))
 
         try:
             plugin = self.charm.plugin_manager.get_plugin(OpenSearchBackupPlugin)
@@ -500,9 +508,7 @@ class OpenSearchBackup(Object):
             event.defer()
             return
         except OpenSearchError as e:
-            self.charm.status.set(
-                BlockedStatus("Unexpected error during plugin configuration, check the logs")
-            )
+            self.charm.status.set(BlockedStatus(PluginConfigError))
             # There was an unexpected error, log it and block the unit
             logger.error(e)
             event.defer()
@@ -512,15 +518,17 @@ class OpenSearchBackup(Object):
             PluginState.ENABLED,
             PluginState.WAITING_FOR_UPGRADE,
         ]:
-            self.charm.status.set(WaitingStatus(f"Plugin in state {self._plugin_status}"))
             event.defer()
             return
 
         if not self.charm.unit.is_leader():
             # Plugin is configured locally for this unit. Now the leader proceed.
-            self.charm.status.set(ActiveStatus())
+            self.charm.status.clear(PluginConfigError)
+            self.charm.status.clear(BackupSetupStart)
             return
         self.apply_api_config_if_needed()
+        self.charm.status.clear(PluginConfigError)
+        self.charm.status.clear(BackupSetupStart)
 
     def apply_api_config_if_needed(self) -> None:
         """Runs the post restart routine and API calls needed to setup/disable backup.
@@ -530,14 +538,16 @@ class OpenSearchBackup(Object):
         # Backup relation has been recently made available with all the parameters needed.
         # Steps:
         #     (1) set up as maintenance;
-        self.charm.status.set(MaintenanceStatus("Configuring backup service..."))
+        self.charm.status.set(MaintenanceStatus(BackupConfigureStart))
         #     (2) run the request; and
         state = self._register_snapshot_repo()
         #     (3) based on the response, set the message status
         if state != BackupServiceState.SUCCESS:
-            self.charm.status.set(BlockedStatus(f"Backup setup failed with {state}"))
-        else:
-            self.charm.status.set(ActiveStatus())
+            logger.error(f"Failed to setup backup service with state {state}")
+            self.charm.status.set(BlockedStatus(BackupSetupFailed))
+            return
+        self.charm.status.clear(BackupSetupFailed)
+        self.charm.status.clear(BackupConfigureStart)
 
     def _on_s3_broken(self, event: EventBase) -> None:  # noqa: C901
         """Processes the broken s3 relation.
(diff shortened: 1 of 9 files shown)


phvalguima:
The CI error is caused by a timeout, while waiting for restart: https://github.com/canonical/opensearch-operator/actions/runs/7933066738/job/21680561224

It is unrelated to this change.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -481,38 +490,45 @@ def _on_s3_credentials_changed(self, event: EventBase) -> None:
...
+            logger.warning("s3-changed: cluster not ready yet")
+            event.defer()
+            return
+        except OpenSearchPluginError as e:
+            # This is only possible if we could not apply keys to the store.
+            # That is generally transient.
+            logger.error(e)
+            event.defer()
+            return
         except OpenSearchError as e:
```
Mehdi-Bendriss, line 510: It seems like these 3 `except`s can be merged in one `except` (they log / defer / return)? and set the status if it's of `PluginError` type?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -583,18 +599,22 @@ def _on_s3_broken(self, event: EventBase) -> None:  # noqa: C901
...
                 self.charm.plugin_manager.apply_config(plugin.disable())
+        except OpenSearchNotFullyReadyError:
+            logger.warning("s3-credentials: cluster not ready yet")
+            event.defer()
+            return
+        except OpenSearchPluginError as e:
+            logger.error(e)
+            event.defer()
+            return
         except OpenSearchError as e:
```
Mehdi-Bendriss, line 610: Same comment as above, these 3 may be merged into a single except

Mehdi-Bendriss (commented):
Thanks Pedro, i have a couple of comments

juditnovak (approved):
LGTM

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -458,6 +470,9 @@ def config(self) -> OpenSearchPluginConfig:
             secret_entries_to_del = [...]|{...},
         )
         """
+        if not self._extra_config.get("access-key") or not self._extra_config.get("secret-key"):
+            logger.error("Missing AWS access-key and secret-key configuration")
+            return OpenSearchPluginConfig()
```
Mehdi-Bendriss, line 475: What will happen in this case? where are those missing pre-confitions handled?

Mehdi-Bendriss (commented):
Thanks Pedro - I have a question

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: Hi @Mehdi-Bendriss, handled [here, in opensearch_backups lib](https://github.com/canonical/opensearch-operator/blob/b7187bfeda061a641d8c8480a5b3836db38ea61b/lib/charms/opensearch/v0/opensearch_backups.py#L658). This lib has the view of the entire charm, opensearch, etc; it can decide if this is critical enough or not and set the juju status. The plugin, on the other hand, can only inform the plugin manager "I am not ready to be configured / disabled yet" and that can happen every time the plugin_manager hit `run()`, as it will try to check each plugin in its list.

phvalguima: @Mehdi-Bendriss it should actually raise an error. I fixed that

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -278,32 +294,42 @@ def _on_update_status(self, event):
...
     """Exception thrown when an opensearch plugin is invalid."""
 
-
-class OpenSearchPluginRelationClusterNotReadyError(OpenSearchPluginError):
-    """Exception thrown when making API calls and cluster is not yet ready.
-
-    This exception in specific should be handled by classes managing the relations of plugins.
-    """
+    def __init__(self, msg):
+        super().__init__(msg)
```
Mehdi-Bendriss, line 298: safe to delete

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -278,32 +294,42 @@ def _on_update_status(self, event):
...
 
 
 class OpenSearchPluginMissingDepsError(OpenSearchPluginError):
     """Exception thrown when an opensearch plugin misses installed dependencies."""
 
+    def __init__(self, name, deps):
+        super().__init__(f"Failed to install plugin: {name} - missing dependencies {deps}")
+        self.deps = deps
```
Mehdi-Bendriss, line 306: safe to delete - unless you see a use for them besides logging?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -278,32 +294,42 @@ def _on_update_status(self, event):
...
+
 
 class OpenSearchPluginInstallError(OpenSearchPluginError):
     """Exception thrown when opensearch plugin installation fails."""
 
+    def __init__(self, name):
+        super().__init__("Failed to install plugin: {}".format(name))
```
Mehdi-Bendriss, line 313: This is fine I suppose, though redundant. 
It is already implied by the type of the exception itself on the stacktrace that a plugin failed to install. 
So: `raise OpenSearchPluginInstallError(f"Plugin: {plugin.name}")` means the same thing as `raise OpenSearchPluginInstallError(f"Failed to install plugin: {plugin.name}")`

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -278,32 +294,42 @@ def _on_update_status(self, event):
...
 
 class OpenSearchPluginMissingConfigError(OpenSearchPluginError):
     """Exception thrown when config() or disable() fails to find a config key.
 
     The plugin itself should raise a KeyError, to avoid burden in the plugin development.
     """
 
+    def __init__(self, name, configs: List[str]):
+        super().__init__(f"Plugin {name} is missing configs: {configs}")
+        self.configs = configs
```
Mehdi-Bendriss, line 331: same as above, safe to delete unless needed for something?

Mehdi-Bendriss (commented):
Thanks Pedro - left a few minor comments.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: Keeping only the fields, so we not forget them

Mehdi-Bendriss: It seems that we don't need to keep them. They're only used for logging. Let's remove this constructor altogether and just set the right message with the list of dependencies in the str.

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -279,31 +295,38 @@ class OpenSearchPluginError(OpenSearchError):
...
+    def __init__(self, name, deps):
+        self.name = name
+        self.deps = deps
+
 
 class OpenSearchPluginInstallError(OpenSearchPluginError):
     """Exception thrown when opensearch plugin installation fails."""
 
+    def __init__(self, name):
+        self.name = name
```
Mehdi-Bendriss, line 310: It doesn't seem like we need this constructor?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -279,31 +295,38 @@ class OpenSearchPluginError(OpenSearchError):
...
+
 
 class OpenSearchPluginRemoveError(OpenSearchPluginError):
     """Exception thrown when opensearch plugin removal fails."""
 
+    def __init__(self, name):
+        self.name = name
```
Mehdi-Bendriss, line 317: It doesn't seem like we need this constructor?

lib/charms/opensearch/v0/opensearch_plugins.py
Mehdi-Bendriss: It doesn't seem like we need this constructor? we don't need to keep these attributes as they're only used for logging. Let's remove this constructor altogether and just set the right message with the list of configs in the str.

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -458,6 +481,16 @@ def config(self) -> OpenSearchPluginConfig:
             secret_entries_to_del = [...]|{...},
         )
         """
+        if not self._extra_config.get("access-key") or not self._extra_config.get("secret-key"):
+            raise OpenSearchPluginMissingConfigError(
```
Mehdi-Bendriss, line 485: Let's just construct a string here.

Mehdi-Bendriss (changes requested):
Thanks Pedro. I left a few comments.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: So, I personally do not like to have things undefined. I like well-defined constructs and anyone trying to use these classes will have to follow that pattern. But you're the reviewer, so I will just follow that and remove this definition.

Mehdi-Bendriss (approved):

juditnovak (approved):
LGTM (with the very limited understanding that I've got on the topic).

I'd feel more re-assured though with integration tests added...

PR #197, opened 2024-03-08
[DPE-3647] Pin snap to rev37 (2.10.0-ubuntu1)

### LINKED ISSUE #123: Add snap hold and pin version
We should set snap version and hold settings per juju revision.

phvalguima:
Closes #123

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -29,6 +29,7 @@ from .helpers_deployments import Status, get_application_units, get_unit_hostnam
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 APP_NAME = METADATA["name"]
 
+EXPECTED_SNAP_REVISION = 37
 SERIES = "jammy"
 UNIT_IDS = [0, 1, 2]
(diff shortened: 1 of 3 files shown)


tests/integration/helpers.py
```diff
@@ -29,6 +29,7 @@
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 APP_NAME = METADATA["name"]
 
+EXPECTED_SNAP_REVISION = 37
```
delgod, line 32: just to simplify future changes, I would import revision directly from `src/opensearch.py`.
something like
```
from opensearch import OPENSEARCH_SNAP_REVISION
```
but, it is not blocker

delgod (approved):

tests/integration/helpers.py
phvalguima: Hi @delgod I am not sure that works any longer with the new CI, but I will try it out and make a PR for that

paulomach (approved):
nice test, haven't seen anyone testing for that

PR #200, opened 2024-03-11
[DPE-3784] Fix backup ID format

phvalguima:
Currently, OpenSearch charm creates backups with ID as integer (1, 2, etc). However, that is very misleading for users when it comes to selecting the appropriate backup to restore from.

This PR implements [DPE-3784](https://warthogs.atlassian.net/browse/DPE-3784) and moves OpenSearch to a timestamp ID [that follows PSQL charm pattern](https://github.com/canonical/postgresql-operator/blob/28990d4886aee9e8c2ad2de1e8ecd28e7d49a61f/src/constants.py#L6-L7). It defines two types of timestamps:
1) The user ID: that is defined by the `BACKUP_ID_FORMAT` and provides a timestamp that the user can use to restore
2) The internal ID: given backup/restore is communicated to OpenSearch via HTTP requests, we need to avoid special characters, such as `:`. The internal ID is formatted following `OPENSEARCH_BACKUP_ID_FORMAT`.

This PR also extends unit test coverage to cover more cases of `restore` action, besides updating both unit and integration tests to the new backup ID formats.

[DPE-3784]: https://warthogs.atlassian.net/browse/DPE-3784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -377,10 +390,8 @@ def _on_create_backup_action(self, event: ActionEvent) -> None:  # noqa: C901
             event.fail("Failed: backup service is not configured or busy")
             return
 
-        new_backup_id = None
+        new_backup_id = datetime.now().strftime(OPENSEARCH_BACKUP_ID_FORMAT)
```
Mehdi-Bendriss, line 393: I believe the format should be: `BACKUP_ID_FORMAT`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -270,11 +271,21 @@ def _close_indices_if_needed(self, backup_id: int) -> Set[str]:
         return indices_to_close
 
     def _restore(self, backup_id: int) -> Dict[str, Any]:
```
Mehdi-Bendriss, line 273:
```suggestion
    def _restore(self, backup_id: str) -> Dict[str, Any]:
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -270,11 +271,21 @@ def _close_indices_if_needed(self, backup_id: int) -> Set[str]:
...
+        """Runs the restore and processes the response.
+
+        Backups come in the user-formatted string, YYYY-MM-DDTHH:MM:SSZ.
+        """
         backup_indices = self._list_backups()[backup_id]["indices"]
+        # now, fix the id format for the actual request
+        id = self._format_backup_id(
+            backup_id,
+            format_from=BACKUP_ID_FORMAT,
+            format_to=OPENSEARCH_BACKUP_ID_FORMAT,
```
Mehdi-Bendriss, line 283: Why do we do this?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -428,11 +442,26 @@ def _can_unit_perform_backup(self, event: ActionEvent) -> bool:
             return False
         return not self.is_backup_in_progress()
 
+    def _format_backup_id(
```
Mehdi-Bendriss, line 445: why do we need to wrangle 2 different date formats, why not stick to `BACKUP_ID_FORMAT` everywhere?

tests/integration/ha/test_backups.py
```diff
@@ -269,6 +270,7 @@ async def test_restore_cluster(
     await _configure_s3(ops_test, config, cloud_credentials[cloud_name], app)
 
     logger.info("Restoring backup")
```
Mehdi-Bendriss, line 272: can we delete this?

tests/integration/ha/test_backups.py
```diff
@@ -324,6 +326,7 @@ async def test_restore_cluster_after_app_destroyed(
...
```
Mehdi-Bendriss, line 328: can we delete this?

Mehdi-Bendriss (commented):
Thanks Pedro - I left some comments / questions

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: I have converged both formats, the only remark is that opensearch only accepts minor chars as part of its snapshot ID. Therefore, we have to `lower()` and `upper()` in certain points of the code (as `time.time` uses capital letters and they are more visually easier).

tests/integration/ha/helpers.py
```diff
@@ -522,22 +522,21 @@ async def start_and_check_continuous_writes(ops_test: OpsTest, unit_ip: str, app
...
 
 
-async def restore(ops_test: OpsTest, backup_id: int, unit_ip: str, leader_id: int) -> bool:
+async def restore(ops_test: OpsTest, backup_id: str, unit_ip: str, leader_id: int) -> bool:
     """Restores a backup."""
     id = backup_id
-    if not isinstance(backup_id, int):
-        id = int(backup_id)
+    if not isinstance(id, str):
+        id = str(backup_id)
```
Mehdi-Bendriss, line 539: Can you explain this?

tests/integration/ha/test_backups.py
```diff
@@ -240,7 +240,7 @@ async def test_create_backup_and_restore(
             leader_id,
             unit_ip=unit_ip,
         )
-    ) > 0
+    ) != ""
```
Mehdi-Bendriss, line 243: I would change this to the following, so that it throws an exception if it's not a date format or if the date is wrong:
```
from datetime import datetime

date_before_backup = datetime.now()

backup_id = await create_backup(ops_test, leader_id, unit_ip=unit_ip)
assert datetime.strptime(backup_id, OPENSEARCH_BACKUP_ID_FORMAT) > date_before_backup
```

tests/integration/ha/test_backups.py
```diff
@@ -408,7 +408,7 @@ async def test_restore_to_new_cluster(
...
```
Mehdi-Bendriss, line 411: Same as above

Mehdi-Bendriss (commented):
Thanks Pedro. I have a couple of comments.

tests/integration/ha/helpers.py
phvalguima: I guess it is old intermediary that stayed. Removed.

delgod (approved):

Mehdi-Bendriss (approved):

PR #201, opened 2024-03-15
Use `relation_created` instead of `relation_joined`

carlcsaposs-canonical:
The handler is supposed to run when the relation is created, not when a unit joins the relation

See https://github.com/canonical/mysql-router-k8s-operator/pull/101

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -70,7 +70,7 @@ jobs:
     name: Integration test charm
     needs:
       - lint
-#      - unit-test
+      - unit-test
       - build
     uses: canonical/data-platform-workflows/.github/workflows/integration_test_charm.yaml@v13.1.1
     with:


juditnovak (commented):
I get the rationale, indeed makes more sense to request certificates on `relation-joined`.

Out of curiousity: was there any issue with the previous workflows? I.e. is this a fix or rather preventive to correct our incorrect worklows?

Either way, we sould have the tests passing ;-)

Mehdi-Bendriss:
Hmm, I don't think that's accurate. In OpenSearch, **every** opensearch unit must get its own certificates. And this will not be possible with `RelationCreatedEvent` which would have already been invoked initially - RelationJoinedEvent will guarantee that every unit requests its own certificates. Did i misunderstand something?

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -30,7 +30,7 @@
     generate_csr,
     generate_private_key,
 )
-from ops.charm import ActionEvent, RelationBrokenEvent, RelationJoinedEvent
```
phvalguima, line 33: Hi @carlcsaposs-canonical thanks for this.

I reviewed TLS lib docstring (most recent one), and I noticed two things:
1) They are also recommending to use `relation-joined`: we need a bug there asking to update the docs in this case
2) I also noticed they have more events that we should track in opensearch - the [invalidated events](https://github.com/canonical/self-signed-certificates-operator/blob/7f41b9f22feecf55f556c8e80ea7e6a26e646321/lib/charms/tls_certificates_interface/v3/tls_certificates.py#L159-L165)

For (2), I've filed a separate bug here: #204.

phvalguima (changes requested):

Mehdi-Bendriss (approved):

phvalguima (dismissed):

phvalguima (dismissed):

Mehdi-Bendriss (approved):

Mehdi-Bendriss (approved):

PR #30, opened 2023-02-08
Update data_interfaces lib to include opensearch provider

WRFitch:
## Proposal
Updates data_interfaces lib to include objects and events to manage the opensearch provider lib, as well as TLS. 

## Context
- While the changes in the data_interfaces library aren't unique to this repo, I've chosen to add them here first so we can ensure they're fit for purpose, and then we can add the finished library to the data-integrator repo. 
- The data-integrator changes are duplicated across the opensearch and integration test application charm lib/ directories. This should be an exact duplication. 
- rollingops lib has been updated to update locks on leader_elected. If the leader is replaced during a rolling op, the new leader never receives the process_locks event, so it never runs the op. A PR has been opened for this [here](https://github.com/canonical/charm-rolling-ops/pull/1), but again, I'm testing my changes here before merging.
- This PR is *definitely* getting a merge commit into main so I don't nuke our commit history

## Changelog
- Updated data_interfaces lib to include OpenSearchProvider and OpenSearchRequirer objects, and events associated with opensearch. 
- data_interfaces lib has been updated to the correct ops version
- Added TLS to opensearch provider relation
- Added data_interfaces OpenSearchProvider to provider relation
- Updated endpoint update method to guarantee we have the correct number of endpoints in the databag when scaling
- added data_interfaces OpenSearchRequirer to requirer application test charm
- Deferred starting opensearch if the init script fails to connect
- Possible fix for a bug where update-status fails if we remove the tls-operator
- Fixed some typos and formatting
- Updated rollingops lib to update locks on leader_elected
- Updated integration tests

## Testing
- Added unit tests
- Added more extensive integration tests, which test the following:
  - opensearch API access while scaling
  - Checks that index data persists across relations

PR #205, opened 2024-03-18
Remove unused event handler

carlcsaposs-canonical:
Handler is never called

Handler was first added in #30 but a call to the handler was not added


Mehdi-Bendriss (commented):
Thanks Carl! Looks good, but I wonder if this was rather not a bug that this handler was not invoked

phvalguima:
@Mehdi-Bendriss should we transfer this to a bug? It seems we are missing an `observe` here

Mehdi-Bendriss:
Yeah let's close this one. This is now tracked here https://github.com/canonical/opensearch-operator/pull/311

PR #208, opened 2024-03-20
Remove status setting of plugin configuration on first start of cluster

phvalguima:


Mehdi-Bendriss:
Thanks Pedro - the issue is not exclusively about deferrals:
- The status [gets set regardless](https://github.com/canonical/opensearch-operator/pull/208/files#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R487) even though no plugin configuration happened as shown here:
```
try:
    self.status.set(MaintenanceStatus(PluginConfigStart))
    if self.plugin_manager.run():
        ...
```

The status should be set **only** if a plugin configuration change happened.

Mehdi-Bendriss (approved):

dragomirp (approved):

PR #209, opened 2024-03-25
Re-balance roles deterministically instead of randomly

carlcsaposs-canonical:
(for auto-generated roles)

Algorithm:
- if odd # of nodes: all nodes are cluster manager eligible
- if even # of nodes: node with highest unit number is data node, all other nodes are cluster manager eligible

Replaces old algorithm of `random.choice` one of the nodes that should change roles

This will enable in-place upgrades to be ordered from highest to lowest unit number even if roles are re-balanced (for HA [e.g. because of network cut]) while upgrade is in-progress

That allows upgrade/rollback to be coordinated without leader unit setting information in peer databag, which allows partial rollback even if leader unit is in error state

Context:
https://chat.canonical.com/canonical/pl/zpqoyei54ffr5kwgwkbimizhwo
https://warthogs.atlassian.net/browse/DPE-3878

---

Also, this algorithm ensures that if a unit needs to be restarted for HA while upgrade in-progress, a unit (the highest number unit) on the newer version of the charm/workload will be upgraded. This reduces the likelihood that:
- one of the last units on the older workload version restarts and
- the cluster manager switches to operating with the new workload version (i.e. it stops operating in compatability mode with the old version—see https://discuss.elastic.co/t/rolling-upgrades-master-nodes-voting-config-exclusions/320463/2)

which would mean that units on the old version could not re-connect to the cluster

### DIFF
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -77,6 +77,7 @@ class ClusterTopology:
         """Recompute the configuration of all the nodes (cluster set to auto-generate roles)."""
         if not nodes:
             return {}
+        logger.debug(f"Roles before re-balancing {({node.name: node.roles for node in nodes})=}")
         nodes_by_name = {}
         current_cluster_nodes = []
         for node in nodes:
@@ -103,6 +104,9 @@ class ClusterTopology:
                 unit_number=node.unit_number,
                 temperature=node.temperature,
             )
+        logger.debug(
+            f"Roles after re-balancing {({name: node.roles for name, node in nodes_by_name.items()})=}"
+        )
         return nodes_by_name
 
     @staticmethod


lib/charms/opensearch/v0/helper_cluster.py
```diff
@@ -77,106 +75,39 @@ def get_cluster_settings(
...
-        # logic, in which the max difference between CMs and Non-CMs is 1 node (to keep quorum)
-        unbalanced = len(current_cm_eligible) < len(current_cluster_nodes) - 1
-
-        updated_nodes = []
+        highest_unit_number = max(node.unit_number for node in current_cluster_nodes)
         for node in current_cluster_nodes:
             # we do this in order to remove any non-default role / add any missing default role
-            new_roles = full_roles if unbalanced or node.is_cm_eligible() else base_roles
-            updated = Node(
+            if len(current_cluster_nodes) % 2 == 0 and node.unit_number == highest_unit_number:
```
phvalguima, line 94: @carlcsaposs-canonical nice! Really good simplification! 

I see 2x potential issues we need to address:
1) What happens if I deploy 15 nodes at once? I will have a first run with `len(current_cluster_nodes) == 1`, then `len(...)==2`, etc. Also, this logic is called [here](https://github.com/canonical/opensearch-operator/blob/c0225203320a412c3492ac827ac47b3b5b763787/lib/charms/opensearch/v0/opensearch_base_charm.py#L341) and it will be re-executed multiple times while the cluster is deploying. Why not using `planned_units` here instead? That will return `15` every time. Btw, this is going to be an issue at both start-up and any time we do a scale up.
2) If I have a cluster with 2 nodes and let's assume `opensearch/1` is the charm leader; then I add 3 more nodes. Now, `opensearch/1` will get a new role ("cluster_manager"). However, if [you look in this code path](https://github.com/canonical/opensearch-operator/blob/c0225203320a412c3492ac827ac47b3b5b763787/lib/charms/opensearch/v0/opensearch_base_charm.py#L344), the leader never requests a restart. I think we need to update this line from `elif app_data`, to `if app_data`.

I guess we will not see (2) at deployment time, as we keep deferring the `start` event, which at its end, has a restart request. However, as we need to expand the cluster, then I think (2) may become an issue.

phvalguima (commented):
Nice refactoring proposal!

lib/charms/opensearch/v0/helper_cluster.py
phvalguima: @carlcsaposs-canonical makes sense. The `planned_units` is used along the path, so you will get the total number of units anyway.

phvalguima (dismissed):
LGTM

lib/charms/opensearch/v0/helper_cluster.py
Mehdi-Bendriss, line 94: This works fine for simple deployment where we only have 1 cluster. However, in large deployments (multiple clusters) - it doesn't seem like `if len(current_cluster_nodes) % 2 == 0` is enough, this must be tested against the full list of nodes of the fleet where `node.is_cm_eligible()`.

lib/charms/opensearch/v0/helper_cluster.py
```diff
@@ -77,106 +75,39 @@ def get_cluster_settings(
...
+            else:
+                roles = full_roles
+
+            nodes_by_name[node.name] = Node(
                 name=node.name,
-                roles=new_roles,
+                roles=roles,
                 ip=node.ip,
                 app_name=node.app_name,
+                unit_number=node.unit_number,
```
Mehdi-Bendriss, line 104: nit: Given the unit_number can be easily be inferred from `Node.name` as follows: `int(self.name.split("-")[-1])` it would make sense to not have it as part of of the constructor arguments but rather as a `@validator` on the Node model class itself

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1109,7 +1110,7 @@ def unit_name(self) -> str:
     @property
     def unit_id(self) -> int:
         """ID of the current unit."""
-        return int(self.unit.name.split("/")[1])
+        return int(self.unit.name.split("/")[-1])
```
Mehdi-Bendriss, line 1113: Nice

Mehdi-Bendriss (commented):
Thanks Carl! Great work! I have some comments regarding the logic we use to recompute the roles based on the count of only the nodes of the current cluster, whereas we should be based on the entirety of the fleet

lib/charms/opensearch/v0/helper_cluster.py
Mehdi-Bendriss: Yes - the existing implementation is lacking what I described above

Mehdi-Bendriss: Let's address this in another PR.

Mehdi-Bendriss (approved):
Thanks Carl!

phvalguima (approved):

PR #211, opened 2024-04-02
Use OpenSearch for locking with fallback to peer databag

### LINKED ISSUE #183: Deferring start/restart event causes lock to be prematurely released
I am currently seeing failures such as [this error in the early steps of the CI run](https://github.com/canonical/opensearch-operator/actions/runs/7815857925/job/21320390864?pr=177).

The entire description here is based in this PR: https://github.com/canonical/opensearch-operator/tree/DPE-3352-last-passing-tests

Full deployment:
```
Model              Controller           Cloud/Region         Version  SLA          Timestamp
test-backups-23v7  github-pr-2fa19-lxd  localhost/localhost  3.1.7    unsupported  14:16:34Z

App                       Version  Status       Scale  Charm                     Channel  Rev  Exposed  Message
opensearch                         maintenance      3  opensearch                           0  no       Beginning rolling service
s3-integrator                      active           1  s3-integrator             stable    13  no       
self-signed-certificates           active           1  self-signed-certificates  stable    72  no       

Unit                         Workload  Agent      Machine  Public address  Ports  Message
opensearch/0                 waiting   idle       2        10.241.94.24           Waiting for OpenSearch to start...
opensearch/1*                waiting   executing  3        10.241.94.127          Waiting for OpenSearch to start...
opensearch/2                 error     idle       4        10.241.94.7            hook failed: "opensearch-peers-relation-changed"
s3-integrator/0*             active    idle       1        10.241.94.199          
self-signed-certificates/0*  active    idle       0        10.241.94.194          

Machine  State    Address        Inst id        Base          AZ  Message
0        started  10.241.94.194  juju-ad79f3-0  ubuntu@22.04      Running
1        started  10.241.94.199  juju-ad79f3-1  ubuntu@22.04      Running
2        started  10.241.94.24   juju-ad79f3-2  ubuntu@22.04      Running
3        started  10.241.94.127  juju-ad79f3-3  ubuntu@22.04      Running
4        started  10.241.94.7    juju-ad79f3-4  ubuntu@22.04      Running

Integration provider                   Requirer                           Interface            Type     Message
opensearch:opensearch-peers            opensearch:opensearch-peers        opensearch_peers     peer     
opensearch:service                     opensearch:service                 rolling_op           peer     
s3-integrator:s3-credentials           opensearch:s3-credentials          s3                   regular  
s3-integrator:s3-integrator-peers      s3-integrator:s3-integrator-peers  s3-integrator-peers  peer     
self-signed-certificates:certificates  opensearch:certificates            tls-certificates     regular  

Storage Unit  Storage ID         Type        Pool    Mountpoint                   Size    Status    Message
opensearch/0  opensearch-data/0  filesystem  rootfs  /var/snap/opensearch/common  72 GiB  attached  
opensearch/1  opensearch-data/1  filesystem  rootfs  /var/snap/opensearch/common  72 GiB  attached  
opensearch/2  opensearch-data/2  filesystem  rootfs  /var/snap/opensearch/common  72 GiB  attached  
```

Unit opensearch/2 is failing because all shards disappeared at once. At 14:15:01, everything is there:
```
2024-02-07 14:15:01 DEBUG unit.opensearch/2.juju-log server.go:325 opensearch-peers:1: https://10.241.94.7:9200 "GET /_cat/indices?v HTTP/1.1" 200 606
2024-02-07 14:15:01 DEBUG unit.opensearch/2.juju-log server.go:325 opensearch-peers:1: Getting secret app:admin-password
2024-02-07 14:15:01 DEBUG unit.opensearch/2.juju-log server.go:325 opensearch-peers:1: Starting new HTTPS connection (1): 10.241.94.7:9200
2024-02-07 14:15:01 DEBUG unit.opensearch/2.juju-log server.go:325 opensearch-peers:1: https://10.241.94.7:9200 "GET /_cat/shards?v HTTP/1.1" 200 1774
2024-02-07 14:15:01 DEBUG unit.opensearch/2.juju-log server.go:325 opensearch-peers:1: indices status:
[{'health': 'green', 'status': 'open', 'index': '.opensearch-observability', 'uuid': '_pC83dCcQVuYW5VWnbasfg', 'pri': '1', 'rep': '2', 'docs.count': '0', 'docs.deleted': '0', 'store.size': '416b', 'pri.store.size': '208b'}, {'health': 'green', 'status': 'open', 'index': '.plugins-ml-config', 'uuid': 'RXeUPKcXToOBpRQPM4VQEA', 'pri': '1', 'rep': '2', 'docs.count': '1', 'docs.deleted': '0', 'store.size': '7.7kb', 'pri.store.size': '3.8kb'}, {'health': 'green', 'status': 'open', 'index': '.opendistro_security', 'uuid': 'FXZtZFctTWyEjVt9ZDRqcw', 'pri': '1', 'rep': '2', 'docs.count': '10', 'docs.deleted': '2', 'store.size': '105.2kb', 'pri.store.size': '57.8kb'}]
indices shards:
[{'index': '.plugins-ml-config', 'shard': '0', 'prirep': 'r', 'state': 'STARTED', 'docs': '1', 'store': '0b', 'ip': '10.241.94.7', 'node': 'opensearch-2'}, {'index': '.plugins-ml-config', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': '1', 'store': '3.8kb', 'ip': '10.241.94.127', 'node': 'opensearch-1'}, {'index': '.plugins-ml-config', 'shard': '0', 'prirep': 'r', 'state': 'STARTED', 'docs': '1', 'store': '3.9kb', 'ip': '10.241.94.24', 'node': 'opensearch-0'}, {'index': '.opensearch-observability', 'shard': '0', 'prirep': 'r', 'state': 'STARTED', 'docs': '0', 'store': '0b', 'ip': '10.241.94.7', 'node': 'opensearch-2'}, {'index': '.opensearch-observability', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': '0', 'store': '208b', 'ip': '10.241.94.127', 'node': 'opensearch-1'}, {'index': '.opensearch-observability', 'shard': '0', 'prirep': 'r', 'state': 'STARTED', 'docs': '0', 'store': '208b', 'ip': '10.241.94.24', 'node': 'opensearch-0'}, {'index': '.opensearch-sap-log-types-config', 'shard': '0', 'prirep': 'r', 'state': 'STARTED', 'docs': None, 'store': None, 'ip': '10.241.94.7', 'node': 'opensearch-2'}, {'index': '.opensearch-sap-log-types-config', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': None, 'store': None, 'ip': '10.241.94.127', 'node': 'opensearch-1'}, {'index': '.opensearch-sap-log-types-config', 'shard': '0', 'prirep': 'r', 'state': 'STARTED', 'docs': None, 'store': None, 'ip': '10.241.94.24', 'node': 'opensearch-0'}, {'index': '.opendistro_security', 'shard': '0', 'prirep': 'r', 'state': 'STARTED', 'docs': '10', 'store': '0b', 'ip': '10.241.94.7', 'node': 'opensearch-2'}, {'index': '.opendistro_security', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': '10', 'store': '57.8kb', 'ip': '10.241.94.127', 'node': 'opensearch-1'}, {'index': '.opendistro_security', 'shard': '0', 'prirep': 'r', 'state': 'STARTED', 'docs': '10', 'store': '47.4kb', 'ip': '10.241.94.24', 'node': 'opensearch-0'}]
```

Then, it disappears:
```
2024-02-07 14:16:21 DEBUG unit.opensearch/2.juju-log server.go:325 opensearch-peers:1: https://10.241.94.7:9200 "GET /_cat/shards?v HTTP/1.1" 200 555
2024-02-07 14:16:21 DEBUG unit.opensearch/2.juju-log server.go:325 opensearch-peers:1: indices status:
[{'health': 'green', 'status': 'open', 'index': '.opensearch-observability', 'uuid': '_pC83dCcQVuYW5VWnbasfg', 'pri': '1', 'rep': '0', 'docs.count': '0', 'docs.deleted': '0', 'store.size': '208b', 'pri.store.size': '208b'}, {'health': 'green', 'status': 'open', 'index': '.plugins-ml-config', 'uuid': 'RXeUPKcXToOBpRQPM4VQEA', 'pri': '1', 'rep': '0', 'docs.count': '1', 'docs.deleted': '0', 'store.size': '3.9kb', 'pri.store.size': '3.9kb'}, {'health': 'red', 'status': 'open', 'index': '.opendistro_security', 'uuid': 'FXZtZFctTWyEjVt9ZDRqcw', 'pri': '1', 'rep': '0', 'docs.count': None, 'docs.deleted': None, 'store.size': None, 'pri.store.size': None}]
indices shards:
[{'index': '.opensearch-observability', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': '0', 'store': '208b', 'ip': '10.241.94.7', 'node': 'opensearch-2'}, {'index': '.plugins-ml-config', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': '1', 'store': '3.9kb', 'ip': '10.241.94.7', 'node': 'opensearch-2'}, {'index': '.opensearch-sap-log-types-config', 'shard': '0', 'prirep': 'p', 'state': 'UNASSIGNED', 'docs': None, 'store': None, 'ip': None, 'node': None}, {'index': '.opendistro_security', 'shard': '0', 'prirep': 'p', 'state': 'UNASSIGNED', 'docs': None, 'store': None, 'ip': None, 'node': None}]
```

Looking at the logs, I can see that:
* Opensearch/0: stops at: `2024-02-07 14:15:57 DEBUG unit.opensearch/0.juju-log server.go:325 service:2: Rolling Ops Manager: stop_opensearch called`
* Opensearch/1: stops at: `2024-02-07 14:15:34 DEBUG unit.opensearch/1.juju-log server.go:325 service:2: Rolling Ops Manager: stop_opensearch called`

That means, when the node opensearch/2 is running its routine, all nodes are gone and it fails.

It is happening because both opensearch 0 and 1 are deferring their RunWithLocks event:
* Opensearch/0: RunWithLock via OpenSearchOperatorCharm/on/service_run_with_lock[283] appears multiple time
* Opensearch/1: RunWithLock via OpenSearchOperatorCharm/on/service_run_with_lock[318] appears multiple time

Looking into the `rolling-ops ` code: [there is no post-processing, if the event has been deferred or not](https://github.com/canonical/opensearch-operator/blob/main/lib/charms/rolling_ops/v0/rollingops.py#L405C9-L405C17). It will only react to an exception, which will break the entire hook.

I will report a bug with rolling ops as well, asking for the `run_on_lock` to capture *any* exceptions and double check if the event has been marked for deferral post callback returned.

carlcsaposs-canonical:
### Fix issues with current lock implementation
- Fixes #183: Units can start/restart without rollingops lock (if event deferred, lock released but unit (re)starts on next event anyways).
- Unit can scale down (using "ops lock") while another unit is (re)starting (using separate rollingops lock)

### Prepare for in-place upgrades
- In order to maintain HA, only one unit should start, restart, join cluster (scale up), leave cluster (scale down), or upgrade at a time
- Upgrades adds an additional challenge: if the charm code is upgraded (in addition to the workload), it's possible for the leader unit to go into error state—which prevents coordination of locks via peer databag (current rolling ops implementation) and can block rollback (for other units)

## Options considered
1. Use peer relation databag for lock
    - Unit requests lock by adding data to unit databag -> relation-changed event triggered on leader
    - Leader grants lock by adding data to app databag -> relation-changed event triggered on unit
    - Unit proceeds & releases lock from unit databag
2. Don't use lock for upgrades. Make upgrade order deterministic (so that each unit can independently determine upgrade order & upgrade order does not change) and each unit upgrades when it sees the units before it have upgraded
3. Use opensearch index/document for lock
4. Option 3 but fallback to option 1 if ~~no units online~~ less than 2 units online

Cons of each option:
1. if leader unit is in error state (raises uncaught exception), rollback will be blocked for all units
2. a unit can restart for non-upgrade related reasons (role re-balancing from network cut or scale up) while another unit is restarting to upgrade
3. doesn't work if all units offline (initial start, full cluster crash/network cut)
4. implementation complexity, some edge cases not supported e.g. https://github.com/canonical/opensearch-operator/blob/b1b28c10f2bd70bee4270707b616d8d565b8b616/lib/charms/opensearch/v0/opensearch_locking.py#L227-L228

Pros of each option:
1. only one unit will (re)start at a time, for any reason
2. rollback with `juju refresh` will immediately rollback highest unit even if leader/other units in error state
3. only one unit will (re)start at a time, for any reason and rollback with `juju refresh` will quickly rollback highest unit even if leader unit charm code in error state
4. same as 3 and locking (mostly, except aforementioned edge cases) works when all units offline

More context:
Discussion: https://chat.canonical.com/canonical/pl/9fah5tfxd38ybnx3tq7zugxfyh
Option 1 in discussion is option 2 here, option 2 in discussion is option 1 here

**Option chosen:** Option 4

### Opensearch index vs document for lock
Current "ops lock" implementation with opensearch index:
Each unit requests the lock by trying to create an index. If the index does not exist, the "lock" is granted.

However, if a unit requests the lock, charm goes into error state, and error state is resolved (e.g. after rollback) it will not be able to use the lock—no unit will be aware that it has the lock and no unit will be able to release the lock

Solution: use document id 0 that stores "unit-name" as lock
Discussion: https://chat.canonical.com/canonical/pl/biddxzzk3fbpjgbhmatzr8n6bw

### Design
(Option 4): Use opensearch document as lock (for any (re)start, join cluster, leave cluster, or upgrade). Fallback to peer databag if all units offline.

### Implementation
Create custom events `_StartOpensearch` and `_RestartOpensearch`
https://github.com/canonical/opensearch-operator/blob/b1b28c10f2bd70bee4270707b616d8d565b8b616/lib/charms/opensearch/v0/opensearch_base_charm.py#L121-L132

When opensearch should be (re)started, emit the custom event.
Custom event requests the lock. If granted, it (re)starts opensearch.
Once opensearch fully ready, the lock is released.
If opensearch fails to start, the lock is released.
While opensearch is starting or while the lock is not granted, the custom event will be continually deferred.

Note: the original event is not deferred—only the custom event. This is so that any logic that ran before the request to (re)start opensearch does not get re-ran.

By requesting the lock within the custom event, and attempting to reacquire the lock each time the custom event runs (i.e. after every time it's deferred), we solve the design issue with rollingops and deferred events detailed in #183

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_locking.py
+++ b/lib/charms/opensearch/v0/opensearch_locking.py
@@ -7,6 +7,7 @@ import logging
 import typing
 
 import ops
+from charms.opensearch.v0.helper_cluster import ClusterTopology
 from charms.opensearch.v0.opensearch_exceptions import OpenSearchHttpError
 
 if typing.TYPE_CHECKING:
@@ -168,68 +169,91 @@ class OpenSearchNodeLock(ops.Object):
             host = None
         alt_hosts = [host for host in self._charm.alt_hosts if self._opensearch.is_node_up(host)]
         if host or alt_hosts:
-            logger.debug("[Node lock] Using opensearch for lock")
-            # Acquire opensearch lock
-            # Create index if it doesn't exist
+            logger.debug("[Node lock] 1+ opensearch nodes online")
             try:
-                self._opensearch.request(
-                    "PUT",
-                    endpoint=f"/{self._OPENSEARCH_INDEX}",
-                    host=host,
-                    alt_hosts=alt_hosts,
-                    retries=3,
-                    payload={"settings": {"index": {"auto_expand_replicas": "0-all"}}},
+                online_nodes = len(
+                    ClusterTopology.nodes(
+                        self._opensearch, use_localhost=host is not None, hosts=alt_hosts
+                    )
                 )
-            except OpenSearchHttpError as e:
-                if (
-                    e.response_code == 400
-                    and e.response_body.get("error", {}).get("type")
-                    == "resource_already_exists_exception"
-                ):
-                    # Index already created
-                    pass
-                else:
-                    logger.exception("Error creating OpenSearch lock index")
-                    return False
-            # Attempt to create document id 0
-            try:
-                self._opensearch.request(
-                    "PUT",
-                    endpoint=f"/{self._OPENSEARCH_INDEX}/_create/0?refresh=true",
-                    host=host,
-                    alt_hosts=alt_hosts,
-                    retries=3,
-                    payload={"unit-name": self._charm.unit.name},
-                )
-            except OpenSearchHttpError as e:
-                if e.response_code == 409 and "document already exists" in e.response_body.get(
-                    "error", {}
-                ).get("reason", ""):
-                    # Document already created
-                    if (unit := self._unit_with_lock(host)) != self._charm.unit.name:
-                        # Another unit has lock
-                        # (Or document deleted after last request & before request in
-                        # `self._lock_acquired()`)
-                        logger.debug(
-                            f"[Node lock] Not acquired. Unit with opensearch lock: {unit}"
-                        )
+            except OpenSearchHttpError:
+                logger.exception("Error getting OpenSearch nodes")
+                return False
+            logger.debug(f"[Node lock] Opensearch {online_nodes=}")
+            assert online_nodes > 0
+            if online_nodes >= 2:
+                logger.debug("[Node lock] Attempting to acquire opensearch lock")
+                # Acquire opensearch lock
+                # Create index if it doesn't exist
+                try:
+                    self._opensearch.request(
+                        "PUT",
+                        endpoint=f"/{self._OPENSEARCH_INDEX}",
+                        host=host,
+                        alt_hosts=alt_hosts,
+                        retries=3,
+                        payload={"settings": {"index": {"auto_expand_replicas": "0-all"}}},
+                    )
+                except OpenSearchHttpError as e:
+                    if (
+                        e.response_code == 400
+                        and e.response_body.get("error", {}).get("type")
+                        == "resource_already_exists_exception"
+                    ):
+                        # Index already created
+                        pass
+                    else:
+                        logger.exception("Error creating OpenSearch lock index")
                         return False
-                else:
-                    logger.exception("Error creating OpenSearch lock document")
-                    return False
-            # Lock acquired
-            # Release peer databag lock, if any
-            logger.debug("[Node lock] Acquired via opensearch")
-            self._peer.release()
-            logger.debug("[Node lock] Released redundant peer lock (if held)")
-            return True
-        else:
-            logger.debug("[Node lock] Using peer databag for lock")
-            # Request peer databag lock
-            # If return value is True:
-            # - Lock granted in previous Juju event
-            # - OR, unit is leader & lock granted in this Juju event
-            return self._peer.acquired
+                # Attempt to create document id 0
+                try:
+                    self._opensearch.request(
+                        "PUT",
+                        endpoint=f"/{self._OPENSEARCH_INDEX}/_create/0?refresh=true",
+                        host=host,
+                        alt_hosts=alt_hosts,
+                        retries=3,
+                        payload={"unit-name": self._charm.unit.name},
+                    )
+                except OpenSearchHttpError as e:
+                    if e.response_code == 409 and "document already exists" in e.response_body.get(
+                        "error", {}
+                    ).get("reason", ""):
+                        # Document already created
+                        pass
+                    else:
+                        logger.exception("Error creating OpenSearch lock document")
+                        return False
+            unit = self._unit_with_lock(host)
+            if unit == self._charm.unit.name:
+                # Lock acquired
+                # Release peer databag lock, if any
+                logger.debug("[Node lock] Acquired via opensearch")
+                self._peer.release()
+                logger.debug("[Node lock] Released redundant peer lock (if held)")
+                return True
+            if unit or online_nodes >= 2:
+                # Another unit has lock
+                # (Or document deleted after request to create document & before request in
+                # `self._unit_with_lock()`)
+                logger.debug(f"[Node lock] Not acquired. Unit with opensearch lock: {unit}")
+                return False
+            # If online_nodes == 1, we should acquire the lock via the peer databag.
+            # If we acquired the lock via OpenSearch and this unit was stopping, we would be unable
+            # to release the OpenSearch lock. For example, when scaling to 0.
+            # Then, when 1+ OpenSearch nodes are online, a unit that no longer exists could hold
+            # the lock.
+            # Note: if online_nodes > 1, this situation is still possible (e.g. if this unit was
+            # stopping and another unit went offline simultaneously)—but it's an edge case we don't
+            # support (to reduce complexity & improve robustness in other cases).
+            # If online_nodes > 1, we should re-attempt to acquire the OpenSearch lock.
+            logger.debug("[Node lock] No unit has opensearch lock")
+        logger.debug("[Node lock] Using peer databag for lock")
+        # Request peer databag lock
+        # If return value is True:
+        # - Lock granted in previous Juju event
+        # - OR, unit is leader & lock granted in this Juju event
+        return self._peer.acquired
 
     def release(self):
         """Release lock.


phvalguima (changes requested):
If the locking index will be widely used, we should exclude it from a backup / recovery scenario. @carlcsaposs-canonical can you add the correct index name to [this constant value here?](https://github.com/canonical/opensearch-operator/blob/c0225203320a412c3492ac827ac47b3b5b763787/lib/charms/opensearch/v0/opensearch_backups.py#L95)

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -604,31 +613,29 @@ def _is_tls_fully_configured(self) -> bool:
...
         if not self._can_service_start():
-            self.peers_data.delete(Scope.UNIT, "starting")
+            self.node_lock.release()
             event.defer()
             return
 
-        if self.peers_data.get(Scope.UNIT, "starting", False) and self.opensearch.is_failed():
-            self.peers_data.delete(Scope.UNIT, "starting")
+        # TODO: need to check if not first execution after acquiring lock?
+        if self.opensearch.is_failed():
```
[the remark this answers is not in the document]
Mehdi-Bendriss: The `starting` flag is useful when the start process took so long that it triggered a `OpenSearchStartTimeoutError` or when the node started but is not ready yet and raised a 503 error `OpenSearchNotFullyReadyError` thus event deferred - but no other unit should start because this unit hasn't completed `starting` and a subsequent check may confirm it has indeed fully started after a few moments.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -718,18 +724,23 @@ def _stop_opensearch(self) -> None:
...
+            event.defer()
+            return
 
-        self._start_opensearch(event)
+        try:
+            self._stop_opensearch()
+        except OpenSearchStopError as e:
+            logger.exception(e)
+            # TODO: should lock be released?
+            event.defer()
```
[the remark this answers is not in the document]
Mehdi-Bendriss: we can safely release release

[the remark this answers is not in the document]
Mehdi-Bendriss: Yes - we should release, and it is safe because: the service would have failed to stop so it is not needed for the lock to be retained - a later retry (defer) should be enough.

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -279,6 +279,7 @@ def call(url: str) -> requests.Response:
                 raise OpenSearchHttpError(response_text=str(e))
 
             if resp_status_code:
+                logger.exception(f"FOO {e.response.status_code}\n{e.response.json()}\n")
```
phvalguima, line 282: Maybe adding some more explicit text!

tests/unit/lib/test_helper_cluster.py
```diff
@@ -124,76 +124,77 @@ def setUp(self) -> None:
...
-            app_name=self.cluster1, nodes=list(old_cluster_conf.values())
-        )
-        assert new_cluster_conf["cm5"].roles == self.base_roles
-        # Assert other remaining nodes unchanged
-        old_cluster_conf.pop("cm5")
-        new_cluster_conf.pop("cm5")
-        assert old_cluster_conf == new_cluster_conf
+    # TODO: remove in https://github.com/canonical/opensearch-operator/issues/230
+    # def test_topology_roles_suggestion_odd_number_of_planned_units(self):
+    #     """Test the suggestion of roles for a new node and odd numbers of planned units."""
```
phvalguima, line 129: Out of curiosity, these tests are commented not because of the lock, right?

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
+        return value
+
+    @property
+    def _unit_with_lock(self):
+        if self._relation:
+            return self._relation.data[self._charm.app].get("unit-with-lock")
+
+    @_unit_with_lock.setter
+    def _unit_with_lock(self, value: str):
+        assert self._relation
```
phvalguima, line 89: `assert` will throw an exception and I cannot see any case where you're handling that with a try/catch. I think this condition never *actually* happens anyways, but we should have a "if" and return None instead of an exception.

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
+        self._relation.data[self._charm.app]["unit-with-lock"] = value
+
+    @_unit_with_lock.deleter
+    def _unit_with_lock(self):
+        assert self._relation
```
phvalguima, line 94: Same

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
+    def _relation(self):
+        # Use property instead of `self._relation =` in `__init__()` because of ops Harness unit
+        # tests
+        return self._charm.model.get_relation(self._NAME)
+
+    def _on_peer_relation_changed(self, _=None):
+        """Grant & release lock."""
+        if not self._charm.unit.is_leader():
+            return
+        assert self._relation
```
phvalguima, line 107: Same

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -776,6 +787,7 @@ def _post_start_init(self, event: EventBase):
         # may be thrown when calling a node - we want to ensure this node is perfectly ready
         # before marking it as ready
         if not self.opensearch.is_node_up():
+            logger.warning("FOO not ready")
```
phvalguima, line 790: nit: same - meaningful logs!

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
 
-    @retry(stop=stop_after_attempt(3), wait=wait_fixed(0.5), reraise=True)
-    def acquire(self):
-        """Method for Acquiring the "ops" lock."""
-        # no lock acquisition needed if only 1 unit remaining
-        if len(self._charm.model.get_relation(PeerRelationName).units) == 1:
+        Returns:
+            Whether lock was acquired
+        """
+        if not self._relation:
```
phvalguima, line 45: I think this condition never really happens, but if it does, then we should return True instead, no? It means you are alone in the cluster. Maybe we can then check `planned_units > 1`.

phvalguima (commented):
Nice work @carlcsaposs-canonical!

I tried to look for deadlocks did not see any. I think we need to be careful about always having a "release" in the same method we run an "acquired", and before every return call to avoid one.

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
+            # - OR, unit is leader & lock granted in this Juju event
+            return self._peer.acquired
 
-    def _is_lock_in_peer_data(self) -> bool:
-        """Method checking if lock acquired from the peer rel data."""
-        rel = self._charm.model.get_relation(PeerRelationName)
-        for unit in rel.units:
-            if rel.data[unit].get(OpenSearchOpsLock.PEER_DATA_LOCK_FLAG) == "True":
-                return True
+    def release(self):
```
phvalguima, line 232: You also need to extend `_on_peer_relation_changed`: the leader should not be conceding the peer relation lock if the opensearch lock is being used.

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
+                break
+        else:
+            logger.debug("[Node lock] (leader) cleared peer lock")
+            del self._unit_with_lock
+
 
-        # we check first on the peer data bag if the lock is already acquired
-        if self._is_lock_in_peer_data():
-            raise OpenSearchOpsLockAlreadyAcquiredError("Another unit is being removed.")
+class OpenSearchNodeLock(ops.Object):
```
phvalguima, line 125: I think we should add a "token" to the peer relation once the leader successfully acquires the node lock the first time. This "token" would force all the units to never use the peer relation again.

If we have a cluster that got so disconnected it needs to fallback to the peer relation, then we have waaay more problems to deal before dealing with the lock itself. And if we manage to recover the cluster, then we have a functional cluster again.

phvalguima (changes requested):

lib/charms/opensearch/v0/opensearch_locking.py
phvalguima: @Mehdi-Bendriss , wdyt?

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
+    def _unit_requested_lock(self, unit: ops.Unit):
+        """Whether unit requested lock."""
+        assert self._relation
+        value = self._relation.data[unit].get("lock-requested")
+        if not value:
+            return False
+        value = json.loads(value)
+        if not isinstance(value, bool):
+            raise ValueError
+        return value
```
Mehdi-Bendriss, line 80: Perhaps - safer and more concise (asserts are strongly discouraged outside of testing code - they can even be disabled when running with `-O`. Raise a specific exception if needed instead):
```suggestion
    def _unit_requested_lock(self, unit: ops.Unit) -> bool:
        """Whether unit requested lock."""
        if not self._relation:
            return False
            
        return json.loads(
            self._relation.data[unit].get("lock-requested", "false")
        )
```

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: @carlcsaposs-canonical we do not need the `assert` - as the subsequent `self._relation.data` will raise an exception anyway

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
+
+    @property
+    def _unit_with_lock(self):
+        if self._relation:
+            return self._relation.data[self._charm.app].get("unit-with-lock")
```
Mehdi-Bendriss, line 85:
```suggestion
    def _unit_with_lock(self) -> Optional[str]:
        if self._relation:
            return self._relation.data[self._charm.app].get("unit-with-lock")
        return None
```

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: Similar to a previous point, if the aim is to raise an exception, we do not need an assert, the next line will raise one

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: Yeah I think the only case where this might not work is on a full cluster crash of the units (not the services, because systemd services restart automatically) - which would essentially be treated the same as a "new cluster" (lock wise)

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
+                "GET",
+                endpoint=f"/{self._OPENSEARCH_INDEX}/_source/0",
+                host=host,
+                alt_hosts=self._charm.alt_hosts,
+                retries=3,
+            )
+        except OpenSearchHttpError as e:
+            if e.response_code == 404:
+                # No unit has lock
+                return
```
Mehdi-Bendriss, line 152: nit:
```suggestion
                return None
```

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
+                else:
+                    logger.exception("Error creating OpenSearch lock document")
+                    return False
+            # Lock acquired
+            # Release peer databag lock, if any
+            logger.debug("[Node lock] Acquired via opensearch")
+            self._peer.release()
+            logger.debug("[Node lock] Released redundant peer lock (if held)")
+            return True
+        else:
```
Mehdi-Bendriss, line 224: nit: could we please invert the if/else so that we reduce a level of nestedness?
```
if not host and not alt_hosts:
    return self._peer.acquired

# rest of the code
```

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,240 @@
...
-        """Method checking if lock acquired from the peer rel data."""
-        rel = self._charm.model.get_relation(PeerRelationName)
-        for unit in rel.units:
-            if rel.data[unit].get(OpenSearchOpsLock.PEER_DATA_LOCK_FLAG) == "True":
-                return True
+    def release(self):
+        """Release lock.
 
-        return False
+        Limitation: if lock acquired via OpenSearch document and all units offline, OpenSearch
```
Mehdi-Bendriss, line 235: Can this be solved in a subsequent `_StartOpenSearch` event (which will make use of the peer rel lock), as follows:
```
# ... start logic
# now the opensearch node is up
# delete the opensearch lock
```

Mehdi-Bendriss (commented):
Thanks Carl! I have a few minor questions?

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: Type hinting are for the IDE but also for readability

phvalguima (dismissed):

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: Exactly, so what i'm describing is the below flow: 
1. lock acquired via OpenSearch
2. scale-down => all units offline now 
3. scale-up
4. lock acquired with peer data, to start 
5. after starting, and before the lock gets released, `delete the opensearch lock if any` (from step 1)

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: Okay - but just for the sake of clarity and completeness - `failing silently` will not happen, as what I'm advocating for is to let `self._relation.data[unit]` fail in the next line.

The other comment was with regards to the check:
```
value = json.loads(value)
if not isinstance(value, bool):
```
when would this check evaluate to true?

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -25,78 +25,242 @@
...
+                host=host,
+                alt_hosts=self._charm.alt_hosts,
+                retries=3,
+            )
+        except OpenSearchHttpError as e:
+            if e.response_code == 404:
+                # No unit has lock
+                return
+            raise
+        return document_data["unit-name"]
```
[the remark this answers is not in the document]
phvalguima: Actually, it would be better that only the charm app leader that is the "cluster_manager" to do this check, and it would check against all the units related to it via peer && large deployment clusters. This way we do not delete that entry for no reason.

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: I think in this case we should have a bit more predictability and trust in the code. If we do not expect unpredictable input on the `lock-requested` entry of the peer data, we should trust that and not be overly defensive here

phvalguima (approved):

PR #213, opened 2024-04-03
Add HTTPError information to debug log

carlcsaposs-canonical:
Add HTTP response code, json, and text to Exception message so that it's printed in the debug log

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_exceptions.py
+++ b/lib/charms/opensearch/v0/opensearch_exceptions.py
@@ -81,9 +81,11 @@ class OpenSearchHttpError(OpenSearchError):
         except (json.JSONDecodeError, TypeError):
             self.response_body = {}
         self.response_code = response_code
-        super().__init__(
-            f"HTTP error {self.response_code=}\n{self.response_body=}\n{self.response_text=}"
-        )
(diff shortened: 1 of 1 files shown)


lib/charms/opensearch/v0/opensearch_exceptions.py
```diff
@@ -75,13 +75,15 @@ class OpenSearchHttpError(OpenSearchError):
...
-            self.response_text = response_text
             self.response_body = json.loads(response_text)
         except (json.JSONDecodeError, TypeError):
             self.response_body = {}
-        finally:
-            self.response_code = response_code
+        self.response_code = response_code
+        super().__init__(
+            f"HTTP error {self.response_code=}\n{self.response_body=}\n{self.response_text=}"
+        )
```
phvalguima, line 86: @carlcsaposs-canonical I would not log both, but json, if present, otherwise log the response text

phvalguima (changes requested):

phvalguima (approved):
LGTM

Mehdi-Bendriss (approved):

juditnovak (commented):
"post-mortem" LGTM :-)

PR #216, opened 2024-04-07
[DPE-2944] kibanaserver user

juditnovak:
## `kibanaserver` user doesn't exist
Opensearch Charm has a particularly strict user management.

At start ALL users are deleted.
Then, one-by-one they are "re-added" back (such as `admin`) -- or brand-new ones may be created (such as COS `monitor` user).

We are planning to use the system `kibanaserver` user for Dashboard access, as it seems the most stable and resilient solution.

## `kibanaserver` user with password should be available to Dashboard

On a particular, specific relation request (from the Dashboard Charm), the Opensearch Charm should provide the `kibanaserver` 
user credentials (via Juju Secrets).

## Solution

1. We "re-add" the freshly deleted `kibanaserver` user.
NOTE: unlike the original, the new one will have `reserved: false` set, in order to allow for password changes.

2. We alter the charm's current `OpensearchProvider` interface, so that it recognizes the `kibana_server` request as an `extra_user_role`. When this role is requrested, instead of creating a new relation user, the Opensearch Charm returns the `kibanaserver` user's credentials.

**NOTE**: This PR is also providing a fix for 
 - Issue https://github.com/canonical/opensearch-operator/issues/246
 - Libs are updated
 - Unittests not using secrets at all
 - Horizontal scaling tests are known to be broken and are to be fixed outside of the scope of this PR
 
 
image


lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -156,6 +159,21 @@ def __init__(self, charm: CharmBase) -> None:
...
+        result = []
+        for relation in self.opensearch_provides.relations:
+            # Won't work... There's NO trace of any relation details that would
+            # allow for decision-making here :-(
+            #
+            # if KibanaserverRole in self.opensearch_provides.fetch_relation_field(
+            #     relation.id, "extra_user_roles"
+            # ):
+            if any(key.name == "opensearch-dashboards" for key in relation.data.keys()):
+                result.append(relation)
```
juditnovak, line 174: ...is this really what we want...?

If yes, I move the sting to a constant. If not... I'm running out of ideas :-(

Mehdi-Bendriss: Wouldn't what you commented out work here?
```
if KibanaserverRole in self.opensearch_provides.fetch_relation_field(
    relation.id, "extra_user_roles"
)
```
I prefer it over the `key.name` approach

Mehdi-Bendriss:
I think we can keep treat the `kibanaserver` user at the same level of `admin` and not delete it but rather change its password / hash and store it in `internalusers.yml` - we can also keep it as `reserved: true` and we'll still be able to rotate the password later on with the REST Api

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: I added 

```
            # Won't work... There's NO trace of any relation details that would
            # allow for decision-making here :-(
```
comment above the commented code to highlight that it wouldn't work, and why.

Mehdi-Bendriss: Sorry, I could be wrong, but I think the below works:
```
dashboards_relations = []
for relation in self.model.relations[self.relation_name]:
    remote_data = relation.data[relation.app]
    if KibanaServerRole in remote_data.get("extra_user_roles", []):
         dashboard_relations.append(relation.id)
```

juditnovak: https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/opensearch_client/v0#example

juditnovak: Let me double-check in practice if I overlooked something.

I couldn't find the field in the databag.

juditnovak:
To my understanding `admin` IS deleted.

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: All right my bad.

I don't know what I must have overlooked.

But the field is there indeed

```
  - relation-id: 5
    endpoint: opensearch-client
    related-endpoint: opensearch_client
    application-data:
      extra-user-roles: kibanaserver
      index: .opensearch-dashboards
      requested-secrets: '["username", "password", "tls", "tls-ca", "uris"]'
    related-units:
      opensearch-dashboards/0:
        in-scope: true
        data:
          data: '{"endpoints": "10.196.115.48:9200", "index": ".opensearch-dashboards",
            "secret-tls": "secret://4a938601-7354-49d8-8f2a-e0f434d2ea48/coc18l3v6d6i9akiaqd0",
            "secret-user": "secret://4a938601-7354-49d8-8f2a-e0f434d2ea48/coc18l3v6d6i9akiaqcg",
            "version": "2.12.0"}'
          egress-subnets: 10.196.115.189/32
          ingress-address: 10.196.115.189
          private-address: 10.196.115.189
```

juditnovak:
Ah! And `admin` isn't reserved...

juditnovak:
No we can't: As you find in my PR they appear with different parameters in [internal_users.yaml](https://github.com/opensearch-project/security/blob/main/config/internal_users.yml), they are quite different.

I don't like the current solution, but it's no worse than hanving a

```
def create_internal_user():
    if username == 'kibanaserver':
        <do this>
    if username == 'admin':
        <do totally different>
```

So far the current approach (that I'm far from happy with) was the most elegant way to unite common code, and distinguish where needed.

juditnovak:
Also note that we treat secrets different for `admin` as for others (as we're also saving the password hash, which we don't need for others)

juditnovak:
@Mehdi-Bendriss FYI I also added a solution for https://github.com/canonical/opensearch-operator/issues/246 in this PR at the end. 

This was the most convenient way, due to how secrets interactions are designed.

Turns out if we don't fetch a charm internal secret, we're not gonna get `secret-changed` events about it. (See https://matrix.to/#/!wJiiHsLipVywuWOyNi:ubuntu.com/$y-BhyxlsJvxTjYp68JgdAbL6DSN8W8nhug6ICSgyi7w?via=ubuntu.com&via=matrix.org&via=beeper.com)

The most convenient way to "initial fetch" secrets in question was to fix the issue. As we initialize `internal_users.yml` correctly, we fetch and use the `<user>-password-hash` secrets for the first time on all units. Thus when rotating these credentials all units will get notified of the secret change, and be able to correctly maintain their internal list of users.

lib/charms/opensearch/v0/opensearch_secrets.py
```diff
@@ -69,18 +70,47 @@ def _on_secret_changed(self, event: SecretChangedEvent):
             logging.info(f"Label {event.secret.label} was meaningless for us, returning")
             return
 
+        system_user_hash_keys = [
+            self._charm.secrets.hash_key(user) for user in OpenSearchSystemUsers
+        ]
+        keys_to_process = system_user_hash_keys + [
+            CertType.APP_ADMIN.val,
+            self._charm.secrets.password_key(KibanaserverUser),
```
Mehdi-Bendriss, line 78: I understand what you're doing (Kibanaserveruser needs to be updated in the client relation), but could we please add a comment here just so it's clear to future readers

Mehdi-Bendriss (commented):
Thanks Judit! Looks good to me!

Mehdi-Bendriss (dismissed):
Great work! Thank you!

.github/workflows/ci.yaml
```diff
@@ -69,8 +69,7 @@ jobs:
   integration-test:
     name: Integration test charm
     needs:
-      - lint
-      - unit-test
+      # - unit-test
```
phvalguima, line 72: @juditnovak we need to revert this

phvalguima (commented):
Only 2x comments:
1. We need to revert the change in the ci.yaml. I'm happy with another PR on top of `DPE-2944_kibanaserver_user` instead of changing this one: and linking it here
2. Missing unit / integration tests: the changes below will impact way more than Kibana's users - we will need some test coverage at least in the pwd rotation: can you create a ticket and work on that?

.github/workflows/ci.yaml
juditnovak: Yes, this was a leftover, by mistake. Thx 4 spotting it :-)

phvalguima (approved):
LGTM

We need the unit / int. tests but we can add it as part of another PR

Mehdi-Bendriss (approved):

Issue #227, opened 2024-04-12
ha/test_ha_networking.py fails to run lxc command since lxd 5.21


carlcsaposs-canonical:
test passed on 2024-04-11 with lxd 5.20-f3dd836 snap rev 27049 https://github.com/canonical/opensearch-operator/actions/runs/8640068334/job/23687547198
test failed on 2024-04-12 with lxd 5.21.1-43998c6 snap rev 28155 https://github.com/canonical/opensearch-operator/actions/runs/8655512868/job/23734702946#step:21:1686

```
 _________ test_full_network_cut_without_ip_change_node_with_elected_cm _________
Traceback (most recent call last):
  File "/home/runner/work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/runner.py", line 341, in from_call
    result: Optional[TResult] = func()
  File "/home/runner/work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/runner.py", line 262, in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
  File "/home/runner/work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_hooks.py", line 501, in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
  File "/home/runner/work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_manager.py", line 119, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  File "/home/runner/work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 181, in _multicall
    return outcome.get_result()
  File "/home/runner/work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_result.py", line 99, in get_result
    raise exc.with_traceback(exc.__traceback__)
  File "/home/runner/work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 102, in _multicall
    res = hook_impl.function(*args)
  File "/home/runner/work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/runner.py", line 177, in pytest_runtest_call
    raise e
[...]

PR #228, opened 2024-04-12
Pin lxd to 5.20 in integration tests

carlcsaposs-canonical:
Workaround for https://github.com/canonical/opensearch-operator/issues/227

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -77,6 +77,8 @@ jobs:
       artifact-prefix: packed-charm-cache-true
       cloud: lxd
       juju-snap-channel: 3.3/stable
+      # Workaround for https://github.com/canonical/opensearch-operator/issues/227
+      lxd-snap-channel: 5.20/stable
     secrets:
       integration-test: |
         {


Mehdi-Bendriss (approved):
Thank you!

.github/workflows/ci.yaml
```diff
@@ -77,6 +77,8 @@ jobs:
       artifact-prefix: packed-charm-cache-true
       cloud: lxd
       juju-snap-channel: 3.3/stable
+      # Workaround for https://github.com/canonical/opensearch-operator/issues/227
+      lxd-snap-channel: 5.20/stable
```
phvalguima, line 81: @carlcsaposs-canonical this is unnecessary. Check my comment on #227: (1) I fixed the last `limits.network.priority` command and (2) we need to figure out another way to implement this network cut.

phvalguima (changes requested):

.github/workflows/ci.yaml
phvalguima: @carlcsaposs-canonical @Mehdi-Bendriss we don't need to pin, check here: https://github.com/canonical/opensearch-operator/pull/239

And this is a permanent fix moving forward with 5.21+

PR #231, opened 2024-04-12
Handle leader raising uncaught exception while acquiring peer databag lock

carlcsaposs-canonical:
If the leader unit raises an uncaught exception in the same Juju event that it acquires the peer databag lock, the changes to the peer databag will be reverted—which could allow another unit to get the lock.

Therefore, we should wait until the next Juju event to use the lock.

Note: for the leader unit, instead of the lock being requested & granted in the same event, the process is now:
- leader unit requests lock & provisionally grants itself lock
- non-leader units get peer-relation-changed and put a unique value in their databag to trigger peer-relation-changed on leader
- leader gets peer-relation-changed (or any other event) and proceeds to use lock and start/stop opensearch

Context: https://matrix.to/#/!xdClnUGkurzjxqiQcN:ubuntu.com/$dRO72RvyFD4EI65_n83hIQPAMFtHmVOQtjG5bMENaMk?via=ubuntu.com&via=matrix.org&via=cutefunny.art

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_locking.py
+++ b/lib/charms/opensearch/v0/opensearch_locking.py
@@ -158,7 +158,7 @@ class _PeerRelationLock(ops.Object):
                 # (Value should never be read)
                 # (If we set the same value that is currently in the databag, a peer relation
                 # changed event will not be triggered)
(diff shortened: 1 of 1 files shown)


Mehdi-Bendriss (dismissed):
thank you

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #240, opened 2024-04-17
[DEBUG][DPE-2944] kibanaserver user debug

juditnovak:


juditnovak:
Debugging finished, not needed anymore

PR #241, opened 2024-04-17
[DPE-4115] Performance Profiles

phvalguima:
Adds performance profile to the charm and an equivalent option that allows choosing which one.

Currently, all jvm options are hardcoded in the charm. This PR extends the charm to load `meminfo` and `cpuinfo` and use it to create a performance-oriented profile. This profile is named "production".

The profile itself is based on comments in: https://opensearch.org/docs/latest/tuning-your-cluster/performance/

And the compaction algorithm that has been chosen instead of "best_compression" is `zstd`, as improved results showed in [here](https://opensearch.org/docs/latest/im-plugin/index-codecs/).

# OPEN POINTS

* Right now, I am applying this setting once, at install. I think it should move to config-changed instead.

* Not yet clear how to add this option to config-changed routine. My idea was to have something similar to `_compute_settings` and `apply_config` from plugin-manager: we load all configs and compare with the profile that has been requested by the user. 

* To create a "testing" profile, with smaller memory consumption (bottom "1g") and default max(..min()) for merge settings

* I think we should create a dataclass that hardcodes each option this profile touches. This way it is easy to: (1) check if a given option that is being overriden by some other logic, actually bypasses the profile; and (2) make sure we remember to tune the same parameters


config.yaml
```diff
@@ -36,3 +36,13 @@ options:
     default: true
     type: boolean
     description: Enable opensearch-knn
+
+  profile:
+   description: |
+      Profile representing the scope of deployment, and used to tune resource allocation.
+      Allowed values are: "production" or "default".
```
delgod, line 43: only `production`, `testing and `staging` are approved https://docs.google.com/document/d/1bq7KyUAPIIcUWGgifW05p40v7lGXs3PhMznM2ECaMKY/edit
```suggestion
      Allowed values are: "production" or "testing".
```

delgod (changes requested):

PR #242, opened 2024-04-18
Add in-place upgrades

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-router-operator/pull/88

upgrade.py, machine_upgrade.py and event handlers in charm.py are mostly shared with upgrade.py, machine_upgrade.py and event handlers in abstract_charm.py/machine_charm.py in MySQL Router VM charm

Non-upgrade event handler guards implemented per https://docs.google.com/spreadsheets/d/1_qjINUoj5Mi7PlXhCdpwQ0170bIZSgpvNFMAnTOtuQw/edit#gid=0

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -76,8 +76,3 @@ force-upgrade:
     Use to
     - force incompatible upgrade and/or
     - continue upgrade if 1+ upgraded units have non-active status
-  params:
-    ignore-lock:
-      type: boolean
-      default: false
-      description: Allow unit to upgrade while another unit is starting or stopping
--- a/src/charm.py
+++ b/src/charm.py
@@ -167,7 +167,10 @@ class OpenSearchOperatorCharm(OpenSearchBaseCharm):
             return
         logger.debug("Forcing upgrade")
         event.log(f"Forcefully upgrading {self.unit.name}")
-        self._upgrade_opensearch_event.emit(ignore_lock=event.params["ignore-lock"])
+        # TODO: replace `ignore_lock=False` with `event.params["ignore-lock"]` if specification
+        # DA091 approved
+        # (https://docs.google.com/document/d/1rwnS-deJU9Mzc8BFkl3UGgjZiBa6e3bxoT-6BQo9e3E/edit)
+        self._upgrade_opensearch_event.emit(ignore_lock=False)
         event.set_results({"result": f"Forcefully upgraded {self.unit.name}"})
         logger.debug("Forced upgrade")
 


phvalguima (changes requested):
Are you adding the `dependency.json` later or is it gone on newer upgrade processes?

actions.yaml
```diff
@@ -62,4 +62,14 @@ restore:
...
         A backup-id to identify the backup to restore. Format: backup-id=<string>.
   required:
-    - backup-id
\ No newline at end of file
+    - backup-id
+
+resume-upgrade:
+  description: Upgrade remaining units (after you manually verified that upgraded units are healthy).
+force-upgrade:
+  description: Force upgrade of this unit.
```
phvalguima, line 70: I think we need to be more clear, e.g.:
```suggestion
  description: |
        WARNING: running force-upgrade may result in breaking the cluster or data loss. Evaluate very carefully before executing it.
        Force upgrade of this unit. Use this action to override the upgrade constraints, e.g. version mismatch errors.
```

phvalguima (changes requested):
Left some extra suggestions. I will review the core of the upgrade lib later today.

src/charm.py
```diff
@@ -41,6 +64,101 @@ def _on_install(self, _: InstallEvent) -> None:
...
+            message = "No upgrade in progress"
+            logger.debug(f"Force upgrade event failed: {message}")
+            event.fail(message)
+            return
+        if not self._upgrade.upgrade_resumed:
+            message = f"Run `juju run {self.app.name}/leader resume-upgrade` before trying to force upgrade"
+            logger.debug(f"Force upgrade event failed: {message}")
+            event.fail(message)
+            return
+        if self._upgrade.unit_state != "outdated":
```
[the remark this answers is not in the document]
phvalguima: @carlcsaposs-canonical then make it a constant. As long as we don't have to search across the code if we ever decide to change it.

src/upgrade.py
```diff
@@ -0,0 +1,222 @@
...
+        self._unit: ops.Unit = charm_.unit
+        self._unit_databag = self._peer_relation.data[self._unit]
+        self._app_databag = self._peer_relation.data[charm_.app]
+        self._app_name = charm_.app.name
+        self._current_versions = {}  # For this unit
+        for version, file_name in {
+            "charm": "charm_version",
+            "workload": "workload_version",
+        }.items():
+            self._current_versions[version] = pathlib.Path(file_name).read_text().strip()
```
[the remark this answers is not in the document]
phvalguima: @carlcsaposs-canonical added the integration test mentioned here in #258.

src/upgrade.py
```diff
@@ -0,0 +1,227 @@
...
+import pathlib
+import typing
+
+import ops
+import poetry.core.constraints.version as poetry_version
+from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
+
+logger = logging.getLogger(__name__)
+
+PEER_RELATION_ENDPOINT_NAME = "upgrade-version-a"
```
Mehdi-Bendriss, line 24: Could we rename this to include `upgrade` in it?

src/upgrade.py
```diff
@@ -0,0 +1,227 @@
...
+def unit_number(unit_: ops.Unit) -> int:
+    """Get unit number"""
+    return int(unit_.name.split("/")[-1])
+
+
+class PeerRelationNotReady(Exception):
+    """Upgrade peer relation not available (to this unit)"""
+
+
+class UnitState(str, enum.Enum):
```
Mehdi-Bendriss, line 37: this ideally is part of `helper_charm.py`

src/upgrade.py
```diff
@@ -0,0 +1,227 @@
...
+import poetry.core.constraints.version as poetry_version
+from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
+
+logger = logging.getLogger(__name__)
+
+PEER_RELATION_ENDPOINT_NAME = "upgrade-version-a"
+RESUME_ACTION_NAME = "resume-upgrade"
+
+
+def unit_number(unit_: ops.Unit) -> int:
```
Mehdi-Bendriss, line 28: this ideally is part of `helper_charm.py`

src/upgrade.py
Mehdi-Bendriss, line 1: Could we move this module to the charm libs under `opensearch_upgrade.py` and keep the `machine_upgrade.py` implementation on the `src` dir?

Mehdi-Bendriss (commented):
Thanks Carl! Great work! I do not have major concerns here, only a few comments

actions.yaml
```diff
@@ -62,4 +62,22 @@ restore:
...
+  description: |
+    Potential of *data loss* and *downtime*
+        
+    Force upgrade of this unit.
+    
+    Use to
+    - force incompatible upgrade and/or
+    - continue upgrade if 1+ upgraded units have non-active status
+  params:
+    ignore-lock:
```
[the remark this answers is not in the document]
Mehdi-Bendriss: we can find something a bit more "user" oriented, perhaps: `apply-now` or `force-now` or something

src/upgrade.py
Mehdi-Bendriss: It's fine, but keep in mind, another developer may later during a refactor remove the module name on the reference - and then this might be prone to confusions and strange errors if reused elsewhere thinking it was `PeerRelationName`

actions.yaml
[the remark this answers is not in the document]
Mehdi-Bendriss: It could also be: `ignore-order/priority/turn` or similar, feel free to come up with something that's as meaningful to the end user as `lock` is to us charm developers

phvalguima (dismissed):

PR #248, opened 2024-04-22
[DPE-3661] Add support for large deployments backup

phvalguima:
Implements [DPE-3661](https://warthogs.atlassian.net/browse/DPE-3661): extend backup feature to support large deployment scenarios.

Currently, there are 4x type of scenarios:
1) Small deployments: the cluster performs all the different types of node roles
2) Large deployments - orchestrator: the app is in charge of not only its own application units but also to coordinate the across the different clusters
3) Large deployments - failover orchestrator: very similar to (2), this app must also publish its information in the peer relation, although all the clusters will only listen to the active manager
4) Large deployments - data only: do not perform any management tasks and should receive any relevant information via peer relation

For backups, clusters of type (3) and (4) have a special behavior: they will receive the backup data via peer-cluster relation and should refuse: i. to execute backup-related actions; and ii. to execute the s3-relation events themselves. The latter avoids confusions, e.g. an user inadvertently relates the cluster to different s3-integrators. 

The implementation of (1) and (2) are very similar. 

It contains the same fix as: https://github.com/canonical/opensearch-operator/pull/253

Adds following fixes related to testing in general:
1. `ContinuousWrites` is updated to hold the right count of documents in `writes_value` internally
2. Adds an `is_burst` option to `ContinuousWrites`: a test may choose to send 100-burst docs vs. doc-by-doc - `is_burst` defaults to `True`
3. The `ContinuousWrites` terminates its process as part of `stop`, avoiding stranded process generating docs to `ContinuousWrites.INDEX_NAME` post a given test
4. The `start_and_check_continuous_writes` updated to `assert_start_and_check_continuous_writes`

# How To

## Setup a large scale deployment

Deploy a large scale environment with:
```
juju deploy tls-certificates-operator --channel stable --show-log --verbose
juju config tls-certificates-operator generate-self-signed-certificates=true ca-common-name="CN_CA"

# deploy main-orchestrator cluster 
juju deploy -n 3 ./opensearch.charm \
    main \
    --config cluster_name="log-app" --config init_hold=false --config roles="cluster_manager"

# deploy failover-orchestrator cluster
juju deploy -n 2 ./opensearch.charm \
    failover \
    --config cluster_name="log-app" --config init_hold=true --config roles="cluster_manager"

# deploy data-hot cluster
juju deploy -n 2 ./opensearch.charm \
    data-hot \
    --config cluster_name="log-app" --config init_hold=true --config roles="data.hot"

# integrate TLS
juju integrate tls-certificates-operator main
juju integrate tls-certificates-operator failover
juju integrate tls-certificates-operator data-hot

# integrate the "main"-orchestrator with all clusters:
juju integrate main:peer-cluster-orchestrator failover:peer-cluster
juju integrate main:peer-cluster-orchestrator data-hot:peer-cluster

# integrate the "failover"-orchestrator with rest of clusters:
juju integrate failover:peer-cluster-orchestrator data-hot:peer-cluster
```

## Connect with S3

This step assumes a [s3-integrator](https://discourse.charmhub.io/t/s3-integrator-documentation/10947) charm has been successfully deployed and configured. The large deployments backup is set with a `s3-integrator` connected solely to the main charm. It must be set as follows:

```
juju integrate s3-integrator main
```

Wait until the cluster deployment set. Now, backup / restore actions can be executed by running them against the main orchestrator's leader:

```
juju run main/leader create-backup
```

## In case of failover

In case a failover must be triggered, besides the process described for failover of large deployments, also move the current s3-integrator from one cluster to another:

```
# If the cluster still exists
juju remove-relation main s3-integrator

# Then, connect with the new cluster manager
juju relate s3-integrator failover
```

# Implementation Details

For developers, there is no meaningful difference between small and large deployments.
They both use the same backup_factory() to return the correct object for their case.

The large deployments expands the original concept of OpenSearchBackup to include other
juju applications that are not cluster_manager. This means a cluster may be a data-only or
even a failover cluster-manager and still interacts with s3-integrator at a certain level.

The baseline is that every unit in the cluster must import the S3 credentials. The main
orchestrator will share these credentials via the peer-cluster relation. Failover and data
clusters will import that information from the peer-cluster relation.

To implement the points above without causing too much disruption to the existing code,
a factory pattern has been adopted, where the main charm receives a OpenSearchBackupBase
object that corresponds to its own case (cluster-manager, failover, data, etc).
"""

[DPE-3661]: https://warthogs.atlassian.net/browse/DPE-3661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -45,6 +45,21 @@ def __init__(self, charm):
...
+        #         self._opensearch._run_cmd(
+        #             self._keytool,
+        #             f"-import -alias {key} "
+        #             f"-file {filename} -storetype JKS "
+        #             f"-storepass {self.password} "
+        #             f"-keystore {self._keystore} -noprompt",
+        #         )
+        #     except OpenSearchCmdError as e:
+        #         raise OpenSearchKeystoreError(str(e))
+
```
juditnovak, line 62: "...should I stay or should I go?"

juditnovak (dismissed):
All right...
I think I've got the base concept.
I keep wondering if this should rather get "library-zed"...? Or, if we're conceptually dealing with cross-charm relations, we may really wanna move this code towards `data_interfaces`...?

Overall, the concept seems good, but it really feels like it's beyond that charm's own code...

Furthermore, secrets seem essential ;-)

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -800,3 +1095,27 @@ def get_snapshot_status(self, response: Dict[str, Any] | None) -> BackupServiceS
...
+
+    If this cluster is an orchestrator or failover cluster, then return the OpenSearchBackup.
+    Otherwise, return the OpenSearchNonOrchestratorBackup.
+
+    There is also the condition where the deployment description does not exist yet. In this
+    case, return the base class OpenSearchBackupBase. This class solely defers all s3-related
+    events until the deployment description is available and the actual S3 object is allocated.
+    """
+
+    def __new__(cls, charm, *args, **kwargs):
```
Mehdi-Bendriss, line 1113: I would prefer that we're clear here and make it into a static method `.create()` or similar. Instantiating a factory is not very intuitive

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -315,6 +316,23 @@ def _rel_data(
...
+                    secret_key=self.charm.backup.s3_client.get_s3_connection_info().get(
+                        "secret-key", ""
+                    ),
+                )
+            else:
+                # Return what we have received from the peer relation
+                s3_credentials = S3RelDataCredentials(
+                    access_key=secrets.get(Scope.APP, "access-key", default=""),
+                    secret_key=secrets.get(Scope.APP, "secret-key", default=""),
+                )
```
Mehdi-Bendriss, line 335: Can you take this out in a separate method: `_s3_credentials()`?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -490,6 +513,9 @@ def _set_security_conf(self, data: PeerClusterRelData) -> None:
         self.charm.peers_data.put(Scope.APP, "admin_user_initialized", True)
         self.charm.peers_data.put(Scope.APP, "security_index_initialised", True)
 
+        self.charm.secrets.put(Scope.APP, "access-key", data.s3_credentials.access_key)
+        self.charm.secrets.put(Scope.APP, "secret-key", data.s3_credentials.secret_key)
```
Mehdi-Bendriss, line 517: can't this be grouped into an object? `s3_credentials` or similar and simply:
```
self.charm.secrets.put(Scope.APP, "s3-credentials", data.s3_credentials.to_dict())
```

tests/integration/ha/continuous_writes.py
```diff
@@ -211,6 +212,8 @@ async def _run(  # noqa: C901
         event: Event, data_queue: Queue, starting_number: int, is_bulk: bool
     ) -> None:
         """Continuous writing."""
+        run_logger = log_to_stderr()
```
Mehdi-Bendriss, line 215:
```suggestion
        proc_logger = log_to_stderr()
```

tests/integration/ha/continuous_writes.py
```diff
@@ -256,7 +263,7 @@ def _client(_data) -> OpenSearch:
         # write last expected written value on disk
         with open(ContinuousWrites.LAST_WRITTEN_VAL_PATH, "w") as f:
             if is_bulk:
-                write_value = (100 * write_value) + 99
+                write_value += 99
```
Mehdi-Bendriss, line 266: What we had was correct, we're ensuring that this **only** gets incremented if `write_value` has been successfully incremented.

tests/integration/ha/continuous_writes.py
```diff
@@ -269,7 +276,7 @@ def _bulk(client: OpenSearch, write_value: int) -> None:
         """Bulk Index group of docs."""
         data = []
         for i in range(100):
-            val = (100 * write_value) + i
+            val = write_value + i
```
Mehdi-Bendriss, line 279: What we had was correct, we're ensuring that this **only** gets incremented if `write_value` has been successfully incremented.

tests/integration/ha/helpers.py
```diff
@@ -514,54 +519,59 @@ async def start_and_check_continuous_writes(ops_test: OpsTest, unit_ip: str, app
         f"Index {ContinuousWrites.INDEX_NAME} has {initial_count} documents, starting there"
     )
     writer = ContinuousWrites(ops_test, app, initial_count=initial_count)
-    await writer.start()
-    time.sleep(60)
+    await writer.start(is_bulk=True)
```
Mehdi-Bendriss, line 522: nit: `is_bulk` set to True by default. No need to specify it. We generally don't use the "non bulk" write as it's not representative of how OpenSearch is used.

Mehdi-Bendriss (commented):
Thanks Pedro. Good work. 
We need to add a test which checks the output of various  actions when not valid to run them, i.e: running a backup on an ML or data cluster

tests/integration/ha/continuous_writes.py
phvalguima: @Mehdi-Bendriss not sure I follow?? Originally, we [used to have write_value starting as 0 (an int)](https://github.com/canonical/opensearch-operator/blob/de29af17d395d50e4c3fa2c97f1cdc4178d4981d/tests/integration/ha/continuous_writes.py#L181C45-L181C46):
```
>>> write_value = 0
>>> write_value = (100 * write_value) + 99
>>> write_value
99
```

This `write_value += 99` does effectively the same thing, BUT it fixes a problem: if we have a backup recovered, `series_index` will start from a non-zero value and w we want to have: `max-id`, `write_value` and `last_expected_id` all matching nonetheless as soon as we restart `cwrites`.

If a failure happens and no docs are written, then `written_value (99 now) != 0`, it will fail.

So, I was having a situation where a recovered `series_index` starts with, say, 101 docs. If we add 10 more docs and run the original formula: `write_value == 11100` now.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -126,6 +176,26 @@ class OpenSearchRestoreIndexClosingError(OpenSearchRestoreError):
     """Exception thrown when restore fails to close indices."""
 
 
+def _load_s3_credentials_data(raw_relation_data: RelationDataContent) -> dict[str, str]:
+    """Loads the relation data from the peer-cluster relation from a sub-key."""
+    connection_data = {}
+    relation_data = (
+        raw_relation_data.get("data", {})
+        if isinstance(raw_relation_data.get("data", {}), dict)
```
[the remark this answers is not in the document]
phvalguima: Not really, I am getting both there.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -149,15 +219,229 @@ class BackupServiceState(BaseStrEnum):
...
+                f"Some mandatory fields: {missing_options} are not present, do not emit credential change event!"
+            )
+
+    def _load_relation_data(self, raw_relation_data: RelationDataContent) -> dict[str, str]:
+        """Loads the relation data from the peer-cluster relation from a sub-key."""
+        return _load_s3_credentials_data(raw_relation_data)
+
+
+class OpenSearchBackupBase(Object):
+    """Works as parent for all backup classes.
```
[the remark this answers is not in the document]
phvalguima: That behavior is used on other classes later on as well.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -149,15 +219,229 @@ class BackupServiceState(BaseStrEnum):
...
+    on = S3CredentialRequiresEvents()  # pyright: ignore [reportGeneralTypeIssues]
+
+    S3_REQUIRED_OPTIONS = ["access-key", "secret-key"]
+
+    def _on_relation_joined(self, _) -> None:
+        """Reimplements the relation joined event to defer it."""
+        # We do not need this method, so we just overload it
+        pass
+
+    def _on_relation_changed(self, event) -> None:
```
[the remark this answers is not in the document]
phvalguima: Yes, I am simplifying it following your feedback about AliastGenerators.

lib/charms/opensearch/v0/opensearch_backups.py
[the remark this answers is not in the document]
Mehdi-Bendriss: Yes, but what I'm saying is that `RelationDataContent` is a Dict[str, str], the values are always strings in the dictionary that you retrieve with `self.relation.data[self.relation.app]`. 
So: `json.loads(raw_relation_data.get("data", {})` is enough here

tests/integration/ha/continuous_writes.py
Mehdi-Bendriss: hmmm i understand, this was indeed not designed to have a "resume" functtion

lib/charms/opensearch/v0/opensearch_backups.py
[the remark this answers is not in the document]
Mehdi-Bendriss: Were AliasGenerators implemented? I see the wrangling of `-` / `_` on the name, still done somewhere

Mehdi-Bendriss (commented):
Thanks Pedro, I left some comments

lib/charms/opensearch/v0/constants_charm.py
```diff
@@ -61,6 +61,11 @@
 )
 PluginConfigError = "Unexpected error during plugin configuration, check the logs"
 BackupSetupFailed = "Backup setup failed, check logs for details"
+S3RelMissing = "Backup failover cluster missing S3 relation."
+S3RelShouldNotExist = "This unit should not be related to S3"
+S3RelMismatch = "S3 relation mismatch between {} and clusters"
```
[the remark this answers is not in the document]
phvalguima: No, reviewing the states I have, {S3RelMismatch, S3RelShouldNotExist} replace it

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -153,15 +211,156 @@ class BackupServiceState(BaseStrEnum):
...
+            self.charm.on[S3_RELATION].relation_broken,
+        ]:
+            self.framework.observe(event, self._on_s3_relation_event)
+
+    @abc.abstractmethod
+    def _on_s3_relation_event(self, event: EventBase) -> None:
+        pass
+
+
+class OpenSearchWaitingDeploymentDescBackup(OpenSearchBackupBase):
```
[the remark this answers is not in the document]
phvalguima: We have to account for the period where we may have a `s3-relation` but we do not have yet a `DeploymentDescription`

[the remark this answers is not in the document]
phvalguima: Reverting it back

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -136,15 +136,11 @@ def _extra_conf(self, plugin_data: Dict[str, Any]) -> Optional[Dict[str, Any]]:
...
-        return (
-            self._charm.opensearch.is_node_up()
-            and len(self._charm._get_nodes(True)) == self._charm.app.planned_units()
-            and self._charm.health.apply()
-            in [
-                HealthColors.GREEN,
-                HealthColors.YELLOW,
-            ]
-        )
+        return self._charm.opensearch.is_node_up() and self._charm.health.apply() in [
```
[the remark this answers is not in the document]
phvalguima: Good catch!

Sorry, that is actually an old fixes (I had to co-exist with the PR fixing and this one to move forward).

I had this PR: https://github.com/canonical/opensearch-operator/pull/253/files

But given the fix is very small and I need to unblock, I am adding it here manually.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -129,6 +131,14 @@ def __init__(self, charm: "OpenSearchBaseCharm"):
             charm.on[self.relation_name].relation_departed,
             self._on_peer_cluster_relation_departed,
         )
+        self.framework.observe(
+            self.charm.on[TLS_RELATION].relation_changed,
+            self._on_certificate_changed,
+        )
```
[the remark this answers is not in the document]
phvalguima: It is impossible. This PR will fail waiting for TLS because, some times, the TLS certs get set *after* peer-cluster relations.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -412,6 +448,37 @@ def __init__(self, charm: "OpenSearchBaseCharm"):
...
```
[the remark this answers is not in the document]
phvalguima: Same comment as above.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -423,6 +490,24 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):
...
+        # fetch the success data
+        data = PeerClusterRelData.from_str(data["data"])
+
+        # process the peer cluster data
+        self._process_peer_cluster_data(event, data)
+
+    def _process_peer_cluster_data(self, event: RelationChangedEvent, data) -> None:
+        if not self.charm.unit.is_leader():
+            return
+
```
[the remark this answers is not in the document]
phvalguima: Same comment as above

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -431,19 +516,13 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):
             event.defer()
             return
 
-        if not (data := event.relation.data.get(event.app)):
-            return
```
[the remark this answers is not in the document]
phvalguima: It is received as a parameter from `_process_peer_cluster_data`; and I need `_process_peer_cluster_data` because I also need to process TLS events.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -431,19 +516,13 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):
...
         # fetch main and failover clusters relations ids if any
         orchestrators = self._orchestrators(event, data, deployment_desc)
 
         # check errors sent by providers
         if self._error_set_from_providers(orchestrators, data, event.relation.id):
             return
 
-        # fetch the success data
-        data = PeerClusterRelData.from_str(data["data"])
-
```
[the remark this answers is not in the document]
phvalguima: Because I am processing it at the function caller level.

lib/charms/opensearch/v0/opensearch_backups.py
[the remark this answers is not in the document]
phvalguima: Sorry, this code is gone. Closing this comment.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -164,6 +165,7 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):
 
         if not (candidate_failover_app := data.get("candidate_failover_orchestrator_app")):
             self.refresh_relation_data(event)
+            event.defer()
```
Mehdi-Bendriss, line 168: I don't think that's needed, we don't have to have a failover, a main orchestrator is plenty and the only required orchestrator.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -172,6 +174,7 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):
         if orchestrators.failover_app and orchestrators.failover_rel_id in target_relation_ids:
             logger.info("A failover cluster orchestrator is already registered.")
             self.refresh_relation_data(event)
+            event.defer()
```
Mehdi-Bendriss, line 177: why is this needed? this condition simply means that the failover candidacy got rejected from a candidate, because one was already registered

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -823,7 +826,17 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
         if self.opensearch.is_started():
             try:
                 self._post_start_init(event)
-            except (OpenSearchHttpError, OpenSearchNotFullyReadyError):
+            except (
+                OpenSearchHttpError,
+                OpenSearchStartTimeoutError,
```
Mehdi-Bendriss, line 831: this is not raised by `post_start_init`

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -823,7 +826,17 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
...
+                OpenSearchNotFullyReadyError,
+            ):
+                event.defer()
+            except (OpenSearchStartError, OpenSearchUserMgmtError) as e:
```
Mehdi-Bendriss, line 835:
```suggestion
            except OpenSearchUserMgmtError as e:
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -868,9 +881,10 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
                 )
             )
             self._post_start_init(event)
-        except (OpenSearchStartTimeoutError, OpenSearchNotFullyReadyError):
+        except (OpenSearchHttpError, OpenSearchStartTimeoutError, OpenSearchNotFullyReadyError):
             event.defer()
-        except OpenSearchStartError as e:
+        except (OpenSearchStartError, OpenSearchUserMgmtError) as e:
+            # Either generic start failure or cluster is not read to create the internal users
```
Mehdi-Bendriss, line 887: when would the cluster not be ready to create the users?

Mehdi-Bendriss (commented):
Thanks Pedro, I believe this is the last pass to polish things up

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: If we do not have enough peers to form a cluster? E.g. this is the first unit booting up

Mehdi-Bendriss (approved):

juditnovak (approved):
My second reveiw is a bit superficial yet with the sold level of testing (HA!) I'm confident that the changes (introduced as a team) since my last reveiw only got the code more solid :-)

Issue #104, opened 2023-08-03
Juju status does not show open ports


gustavosr98:
## Steps to reproduce
1. Deploy
2. Run:
```bash
$ juju status
```

## Expected behavior
I would expect to see the workload open ports under the column ports when running `juju status`  

## Actual behavior

```
$ juju status
Model       Controller         Cloud/Region     Version  SLA          Timestamp
opensearch  google-controller  google/us-east1  2.9.44   unsupported  16:38:20-05:00

App                        Version  Status   Scale  Charm                      Channel      Rev  Exposed  Message
opensearch                          blocked      1  opensearch                 2/edge        28  no       1 or more 'replica' shards are not assigned, please scale your application up.
opensearch-di-admin                 blocked      1  data-integrator            latest/edge   13  no       Please specify either topic, index, or database name
tls-certificates-operator           active       1  tls-certificates-operator  latest/edge   27  no       

Unit                          Workload  Agent  Machine  Public address  Ports  Message
opensearch-di-admin/0*        blocked   idle   0        34.23.69.175           Please specify either topic, index, or database name
opensearch/0*                 active    idle   0        34.23.69.175           
tls-certificates-operator/0*  active    idle   0        34.23.69.175           

Machine  State    Address       Inst id        Series  AZ          Message
0        started  34.23.69.175  juju-952c6f-0  jammy   us-east1-b  RUNNING
```

## Versions
Operating system: Ubuntu Jammy
Juju CLI: 2.9.44-ubuntu-amd64
Juju Controller: 2.9.44
Charm revision: 2/edge (rev 2)
Cloud Substrate: GCP

## Additional context
Workaround
```
juju run -u opensearch/0 'open-port 9200/TCP'
```

reneradoi:
This issue was fixed with https://github.com/canonical/opensearch-operator/pull/249:

```
juju deploy opensearch --channel 2/edge
Located charm "opensearch" in charm-hub, revision 64
Deploying "opensearch" from charm-hub charm "opensearch", revision 64 in channel 2/edge on ubuntu@22.04/stable

juju status
Model       Controller      Cloud/Region      Version  SLA          Timestamp
opensearch  aws-opensearch  aws/eu-central-1  3.1.8    unsupported  08:33:54Z

App                        Version  Status       Scale  Charm                      Channel  Rev  Exposed  Message
opensearch                          maintenance      1  opensearch                 2/edge    64  no       Plugin configuration check.
[...]

PR #249, opened 2024-04-22
DPE-2333 show open ports

reneradoi:
## Issue
When starting opensearch, port 9200 was not shown as open port in `juju status` or `juju show-unit`, as described in https://github.com/canonical/opensearch-operator/issues/104

## Solution
After startup of opensearch was completed, `self.unit.open_port`is called to set tcp port 9200 as open for the unit.

It is now displayed in `juju status`:
```
Model       Controller  Cloud/Region         Version  SLA          Timestamp
opensearch  opensearch  localhost/localhost  3.1.8    unsupported  09:24:11Z

App                        Version  Status  Scale  Charm                      Channel  Rev  Exposed  Message
opensearch                          active      1  opensearch                            7  no       
tls-certificates-operator           active      1  tls-certificates-operator  stable    22  no       

Unit                          Workload  Agent  Machine  Public address  Ports     Message
opensearch/5*                 active    idle   6        10.183.254.77   9200/tcp  
[...]
```

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -721,7 +721,7 @@ class OpenSearchBaseCharm(CharmBase):
             return
         self.peers_data.delete(Scope.UNIT, "started")
         if self.opensearch.is_started():
-            self.unit.open_port('tcp', 9200)
+            self.unit.open_port("tcp", 9200)
             try:
                 self._post_start_init(event)
             except (OpenSearchHttpError, OpenSearchNotFullyReadyError):
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -721,6 +721,7 @@ class OpenSearchBaseCharm(CharmBase):
             return
         self.peers_data.delete(Scope.UNIT, "started")
         if self.opensearch.is_started():
+            self.unit.set_ports(9200)
             try:
                 self._post_start_init(event)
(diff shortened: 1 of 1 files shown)


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -721,6 +721,7 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
             return
         self.peers_data.delete(Scope.UNIT, "started")
         if self.opensearch.is_started():
+            self.unit.open_port("tcp", 9200)
```
Mehdi-Bendriss, line 724: This would only apply in case a previous start had timed-out for ex [here](https://github.com/canonical/opensearch-operator/pull/249/files#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R773). A safer place would be at the end of `_post_start_init`, as it is guaranteed to be invoked after the current opensearch service stabilizes on the first try or a subsequent retry

Mehdi-Bendriss (commented):
Thank you! and welcome to your first contribution to the product 🎉 
I have one comment

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: Thanks for the comment, Mehdi, I got this suggestion also from Pedro, and it's now moved to _post_start_init.

phvalguima (approved):

Mehdi-Bendriss (approved):

PR #251, opened 2024-04-23
[DPE-4192] Update test_opensearch_tls

### LINKED ISSUE #250: Unit Test `test_opensearch_tls.py::test_get_sans` fail with wrong `sans_ip`
It is not 100% reproducible, but I can see `test_get_sans` failing most of the times in PR #248 with:

```
AssertionError: {'san[34 chars]: ['192.0.2.0', 'XX.XXX.XX.XXX', 'address1', '[55 chars]-0']} != {'san[34 chars]: ['1.1.1.1', 'XX.XXX.XX.XXX', 'address1', 'ad[53 chars]-0']}
  {'sans_dns': ['alias', 'nebula', 'opensearch-0'],
-  'sans_ip': ['192.0.2.0', 'XX.XXX.XX.XXX', 'address1', 'address2'],
?                -- ^ ^ ^

+  'sans_ip': ['1.1.1.1', 'XX.XXX.XX.XXX', 'address1', 'address2'],
?                 ^ ^ ^

   'sans_oid': ['1.2.3.4.5.5']}
```

Looking into the origins of this IP, I noticed that newer ops framework has [this commit](https://github.com/canonical/operator/commit/5804652253926fea5c2aae5952d3032cea12ca5f), that adds a network `192.0.2.0` if we do not have any existing networks.

We were originally using `patch_network_get` decorator on this test. I think we should move to `_TestModelBackend.add_network` instead.

phvalguima:
There was a recent update on the ops framework and it now explicitly checks for the `_TestModelBackend.networks` for networks set. If no networks are set, it will "create" one: `192.0.2.0`.

This PR adds the network 1.1.1.1 to resolve this issue in the TLS.

Closes #250

### DIFF
--- a/tests/unit/lib/test_opensearch_tls.py
+++ b/tests/unit/lib/test_opensearch_tls.py
@@ -33,6 +33,8 @@ class TestOpenSearchTLS(unittest.TestCase):
     def setUp(self, _put_admin_user) -> None:
         self.harness = Harness(OpenSearchOperatorCharm)
         self.addCleanup(self.harness.cleanup)
+        self.harness.add_network("1.1.1.1", endpoint=PeerRelationName)
+        self.harness.add_network("1.1.1.1", endpoint=TLS_RELATION)
         self.harness.begin()
 
         self.charm = self.harness.charm


zmraul (approved):

paulomach (approved):
excelente

PR #253, opened 2024-04-23
Make check_plugin_manager_ready more large-deployment friendly

phvalguima:
If we are running in a large deployments, we cannot assume anymore that `_get_nodes()` == `planned_units` for any given cluster.

Besides, the charm health may apply to certain nodes and not others. Given this is dealing with plugins, we add the HealthColors.IGNORE as an option.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_plugin_manager.py
+++ b/lib/charms/opensearch/v0/opensearch_plugin_manager.py
@@ -136,15 +136,11 @@ class OpenSearchPluginManager:
 
     def check_plugin_manager_ready(self) -> bool:
         """Checks if the plugin manager is ready to run."""
-        return (
-            self._charm.opensearch.is_node_up()
-            and self._charm.health.apply()
-            in [
-                HealthColors.GREEN,
-                HealthColors.YELLOW,
-                HealthColors.IGNORE,
-            ]
-        )
+        return self._charm.opensearch.is_node_up() and self._charm.health.apply() in [
+            HealthColors.GREEN,
+            HealthColors.YELLOW,
+            HealthColors.IGNORE,
+        ]
 
     def run(self) -> bool:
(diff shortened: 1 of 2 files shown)


lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -138,11 +138,19 @@ def check_plugin_manager_ready(self) -> bool:
         """Checks if the plugin manager is ready to run."""
         return (
             self._charm.opensearch.is_node_up()
-            and len(self._charm._get_nodes(True)) == self._charm.app.planned_units()
+            and len(
+                [
+                    x
+                    for x in self._charm._get_nodes(True)
+                    if x.name.startswith(self._charm.app.name)
```
Mehdi-Bendriss, line 145: This is risky, in case 2 apps: `my-app` and `my-app-2`
```suggestion
                    if "-".join(x.name.split("-")[:-1]) == self._charm.app.name
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -138,11 +138,19 @@ def check_plugin_manager_ready(self) -> bool:
         """Checks if the plugin manager is ready to run."""
         return (
             self._charm.opensearch.is_node_up()
```
Mehdi-Bendriss, line 140: Let's add this as a first check:
```
self._charm.peers_data.get(Scope.APP, "security_index_initialised", False)
```

This should save us a few subsequent checks when the first cluster spins up, or when the large relation data hasn't completely been populated.

Mehdi-Bendriss (changes requested):
Thanks Pedro - left a few comments.

phvalguima:
Merged alongside DPE-3661 instead

PR #264, opened 2024-04-26
fix DPE-2119 snap installation error when re-using storage

reneradoi:


reneradoi:
not relevant anymore, will be solved with https://github.com/canonical/opensearch-operator/pull/272

Issue #243, opened 2024-04-18
Destroying opensearch application results in 2/3 errored units


phvalguima:
Seems it is not possible to remove opensearch application without `--force` anymore. In the end, it ends with 2x openseach units, both in error with:

```
Model       Controller         Cloud/Region   Version  SLA          Timestamp
opensearch  aws-tf-controller  aws/us-east-1  3.4.2    unsupported  14:16:53+02:00

SAAS                             Status  Store              URL
alertmanager-karma-dashboard     active  aws-tf-controller  admin/cos.alertmanager-karma-dashboard
grafana-grafana-dashboard        active  aws-tf-controller  admin/cos.grafana-grafana-dashboard
loki-logging                     active  aws-tf-controller  admin/cos.loki-logging
prometheus-receive-remote-write  active  aws-tf-controller  admin/cos.prometheus-receive-remote-write

App                       Version  Status   Scale  Charm                     Channel        Rev  Exposed  Message
grafana-agent                      unknown      0  grafana-agent             latest/stable   65  no       
opensearch                         active       2  opensearch                2/edge          60  no       
self-signed-certificates           active       1  self-signed-certificates  latest/stable   72  no       
ubuntu                    22.04    active       1  ubuntu                    latest/stable   24  no       

Unit                         Workload  Agent  Machine  Public address   Ports  Message
opensearch/1*                error     idle   2        192.168.235.125         hook failed: "storage-detaching"
opensearch/2                 error     idle   3        192.168.235.252         hook failed: "storage-detaching"
self-signed-certificates/0*  active    idle   0        192.168.235.97          
ubuntu/0*                    active    idle   4        192.168.235.243         

Machine  State    Address          Inst id              Base          AZ          Message
0        started  192.168.235.97   i-071c5e7d69b1c481e  ubuntu@22.04  us-east-1a  running
2        started  192.168.235.125  i-088477c2ef3b121f9  ubuntu@22.04  us-east-1a  running
3        started  192.168.235.252  i-01d8ebfd3ae8d2a24  ubuntu@22.04  us-east-1a  running
4        started  192.168.235.243  i-074de4290a1500f16  ubuntu@22.04  us-east-1a  running
```

Full logs: https://pastebin.ubuntu.com/p/vHxJX9rWdr/

Core error being:

```
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-opensearch-1/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 272, in request
[...]

PR #272, opened 2024-04-30
DPE-2119-fix-snap-issue-when-reusing-storage

reneradoi:
## Issue
When attaching an existing storage to a new unit, 2 issues happen:

- Snap install failed because of permissions / ownership of directories 
- snap_common gets completely deleted

## Solution
- bump snap version, use the fixed one
- enhance test coverage for integration tests

### DIFF
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -101,7 +101,7 @@ KibanaserverUser = "kibanaserver"
 KibanaserverRole = "kibana_server"
 
 # Opensearch Snap revision
-OPENSEARCH_SNAP_REVISION = 40  # Keep in sync with `workload_version` file
+OPENSEARCH_SNAP_REVISION = 47  # Keep in sync with `workload_version` file
 
 # User-face Backup ID format
 OPENSEARCH_BACKUP_ID_FORMAT = "%Y-%m-%dT%H:%M:%SZ"


reneradoi:
This PR updates the installed snap to the fixed version which can install on existing storage and does not overwrite the data in snap_common.

Other content of this PR:

The integration test `test_storage_reuse_after_scale_down` is now working. I've:
- created a storage pool on lxd
- deployed the application with storage from that pool
- adjusted the test workflow a little to make it more robust
- re-attaching storage to a new unit now tests correct now

The test case `test_storage_reuse_in_new_cluster_after_app_removal` is being skipped. Currently this test fails all the time, I first need to understand how this is supposed to work before fixing this test.

tests/integration/ha/test_storage.py
```diff
@@ -30,11 +30,14 @@ async def test_build_and_deploy(ops_test: OpsTest) -> None:
...
     await ops_test.model.set_config(MODEL_CONFIG)
+    # this assumes the test is run on a lxd cloud
+    await ops_test.model.create_storage_pool("opensearch-pool", "lxd")
+    storage = {"opensearch-data": {"pool": "opensearch-pool", "size": 2048}}
     # Deploy TLS Certificates operator.
     config = {"ca-common-name": "CN_CA"}
     await asyncio.gather(
         ops_test.model.deploy(TLS_CERTIFICATES_APP_NAME, channel="stable", config=config),
-        ops_test.model.deploy(my_charm, num_units=1, series=SERIES),
+        ops_test.model.deploy(my_charm, num_units=1, series=SERIES, storage=storage),
```
Mehdi-Bendriss, line 40: nice

tests/integration/ha/test_storage.py
```diff
@@ -61,33 +64,29 @@ async def test_storage_reuse_after_scale_down(
...
 
+    # scale up to 2 units
+    await ops_test.model.applications[app].add_unit(count=1)
+    await ops_test.model.wait_for_idle(
+        apps=[app],
+        status="active",
+        timeout=1000,
+        wait_for_exact_units=2,
+    )
+
```
Mehdi-Bendriss, line 77: I believe we could move the line `writes_result = await c_writes.stop()` to here, so that enough writes get written by then (the scale up substitutes the need for time.sleep, this will be useful in the future when we'd like to ensure that the data itself has been reused)

tests/integration/ha/test_storage.py
```diff
@@ -61,33 +64,29 @@ async def test_storage_reuse_after_scale_down(
...
     # get unit info
-    unit_id = get_application_unit_ids(ops_test, app)[0]
+    unit_id = get_application_unit_ids(ops_test, app)[1]
     unit_storage_id = storage_id(ops_test, app, unit_id)
```
Mehdi-Bendriss, line 81: Could we add a text file to `SNAP_COMMON/..` (/var/snap/opensearch/common/..) the unit we're about to destroy? so that we check that it's been restored and not ovewritten on_install?

tests/integration/ha/test_storage.py
```diff
@@ -113,6 +112,7 @@ async def test_storage_reuse_after_scale_down(
 
 @pytest.mark.group(1)
 @pytest.mark.abort_on_fail
+@pytest.mark.skip(reason="This test does not work currently, need to clarify the functionality.")
```
Mehdi-Bendriss, line 115: The goal here is essentially to destroy a cluster while keeping the storage, then to deploy a new cluster with the previously kept storage from the previous cluster, and then see if the data of the continuous writes has restored.

Mehdi-Bendriss (commented):
Thanks René! Could we add a test which does:
- stopping continuous writes 
- scale-down to 0 units (while keeping storage around)
- scale-up to N units with storage attach
- check if the continuous writes data has been indeed been kept and is usable

tests/integration/ha/test_storage.py
reneradoi: addressed in this commit: https://github.com/canonical/opensearch-operator/pull/272/commits/c7081372247dfdc55b52cf3bccef9d7f67f598f6

tests/integration/ha/test_storage.py
reneradoi: done

tests/integration/ha/test_storage.py
```diff
@@ -98,21 +103,77 @@ async def test_storage_reuse_after_scale_down(
...
+    await ops_test.model.wait_for_idle(
+        apps=[app],
+        status="active",
+        timeout=1000,
+        wait_for_exact_units=len(unit_ids),
+    )
+
+    # check if data is also imported
+    assert writes_result.count == (await c_writes.count())
+    assert writes_result.max_stored_id == (await c_writes.max_stored_id())
```
phvalguima, line 171: Can you also add the following test:

```suggestion
    assert writes_result.max_stored_id == (await c_writes.max_stored_id())

    # Restart it, so we can validate the cluster is still working
    c_writes = ContinuousWrites(ops_test, app, initial_count=writes_result.count)
    await c_writes.start()
    await assert_continuous_writes_increasing(c_writes)
    # final validation
    await assert_continuous_writes_consistency(ops_test, c_writes, app)
```

phvalguima, line 171: Can you also add the following test:

```suggestion
    assert writes_result.max_stored_id == (await c_writes.max_stored_id())

    # Restart it, so we can validate the cluster is still working
    c_writes = ContinuousWrites(ops_test, app, initial_count=writes_result.count)
    await c_writes.start()
    await assert_continuous_writes_increasing(c_writes)
    # final validation
    await assert_continuous_writes_consistency(ops_test, c_writes, app)
```

tests/integration/ha/test_storage.py
```diff
@@ -98,21 +103,77 @@ async def test_storage_reuse_after_scale_down(
...
     assert writes_result.max_stored_id == (await c_writes.max_stored_id())
 
+    # check if the testfile is still there or was overwritten on installation
+    check_testfile_cmd = f"juju ssh {app}/{new_unit_id} -q sudo ls {testfile}"
+    assert testfile == subprocess.getoutput(check_testfile_cmd)
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+@pytest.mark.skip(reason="scaling down to zero and scaling back up doesn't work currently")
```
phvalguima, line 128: We need to make this test work, let's remove this skip.

tests/integration/ha/test_storage.py
```diff
@@ -124,7 +185,7 @@ async def test_storage_reuse_in_new_cluster_after_app_removal(
             "reuse of storage can only be used on deployments with persistent storage not on rootfs deployments"
         )
 
-    # scale-down to 1 if multiple units
+    # scale-up to 3 to make it a cluster
```
phvalguima, line 188: Add the same final checks

tests/integration/ha/test_storage.py
```diff
@@ -148,7 +209,7 @@ async def test_storage_reuse_in_new_cluster_after_app_removal(
         storage_ids.append(storage_id(ops_test, app, unit_id))
 
     # remove application
-    await ops_test.model.applications[app].destroy()
+    await ops_test.model.applications[app].destroy(force=True, no_wait=True)
```
phvalguima, line 212: app with `force=True` does not mean "removing machines with force=True". This is important because we have currently #243, which does not allow us to cleanly remove the app with a single `remove-application` command.

I'd recommend you replace this for [this logic](https://github.com/canonical/opensearch-operator/blob/4cc224a57d22f83ce4dee13ef3695319e151e2b8/tests/integration/ha/test_backups.py#L580-L588). Remove the machines one-by-one, then remove the app.

Also, you should wait for it (`no_wait=False`).

phvalguima (changes requested):
Looks good! Let's give this logic a try with the integration tests and see how it goes.

tests/integration/ha/test_storage.py
reneradoi: Thanks for providing that input, I've seen that error too and didn't fully understand yet the reason why removal failed. That's why I've skipped this test.

reneradoi: adjusted the logic

tests/integration/ha/test_storage.py
reneradoi: done

tests/integration/ha/test_storage.py
reneradoi: Removed the skip. The test fails each time I run it currently with "hook failure: leader elected". I've encountered that with the 2/edge version of charm and snap as well, so this is something we need to follow up.

reneradoi:
The PR is now ready for review again.

tests/integration/ha/test_storage.py
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
-# Copyright 2024 Canonical Ltd.
+# Copyright 2023 Canonical Ltd.
```
phvalguima, line 2: Rollback this line's change

tests/integration/ha/test_storage.py
```diff
@@ -97,23 +102,81 @@ async def test_storage_reuse_after_scale_down(
...
+
+    await ops_test.model.wait_for_idle(
+        apps=[app],
+        status="active",
+        timeout=1000,
+        wait_for_exact_units=len(unit_ids),
+    )
+
+    # check if data is also imported
+    assert writes_result.count == (await c_writes.count())
```
phvalguima, line 172: We are missing a check that we can continue to write to this cluster, as c_writes was stopped before

phvalguima (changes requested):

tests/integration/ha/test_storage.py
reneradoi: Yes you're right, somehow that got lost in the process. Have added it again.

lib/charms/opensearch/v0/constants_charm.py
```diff
@@ -108,7 +108,7 @@
 KibanaserverRole = "kibana_server"
 
 # Opensearch Snap revision
-OPENSEARCH_SNAP_REVISION = 51  # Keep in sync with `workload_version` file
```
phvalguima, line 111: Revert this change, please

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -205,8 +205,8 @@ def _unit_with_lock(self, host) -> str | None:
                 retries=3,
             )
         except OpenSearchHttpError as e:
-            if e.response_code == 404:
-                # No unit has lock
+            if e.response_code in [404, 503]:
+                # No unit has lock or index not available
```
phvalguima, line 209: As we've discussed, I think this is actually masking the problem. 503 means shards have not been alllocated for whatever reason.

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -240,7 +240,10 @@ def acquired(self) -> bool:  # noqa: C901
                 unit = self._unit_with_lock(host)
             except OpenSearchHttpError:
                 logger.exception("Error checking which unit has OpenSearch lock")
-                return False
+                # if the node lock cannot be acquired, fall back to peer databag lock
+                # this avoids hitting deadlock situations in cases where
+                # the .charm_node_lock index is not available
+                return self._peer.acquired
```
phvalguima, line 246: The `self._peer.acquired` is a last-resort. We should not be using it just because we had an HTTP error. It is supposed to be used on special cases only, e.g. we have 1 or 2 units only. I think it does fit your case, as you are destroying the application.

In reality, we should use it if unit count in peer relation is <=2.

phvalguima (changes requested):
Hey @reneradoi I had a similar issue as yours, where there were units initially available for shards but they were still hanging as "UNASSIGNED". Although that was not about the lock itself, I think we can troubleshoot your problem here the same way.

I've added this [logic here](https://github.com/canonical/opensearch-operator/pull/319/files#diff-29386b01645a55e8571c903122c13345e969c7b15d67aa82169e1d95dee82e43) to give me more logs. You have to tweak the "explain" API to target the charm lock itself.

lib/charms/opensearch/v0/constants_charm.py
reneradoi: Sure, this was not intended to go through, just to confirm the tests are passing with 2.13.0.

lib/charms/opensearch/v0/opensearch_locking.py
reneradoi: I agree. The idea of this change here is just to avoid raising an error to avoid deadlocks when removing the application.

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: Why should it be `<=2` and not `<=1`?

phvalguima: As we've discussed, 2-node deployment does not apply the automated role management feature.

lib/charms/opensearch/v0/constants_charm.py
phvalguima: @reneradoi the goal is to merge with 2.13.0 or 2.14.0 pinned?

phvalguima (approved):
LGTM. My only remark is that we need to dig more into what is causing the 503.

On a separate PR, we should log 503 errors and its explanation, as follows:
```
GET _cluster/allocation/explain
{
  "index": ".charm-node-lock",
}
```

reneradoi:
> LGTM. My only remark is that we need to dig more into what is causing the 503.
> 
> On a separate PR, we should log 503 errors and its explanation, as follows:
> 
> ```
> GET _cluster/allocation/explain
> {
>   "index": ".charm-node-lock",
> }
> ```

Yes that's true, I'll add another issue or draft PR, will think about it.

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: For completeness: we no longer do automated role management. To clarify further, `<= 2` would have been a temporary workaround for the edge case where a cluster finds itself with only 2 nodes both of which having the `cluster_manager` role. One of these nodes should be added to the voting exclusion list in order to guarantee the possibility of an election in the event where something takes them both offline at the same time.

I prefer that we keep `<= 1` and address the voting exclusion separately. We're more likely to face the locking problems due to `<= 2` in the day to day development than the voting exclusion issue.

Let's please create an issue for the voting exclusion.

reneradoi: <= 1 is updated, issue is created here: https://github.com/canonical/opensearch-operator/issues/327

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #279, opened 2024-05-01
[DPE-4138] Fix failures raised by integration tests on CI

Mehdi-Bendriss:
## Issue
This PR addresses [DPE-4138](https://warthogs.atlassian.net/browse/DPE-4138), namely, this PR:
- reworks the health API
- reworks applying health on various parts of the charm
- fixes the replication of each "built-in" index
- reworks "peer-rel" trigger logic
- removes the roles re-computing from the charm / tests
- fixes various bug
- fixes integration tests
- fixes client-integration test with failing secrets / requests 
- updates libs 

[DPE-4138]: https://warthogs.atlassian.net/browse/DPE-4138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -115,3 +118,22 @@ def relation_departure_reason(charm: CharmBase, relation_name: str) -> RelDepart
         return RelDepartureReason.SCALE_DOWN
 
     return RelDepartureReason.REL_BROKEN
+
+
+def trigger_peer_rel_changed(
```
phvalguima, line 123: This is a potential source of noise, why is this needed?

lib/charms/opensearch/v0/opensearch_health.py
```diff
@@ -58,42 +61,87 @@ def apply(
...
+        # is being relocated. Examples are: simple stop, unit removal, upgrade
+        raise OpenSearchHAError("Shards haven't completed relocating.")
+
+    def _apply_for_app(self, status: str) -> None:
         """Cluster wide / app status."""
         if not self._charm.unit.is_leader():
-            # this is needed in case the leader is in an error state and doesn't
-            # report the status itself
-            self._charm.peers_data.put(Scope.UNIT, "health", status)
+            trigger_peer_rel_changed(self._charm, on_other_units=True)
```
phvalguima, line 144: If we will apply "something" in the peer relation, why not applying the `status`?

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -88,6 +89,11 @@ def run(self) -> None:
                 Scope.APP, "deployment-description", deployment_desc.to_dict()
             )
 
+        if deployment_desc.start == StartMode.WITH_GENERATED_ROLES:
+            # trigger roles change on the leader, other units will have their peer-rel-changed
+            # event triggered
+            trigger_peer_rel_changed(self._charm, on_other_units=False, on_current_unit=True)
```
phvalguima, line 95: There is already a "put" above. Is that not enough?

phvalguima (approved):

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -603,40 +607,49 @@ def _on_update_status(self, event: UpdateStatusEvent):
...
-            if self.plugin_manager.run():
+
+            if self.plugin_manager.run() and not restart_requested:
                 if self.upgrade_in_progress:
                     logger.warning(
                         "Changing config during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
                     )
                     event.defer()
                     return
+
```
phvalguima, line 652: We should move any "plugin_manager / defer" *before* the line 629:
```
self.opensearch_peer_cm.run()
```
As that `run()` will make a call to `trigger_peer_rel_changed`

phvalguima: Please, also a very big comment banner informing: "no defer after this line", with an explanation on *why* we need that precaution.

lib/charms/opensearch/v0/helper_charm.py
Mehdi-Bendriss: It's needed for 2 things: 
1. "roles changes" when manually requested by user (config change) - this is invoked, and triggers a peer rel changed event, where the whole topology is again recomputed and units restart
2. `health.apply(app=True)`, when called from a non leader unit - this unit computes the health but since it cannot write on the app status, it will trigger a peer rel changed event to be caught by the leader who will then compute the status and display it on the app level

lib/charms/opensearch/v0/opensearch_health.py
```diff
@@ -58,42 +61,87 @@ def apply(
...
-            if not compute_health:
-                return HealthColors.IGNORE
+        status = self.get(
+            wait_for_green_first=wait_for_green_first,
+            use_localhost=use_localhost,
+        )
+        logger.info(f"Current health of cluster: {status}")
+
+        if app:
+            self._apply_for_app(status)
```
phvalguima, line 71: So, this OpenSearchHealth.apply, if calling apply, there is a potential to call the "trigger defer" method. We must make sure if: (1) we need this "trigger defer" in "_apply_for_app"; or (2) if we do, then we need to ensure there is no risk for any OpenSearchHealth.apply() to be called before a defer

lib/charms/opensearch/v0/opensearch_health.py
phvalguima: Actually, this is never going to be called:
`_apply_for_app` only runs for app-level, this is supposed to never run at app-level (not leader).

Mehdi-Bendriss: Because the status is a variable that changes with time, setting/getting it from the peer rel data may lead to outdated status reporting (we used to do it like this on scale-down, it didn't work well)

phvalguima (changes requested):
My main concern is the trigger mechanism. There is a potential of getting trapped between deferrals and continuously changing the peer relation data.

lib/charms/opensearch/v0/opensearch_health.py
phvalguima: @Mehdi-Bendriss not sure if you saw my 2nd comment here:

> Actually, this is never going to be called:
_apply_for_app only runs for app-level, this is supposed to never run at app-level (not leader).

(It was 1 min apart from yours)

In a nutshell, if I understood correctly that should never be called.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
Mehdi-Bendriss: it would be but only if `deployment_desc.state.value == State.ACTIVE` - otherwise not

lib/charms/opensearch/v0/helper_charm.py
phvalguima: If we are so coupled, why not running it all under the same config-changed?

phvalguima: If these events are so entangled together, why not running it all under the same config-changed?

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: Not sure I understand what you mean exactly here.

> As that run() will make a call to trigger_peer_rel_changed

No - this will only run if there was a change to the 2 config options of the app:
- init-hold
- roles 
You can see this on the logic of `self.opensearch_peer_cm.run()` - it only concerns `exisiting clusters` which we're updating

lib/charms/opensearch/v0/opensearch_health.py
```diff
@@ -58,42 +61,87 @@ def apply(
         app: bool = True,
```
phvalguima, line 61: @Mehdi-Bendriss we should default this to False, as `self._apply_for_app()` changes the peer relation directly

lib/charms/opensearch/v0/opensearch_health.py
Mehdi-Bendriss: This will only happen in the case where, a NON-leader unit wants to apply the health to application status (which it cannot, so it emits this event so that the leader is notified that it must compute it). 

and this happens in exactly 2 cases:
- `_on_opensearch_data_storage_detaching`
- `_post_start_init`

lib/charms/opensearch/v0/opensearch_health.py
Mehdi-Bendriss: Same answer as https://github.com/canonical/opensearch-operator/pull/279#discussion_r1589336622

lib/charms/opensearch/v0/helper_charm.py
Mehdi-Bendriss: because this coming from when we were doing automated roles-rebalancing, which we're no longer doing sincea couple of weeks. This is something we're planning to address in the next cycle

phvalguima (approved):
The PR fixes most of the tests. The remaining will be investigated on `large-deployments-tests` PR. 

There is one concern I have we must keep a close watch: `trigger_peer_rel_changed` has the potential of endless calls to `opensearch-peers-changed` hook. Specially with `_post_start_init`.

PR #282, opened 2024-05-02
[DPE-4196] Plugin Management Refactor

### LINKED ISSUE #252: [RFE] Extend plugin manager
The plugin manager should support more use cases:
* More than one relation per plugin
* Same relation for multiple plugins
* A mix of configs and relations

### LINKED ISSUE #280: Plugin manager doesn't handle case where charm deployed with config options set
The charm should handle from the get-go the possibility to deploy the charm with config options set

phvalguima:
The main goal is to: (1) separate API call (`/_cluster`) between plugin-specific calls and cluster settings; (2) curb the requirements for restart for any plugin changes; (3) get a faster response time using cached entries wherever possible; (4) add new use-case with secret handling within the plugin logic itself; and (5) define models so we can standardize the plugin data exchanged between the objects and via relation.

# Dev experience

The idea is to make it easier to add management for separated plugins without impacting the rest of the code. A dev willing to add a new plugin must decide: do we need to manage a relation or just config options on the charm?
    If not, then we can add a `config-only plugin`
    If yes,  then we will need new a new object to encapsulate the plugin handling: the "DataProvider"

## The config-only plugin

These are plugin configured via config options. In this case, it is only needed to add a new `OpenSearchPlugin` -child class that manages the config options to be added or removed from the cluster.

For example, `opensearch-knn` receives the config from the charm and returns the options to be set in the `opensearch.yml`.

## The relation-based plugins

These plugins are more elaborate, as they have to process events specific to the given plugin. We also must consider the case of large deployments, where data may come via dedicated relation or the `peer-cluster` relation.

These plugins should be managed by a separate entity, named the `relation manager`. Defining a common structure for the relation manager is outside of the scope of this PR.

For example, `repository-s3` and OpenSearch backup.

# New Plugin Manager Infra

Now, the plugin manager is able to manage plugins that depend on config options, API calls and secrets. Whenever adding a new plugin, we should consider:
1) `opensearch_plugins.py`: this plugin should have a representation that is consumable by plugin_manager; it should be composed of all the configurations and keys to be added or removed to the cluster's main configuration
2) `opensearch_plugin_manager.py`: add the new plugin to the plugin dict; the manager must be able to instantiate this new plugin
3) `opensearch_{plugin-name}.py`: if the plugin is managed by a given relation, this lib will implement the relation manager and interface with OpenSearch's plugin-specific APIs
4) `models.py`: add any relation data model to this lib

## Using the new plugin data provider

While specific classes takes care of the plugin's API calls (e.g. `/_snapshot` API for the backup plugin is done by `OpenSearchBackup` class), the `data provider` facilitates the exchange of relation data between the specific class and the plugin_manager itself. This way, the plugin manager can apply any cluster-wide configurations that are needed for that plugin.

We need a class to do deal with relation specifics as some plugins may expect different relations depending on their deployment description, e.g. `OpenSearchBackupPlugin`. The `OpenSearchPluginDataProvider` encapsulates that logic away from the main plugin classes.

## Secret Management

Each plugin that handles the specific secrets must implement the secret management logic in its operation. The goal is to avoid filling the `opensearch_secrets.py` methods with `if`s for each plugin case and separating / isolating each plugin code.

# Remove unneeded restarts and add caching

We ensure that any configuration changes that come from plugin management are applied via API before being persisted on config files. If the API responds with a `200` status, then we should only write the new value to the configuration and finish without a need for restart.

In case the service is down and API is not available, we can assume we will eventually start the service back up. In this case, it suffices to write the config entries to the files and leave to the next start to pick them up.

This task is going to be divided into 3x parts:
1) Addresses low ranging fruits where we reduce the number of restarts and add caching support
2) Three main actions: (i) Merge `{add,delete}_plugin` together and its equivalents in `OpenSearchPluginConfig` class; (ii) we receive one big dictionary where a `key: None` means we want to simply delete that entry; and (iii) the main `OpenSearchKeystore` must observe secret changes and update its values accordingly
3) Returns unit tests: this is going to be commented out whilst Parts 1 and 2 happen, given this part of the code was covered with extensive testing

The current implementation of `plugin_manager.run` waits for the cluster to be started before processing its config changed. We relax this demand and open to the option where the cluster is not yet ready, so we can modify the configuration without issuing a restart request.

#252 is closed with `OpenSearchPluginRelationsHandler` interface. It allows plugins to define how they will handle its relation(s). `opensearch_backup` module extends this interface and defines a checker to process either small or large deployments details.

Other relevant changes:
1. Renaming method `check_plugin_manager_ready` to `check_plugin_manager_ready_for_api`
3. Any plugin that needs to manage things via API call should check the health of the cluster using `check_plugin_manager_ready_for_api`
4. Moving opensearch_distro.version to load the workload_version file we have present instead of an API call: this is two fold, 1. removes the dependency to a cluster to be ready and 2. makes this method in-sync with recent changes for upgrades logic
5. Waive the need of loading the default settings if this particular unit is powered down: which makes sense, in this moment we can do any config changes as we will eventually powered it back up later
6. If `/_cluster/settings` is available: apply the configs via API and do not add a restart request
7. On config-changed handler, the `upgrade_in_progress` check gets precedence and will continuously defer `config-changed` event until upgrade is finished before calling the plugin manager
8. Create a `OpenSearchKeyStoreNotReadyYetError`: responsible to identify the keystore has not been initialized yet across the cluster and hence, we cannot manage any plugins that use it; however, we always apply the `opensearch.yml` changes from that plugin
9. Add `cached_property` whenever it makes sense, also adds logic to clean the cache if there was any relevant changes to its content.

That still frees the config_changed to just call plugin_manager.run() before everything is set, as the run() method changes hard configuration only.

Closes #252, #280


phvalguima:
@carlcsaposs-canonical I have reshuffled the `self.upgrade_in_progress` here. Now, we check and defer before even doing the `plugin_manager.run()` (hence, no options will change until the upgrade is finished). Is this the right way? In another sense, it holds the user from doing any changes while the upgrade is in progress.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -54,11 +53,6 @@
         "config": "plugin_opensearch_knn",
         "relation": None,
     },
-    "repository-s3": {
-        "class": OpenSearchBackupPlugin,
-        "config": None,
-        "relation": "s3-credentials",
-    },
```
phvalguima, line 61: This plugin depends on `keystore`, therefore we cannot set it via plugin_manager. It will be managed by `opensearch-backups`.

.github/workflows/ci.yaml
```diff
@@ -18,19 +18,19 @@ jobs:
...
-    timeout-minutes: 10
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
-      - name: Install tox & poetry
-        run: |
-          pipx install tox
-          pipx install poetry
-      - name: Run tests
-        run: tox run -e unit
```
phvalguima, line 33: Temp.: Will revert it once the main integration tests pass and we are sure of the changes below (there are a lot of unit tests that will need to be updated here)

.github/workflows/ci.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
     name: Integration test charm | 3.4.2
     needs:
       - lint
-      - unit-test
```
phvalguima, line 73: Temp: we will eventually revert this

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -303,32 +330,85 @@ def _compute_settings(
...
+        # Update the configuration files
         if config.config_entries_to_del:
             self._opensearch_config.delete_plugin(config.config_entries_to_del)
         if config.config_entries_to_add:
             self._opensearch_config.add_plugin(config.config_entries_to_add)
-        return True
+
+        if cluster_settings_changed:
+            # We have changed the cluster settings, clean up the cache
+            del self.cluster_config
```
Mehdi-Bendriss, line 393: we only call the cluster_config once, is it worth caching?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -303,32 +330,85 @@ def _compute_settings(
...
+                    "PUT",
+                    "/_cluster/settings?flat_settings=true",
+                    payload={"persistent": config.config_entries_to_add},
+                )
+                cluster_settings_changed = True
 
-        # Update the configuration
+        # Update the configuration files
         if config.config_entries_to_del:
             self._opensearch_config.delete_plugin(config.config_entries_to_del)
```
Mehdi-Bendriss, line 387: why is there a mixing between API calls (above) and the same on the opensearch.yml level? how do you differentiate between the ones related to the file and the ones of the API?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -303,32 +330,85 @@ def _compute_settings(
...
+        # Final conclusion, we return a restart is needed if:
+        # (1) configuration changes are needed and applied in the files; and (2)
+        # the node is not up. For (2), we already checked if the node was up on
+        # _cluster_settings and, if not, cluster_settings_changed=True.
+        return all(
+            [
+                (config.secret_entries_to_add or config.secret_entries_to_del),
+                (
+                    (config.config_entries_to_add or config.config_entries_to_del)
+                    and not cluster_settings_changed
```
Mehdi-Bendriss, line 408: Maybe I'm missing something, but aren't these 2 conditions exclusive? 
`cluster_settings_changed == True` effectively means from the previous logic that `(config.config_entries_to_add or config.config_entries_to_del)`

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -340,28 +420,30 @@ def status(self, plugin: OpenSearchPlugin) -> PluginState:
...
-            return PluginState.DISABLED
-
-        if self._is_enabled(plugin):
-            return PluginState.ENABLED
-
+        try:
+            if not self._is_enabled(plugin):
+                return PluginState.DISABLED
+            elif self._user_requested_to_enable(plugin):
+                return PluginState.ENABLED
```
Mehdi-Bendriss, line 427: i believe here we're mixing 2 different logics, 1 to check whether the user requested to enable and the other whether it's actually enabled

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -340,28 +420,30 @@ def status(self, plugin: OpenSearchPlugin) -> PluginState:
...
-        if not (
-            self._charm.config.get(plugin_data["config"], False)
-            or self._is_plugin_relation_set(plugin_data["relation"])
-        ):
-            # User asked to disable this plugin
-            return False
-        return True
+        return self._charm.config.get(plugin_data["config"], False) or (
+            plugin_data["relation_handler"]
+            and plugin_data["relation_handler"](self._charm).is_relation_set()
```
Mehdi-Bendriss, line 445: doesn't this force all plugins to be enabled by relation?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1062,22 +1075,96 @@ def get_service_status(  # noqa: C901
...
+            relation = self._charm.model.get_relation(self._relation_name)
+            return relation is not None and relation.units != {}
+
+        # We are not not the MAIN_ORCHESTRATOR
+        # The peer-cluster relation will always exist, so we must check if
+        # the relation has the information we need or not.
+        return (
+            self._get_peer_rel_data is not None
+            and self._get_peer_rel_data.access_key
+            and self._get_peer_rel_data.secret_key
```
Mehdi-Bendriss, line 1123: similarly, this should be secrets based

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1062,22 +1075,96 @@ def get_service_status(  # noqa: C901
...
+        There is also the condition where the deployment description does not exist yet. In this
+        case, return the base class OpenSearchBackupBase. This class solely defers all s3-related
+        events until the deployment description is available and the actual S3 object is allocated.
+        """
+        if not self._charm.opensearch_peer_cm.deployment_desc():
+            # Temporary condition: we are waiting for CM to show up and define which type
+            # of cluster are we. Once we have that defined, then we will process.
+            return OpenSearchBackupBase(self._charm)
+        elif (
+            self._charm.opensearch_peer_cm.deployment_desc().typ
```
Mehdi-Bendriss, line 1166: this will need to change to `self._charm.opensearch_peer_cm.is_provider(typ="main")`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1062,22 +1075,96 @@ def get_service_status(  # noqa: C901
...
+        # The peer-cluster relation will always exist, so we must check if
+        # the relation has the information we need or not.
+        return (
+            self._get_peer_rel_data is not None
+            and self._get_peer_rel_data.access_key
+            and self._get_peer_rel_data.secret_key
+        )
+
+    @property
+    def _relation_name(self) -> str:
```
Mehdi-Bendriss, line 1127: same comment

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -332,6 +332,41 @@ class PluginState(BaseStrEnum):
     WAITING_FOR_UPGRADE = "waiting-for-upgrade"
 
 
+class OpenSearchPluginRelationsHandler(ABC):
+    """Implements the relation manager for each plugin.
+
+    Plugins may have one or more relations tied to them. This abstract class
```
Mehdi-Bendriss, line 338: I guess this is the case for the backup plugin with Azure or AWS S3, right?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -332,6 +332,41 @@ class PluginState(BaseStrEnum):
...
+    enables different modules to implement a class that can specify which
+    relations should plugin manager listen to.
+    """
+
+    _singleton = None
```
Mehdi-Bendriss, line 343: can you explain why?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -332,6 +332,41 @@ class PluginState(BaseStrEnum):
     WAITING_FOR_UPGRADE = "waiting-for-upgrade"
 
 
+class OpenSearchPluginRelationsHandler(ABC):
```
Mehdi-Bendriss, line 335: at which layer will this class be extended? given we already have the plugins classes

Mehdi-Bendriss (commented):
Thanks Pedro, I left some comments and questions

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: Yes

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: So, I am using the relation to prototype the plugin with a relation handler

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -133,23 +112,33 @@ def __init__(self, charm):
...
+        if not os.path.exists(self.keystore):
+            raise OpenSearchKeystoreNotReadyYetError()
+
         if not entries:
             return  # no key/value to remove, no need to request reload of keystore either
         for key in entries:
             self._delete(key)
 
-    def list(self, alias: str = None) -> List[str]:
+    @functools.cached_property
```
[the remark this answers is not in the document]
phvalguima: Well, we never faced an instance where we install / remove a plugin, so this function has always the same output. It is also very verbose in the debug log level. This cache allows to reduce a lot of that verbosity.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: Because we cannot have two or more objects created in the same hook, pointing to the same relation. Having a singleton makes this class more flexible, so the plugin manager can be called several times, each time it either allocates or reuses the same singleton, when needed. Otherwise the plugin manager would have to allocate this handler for each plugin case.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: The plugin manager, opensearch backups, i.e. objects the charm creates. However, we may call this handler's creation several times.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: @Mehdi-Bendriss is this part of the new merged PR?

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -34,46 +34,25 @@ class OpenSearchKeystoreError(OpenSearchError):
...
-            )
-        except OpenSearchCmdError as e:
-            raise OpenSearchKeystoreError(str(e))
 
     def list(self, alias: str = None) -> List[str]:
         """Lists the keys available in opensearch's keystore."""
         try:
             # Not using OPENSEARCH_BIN path
-            return self._opensearch._run_cmd(self._keytool, f"-v -list -keystore {self._keystore}")
+            return self._opensearch.run_bin(self._keytool, f"-v -list -keystore {self._keystore}")
```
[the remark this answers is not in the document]
phvalguima: The `opensearch-keystore` also sits in that bin folder:
```
sudo ls /snap/opensearch/current/usr/share/opensearch/bin/
opensearch  opensearch-cli  opensearch-env  opensearch-env-from-file  opensearch-keystore ...
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1062,22 +1075,96 @@ def get_service_status(  # noqa: C901
...
-    """
-    if not charm.opensearch_peer_cm.deployment_desc():
-        # Temporary condition: we are waiting for CM to show up and define which type
-        # of cluster are we. Once we have that defined, then we will process.
-        return OpenSearchBackupBase(charm)
-    elif charm.opensearch_peer_cm.deployment_desc().typ == DeploymentType.MAIN_ORCHESTRATOR:
-        return OpenSearchBackup(charm)
-    return OpenSearchNonOrchestratorClusterBackup(charm)
+    @property
+    def _get_peer_rel_data(self) -> Optional[S3RelDataCredentials]:
```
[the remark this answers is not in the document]
phvalguima: So, I am using the relation to prototype the plugin with a relation handler

lib/charms/opensearch/v0/opensearch_keystore.py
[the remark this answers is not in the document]
phvalguima: Okay, for the `Keystore` class, it indeed makes sense to use `run_cmd`.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -123,41 +136,31 @@ def get_plugin_status(self, plugin_class: Type[OpenSearchPlugin]) -> PluginState
...
-            == self._charm.app.planned_units()
-            and self._charm.health.apply()
-            in [
-                HealthColors.GREEN,
-                HealthColors.YELLOW,
-                HealthColors.IGNORE,
-            ]
-        )
+        return self._charm.peers_data.get(
+            Scope.APP, "security_index_initialised", False
```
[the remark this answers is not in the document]
phvalguima: Yes. We are not interested in this node being up, necessarily. So I am using that check + `health` to validate we are ready for API calls in general.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -258,13 +282,10 @@ def _configure_if_needed(self, plugin: OpenSearchPlugin) -> bool:
...
         try:
-            if self._user_requested_to_enable(plugin) or self.status(plugin) not in [
-                PluginState.ENABLED,
-                PluginState.WAITING_FOR_UPGRADE,
-            ]:
-                # Only considering "INSTALLED" or "WAITING FOR UPGRADE" status as it
-                # represents a plugin that has been installed but either not yet configured
-                # or user explicitly disabled.
+            if (
+                self._user_requested_to_enable(plugin)
```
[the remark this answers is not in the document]
phvalguima: We don't need to disable if we have an user requesting to enable this plugin or the plugin is already disabled.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -303,32 +330,85 @@ def _compute_settings(
...
+            # Postpone the exception now and set the remaining config.
+            keystore_ready = False
 
         current_settings, new_conf = self._compute_settings(config)
-        if current_settings == new_conf:
-            # Nothing to do here
-            logger.info("apply_config: nothing to do, return")
-            return False
+        if current_settings and new_conf and current_settings != new_conf:
+            if config.config_entries_to_del:
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss yes, and I will postpone this comment, as I am planning to shrink the adds / deletes of plugin manager on the 2nd part of this work anyways. Leaving it open so we can remember

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: Every plugin change from the plugin manager is persisted on file.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: Not necessarily. We may have one case (now, more an exception), where:
1) we have a change (i.e. config_entries_to_{add,del} are not None)
2) the cluster is active, but we did not use the API calls to apply the configs from (1)

This is the only situation that we must restart, as the node must be restart so the changes take place.

phvalguima: That means:
```
        # Update the configuration files
        if config.config_entries_to_del:
            self._opensearch_config.delete_plugin(config.config_entries_to_del)
        if config.config_entries_to_add:
            self._opensearch_config.add_plugin(config.config_entries_to_add)
```

may do a change, but `cluster_settings_changed = False`.

phvalguima: However, that expression I wrote at the end is indeed very convoluted. I will simplify that.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: In short, I am not right now making any differences. What I am doing there is making sure the configs are persisted in file and, if possible, we apply everything via API.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: That `elif` means: `self._is_enabled(plugin)==True` and the user still wants to keep that plugin - `self._user_requested_to_enable(plugin)==True`.

So, the full matrix would be:
```
self._is_enabled(plugin)==False and self._user_requested_to_enable(plugin)==<does not matter>
    return DISABLED

self._is_enabled(plugin)==True and self._user_requested_to_enable(plugin)==True
    return ENABLED

self._is_enabled(plugin)==True and self._user_requested_to_enable(plugin)==False
    return INSTALLED
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: No, KNN for example does not depend on relation_handler:
```
    "opensearch-knn": {
        "class": OpenSearchKnn,
        "config": "plugin_opensearch_knn",
        "relation_handler": None,
    },
```
That 2nd check is false as `plugin_data["relation_handler"]==None`

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: As discussed, in case we are interested in the large deployments only. The new API set: `self._charm.opensearch_peer_cm.is_{provider|consumer}` is more suitable.

If we *do not* want to differ between them, then we use the `self._charm.opensearch_peer_cm.deployment_desc().typ`.

The advantage of the provider / consumer: checks if the relation is stablished, has data, ready for consumption, etc.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -123,41 +136,31 @@ def get_plugin_status(self, plugin_class: Type[OpenSearchPlugin]) -> PluginState
...
-            return {
-                **relation.data[relation.app],
-                **self._charm_config,
-                "opensearch-version": self._opensearch.version,
-            }
-        return {**self._charm_config, "opensearch-version": self._opensearch.version}
-
-    def check_plugin_manager_ready(self) -> bool:
+        relation_handler = plugin_data.get("relation_handler")
+        data = relation_handler(self._charm).get_relation_data() if relation_handler else {}
```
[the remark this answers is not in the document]
phvalguima: As discussed, we should use the `s3_creds` model, which will contain the final values for the secrets, either picked up from the s3-integrator itself or from the secrets in the peer relation cluster.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: We will add another state: `ENABLING`, which represents:
```
self._is_enabled(plugin)==False and self._user_requested_to_enable(plugin)==False
    return DISABLED
self._is_enabled(plugin)==False and self._user_requested_to_enable(plugin)==True
    return ENABLING_NEEDED
self._is_enabled(plugin)==True and self._user_requested_to_enable(plugin)==True
    return ENABLED
self._is_enabled(plugin)==True and self._user_requested_to_enable(plugin)==False
    return DISABLING_NEEDED
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -332,6 +332,41 @@ class PluginState(BaseStrEnum):
...
+
+        It can mean the relation exists or not, simply put; or it can also mean a subset of data
+        exists within a bigger relation. One good example, peer-cluster is a single relation that
+        contains a lot of different data. In this case, we'd be interested in a subset of
+        its entire databag.
+        """
+        return NotImplementedError()
+
+    def get_relation_data(self) -> Dict[str, Any]:
+        """Returns the relation that the plugin manager should listen to.
```
phvalguima, line 362: Rename it to `data()`

lib/charms/opensearch/v0/opensearch_plugin_manager.py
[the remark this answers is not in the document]
phvalguima: Also added a generic `S3RelData` model, that can be used to load and validate `s3` relation data.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: This class will act as a connection between the `plugin_manager`, which only leads with OpenSearch overall APIs; and any other plugin-specific APIs.

phvalguima: The main issue here is that I need the databag at `plugin_manager` to do some configurations. However, we now have `small` and `large` deployments, which means more than one relation is possible for the same plugin.

That is becoming messy from the plugin_manager's perspective.

This class allows a given plugin (e.g. OpenSearchBackup) to instantiate a singleton that holds the databag according to its current config (e.g. using the s3-credentials or the peer-cluster relation) and the `plugin_manager` does not need to know anything else.

Issue #287, opened 2024-05-03
`_post_start_init` triggers continuous password updates on `COSUser`


phvalguima:
The _post_start_init routine may be re-executed multiple times on a single start, if conditions are not yet met for the node to start. [In this scenario, there is a risk of COSUser being recreated](https://github.com/canonical/opensearch-operator/blob/fda46cbf574febc88154df16706e3a59da5a866c/lib/charms/opensearch/v0/opensearch_base_charm.py#L970) multiple times, resulting in new passwords and, hence, `secret-changed` events.

We should adapt `_put_or_update_internal_user` to: (1) check for existing users if password was not provided, if user already present: leave the method; and (2) if password provided, then proceed with the update.

Also, the `typing` of the return value of `get_user` and its [docstring do not match](https://github.com/canonical/opensearch-operator/blob/fda46cbf574febc88154df16706e3a59da5a866c/lib/charms/opensearch/v0/opensearch_users.py#L137)

Mehdi-Bendriss:
Fixed on https://github.com/canonical/opensearch-operator/pull/289

PR #288, opened 2024-05-03
[DPE-4278] Avoid COSUser password re-creation on each _post_start_init run

phvalguima:
Closes: #287


Mehdi-Bendriss:
Fixed in https://github.com/canonical/opensearch-operator/pull/289

PR #291, opened 2024-05-04
Clean-up peer_cm status

phvalguima:


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -605,6 +605,14 @@ def _on_update_status(self, event: UpdateStatusEvent):
         # handle when/if certificates are expired
         self._check_certs_expiration(event)
 
+        if not (deployment_desc := self.opensearch_peer_cm.deployment_desc()):
+            # the deployment description hasn't finished being computed by the leader
+            return
+
+        # check if peer status needs to be cleaned
+        if self.unit.is_leader():
+            self.opensearch_peer_cm.apply_status_if_needed(deployment_desc)
```
Mehdi-Bendriss, line 614: Could we move this to way before in this method, exactly to the `else` of: 
```
if (health := self.health.apply(wait_for_green_first=True)) not in [
    HealthColors.GREEN,
    HealthColors.IGNORE,
]:
    event.defer()
else:
    self.opensearch_peer_cm.apply_status_if_needed(deployment_desc)
```

Mehdi-Bendriss (commented):
Thanks Pedro! 2 minor comments mainly about the positioning

Mehdi-Bendriss (approved):

Issue #309, opened 2024-05-24
Lock is not released when scaling down to zero units


reneradoi:
## Steps to reproduce
- juju add-model opensearch
- juju model-config --file ./cloudinit-userdata.yaml
- juju create-storage-pool opensearch-storage lxd volume-type=standard
- juju deploy opensearch-n 2 --channel 2/edge --storage opensearch-data=opensearch-storage,1G,1
- juju deploy self-signed-certificates
- juju config self-signed-certificates ca-common-name="CN_CA"
- juju relate self-signed-certificates opensearch
- juju remove-unit opensearch/0
- wait until opensearch/0 is gone
- juju remove-unit opensearch/1
- juju add-unit opensearch --attach-storage=opensearch-data/0
- wait until opensearch/2 is started
- juju add-unit opensearch --attach-storage=opensearch-data/1

## Expected behavior
both new units start up correctly

## Actual behavior
2nd new unit can't aquire lock and does not start

```
$ juju status --relations --storage
Model       Controller  Cloud/Region         Version  SLA          Timestamp
opensearch  opensearch  localhost/localhost  3.1.8    unsupported  08:08:21Z

App                       Version  Status  Scale  Charm                     Channel  Rev  Exposed  Message
opensearch                         active      2  opensearch                           0  no       
self-signed-certificates           active      1  self-signed-certificates  stable    72  no       

Unit                         Workload  Agent  Machine  Public address  Ports     Message
opensearch/2*                active    idle   3        10.27.170.215   9200/tcp  
opensearch/3                 waiting   idle   4        10.27.170.204             Requesting lock on operation: start
self-signed-certificates/0*  active    idle   2        10.27.170.207             

Machine  State    Address        Inst id        Base          AZ  Message
2        started  10.27.170.207  juju-a1b76d-2  ubuntu@22.04      Running
3        started  10.27.170.215  juju-a1b76d-3  ubuntu@22.04      Running
4        started  10.27.170.204  juju-a1b76d-4  ubuntu@22.04      Running

Integration provider                   Requirer                       Interface           Type     Message
opensearch:node-lock-fallback          opensearch:node-lock-fallback  node_lock_fallback  peer     
opensearch:opensearch-peers            opensearch:opensearch-peers    opensearch_peers    peer     
opensearch:upgrade-version-a           opensearch:upgrade-version-a   upgrade             peer     
self-signed-certificates:certificates  opensearch:certificates        tls-certificates    regular  

[...]

PR #312, opened 2024-05-24
[DPE-4421] Release stale node locks

reneradoi:
## Issue
https://github.com/canonical/opensearch-operator/issues/309

## Solution
Also release stale locks in the opensearch database from units no longer existing when releasing the lock a unit is currently holding.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_locking.py
+++ b/lib/charms/opensearch/v0/opensearch_locking.py
@@ -187,7 +187,6 @@ class OpenSearchNodeLock(ops.Object):
     """
 
     OPENSEARCH_INDEX = ".charm_node_lock"
-    _ENDPOINT_NAME = "node-lock-fallback"
 
     def __init__(self, charm: "opensearch_base_charm.OpenSearchBaseCharm"):
         super().__init__(charm, "opensearch-node-lock")
--- a/lib/charms/opensearch/v0/opensearch_locking.py
+++ b/lib/charms/opensearch/v0/opensearch_locking.py
@@ -186,6 +186,7 @@ class OpenSearchNodeLock(ops.Object):
     """
 
     OPENSEARCH_INDEX = ".charm_node_lock"
+    _ENDPOINT_NAME = "node-lock-fallback"
 
     def __init__(self, charm: "opensearch_base_charm.OpenSearchBaseCharm"):
         super().__init__(charm, "opensearch-node-lock")
@@ -343,7 +344,12 @@ class OpenSearchNodeLock(ops.Object):
         if host or alt_hosts:
             logger.debug("[Node lock] Checking which unit has opensearch lock")
             # Check if this unit currently has lock
-            if self._unit_with_lock(host) == self._charm.unit.name:
+            # or if there is a stale lock from a unit no longer existing
+            if (
+                self._unit_with_lock(host) == self._charm.unit.name
+                or self._unit_with_lock(host)
+                not in self._charm.model.get_relation(self._ENDPOINT_NAME).units
+            ):
                 logger.debug("[Node lock] Releasing opensearch lock")
                 # Delete document id 0
                 try:
--- a/lib/charms/opensearch/v0/opensearch_locking.py
+++ b/lib/charms/opensearch/v0/opensearch_locking.py
@@ -8,6 +8,7 @@ import os
 from typing import TYPE_CHECKING, List, Optional
 
 import ops
+from charms.opensearch.v0.constants_charm import PeerRelationName
 from charms.opensearch.v0.helper_cluster import ClusterState, ClusterTopology
 from charms.opensearch.v0.opensearch_exceptions import OpenSearchHttpError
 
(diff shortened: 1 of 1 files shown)


lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -343,7 +344,12 @@ def release(self):
         if host or alt_hosts:
             logger.debug("[Node lock] Checking which unit has opensearch lock")
             # Check if this unit currently has lock
-            if self._unit_with_lock(host) == self._charm.unit.name:
+            # or if there is a stale lock from a unit no longer existing
+            if (
+                self._unit_with_lock(host) == self._charm.unit.name
+                or self._unit_with_lock(host)
+                not in self._charm.model.get_relation(self._ENDPOINT_NAME).units
```
phvalguima, line 351: hey @reneradoi, you do not want to do:
```
self._unit_with_lock(host) not in self._charm.model.get_relation(self._ENDPOINT_NAME).units
```

OpenSearch clusters may be composed of several `juju applications`. This is the *large deployments* scenario, where we have each `juju app` running different node roles.
So, that endpoint will not always have all nodes available.

I'd recommend instead:
```
                self._unit_with_lock(host) == self._charm.unit.name
                or (
                      self.opensearch_peer_cm.deployment_desc().typ == DeploymentType.MAIN_ORCHESTRATOR
                      and self._charm.unit.is_leader()
                      and self._unit_with_lock(host)
                      not in self._A_UNIT_LIST
                )
```

Now, `self._A_UNIT_LIST` should be something similar to:
```
[self._charm.unit]
 + [unit for unit in rel.units for rel in self._charm.model.get_relation("peer-cluster-orchestrator")]
 + [unit for unit in rel.units for rel in self._charm.model.get_relation("peer-cluster")]
 + self._charm.model.get_relation(self._ENDPOINT_NAME).units
```

@Mehdi-Bendriss likely has simpler way to build this list above.

phvalguima (commented):
Hi @reneradoi: really welcome change! I left one comment as we need to think a bit more on the `if()` at the core of `release` method.

lib/charms/opensearch/v0/opensearch_locking.py
reneradoi: > OpenSearch clusters may be composed of several `juju applications`. This is the _large deployments_ scenario, where we have each `juju app` running different node roles.
> So, that endpoint will not always have all nodes available.

Good point, didn't think about that!

reneradoi: Testing this manually (with the setup described in the issue) and with the integration tests in CI pipeline was successfull. Please let me know if this solution is appropriate.

Mehdi-Bendriss: This is a bit tricky in the case of large deployments, as we don't have visibility over non-online units in other applications - which may be holding the lock.
it should be around:
```
unit_with_lock = self._unit_with_lock(host)

current_app_units = [u.name for u in charm.model.get_relation(PeerRelationName).units]
other_apps_units = ?? # for large deployments, we don't have this mechanism yet - for units that are not online

if (
    unit_with_lock
    and (
        unit_with_lock == self._charm.unit.name
        or unit_with_lock not in current_app_units
        or unit_with_lock not in other_apps_units
    )
)
```

I think that for large deployments we're missing the `other_apps_units` part

reneradoi:
> I believe the issue was caused by this change: https://github.com/canonical/opensearch-operator/pull/279/files#diff-f0ef92b3c155cc488340483aeb7a14f44e6577ccb3a328d8856fab4984072e57L248-R248
> 
> Proposal: solve the issue by reverting that change
> 
> also (for context), related: [#243 (comment)](https://github.com/canonical/opensearch-operator/issues/243#issuecomment-2071840157)

I've tested the proposed solution, but it only seems to work for the first unit to start up. As soon as the second unit is requesting the lock, it can't get it. This is from the `juju debug-log`:

```
unit-opensearch-1: 07:01:13 DEBUG unit.opensearch/1.juju-log [Node lock] No unit has opensearch lock
unit-opensearch-1: 07:01:13 DEBUG unit.opensearch/1.juju-log [Node lock] Using peer databag for lock
unit-opensearch-1: 07:01:13 DEBUG unit.opensearch/1.juju-log [Node lock] Not acquired. Unit with peer databag lock: None
unit-opensearch-1: 07:01:13 DEBUG unit.opensearch/1.juju-log Lock to start opensearch not acquired. Will retry next event
```
And it goes on like this with no end.

reneradoi:
Hi Carl, here's the full debug-log: [opensearch_locking_debug_log.log](https://github.com/canonical/opensearch-operator/files/15465641/opensearch_locking_debug_log.log)

For my understanding: The condition is `if not unit and online_nodes >= 2:`, where `online_nodes` are the ones in the `/_nodes` endpoint of the cluster. But if the 2nd node is not started yet, it is not in this list, and so length of this list is only 1?

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: Tracked here https://github.com/canonical/opensearch-operator/issues/314

lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -186,6 +186,7 @@ class OpenSearchNodeLock(ops.Object):
     """
 
     OPENSEARCH_INDEX = ".charm_node_lock"
+    _ENDPOINT_NAME = "node-lock-fallback"
```
Mehdi-Bendriss: I believe we can directly make use of the peer relation: `PeerRelationName`

lib/charms/opensearch/v0/opensearch_locking.py
Mehdi-Bendriss: Similarly, we can directly use `PeerRelationName` instead of `self._ENDPOINT_NAME`

Mehdi-Bendriss (commented):
Thanks René. It looks good. I left a couple of minor comments.

Mehdi-Bendriss (approved):
Thanks René!

phvalguima (approved):

PR #315, opened 2024-05-28
[DPE-4305][DPE-4454] Update to 2.14.0 and adds support for multi versions in small upgrade tests

phvalguima:
Update pinning and workload_version to 2.14.0.

Also extends integration testing to cover the jump between 2.12, 2.13 and current 2.14.

Adds a skip to large deployments upgrade test while waiting for DPE-4528.

Accompanying PR to update libs: https://github.com/canonical/opensearch-operator/pull/319


tests/integration/test_small_deployment_upgrades.py
```diff
@@ -19,10 +19,18 @@
...
+OPENSEARCH_CHANNEL = "2/edge"
+
 MACHINE_ID = 0
 
 
+VERSION_TO_REVISION = {
+    "2.13.0": 91,
+}
+FIRST_VERSION = "2.12.0"
+FIRST_REVISION = 90
```
Mehdi-Bendriss, line 31: nit: we could perhaps add these in the above dict, and only keep the `FIRST_VERSION` variable

Mehdi-Bendriss (approved):
Thanks Pedro

tests/integration/test_small_deployment_upgrades.py
phvalguima: I will do an following PR to add "2.14.0". I will merge this one thou

phvalguima: Filed ticket: https://warthogs.atlassian.net/browse/DPE-4560 to track that

Issue #204, opened 2024-03-16
[TLS] OpenSearch is not tracking invalidated certificates event


phvalguima:
We need to extend our logic to manage certificates to track the case where certificates are partially or all of them are invalidated by the TLS charms.

reneradoi:
Resolved with https://github.com/canonical/opensearch-operator/pull/318

PR #318, opened 2024-05-31
[DPE-3822] handle invalidated certificates

reneradoi:
## Issue

https://github.com/canonical/opensearch-operator/issues/204

## Solution
- implement handler for the `on.certificate_invalidated` event
- request new certificate for the invalidated one
- add unit test coverage in `tests/unit/lib/test_opensearch_tls.py:test_on_certificate_invalidated`


phvalguima (approved):
LGTM

Only remark is to add integration tests, but we can do it as a 2nd PR.

Mehdi-Bendriss (approved):

PR #319, opened 2024-06-02
[DPE-4560] Rollback test scenario from consider local built -> original charm rev + update libs

### LINKED ISSUE #324: One or more replica shards...
## Steps to reproduce

1. _Particularly_ loaded host system
2. Check out `opensearch-dsahboards-operator` and run pipeline:
```
 tox run -e integration -- tests/integration/test_upgrade.py --model testing --keep-models
```

## Expected behavior
No errorrs

## Actual behavior

See attached screenshots. The problem was permanent, the system didn't recover state (as timestamp on the top indicates).

Screenshot from 2024-06-07 15-31-41

Screenshot from 2024-06-07 15-45-11

## Versions

Operating system:  jammy

Juju CLI: 3.1.8-genericlinux-amd64

Juju agent: 3.1.8

Charm revision: Most likely 90 or 99 (in case caching may be applied on charmhub, 98 has a chance too)

LXD: 5.0.3 (?)

## Log output

Screenshot from 2024-06-07 15-47-50
Screenshot from 2024-06-07 15-44-24
Screenshot from 2024-06-07 15-41-58
Screenshot from 2024-06-07 15-39-24
Screenshot from 2024-06-07 15-36-06
Screenshot from 2024-06-07 15-35-13
Screenshot from 2024-06-07 15-33-37
Screenshot from 2024-06-07 15-21-01

phvalguima:
Update libs following changes from DPE-4305

[DPE-4560](https://warthogs.atlassian.net/browse/DPE-4560): updates the rollback test scenario to test each minor workload version (and its latest charm revision upstream) against the currently built charm.

Also closes an issue related to shard reassignment and upgrades by disabling `disk watermark decider`.

Closes #324

[DPE-4560]: https://warthogs.atlassian.net/browse/DPE-4560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -23,6 +23,7 @@ from tenacity import (
 
 from ..helpers import (
     APP_NAME,
+    IDLE_PERIOD,
     get_application_unit_ids,
     get_application_unit_ids_hostnames,
     get_application_unit_ids_ips,
@@ -30,7 +31,6 @@ from ..helpers import (
     http_request,
     juju_version_major,
     run_action,
-    IDLE_PERIOD,
 )
 from ..helpers_deployments import get_application_units, wait_until
 from .continuous_writes import ContinuousWrites
--- a/tests/integration/ha/helpers.py
+++ b/tests/integration/ha/helpers.py
@@ -30,7 +30,9 @@ from ..helpers import (
     http_request,
     juju_version_major,
     run_action,
+    IDLE_PERIOD,
 )
+from ..helpers_deployments import get_application_units, wait_until
 from .continuous_writes import ContinuousWrites
 from .helpers_data import index_docs_count
 
@@ -593,3 +595,65 @@ async def assert_restore_indices_and_compare_consistency(
     # We expect that new_count has a loss of documents and the numbers are different.
     # Check if we have data but not all of it.
     assert 0 < new_count < original_count
+
+
+async def assert_upgrade_to_local(
+    ops_test: OpsTest, cwrites: ContinuousWrites, local_charm: str
+) -> None:
+    """Does the upgrade to local and asserts continuous writes."""
+    app = (await app_name(ops_test)) or APP_NAME
+    units = await get_application_units(ops_test, app)
+    leader_id = [u.id for u in units if u.is_leader][0]
+
+    application = ops_test.model.applications[app]
+    action = await run_action(
+        ops_test,
+        leader_id,
+        "pre-upgrade-check",
+        app=app,
+    )
+    assert action.status == "completed"
+
+    async with ops_test.fast_forward():
+        logger.info("Refresh the charm")
+        await application.refresh(path=local_charm)
+
+        await wait_until(
+            ops_test,
+            apps=[app],
+            apps_statuses=["blocked"],
+            units_statuses=["active"],
+            wait_for_exact_units={
+                APP_NAME: 3,
+            },
+            timeout=1400,
+            idle_period=IDLE_PERIOD,
+        )
+
+        logger.info("Upgrade finished")
+        # Resume the upgrade
+        action = await run_action(
+            ops_test,
+            leader_id,
+            "resume-upgrade",
+            app=app,
+        )
+        logger.info(action)
+        assert action.status == "completed"
+
+        logger.info("Refresh is over, waiting for the charm to settle")
+        await wait_until(
+            ops_test,
+            apps=[app],
+            apps_statuses=["active"],
+            units_statuses=["active"],
+            wait_for_exact_units={
+                APP_NAME: 3,
+            },
+            timeout=1400,
+            idle_period=IDLE_PERIOD,
+        )
+
+    # continuous writes checks
+    await assert_continuous_writes_increasing(cwrites)
+    await assert_continuous_writes_consistency(ops_test, cwrites, [app])
--- a/tests/integration/ha/test_small_deployment_upgrades.py
+++ b/tests/integration/ha/test_small_deployment_upgrades.py
@@ -8,12 +8,12 @@ import subprocess
 import pytest
 from pytest_operator.plugin import OpsTest
 
-from .ha.continuous_writes import ContinuousWrites
-from .ha.helpers import app_name, assert_continuous_writes_consistency
-from .ha.test_horizontal_scaling import IDLE_PERIOD
-from .helpers import APP_NAME, MODEL_CONFIG, SERIES, run_action
-from .helpers_deployments import get_application_units, wait_until
-from .tls.test_tls import TLS_CERTIFICATES_APP_NAME
+from ..helpers import APP_NAME, MODEL_CONFIG, SERIES, run_action
+from ..helpers_deployments import get_application_units, wait_until
+from ..tls.test_tls import TLS_CERTIFICATES_APP_NAME
+from .continuous_writes import ContinuousWrites
+from .helpers import app_name, assert_upgrade_to_local
+from .test_horizontal_scaling import IDLE_PERIOD
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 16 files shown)


reneradoi (approved):
The test looks good!

tests/integration/test_small_deployment_upgrades_edge_channel.py
```diff
@@ -25,10 +25,9 @@
 
 
 VERSION_TO_REVISION = {
+    "2.12.0": 90,
     "2.13.0": 91,
 }
-FIRST_VERSION = "2.12.0"
```
Mehdi-Bendriss, line 30: I think we should keep this one as we're hardcoding `2.12.0` in other parts of the code.

tests/integration/test_small_deployment_upgrades_edge_channel.py
```diff
@@ -51,7 +50,7 @@ async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
 
 
 @pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "xlarge"])
-@pytest.mark.group(1)
+@pytest.mark.group("upgrade")
 @pytest.mark.abort_on_fail
 @pytest.mark.skip_if_deployed
 async def test_deploy_latest_from_channel(ops_test: OpsTest) -> None:
```
Mehdi-Bendriss, line 56: For consistency with other tests:

```suggestion
async def test_build_and_deploy(ops_test: OpsTest) -> None:
```

tests/integration/test_small_deployment_upgrades_edge_channel.py
```diff
@@ -63,7 +62,7 @@ async def test_deploy_latest_from_channel(ops_test: OpsTest) -> None:
         application_name=APP_NAME,
         num_units=3,
         channel=OPENSEARCH_CHANNEL,
-        revision=FIRST_REVISION,
+        revision=VERSION_TO_REVISION["2.12.0"],
```
Mehdi-Bendriss, line 65:
```suggestion
        revision=VERSION_TO_REVISION[FIRST_VERSION],
```

tests/integration/test_small_deployment_upgrades_edge_channel.py
```diff
@@ -76,74 +75,14 @@ async def test_deploy_latest_from_channel(ops_test: OpsTest) -> None:
     await ops_test.model.wait_for_idle(
         apps=[TLS_CERTIFICATES_APP_NAME, APP_NAME],
         status="active",
-        timeout=1400,
+        timeout=3600,
```
Mehdi-Bendriss, line 78: Why are we increasing this timeout by this much? I'm afraid we hide issues that are visible to the user if we do this.

tests/integration/test_small_deployment_upgrades_edge_channel.py
```diff
@@ -154,6 +93,10 @@ async def test_upgrade_between_versions(
     leader_id = [u.id for u in units if u.is_leader][0]
 
     for version, rev in VERSION_TO_REVISION.items():
+        if version == "2.12.0":
```
Mehdi-Bendriss, line 96:
```suggestion
        if version == FIRST_VERSION:
```

tests/integration/test_small_deployment_upgrades_edge_channel.py
```diff
@@ -181,6 +124,7 @@ async def test_upgrade_between_versions(
                 wait_for_exact_units={
                     APP_NAME: 3,
                 },
+                timeout=3600,
```
Mehdi-Bendriss, line 127: same remark about very high timeouts

tests/integration/test_small_deployment_upgrades_edge_channel.py
```diff
@@ -268,8 +214,154 @@ async def test_upgrade_to_local(
...
+            version,
+            id=f"{ROLLBACK_PREFIX}{version}",
+            marks=pytest.mark.group(f"{ROLLBACK_PREFIX}{version}"),
+        )
+    )
+    for version in VERSION_TO_REVISION.keys()
+]
+
+
+@pytest.mark.parametrize("version", ROLLBACK_TEST_TYPE)
```
Mehdi-Bendriss, line 248: Why do we have separate groups? Why not all the upgrade / rollback flow in a single group?

Mehdi-Bendriss (commented):
Thanks Pedro - I left a few questions

tests/integration/test_small_deployment_upgrades_edge_channel.py
phvalguima: @Mehdi-Bendriss but we are not building the local charm, rather downloading from charmhub and deploying an "old" version of the cluster, then on further tests we upgrade to the locally built version.

tests/integration/test_small_deployment_upgrades_edge_channel.py
phvalguima: I've noticed the upgrade is taking longer than 20 mins, including in my own workstation. It eventually works, it just takes longer.

tests/integration/test_small_deployment_upgrades_edge_channel.py
phvalguima: PS: there is no point at testing rollback of, e.g. 2.12.0 > 2.13.0 > 2.12.0, as this has been already tested when 2.13.0 was being released (i.e. it was the locally built charm)

phvalguima:
Shard allocation is failing during upgrade using GH runners. The explanation given by `_cluster/explain` is:
```
        {
          "decider": "disk_threshold",
          "decision": "NO",
          "explanation": "the node is above the low watermark cluster setting [cluster.routing.allocation.disk.watermark.low=85%], using more disk space than the maximum allowed [85.0%], actual free: [14.291205595682438%]"
        },
```

Full output: https://pastebin.canonical.com/p/G3XGCFyzk5/

tests/integration/test_small_deployment_upgrades_edge_channel.py
phvalguima: Yes, kept it AND use `FIRST_VERSION` as part of the main version-2-revision dict.

reneradoi (approved):

tests/integration/ha/helpers.py
```diff
@@ -593,3 +595,65 @@ async def assert_restore_indices_and_compare_consistency(
     # We expect that new_count has a loss of documents and the numbers are different.
     # Check if we have data but not all of it.
     assert 0 < new_count < original_count
+
+
+async def assert_upgrade_to_local(
```
Mehdi-Bendriss, line 600: In a next PR - Please take this out of the HA helpers / modules. We should have a separate package for upgrades.

tests/integration/ha/test_small_deployment_upgrades.py
```diff
@@ -81,69 +83,43 @@ async def test_deploy_latest_from_channel(ops_test: OpsTest) -> None:
...
-@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "xlarge"])
-@pytest.mark.group(1)
-@pytest.mark.abort_on_fail
-async def test_upgrade_rollback(
-    ops_test: OpsTest, c_writes: ContinuousWrites, c_writes_runner
-) -> None:
-    """Test upgrade from upstream to currently locally built version."""
+
+@pytest.fixture()
+async def c_writes(ops_test: OpsTest):
```
Mehdi-Bendriss, line 90: This is already part of the `conftest.py` - there is no need to define it again.

tests/integration/ha/test_small_deployment_upgrades.py
```diff
@@ -81,69 +83,43 @@ async def test_deploy_latest_from_channel(ops_test: OpsTest) -> None:
...
-        ops_test,
-        leader_id,
-        "pre-upgrade-check",
-        app=app,
-    )
-    assert action.status == "completed"
 
-    new_rev = list(VERSION_TO_REVISION.values())[-1]
+@pytest.fixture()
+async def c_writes_runner(ops_test: OpsTest, c_writes: ContinuousWrites):
```
Mehdi-Bendriss, line 97: This is already part of the `conftest.py` - there is no need to define it again.

tests/integration/ha/test_small_deployment_upgrades.py
```diff
@@ -8,62 +8,64 @@
...
 VERSION_TO_REVISION = {
+    FIRST_VERSION: 90,
     "2.13.0": 91,
 }
-FIRST_VERSION = "2.12.0"
-FIRST_REVISION = 90
 
 
-charm = None
+FROM_VERSION_PREFIX = "from_v"
```
Mehdi-Bendriss, line 34: We should clarify from where to where we are upgrading - it is not immediately clear on the CI when going through the runs. Perhaps; 
```suggestion
GROUP_NAME = "from_v:{}-to_v:local"
```
then use the `.format(version)`  later

tests/integration/ha/test_small_deployment_upgrades.py
```diff
@@ -154,6 +130,10 @@ async def test_upgrade_between_versions(
     leader_id = [u.id for u in units if u.is_leader][0]
 
     for version, rev in VERSION_TO_REVISION.items():
+        if version == FIRST_VERSION:
```
Mehdi-Bendriss, line 133: Can we please rename `FIRST_VERSION` to something along the lines of: `CURRENT_VERSION_MINUS_2` or similar?

tests/integration/helpers.py
```diff
@@ -112,6 +112,26 @@ async def run_action(
...
+) -> None:
+    """Set watermark on the application."""
+    unit_ip = await get_leader_unit_ip(ops_test, app=app)
+    await http_request(
+        ops_test,
+        "PUT",
+        f"https://{unit_ip}:9200/_cluster/settings",
+        {
+            "persistent": {
+                "cluster.routing.allocation.disk.threshold_enabled": "false",
```
Mehdi-Bendriss, line 128: Can we do something less radical? and just slightly increase it to say 90%

Mehdi-Bendriss (changes requested):
Thanks Pedro - I have a few comments.

tests/integration/test_small_deployment_upgrades_edge_channel.py
Mehdi-Bendriss: One thing to keep in mind, is that those integration tests will run for future versions too. So it's less about `2.12.0` or `2.13.0` but more about `N-1, ..., N-m`

phvalguima: Yes, that is correct @Mehdi-Bendriss: as we add new versions, we will put them in the "version_to_revision" dict we have here.

My point was that, in a PR, we are only interested on how `upgrade from N-m to locally built` or `rollback from N-m to locally built` really play off.

There is an exception of what we call "the happy path", which tries to simulate the lifecycle of a deployment, starting on version `2.12.0` and doing all the right moves all the way to "locally built" revision.

tests/integration/ha/test_small_deployment_upgrades.py
phvalguima: @Mehdi-Bendriss can we rename to `STARTING_VERSION` ? I don't want to pin we will only test down to `N-2` here, but rather on the `VERSION_TO_REVISION` dict.

tests/integration/ha/test_small_deployment_upgrades.py
phvalguima: @Mehdi-Bendriss I cannot use "-" but I will adopt this change.

tests/integration/ha/helpers.py
phvalguima: Btw, did it now

Mehdi-Bendriss (approved):

reneradoi (approved):

Issue #320, opened 2024-06-05
"Initializing the security index" times out when restarting a cluster with existing storage


reneradoi:
## Steps to reproduce
1. form a cluster of 3 nodes
2. remove the application, but keep the storage
3. deploy a new application, attaching the existing storage

## Expected behavior
The cluster should start.

## Actual behavior
When the first new unit (`opensearch/3` in this case) tries to start, it occasionally runs into a timeout when initializing the security index. This can be seen in `juju status`:
```
$ juju status --storage
Model              Controller      Cloud/Region         Version  SLA          Timestamp
test-storage-bg2j  lxd-controller  localhost/localhost  3.4.2    unsupported  11:24:15Z

App                       Version  Status       Scale  Charm                     Channel        Rev  Exposed  Message
opensearch                         maintenance      3  opensearch                                 1  no       Initializing the security index...
self-signed-certificates           active           1  self-signed-certificates  latest/stable   72  no       

Unit                         Workload     Agent      Machine  Public address  Ports  Message
opensearch/3*                maintenance  executing  4        10.27.170.111          Initializing the security index...
opensearch/4                 waiting      idle       5        10.27.170.202          Requesting lock on operation: start
opensearch/5                 waiting      executing  6        10.27.170.47           Requesting lock on operation: start
self-signed-certificates/0*  active       idle       0        10.27.170.145          
```
Running the `securityadmin.sh` manually also runs into timeout:
```
# /var/snap/opensearch/current/usr/share/opensearch/plugins/opensearch-security/tools/securityadmin.sh -cd /var/snap/opensearch/current/etc/opensearch/opensearch-security/ -cn opensearch-tsky -h 10.27.170.111 -cacert /var/snap/opensearch/current/etc/opensearch/certificates/root-ca.cert -cert /var/snap/opensearch/current/etc/opensearch/certificates/app-admin.cert -key /var/snap/opensearch/current/etc/opensearch/certificates/app-admin.key
**************************************************************************
** This tool will be deprecated in the next major release of OpenSearch **
** https://github.com/opensearch-project/security/issues/1755           **
**************************************************************************
Security Admin v7
Will connect to 10.27.170.111:9200 ... done
Connected as "O=opensearch-w3ty,CN=admin"
OpenSearch Version: 2.13.0
Contacting opensearch cluster 'opensearch-tsky' and wait for YELLOW clusterstate ...
[...]

PR #321, opened 2024-06-06
[DPE-4557] fix timeout when initializing the security index

reneradoi:
## Issue
https://github.com/canonical/opensearch-operator/issues/320

## Solution
When the last unit in a cluster is stopping, it adds the `voting_config_exclusion`, but doesn't delete it anymore (because all units are already stopped). It is then persisting on disk. If the storage is reused and the first new unit starts, it may happen that the new unit can't become cluster manager because it's not possible to reach quorum.

Therefore the last unit stopping should not add a voting exclusion.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -1042,15 +1042,18 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
         self.status.set(WaitingStatus(ServiceIsStopping))
 
         if self.opensearch.is_node_up():
-            nodes = self._get_nodes(True)
-            # do not add exclusions if it's the last unit to stop
-            # otherwise cluster manager election will be blocked when starting up again
-            # and re-using storage
-            if len(nodes) > 1:
-                # TODO: we should probably NOT have any exclusion on restart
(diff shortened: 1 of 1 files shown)


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1042,10 +1042,15 @@ def _stop_opensearch(self, *, restart=False) -> None:
         self.status.set(WaitingStatus(ServiceIsStopping))
 
         if self.opensearch.is_node_up():
-            # TODO: we should probably NOT have any exclusion on restart
-            # https://chat.canonical.com/canonical/pl/bgndmrfxr7fbpgmwpdk3hin93c
-            # 1. Add current node to the voting + alloc exclusions
-            self.opensearch_exclusions.add_current()
+            nodes = self._get_nodes(True)
```
Mehdi-Bendriss, line 1045: we should catch the `OpenSearchHttpError` here

Mehdi-Bendriss (commented):
Thanks René - left a comment

lib/charms/opensearch/v0/opensearch_base_charm.py

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #328, opened 2024-06-13
[DPE-4553][DPE-4208][DPE-3970] Extend large deployments to test_plugins.py &&  Fix dashboards

phvalguima:
Renames our dashboard to "Charmed OpenSearch". This PR addresses the following needs in our COS integration for large deployments:
1) **Fixes data node integration with COS**: COSAgentProvider was not tracking peer-cluster-* events, which means it was not receiving changes done to the COSUser by the main orchestrator
2) **Add integration tests to COS**: large deployments integration tests for COS were missing, this PR makes `test_plugins` more flexible, so it can run both small and large deployments
3) **Manage user passwords**: large deployments will now answer to `set-password` action and `MAIN_ORCHESTRATOR` propagates the new password value down to the other apps
4) **Adds Role to Dashboard**: adds "Roles" as a dropdown option to Grafana dashboard, allowing to filter each cluster also by OpenSearch "construct" node role.

## Empty Dashboard Fix

Some dashboards are being set with `$interval`, which defaults to `1m` and does not allow to run the actual calculation. This PR fixes it by using [__rate_interval](https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/) instead, which is a fixed period of X sample times for any `rate()` calculations.

### Before
Screenshot from 2024-06-13 20-42-30

### After
Screenshot from 2024-06-13 20-42-42

This is a graph for `rate()`, which means we want to compare sample deltas over a window of time. The later graph makes more sense in this case.

### DIFF
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -114,4 +114,12 @@ OPENSEARCH_SNAP_REVISION = 51  # Keep in sync with `workload_version` file
 OPENSEARCH_BACKUP_ID_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 
 # Roles that we can add as tags to prometheus data
-COS_TAGGABLE_ROLES = ["data", "cluster_manager", "voting", "cluster_manager_elegible", "ml"]
+COS_TAGGABLE_ROLES = [
+    "data",
+    "cluster_manager",
+    "voting",
+    "coordinating_only",
+    "ingest",
+    "cluster_manager_elegible",
+    "ml",
+]
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -112,3 +112,6 @@ OPENSEARCH_SNAP_REVISION = 51  # Keep in sync with `workload_version` file
 
 # User-face Backup ID format
 OPENSEARCH_BACKUP_ID_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+
+# Roles that we can add as tags to prometheus data
+COS_TAGGABLE_ROLES = ["data", "cluster_manager", "voting", "cluster_manager_elegible", "ml"]
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -1548,13 +1548,11 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             Scope.UNIT, "certs_exp_checked_at", datetime.now().strftime(date_format)
         )
 
-    def _get_prometheus_labels(self) -> Dict[str, str]:
+    def _get_prometheus_labels(self) -> Optional[Dict[str, str]]:
         """Return the labels for the prometheus scrape."""
         try:
             if not self.opensearch.roles:
-                return {
-                    "roles": "unrecognized",
-                }
+                return None
             roles = set(
                 role if role in COS_TAGGABLE_ROLES else "unrecognized"
                 for role in self.opensearch.roles
@@ -1565,9 +1563,7 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             # At very early stages of the deployment, "node.roles" may not be yet present
             # in the opensearch.yml, nor APIs is responding. Therefore, we need to catch
             # the KeyError here and report the appropriate response.
-            return {
-                "roles": "unrecognized",
-            }
+            return None
 
     def _scrape_config(self) -> List[Dict]:
         """Generates the scrape config as needed."""
@@ -1575,6 +1571,7 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             not (app_secrets := self.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val))
             or not (ca := app_secrets.get("ca-cert"))
             or not (pwd := self.secrets.get(Scope.APP, self.secrets.password_key(COSUser)))
+            or not self._get_prometheus_labels()
         ):
             # Not yet ready, waiting for certain values to be set
             return []
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -12,6 +12,7 @@ from typing import Any, Dict, List, Optional, Type
 
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from charms.opensearch.v0.constants_charm import (
+    COS_TAGGABLE_ROLES,
     AdminUser,
     AdminUserInitProgress,
     AdminUserNotConfigured,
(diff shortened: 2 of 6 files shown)


src/grafana_dashboards/opensearch.json
```diff
@@ -2630,30 +2630,78 @@
...
+        "type": "custom"
       },
       {
         "datasource": {
           "type": "datasource",
           "uid": "${prometheusds}"
         },
         "hide": 0,
-        "includeAll": true,
-        "label": "Node",
```
phvalguima, line 2654: "Node" is not really useful, as we already can filter by juju-related tags like `juju_unit`

src/grafana_dashboards/opensearch.json
```diff
@@ -2721,6 +2769,6 @@
     ]
   },
   "timezone": "browser",
-  "title": "OpenSearch",
+  "title": "Charmed OpenSearch",
```
phvalguima, line 2772: Oh yeah!

src/grafana_dashboards/opensearch.json
```diff
@@ -295,7 +295,7 @@
               "value": "null"
             }
           ],
-          "valueName": "avg"
```
phvalguima, line 298: "avg" means the moving average window across a sample. Issue here is, if one node goes down and up, we will see for a brief time something like "3.89" nodes, instead of 4->3->4.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -246,6 +239,20 @@ def __init__(self, *args, distro: Type[OpenSearchDistribution] = None):
         self.framework.observe(self.on.set_password_action, self._on_set_password_action)
         self.framework.observe(self.on.get_password_action, self._on_get_password_action)
 
+        self.cos_integration = COSAgentProvider(
+            self,
+            relation_name=COSRelationName,
+            metrics_endpoints=[],
+            scrape_configs=self._scrape_config,
+            refresh_events=[
```
MiaAltieri, line 247: Noob question. you reference this change in your PR message:
> Fixes data node integration with COS: COSAgentProvider was not tracking peer-cluster-* events, which means it was not receiving changes done to the COSUser by the main orchestrator

What do you mean by " it was not receiving changes done to the COSUser by the main orchestrator" - what does this do to the dashboard data. Is it needed across all DPE charms?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -689,8 +696,18 @@ def _on_set_password_action(self, event: ActionEvent):
             self._put_or_update_internal_user_leader(user_name, password)
             label = self.secrets.password_key(user_name)
             event.set_results({label: password})
+            # We know we are already running for MAIN_ORCH. and its leader unit
+            self.peer_cluster_provider.refresh_relation_data(event)
```
MiaAltieri, line 700: nice 😎

MiaAltieri (approved):
I'm not sure if you already have [these two tests](https://github.com/canonical/mongodb-operator/blob/5b277c393e68b884568e1a2055eea2c93a29d120/tests/integration/metrics_tests/test_metrics.py#L56-L86) but they might be useful to verify that your dashboard is getting properly updated during password changed events or during self healing during a network cut

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: @MiaAltieri in large deployments, only the `main_orchestrator` changes the password of the system users. That means, user `monitor`, which we use to authenticate COS, will be defined by one of the juju apps involved. Now, `main_orchestrator` then publishes that change to other apps. That is why we listen to `peer-cluster-*`.

reneradoi (approved):
Thanks for showing me the dashboards!

lib/charms/opensearch/v0/constants_charm.py
```diff
@@ -112,3 +112,14 @@
...
 OPENSEARCH_BACKUP_ID_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+
+# Roles that we can add as tags to prometheus data
+COS_TAGGABLE_ROLES = [
+    "data",
+    "cluster_manager",
+    "voting",
+    "coordinating_only",
+    "ingest",
+    "cluster_manager_elegible",
```
phvalguima, line 123: Ouch! I have an incoming next PR to reshuffle grafana dashboards order. I will also remove this unneeded entry there.

Mehdi-Bendriss: I think you can directly use `ClusterTopology.generated_roles()`

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -689,8 +696,18 @@ def _on_set_password_action(self, event: ActionEvent):
...
         except OpenSearchError as e:
             event.fail(f"Failed changing the password: {e}")
+        except RuntimeError as e:
```
Mehdi-Bendriss, line 703: I understand why you're doing it, but it's not very pretty... we can keep it until we find a better way - or maybe we get rid of defer in the refresh_relation_data at some point

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1531,19 +1548,42 @@ def _check_certs_expiration(self, event: UpdateStatusEvent) -> None:
             Scope.UNIT, "certs_exp_checked_at", datetime.now().strftime(date_format)
         )
 
+    def _get_prometheus_labels(self) -> Optional[Dict[str, str]]:
+        """Return the labels for the prometheus scrape."""
+        try:
+            if not self.opensearch.roles:
+                return None
+            roles = set(
+                role if role in COS_TAGGABLE_ROLES else "unrecognized"
```
Mehdi-Bendriss, line 1557: I would perhaps rename to `other`

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1531,19 +1548,42 @@ def _check_certs_expiration(self, event: UpdateStatusEvent) -> None:
...
+                for role in self.opensearch.roles
+            )
+            roles = sorted(roles)
+            return {"roles": ",".join(roles)}
+        except KeyError:
+            # At very early stages of the deployment, "node.roles" may not be yet present
```
Mehdi-Bendriss, line 1563: nice

Mehdi-Bendriss (commented):
Thanks Pedro! Looks good, just a minor comment about roles

lib/charms/opensearch/v0/constants_charm.py
phvalguima: @Mehdi-Bendriss added via this PR: https://github.com/canonical/opensearch-operator/pull/332

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Agreed, I have an `if` down to check if we are dealing with the defer-related Runtime error or not, but still, it depends on the message internals. It is not nice but it is the only way for now

Mehdi-Bendriss (approved):

PR #333, opened 2024-06-19
[DPE-4521] Reload TLS certs without restarting

reneradoi:
## Issue
Currently the OpenSearch application is restarted when TLS certificates are provided (with the `on.certificate_available` handler). This could be simplified by using the "hot reload" API OpenSearch provides for renewing the TLS certs.

## Solution
- activate the `plugins.security.ssl_cert_reload_enabled` setting
- when TLS conf is set, query the API instead of restarting OpenSearch


reneradoi:
The reload-API currently runs into an exception when comparing the list of subject alternative names on the current and new certs, because ordering is different. This may be an upstream bug and is currently discussed in the community forum of OpenSearch ([here](https://forum.opensearch.org/t/exception-when-reloading-self-signed-certs-via-rest-api/19944/2)).

reneradoi:
Will be continued in https://github.com/canonical/opensearch-operator/pull/380.

PR #335, opened 2024-06-24
[DPE-4654] Use keystores and truststore for TLS

reneradoi:
## Issue
In order to enable CA certificate rotation without downtime, OpenSearch needs to use pkcs12 trustores / keystores instead of plain certificate and key files.

## Solution
Description will be added.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -22,6 +22,7 @@ from typing import Dict, List, Optional, Tuple, Union
 
 from charms.opensearch.v0.constants_tls import TLS_RELATION, CertType
 from charms.opensearch.v0.helper_networking import get_host_public_ip
+from charms.opensearch.v0.helper_security import generate_password
 from charms.opensearch.v0.models import DeploymentType
 from charms.opensearch.v0.opensearch_exceptions import OpenSearchError
 from charms.opensearch.v0.opensearch_internal_data import Scope
@@ -131,12 +132,17 @@ class OpenSearchTLS(Object):
             event.defer()
             return
         admin_cert = self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
-        if (
-            self.charm.unit.is_leader()
-            and admin_cert is None
-            and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
-        ):
-            self._request_certificate(Scope.APP, CertType.APP_ADMIN)
+        if self.charm.unit.is_leader:
+            # create passwords for both ca trust_store/admin key_store
+            self._create_keystore_pwd_if_not_exists(Scope.APP, "ca")
+            self._create_keystore_pwd_if_not_exists(Scope.APP, CertType.APP_ADMIN.val)
+
+            if admin_cert is None and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR:
+                self._request_certificate(Scope.APP, CertType.APP_ADMIN)
+
+        # create passwords for both unit-http/transport key_stores
+        self._create_keystore_pwd_if_not_exists(Scope.UNIT, CertType.UNIT_TRANSPORT.val)
+        self._create_keystore_pwd_if_not_exists(Scope.UNIT, CertType.UNIT_HTTP.val)
 
         self._request_certificate(Scope.UNIT, CertType.UNIT_TRANSPORT)
         self._request_certificate(Scope.UNIT, CertType.UNIT_HTTP)
@@ -382,3 +388,9 @@ class OpenSearchTLS(Object):
                 certs[CertType.APP_ADMIN] = admin_secrets["cert"]
 
         return certs
+
+    def _create_keystore_pwd_if_not_exists(self, scope: Scope, alias: str):
+        """Create passwords for the key stores if not already created."""
+        keystore_pwd = self.charm.secrets.get(scope, f"keystore-password-{alias}")
+        if not keystore_pwd:
+            self.charm.secrets.put(scope, f"keystore-password-{alias}", generate_password())


lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +123,38 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
     finally:
         os.unlink(tmp_key.name)
         os.unlink(tmp_pkcs8_key.name)
+
+
+def run_cmd(command: str, args: str = None) -> SimpleNamespace:
```
Mehdi-Bendriss, line 128: i think this can be moved to another module, as it's unrelated to `*_security`. Feel free to create a new module for it if needed.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -407,8 +408,7 @@ def _on_peer_relation_created(self, event: RelationCreatedEvent):
             event.defer()
             return
 
-        # Store the "Admin" certificate, key and CA on the disk of the new unit
```
Mehdi-Bendriss, line 410: could we keep this comment?

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -181,6 +195,22 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
             merge=True,
         )
 
+        # currently only make sure there is a CA
+        # workflow for replacement will be added later
```
Mehdi-Bendriss, line 199: nit: 
```suggestion
        # TODO: workflow for replacement will be added later
```

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -382,3 +412,156 @@ def get_unit_certificates(self) -> Dict[CertType, str]:
...
+    def all_tls_resources_stored(self, only_unit_resources: bool = False) -> bool:
+        """Check if all TLS resources are stored on disk."""
+        cert_types = [CertType.UNIT_TRANSPORT, CertType.UNIT_HTTP]
+        if not only_unit_resources:
+            cert_types.append(CertType.APP_ADMIN)
+
+        for cert_type in cert_types:
+            if not exists(f"{self.certs_path}/{cert_type}.p12"):
+                # ensure backward compatibility to older charm versions still using pem files
+                for extension in ["key", "cert"]:
```
Mehdi-Bendriss, line 533: Can you explain this?

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -382,3 +412,156 @@ def get_unit_certificates(self) -> Dict[CertType, str]:
...
+    def store_new_tls_resources(self, cert_type: CertType, secrets: Dict[str, Any]):
+        """Add key and cert to keystore."""
+        cert_name = cert_type.val
+        if cert_type == CertType.APP_ADMIN:
+            store_pwd = self.charm.secrets.get(Scope.APP, f"keystore-password-{cert_name}")
+        else:
+            store_pwd = self.charm.secrets.get(Scope.UNIT, f"keystore-password-{cert_name}")
+        store_path = f"{self.certs_path}/{cert_type}.p12"
+
+        if not secrets.get("key"):
```
Mehdi-Bendriss, line 481: when would this happen?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -748,12 +748,13 @@ def on_tls_conf_set(
...
-        # Store cert/key on disk - must happen after opensearch stop for transport certs renewal
-        self.store_tls_resources(cert_type, current_secrets)
-
         if scope == Scope.UNIT:
             # node http or transport cert
-            self.opensearch_config.set_node_tls_conf(cert_type, current_secrets)
+            self.opensearch_config.set_node_tls_conf(
+                cert_type,
+                self.secrets.get(Scope.APP, "keystore-password-ca"),
+                self.secrets.get(scope, f"keystore-password-{cert_type}"),
```
Mehdi-Bendriss, line 756: we should group all of these under the same secrets where we store the TLS resources (chain, key etc..), differentiating between the ones that are with unit scope vs the ones with the app scope

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -772,24 +773,9 @@ def on_tls_relation_broken(self, _: RelationBrokenEvent):
 
     def is_tls_fully_configured(self) -> bool:
```
Mehdi-Bendriss, line 774: I believe we can get rid of this method - and directly set / remove the `tls_configured` relation data flag from the TLS module

Mehdi-Bendriss (commented):
Thanks René! Looks good, I have a few small comments

lib/charms/opensearch/v0/opensearch_tls.py
reneradoi: @Mehdi-Bendriss This addresses an issue I found with the upgrade integration tests (see here: https://github.com/canonical/opensearch-operator/actions/runs/9704294103/job/26784432918). When the integration test deploys and older version of the charm, there are key- and cert-files used instead of keystores and truststores. Now when the test uses `assert_continuous_writes_...`, this gets the admin password via the `get-password` action, which executes this check and failes if there are only p12-files. That's why I added this.

lib/charms/opensearch/v0/helper_security.py
reneradoi: has been moved to helper_charm.py

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: Good you bring that up, as I wanted to add a question here: The idea behind having it as a separate secret was that maybe we want to treat keystore-passwords similar to user-passwords, meaning that it could make sense to - at some point - add an action to set keystore-passwords. Then it would be easier with this structure.

Do you think that makes sense or should I treat it like the other TLS resources and move it to the corresponding TLS secret?

lib/charms/opensearch/v0/opensearch_tls.py
Mehdi-Bendriss: We can drop compatibility for pem files starting from this PR, we can luckily afford to do that now that we're still on `2/edge`

reneradoi: done

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: The full background is that all charms only have 1 secret for all passwords - but since we have TLS mandatory in opensearch we have the exception of having 2 secrets, 1 for passwords and 1 for TLS resources.

Since those passwords are exclusively related to TLS - I think it makes sense to treat them the same way we treat TLS resources secrets-wise.

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -382,3 +412,156 @@ def get_unit_certificates(self) -> Dict[CertType, str]:
...
+            return
+
+        alias = "ca"
+        store_path = f"{self.certs_path}/{alias}.p12"
+
+        with tempfile.NamedTemporaryFile(mode="w+t") as ca_tmp_file:
+            ca_tmp_file.write(secrets.get("ca-cert"))
+            ca_tmp_file.flush()
+
+            run_cmd(
```
phvalguima, line 455: @reneradoi we have a module named `opensearch_keystore`, which has a `Keystore` class that already manages the keytool and runs these types of commands. OpenSearchTLS should also inherit from `Keystore` class. This way, we have one single point of touch for `keytool` commands.

We can do on a following PR thou, can you open a ticket for this?

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -382,3 +419,164 @@ def get_unit_certificates(self) -> Dict[CertType, str]:
...
+
+    def _read_stored_ca(self, alias: str = "ca") -> Optional[str]:
+        """Load stored CA cert."""
+        secrets = self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
+
+        ca_trust_store = f"{self.certs_path}/ca.p12"
+        if not (exists(ca_trust_store) and secrets):
+            return None
+
+        stored_certs = run_cmd(
```
phvalguima, line 476: this is OpenSearchTLS specific, for example

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -382,3 +419,164 @@ def get_unit_certificates(self) -> Dict[CertType, str]:
...
+            self.charm.secrets.put_object(
+                scope,
+                cert_type.val,
+                {f"keystore-password-{alias}": generate_password()},
+                merge=True,
+            )
+
+    def store_new_ca(self, secrets: Dict[str, Any]):
+        """Add new CA cert to trust store."""
+        keytool = f"sudo {self.jdk_path}/bin/keytool"
```
phvalguima, line 440: We run as root in charm. No need for sudo.

phvalguima (approved):
LGTM. Left some small comments that we can postpone to another ticket.

Mehdi-Bendriss (approved):

Issue #327, opened 2024-06-11
Shards don't get assigned when the Primary get's removed and only two units are left


reneradoi:
## Steps to reproduce
- start a cluster with three nodes
- remove the application
- after the first unit is removed, occasionally (e.g. when the application leader gets removed first) some shards are in `UNASSIGNED` state and don't get assigned anymore
- the application does not scale down any further because the other units can't get the OpenSearch lock (this is currently worked around with [this patch](https://github.com/canonical/opensearch-operator/blob/DPE-2119-attach-existing-storage/lib/charms/opensearch/v0/opensearch_locking.py#L246-L247))

## Expected behavior
If there are only two units left, one should become the primary for the unassigned shards.

## Actual behavior
No units takes over the primary role for the unassigned shards, thus esp. the index `.charm_node_lock` may stay unavailable and no further locks for scaling down (or up) can be acquired.

## Log output
Juju debug log: 
```
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-opensearch-1/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 272, in request
    resp = call(urls[0])
  File "/var/lib/juju/agents/unit-opensearch-1/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 224, in call
    for attempt in Retrying(
  File "/var/lib/juju/agents/unit-opensearch-1/charm/venv/tenacity/__init__.py", line 347, in __iter__
    do = self.iter(retry_state=retry_state)
  File "/var/lib/juju/agents/unit-opensearch-1/charm/venv/tenacity/__init__.py", line 325, in iter
    raise retry_exc.reraise()
  File "/var/lib/juju/agents/unit-opensearch-1/charm/venv/tenacity/__init__.py", line 158, in reraise
    raise self.last_attempt.result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 451, in result
    return self.__get_result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/var/lib/juju/agents/unit-opensearch-1/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 251, in call
    response.raise_for_status()
  File "/var/lib/juju/agents/unit-opensearch-1/charm/venv/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
[...]

Issue #326, opened 2024-06-10
Initializing the security index times out when re-using storage with OpenSearch 2.14.0


reneradoi:
## Steps to reproduce
- create a cluster of 3 units
- remove the application
- deploy again with 1 unit and re-attach storage

## Expected behavior
Newly deployed application unit starts correctly.

## Actual behavior
Unit does not start up, instead hangs in `Initializing the security index...`

## Versions
OpenSearch 2.14.0 (snap revision 51)

## Log output
from the server log:
```
[2024-06-10T07:09:40,716][INFO ][o.o.a.c.ADDataMigrator   ] [opensearch-3] Start migrating AD data
[2024-06-10T07:09:40,716][INFO ][o.o.a.c.ADDataMigrator   ] [opensearch-3] AD job index doesn't exist, no need to migrate
[2024-06-10T07:09:40,716][INFO ][o.o.a.c.ADClusterEventListener] [opensearch-3] Init AD version hash ring successfully
[2024-06-10T07:09:40,731][INFO ][o.o.g.GatewayService     ] [opensearch-3] recovered [5] indices into cluster_state
[2024-06-10T07:09:40,734][WARN ][o.o.o.i.ObservabilityIndex] [opensearch-3] message: index [.opensearch-observability/lao7cXi9SzCfWdPsE-qpdA] already exists
[2024-06-10T07:09:40,735][INFO ][o.o.s.l.LogTypeService   ] [opensearch-3] Loading builtin types!
[2024-06-10T07:09:40,738][INFO ][o.o.s.l.LogTypeService   ] [opensearch-3] Indexing [429] fieldMappingDocs from logTypes: 24
[2024-06-10T07:09:40,771][WARN ][o.o.c.r.a.AllocationService] [opensearch-3] Falling back to single shard assignment since batch mode disable or multiple custom allocators set
[2024-06-10T07:09:40,777][WARN ][o.o.s.SecurityAnalyticsPlugin] [opensearch-3] Failed to initialize LogType config index and builtin log types
[2024-06-10T07:09:40,863][INFO ][o.o.p.PluginsService     ] [opensearch-3] PluginService:onIndexModule index:[.plugins-ml-config/uYso9eSvSRy3MvABKkgmCg]
[2024-06-10T07:09:41,310][WARN ][o.o.c.r.a.AllocationService] [opensearch-3] Falling back to single shard assignment since batch mode disable or multiple custom allocators set
[...]

Issue #325, opened 2024-06-10
Charm fails to bootstrap after host machine reboot 


Mehdi-Bendriss:
When a the LXD host machine reboots the charm fails to get the opensearch service to run, raising 503 http errors in the process.

Server logs:
```
[2024-06-10T09:06:16,978][INFO ][o.o.s.c.ConfigurationRepository] [main-0_711] Wait for cluster to be available ...
[2024-06-10T09:06:17,208][WARN ][o.o.c.c.ClusterFormationFailureHelper] [main-0_711] cluster-manager not discovered or elected yet, an election requires at least 3 nodes with ids from [poxiRShkTxeXfn_Quoh7-Q, 7YwoLqLUSCqYXqEqSzc1BA, rNb9cX69ThSrIY8sUB0G8A, xL8-9yGaSRWsI80bZAV2FA, tkNsDbIHRYWGnfIV2y-d2A], have discovered [{main-0_711}{7YwoLqLUSCqYXqEqSzc1BA}{vrNYjytWTjqkRdwnrmzfXw}{10.122.32.75}{10.122.32.75:9300}{coordinating_onlydimml}{shard_indexing_pressure_enabled=true, app_id=4cac7b7d-58b3-4c18-80c7-8f14aaeaeff7/main}] which is not a quorum; discovery will continue using [127.0.0.1:9300, 127.0.0.1:9301, 127.0.0.1:9302, 127.0.0.1:9303, 127.0.0.1:9304, 127.0.0.1:9305, [::1]:9300, [::1]:9301, [::1]:9302, [::1]:9303, [::1]:9304, [::1]:9305, 127.0.0.1:9300] from hosts providers and [{main-0_711}{7YwoLqLUSCqYXqEqSzc1BA}{vrNYjytWTjqkRdwnrmzfXw}{10.122.32.75}{10.122.32.75:9300}{coordinating_onlydimml}{shard_indexing_pressure_enabled=true, app_id=4cac7b7d-58b3-4c18-80c7-8f14aaeaeff7/main}] from last-known cluster state; node term 1, last-accepted version 86 in term 1
[2024-06-10T09:06:17,673][ERROR][o.o.s.a.BackendRegistry  ] [main-0_711] Not yet initialized (you may need to run securityadmin)
[2024-06-10T09:06:17,880][ERROR][o.o.s.a.BackendRegistry  ] [main-0_711] Not yet initialized (you may need to run securityadmin
```

[...]

PR #345, opened 2024-07-02
**OLD** [DPE-4575] Add voting settle logic at start and stop service

phvalguima:
Currently, we are having several issues with 2-node clusters as OpenSearch won't automatically manage voting anymore. This PR will add logic to manually manage the 2-node cluster scenario using the `voting_exclusions` API. Whenever we have only two nodes active and registered to the cluster as voting units, `_settle_voting_exclusions` will exclude the non-leader unit from voting. It also excludes the leaving unit, as we need to cover for the scenario where that unit is the cluster_manager and moving from 3->2 units we may end up with stale metadata.

This PR also makes exclusion mandatory to happen at start / stop. Therefore, we are sure the voting count will always be correct in each stage. 

For example, moving from 3->2->1 units results:
3->2) The cluster will set 2x voting exclusions: one for the unit leaving (if this is the cluster manager, that position will move away) and one for one of the 2x remaining units, following a sorted list
2->1) The voting exclusions are removed

Likewise, on scaling up:
1->2) One voting exclusion is added, following the same sorted list of all node names
2->3) Voting exclusions are removed

This PR touches #324, #326, #327 and in #325 this behavior is also observed. This is also linked to issues in our `ha/test_storage.py`, [has one can see in this run](https://github.com/canonical/opensearch-operator/actions/runs/9722933828/job/26840944295?pr=343).

Closes: #324, #326, #327


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1054,31 +1054,82 @@ def _stop_opensearch(self, *, restart=False) -> None:
...
+        cms = ClusterTopology.get_cluster_managers_names(nodes)
+
+        # Review the exclusion list
+        self.opensearch_exclusions.delete_voting(cms)
 
-        # 3. Remove the exclusions
-        # TODO: we should probably NOT have any exclusion on restart
-        # https://chat.canonical.com/canonical/pl/bgndmrfxr7fbpgmwpdk3hin93c
-        self.opensearch_exclusions.delete_current()
+        leader = ClusterTopology.elected_manager_name(
```
reneradoi, line 1108: Should we maybe not call it "leader" as this may be confusing with the juju leader?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1054,31 +1054,82 @@ def _stop_opensearch(self, *, restart=False) -> None:
...
+            if not self.opensearch.is_node_up():
+                # We are still running, there was an error
+                raise
 
-        # 2. stop the service
-        self.opensearch.stop()
-        self.peers_data.delete(Scope.UNIT, "started")
-        self.status.set(WaitingStatus(ServiceStopped))
+        # 2. Check if we have enough voting units
+        self._settle_voting_exclusions(exclude_this_unit=True)
```
reneradoi, line 1075: Is it a good idea to work with the API after the service has been stopped?

reneradoi (changes requested):
Hey @phvalguima I've tested this locally and ran into issues when removing the application with only two units left. It runs into 

```
unit-opensearch-3: 09:41:16 ERROR unit.opensearch/3.juju-log Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-opensearch-3/charm/./src/charm.py", line 267, in <module>
    main(OpenSearchOperatorCharm)
  File "/var/lib/juju/agents/unit-opensearch-3/charm/venv/ops/main.py", line 544, in main
    manager.run()
  File "/var/lib/juju/agents/unit-opensearch-3/charm/venv/ops/main.py", line 520, in run
    self._emit()
  File "/var/lib/juju/agents/unit-opensearch-3/charm/venv/ops/main.py", line 509, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name)
  File "/var/lib/juju/agents/unit-opensearch-3/charm/venv/ops/main.py", line 143, in _emit_charm_event
  File "/var/lib/juju/agents/unit-opensearch-3/charm/venv/ops/framework.py", line 350, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-opensearch-3/charm/venv/ops/framework.py", line 849, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-opensearch-3/charm/venv/ops/framework.py", line 939, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-opensearch-3/charm/lib/charms/opensearch/v0/opensearch_base_charm.py", line 523, in _on_opensearch_data_storage_detaching
    raise Exception("Unable to acquire lock: Another unit is starting or stopping.")
Exception: Unable to acquire lock: Another unit is starting or stopping.
```
and can't get out of it anymore. Furthermore I've seen some failures on the integration tests, so it looks like it does not really work yet.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: This should target every unit

phvalguima: Moved it to be executed before stopping the unit.

reneradoi (approved):
Looks good!

PR #347, opened 2024-07-04
[DPE-4769] Add hardware-observe support

phvalguima:
New snap version 53 adds support for hugepages read on sysfs. This charm bumps the snap version so we can start using it on the charm deployments.


Mehdi-Bendriss (approved):
Could you add a short description to the PR?

reneradoi (approved):

PR #361, opened 2024-07-10
[DPE-4860] helper_networking should check peer relation exists

### LINKED ISSUE #360: Single unit removal fails with ERROR when trying to get the lock at `remove-application`
Trying to remove a single unit renders:
```
  File "/var/lib/juju/agents/unit-opensearch-0/charm/./src/charm.py", line 213, in <module>
    main(OpenSearchOperatorCharm)
  File "/var/lib/juju/agents/unit-opensearch-0/charm/venv/ops/main.py", line 548, in main
    manager.run()
  File "/var/lib/juju/agents/unit-opensearch-0/charm/venv/ops/main.py", line 527, in run
    self._emit()
  File "/var/lib/juju/agents/unit-opensearch-0/charm/venv/ops/main.py", line 516, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name)
  File "/var/lib/juju/agents/unit-opensearch-0/charm/venv/ops/main.py", line 147, in _emit_charm_event
  File "/var/lib/juju/agents/unit-opensearch-0/charm/venv/ops/framework.py", line 348, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-opensearch-0/charm/venv/ops/framework.py", line 860, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-opensearch-0/charm/venv/ops/framework.py", line 950, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-opensearch-0/charm/lib/charms/opensearch/v0/opensearch_base_charm.py", line 481, in _on_opensearch_data_storage_detaching
    if not self.node_lock.acquired:
  File "/var/lib/juju/agents/unit-opensearch-0/charm/lib/charms/opensearch/v0/opensearch_locking.py", line 248, in acquired
    alt_hosts = self._charm.alt_hosts
  File "/var/lib/juju/agents/unit-opensearch-0/charm/lib/charms/opensearch/v0/opensearch_base_charm.py", line 1654, in alt_hosts
    all_units_ips = units_ips(self, PeerRelationName)
  File "/var/lib/juju/agents/unit-opensearch-0/charm/lib/charms/opensearch/v0/helper_networking.py", line 75, in units_ips
    for unit in charm.model.get_relation(peer_relation_name).units:
AttributeError: 'NoneType' object has no attribute 'units'
```

If we have a single unit present.

phvalguima:
Closes #360

### DIFF
--- a/lib/charms/opensearch/v0/helper_networking.py
+++ b/lib/charms/opensearch/v0/helper_networking.py
@@ -71,6 +71,8 @@ def unit_ip(charm: CharmBase, unit: Unit, peer_relation_name: str) -> str:
 def units_ips(charm: CharmBase, peer_relation_name: str) -> Dict[str, str]:
     """Returns the mapping "unit id / ip address" of all units."""
     unit_ip_map = {}
+    if not charm.model.get_relation(peer_relation_name):
+        return unit_ip_map
 
     for unit in charm.model.get_relation(peer_relation_name).units:
         unit_id = unit.name.split("/")[1]


reneradoi (dismissed):
Looks good!

tests/integration/test_charm.py
```diff
@@ -282,3 +282,10 @@ async def test_all_units_have_internal_users_synced(ops_test: OpsTest) -> None:
     for unit in ops_test.model.applications[APP_NAME].units:
         unit_conf = get_conf_as_dict(ops_test, unit.name, filename)
         assert leader_conf == unit_conf
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_remove_application(ops_test: OpsTest) -> None:
+    """Removes the application with a single unit."""
```
Mehdi-Bendriss, line 290: (this test deploys 2 units not just 1) - why do we expect the removal of the application to require testing? is there an issue in this regard?

Mehdi-Bendriss (commented):
Thanks Pedro, looks good, 1 question to clarify

tests/integration/test_charm.py
phvalguima: In short, yes, we had issues with the lock and 1x unit removal.

The 2-node removal is also something good to test, as it may fall on: https://github.com/canonical/opensearch-operator/issues/243

Mehdi-Bendriss: Okay we need to update the docstrings of the test then, as this test runs with 2 opensearch units

phvalguima: @Mehdi-Bendriss yes, also added a test to deploy and remove app for a single unit as well.

reneradoi (approved):

tests/integration/test_charm.py
```diff
@@ -35,6 +35,34 @@
 DEFAULT_NUM_UNITS = 2
 
 
+@pytest.mark.group("test-single-node-removal")
```
Mehdi-Bendriss, line 38: Let's please not create a new group for cases like this. We cannot spin this many workers for each PR run. 
If really needed and we have more than 1 case to test for app removal, please create a new test module: `test_application_removal.py` and move all application removal tests on it (by deploying multiple apps with the constraints you're interested in)

Mehdi-Bendriss (commented):
Thanks Pedro - I have a comment on tests

tests/integration/test_charm.py
phvalguima: okay, it will create and destroy that single-node setup  on the same run. Then the test can proceed normally

tests/integration/test_charm.py
```diff
@@ -282,3 +310,10 @@ async def test_all_units_have_internal_users_synced(ops_test: OpsTest) -> None:
     for unit in ops_test.model.applications[APP_NAME].units:
         unit_conf = get_conf_as_dict(ops_test, unit.name, filename)
         assert leader_conf == unit_conf
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_remove_application(ops_test: OpsTest) -> None:
+    """Removes the application with two units."""
+    ops_test.model.remove_application(APP_NAME, block_until_done=True)
```
reneradoi, line 319: Same comment as above.

reneradoi (dismissed):
Looks good now.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -486,9 +491,6 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
                     if node.name != self.unit_name
                 ]
                 self._compute_and_broadcast_updated_topology(remaining_nodes)
-            elif self.app.planned_units() == 0:
```
phvalguima, line 489: Removing this check as the unit is going away and it is the last one of its own peers. This logic will result in cleaning an empty peer data bag.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -472,7 +472,12 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
                 "Removing units during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
             )
         # acquire lock to ensure only 1 unit removed at a time
-        if not self.node_lock.acquired:
+        # Closes canonical/opensearch-operator#378
+        if (
+            self.app.planned_units() > 1
+            and (self.opensearch.is_node_up() or self.alt_hosts)
+            and not self.node_lock.acquired
+        ):
```
reneradoi, line 480: So that means we are planning to no longer use the locking mechanism if `planned_units()` is 0 or 1?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -512,8 +514,9 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
                 else:
                     raise OpenSearchHAError(ClusterHealthUnknown)
         finally:
-            # release lock
-            self.node_lock.release()
+            if self.app.planned_units() > 0 and (self.opensearch.is_node_up() or self.alt_hosts):
+                # release lock
+                self.node_lock.release()
```
reneradoi, line 519: I think this might not be consistent here, because in line 476-480 you're not getting a lock if `planned_units()` == 1, but here you want to release it.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Yeah, but we may have to add: `online_nodes > 1 and self.app.planned_units() > 1`. This way we have a single node in a cluster not expected to grow...

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Same comment as above. We may want to extend this logic to have a `online_nodes > 1 and self.app.planned_units() > 1`

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -472,7 +472,12 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
                 "Removing units during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
             )
         # acquire lock to ensure only 1 unit removed at a time
-        if not self.node_lock.acquired:
+        # Closes canonical/opensearch-operator#378
+        if (
+            self.app.planned_units() > 1
+            and (self.opensearch.is_node_up() or self.alt_hosts)
+            and not self.node_lock.acquired
```
Mehdi-Bendriss, line 479: What of the case where the nodes are down from a full cluster crash + scale-down for ex? why test on:
```
self.app.planned_units() > 1
and (self.opensearch.is_node_up() or self.alt_hosts)
```

Why not test on the number of units instead of the planned number?

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: I'm not sure that's accurate, the first hook to execute during the teardown flow is `storage_detaching` - we still have the peer relation data bag then. 
(It was proved to successfully work before, no?)

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -512,8 +514,9 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
                 else:
                     raise OpenSearchHAError(ClusterHealthUnknown)
         finally:
-            # release lock
-            self.node_lock.release()
+            if self.app.planned_units() > 0 and (self.opensearch.is_node_up() or self.alt_hosts):
```
Mehdi-Bendriss, line 517: Same comment as above.

lib/charms/opensearch/v0/opensearch_internal_data.py
```diff
@@ -91,7 +91,8 @@ def cast(str_val: str) -> Union[bool, int, float, str]:
     def put_or_delete(data: Dict[str, str], key: str, value: Optional[str]):
         """Put data into the key/val data store or delete if value is None."""
         if value is None:
-            del data[key]
+            if key in data:
+                del data[key]
```
Mehdi-Bendriss, line 95: nit:
```suggestion
            data.pop(key, None)
```

lib/charms/opensearch/v0/opensearch_internal_data.py
```diff
@@ -142,7 +142,9 @@ def has(self, scope: Scope, key: str):
         if scope is None:
             raise ValueError("Scope undefined.")
 
-        return key in self._get_relation_data(scope)
+        if not (data := self._get_relation_data(scope)):
+            return False
+        return key in data
```
Mehdi-Bendriss, line 147: nit: 
```suggestion
        return key in (self._get_relation_data(scope) or {})
```

Mehdi-Bendriss (commented):
Thanks Pedro, I have a couple of questions

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: @Mehdi-Bendriss we can test on the node count, for the local peers of this app. The main issue here was the `self.node_lock.acquired` relying on the peer relation when we reach `node_count == 1`; and that peer relation will not exist.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: @Mehdi-Bendriss you're right if you are going away because of a `remove-unit`. Then, at `storage-detaching`, you will have:
```
2024-08-09 13:34:54,371 DEBUG    ops 2.15.0 up and running.
2024-08-09 13:34:54,450 DEBUG    Emitting Juju event opensearch_data_storage_detaching.
(Pdb) p self.model.get_relation(PeerRelationName)
<ops.model.Relation opensearch-peers:1>
```

Whereas, if we do `remove-application`, then we get:
```
2024-08-09 13:50:33,992 DEBUG    ops 2.15.0 up and running.
2024-08-09 13:50:34,065 DEBUG    Emitting Juju event opensearch_data_storage_detaching.
> /var/lib/juju/agents/unit-opensearch-0/charm/lib/charms/opensearch/v0/opensearch_base_charm.py(479)_on_opensearch_data_storage_detaching()
-> if not self.node_lock.acquired:
(Pdb) p self.model.get_relation(PeerRelationName)
None
```

In this case, the lines below will give:
```
(Pdb) self.peers_data.delete(Scope.APP, "bootstrap_contributors_count")
*** KeyError: 'bootstrap_contributors_count'
```

Not sure what changed before and now. We did update `ops` to 2.15 and `juju` to 3.4.5. I am using juju 3.4.5, status: https://pastebin.ubuntu.com/p/YgVSbqWsGW/

phvalguima: I think we can keep that check and add an extra check if the peer relation still exists.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: @Mehdi-Bendriss check [this comment](https://github.com/canonical/opensearch-operator/pull/361/files/e3588a23397c17f357a1d5ec3254958bca366fa1#r1711519631). What we want to know here is the following:
1) Are we on a `remove-application` case? If yes, then `self.app.planned_units == 0`, likewise for the last unit going away
2) Are we the last node to go? If yes, then there is no relevance in grabbing the lock anyways -> I was checking it with `self.app.planned_units > 1 `, which means we have at least 2 nodes "planned" 
3) Are we going to use the node_lock or the peer_lock, I am trying to avoid the later, so I have: `(self.opensearch.is_node_up() or self.alt_hosts)`.

> What of the case where the nodes are down from a full cluster crash + scale-down for ex?

In this case, `self.app.planned_units >= 1` will pass, then `(self.opensearch.is_node_up() or self.alt_hosts)` will fail, as no node is currently reachable. The lock won't be tried to be picked, the exception below will be skipped and the unit will go away simply, as planned.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: @reneradoi check my comment answer [here](https://github.com/canonical/opensearch-operator/pull/361/files/e3588a23397c17f357a1d5ec3254958bca366fa1#r1711549293)

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Similarly to the comments above. I am just not 100% sure if we should have it as `> 0` or  `> 1`.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Check my previous answer to your comment at `acquired` call.

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: 1. why won't the peer relation exist when `node_count == 1`? on_storage_detaching runs before any other teardown event https://github.com/canonical/opensearch-operator/pull/361#discussion_r1711396636
2. `planned_units == 1` can be even when there are currently 5 units and we just initiated a scale-down to 1, thus inaccurate

phvalguima: @Mehdi-Bendriss I cannot answer the `why` question, as the peer relation object comes from ops and juju. One of these teams decided to not return that anymore post-`remove-application`. I can assure you that, in a `remove-application` and you're the last unit, then it will not exist.

```
2024-08-09 13:50:33,992 DEBUG    ops 2.15.0 up and running.
2024-08-09 13:50:34,065 DEBUG    Emitting Juju event opensearch_data_storage_detaching.
> /var/lib/juju/agents/unit-opensearch-0/charm/lib/charms/opensearch/v0/opensearch_base_charm.py(479)_on_opensearch_data_storage_detaching()
-> if not self.node_lock.acquired:
(Pdb) p self.model.get_relation(PeerRelationName)
None
```

Not sure what changed before and now. We did update ops to 2.15 and juju to 3.4.5. I am using juju 3.4.5, status: https://pastebin.ubuntu.com/p/YgVSbqWsGW/

phvalguima: > planned_units == 1 can be even when there are currently 5 units and we just initiated a scale-down to 1, thus inaccurate

Sure, but I am protecting myself against the `remove-application`, hence I need the `planned_units`, not the current unit count.

phvalguima:
@Mehdi-Bendriss answering the:

> I'm not sure that's accurate, the first hook to execute during the teardown flow is storage_detaching - we still have the peer relation data bag then.
(It was proved to successfully work before, no?)

Only backup tests really test `remove-application`, and that has been taken out because of #243 and our lock not allowing to remove app at once. Once that has been fixed, I rolled back that change and now backup tests hang on remove-application and fail. That is when I started to investigate this particular issue.

Now, according to the `ops` code, `self.model.get_relation` calls `_get_unique`. On git blame, this logic did not change for years now [and does return None](https://github.com/canonical/operator/blame/449793f9867eba11dc2e7e1c28ea3dca1e1da231/ops/model.py#L933) if no relation is found.

So, either we did not really test the `remove-application` before my work on backups; or juju changed its behavior.

phvalguima:
Cont:

Now, this may be caused by a recent change on juju 3.4.4: https://bugs.launchpad.net/juju/+bug/2024583

phvalguima:
Confirmed: it was caused 2 months ago because of that change above and released as part of 3.4.4. Filed a bug on Juju: https://bugs.launchpad.net/juju/+bug/2076599

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Dug into this issue and reported a bug to juju: https://bugs.launchpad.net/juju/+bug/2076599

reneradoi (dismissed):

Mehdi-Bendriss (approved):
Thank you!

reneradoi (approved):

PR #364, opened 2024-07-12
[DPE-4654] do not log command args

reneradoi:
To avoid writing sensitive information to the logs I propose to adjust the run_cmd function and mask known patterns for passwords and secrets.

### DIFF
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -175,7 +175,7 @@ def run_cmd(command: str, args: str = None) -> SimpleNamespace:
 
     # only log the command and no arguments to avoid logging sensitive information
     command = mask_sensitive_information(command_with_args)
-    logger.info(f"Executing command: {command}")
+    logger.debug(f"Executing command: {command}")
 
     try:
         output = subprocess.run(
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -173,6 +173,8 @@ def run_cmd(command: str, args: str = None) -> SimpleNamespace:
 
     command_with_args = " ".join(command_with_args.split())
 
+    # only log the command and no arguments to avoid logging sensitive information
+    command = command.split()[0]
     logger.debug(f"Executing command: {command}")
 
     try:
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -167,16 +167,17 @@ def run_cmd(command: str, args: str = None) -> SimpleNamespace:
         command: can contain arguments
         args: command line arguments
     """
+    command_with_args = command
     if args is not None:
-        command = f"{command} {args}"
+        command_with_args = f"{command} {args}"
 
-    command = " ".join(command.split())
+    command_with_args = " ".join(command_with_args.split())
 
     logger.debug(f"Executing command: {command}")
 
     try:
         output = subprocess.run(
-            command,
+            command_with_args,
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             shell=True,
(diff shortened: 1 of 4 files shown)


lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -167,16 +167,17 @@ def run_cmd(command: str, args: str = None) -> SimpleNamespace:
...
     """
+    command_with_args = command
     if args is not None:
-        command = f"{command} {args}"
+        command_with_args = f"{command} {args}"
 
-    command = " ".join(command.split())
+    command_with_args = " ".join(command_with_args.split())
 
     logger.debug(f"Executing command: {command}")
```
phvalguima, line 176: @reneradoi we may pass a full string with all arguments to these methods. I'd suggest we do right before:
```suggestion
    command = command.split()[0] if command else None
    logger.debug(f"Executing command: {command}")
```

phvalguima: This way, we avoid anyone passing: `command --opt=my_secret` as a string to `run_cmd`.

phvalguima (changes requested):

lib/charms/opensearch/v0/helper_charm.py
reneradoi: Yes that would be even more strict, while my initial suggestion would leave some room for still leaking sensitive information here. On the other hand, this will remove all other (maybe helpful) args too that were given with the first (the command) parameter, effectively resulting in a log line of just `keytool` (e.g.). Then we could even think about completely removing the logging output here.

Mehdi-Bendriss: I like René's approach - I find it to be a good compromise between not losing valuable log details, and leaking secrets. 
We should rather go through our code and see where we might be logging secrets and adapting the logging accordingly

Mehdi-Bendriss (dismissed):

lib/charms/opensearch/v0/helper_charm.py
phvalguima: That is not only a problem for the present, but towards the future where we may not even involved in this project anymore.

I don't like it. We should instead have an extra option, e.g.:
```
def run_cmd(command: str, args: str = None, log_all_args: bool = False)
```
And a disclaimer in the docstring on the risk of secret leaking if arguments containing them are logged.
So, if you want to troubleshoot something, then you can enable it conscious of what is going to happen.

Mehdi-Bendriss: In all applications, preventing secrets leaking is a conscious effort by the developer during programming time.
Our case is no different, the same effort / awareness needs happen, where the developer is responsible for being aware of what they log, and split the args in `run_cmd`.

reneradoi: @phvalguima I've adjusted the code to only log the command itself, no args.

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -167,16 +167,19 @@ def run_cmd(command: str, args: str = None) -> SimpleNamespace:
...
+    command_with_args = command
     if args is not None:
-        command = f"{command} {args}"
+        command_with_args = f"{command} {args}"
 
-    command = " ".join(command.split())
+    command_with_args = " ".join(command_with_args.split())
 
+    # only log the command and no arguments to avoid logging sensitive information
+    command = mask_sensitive_information(command_with_args)
```
phvalguima, line 177: We def. need unit tests here to show this method works as intended.

phvalguima (changes requested):
Hi @reneradoi can you add unit tests showing the regex actually works as expected? Please, include special chars such as: ?, !, %, &, etc

We are also missing coverage in the `opensearch_keystore` lib, which has a class `Keystore` using keytool the same way as `opensearch_tls`.

lib/charms/opensearch/v0/helper_charm.py
reneradoi: Yeah I already thought about unit testing this, esp. to make sure nothing is removed from that pattern without a test failing.

Test has been added (incl. special chars), the `opensearch_keystore` lib is covered, I've checked the `run_bin` and `run_cmd` calls there and added the password args to the pattern.

phvalguima (approved):
I am still really uncomfortable with the `run_cmd` arguments, as both options may carry arguments and that seems really confusing. In any case, I believe our secrets will be handled by the keystores and revolve around it anyways.

PR #367, opened 2024-07-15
[DPE-4575][DPE-4886][DPE-4983] Add voting exclusions management

phvalguima:
Currently, we are having several issues with 2-node clusters as OpenSearch won't automatically manage voting anymore. This PR will add logic to manually manage the 2-node cluster scenario using the voting_exclusions API. Whenever we have only two nodes active and registered to the cluster as voting units, _settle_voting_exclusions will exclude the non-leader unit from voting. It also excludes the leaving unit, as we need to cover for the scenario where that unit is the cluster_manager and moving from 3->2 units we may end up with stale metadata.

This PR also makes exclusion mandatory to happen at start / stop. Therefore, we are sure the voting count will always be correct in each stage.

For example, moving from 3->2->1 units results:
3->2) The cluster will set 2x voting exclusions: one for the unit leaving (if this is the cluster manager, that position will move away) and one for one of the 2x remaining units, following a sorted list
2->1) The voting exclusions are removed

Likewise, on scaling up:
1->2) One voting exclusion is added, following the same sorted list of all node names
2->3) Voting exclusions are removed

## To Consider
Watch our system under it would be:

* 2 nodes cluster
* Unexpected (crash for ex or SIGTERM) Loss of elected CM node.
* HA tests (currently run on 3+ nodes cluster, we should ensure the same behvior remains on the 2 nodes cluster)
* See behavior of the last remaining node (previously added to the voting exclusions list)
* We should ensure that a scaleup / restart removes the voting exclusion of already applied for the current node (1 or 2 nodes clusters).

### Acceptance Criteria

* handle voting exclusion in normal case 
* handle voting exclusion removal (so that on scale-up / down the node can still participate in the election)
* scaling-up/down/ha Integration tests should pass 
* test-storage should also pass (fixing stale metadata issue)
* fix of “shards don’t get assigned“ issue (side effect of the cluster not being able to vote for a CM) 

## Issues involved

This PR touches https://github.com/canonical/opensearch-operator/issues/324, https://github.com/canonical/opensearch-operator/issues/326, https://github.com/canonical/opensearch-operator/issues/327 and in https://github.com/canonical/opensearch-operator/issues/325 this behavior is also observed. This is also linked to issues in our ha/test_storage.py, [has one can see in this run](https://github.com/canonical/opensearch-operator/actions/runs/9722933828/job/26840944295?pr=343).

Closes: https://github.com/canonical/opensearch-operator/issues/324, https://github.com/canonical/opensearch-operator/issues/326, https://github.com/canonical/opensearch-operator/issues/327


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1060,6 +1078,80 @@ def _restart_opensearch(self, event: _RestartOpenSearch) -> None:
...
+                self.opensearch_exclusions.add_voting(
+                    hosts, node_names=[self.unit_name, sorted_cm[0]]
+                )
+            else:
+                # We are adding this unit to the cluster and we've waited until it is present
+                # We only exclude one unit:
+                self.opensearch_exclusions.add_voting(hosts, node_names=[sorted_cm[0]])
+            # Now, we clean up the sorted_cm list, as we want to be sure the new manager is elected
+            # and different than the excluded units.
+            sorted_cm.pop(0)
```
reneradoi, line 1138: I was wondering why you chose to exclude node [0] from voting and remove it from the list? This will result in a new `cluster manager node` being elected when you scale up from 1 unit to more and in the process of removing the application. I just saw this locally when testing, not that it creates that much latency, but wouldn't it be better to use the last one from the list instead?

reneradoi (commented):
Hey Pedro, the PR looks quite good. I've tested esp. the scenario with re-attaching storage locally multiple times and could not reproduce the issue anymore. That's good!

I just left some questions and comments for details, but in general I'd say this is fine.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: so, @reneradoi yes, I noticed that as well. But when I was discussing with @Mehdi-Bendriss, we've agreed to make this list predictable instead of keeping the cluster manager. I could add a check here, for that, but then the other check, right before, gets slightly more complicated:
```
            if unit_is_stopping:
                # Remove both this unit and the first sorted_cm from the voting
                self.opensearch_exclusions.add_voting(
                    hosts, node_names=[self.unit_name, sorted_cm[0]]  ## <<<------ should we also add a check here?
                )
```

Given this is quite an exception (i.e. going from 1->2 or 2->1), I took the simpler approach.

phvalguima: I can improve the comments around here, this logic is pretty brittle tbh.

lib/charms/opensearch/v0/helper_cluster.py
```diff
@@ -139,6 +139,14 @@ def nodes_by_role(nodes: List[Node]) -> Dict[str, List[Node]]:
 
         return result
 
+    @staticmethod
+    def elected_manager(nodes: List[Node]) -> List[str]:
```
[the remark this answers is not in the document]
phvalguima: Thanks! Let me update the signature here

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -537,8 +543,6 @@ def _on_update_status(self, event: UpdateStatusEvent):
 
         # if there are exclusions to be removed
         if self.unit.is_leader():
-            self.opensearch_exclusions.cleanup()
```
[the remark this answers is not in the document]
phvalguima: That is not the case... It is done on every case, except `HealthColors.UNKNOWN`. Indeed, we *defer* the event if it is not green... I put it down there because I need the API to be responsive before configuring voting exclusions. If it is not responsive, we will get UNKNOWN anyways and retry later anyways.

I will add some comments to clarify that.

tests/integration/ha/continuous_writes.py
```diff
@@ -56,6 +57,8 @@ async def start(self, repl_on_all_nodes: bool = False, is_bulk: bool = True) ->
         # create index if custom conf needed
         if repl_on_all_nodes:
             await self._create_fully_replicated_index()
+        else:
+            await self._create()
```
phvalguima, line 61: If we let opensearch auto-create this index, it will create with "1-all" replicas; which for a test of scaling down 1 and back, won't work.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1060,6 +1086,80 @@ def _restart_opensearch(self, event: _RestartOpenSearch) -> None:
...
 
+    def _settle_voting_exclusions(self, unit_is_stopping: bool = False):  # noqa C901
+        """Settle the exclusions for all voting units."""
+        hosts = self.alt_hosts
+        if (
+            self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR
+            and (peer_cm_rel_data := self.opensearch_peer_cm.rel_data()) is not None
+        ):
+            # Also consider peer-relation units
+            hosts.extend([node.ip for node in peer_cm_rel_data.cm_nodes])
```
Mehdi-Bendriss, line 1097: This is not needed. `self.alt_hosts` is enough, it already contains all possible alternative hosts incl. from the large deployment relations.

**Note:** Please do not rely on `rel_data` for as long as you can - we should aim for a clear separation of concerns. The 2 checks can be replaced more robustly with `self.opensearch_peer_cm.is_consumer()`.

```suggestion
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1060,6 +1086,80 @@ def _restart_opensearch(self, event: _RestartOpenSearch) -> None:
...
+                    # if still in the list
+                    nodes.remove(node)
+                    # we can finish the for loop here
+                    break
+                elif not unit_is_stopping and not node:
+                    # This unit is starting, we must assure it shows up in the list
+                    # Let's retry
+                    raise OpenSearchStartError("Node not in the list of nodes")
+                # Any other case is okay to move forward without changes
+                break
```
Mehdi-Bendriss, line 1118:
```suggestion
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1060,6 +1086,80 @@ def _restart_opensearch(self, event: _RestartOpenSearch) -> None:
...
+                if unit_is_stopping and node:
+                    # This is a stop operation, we exclude the unit that is stopping
+                    # if still in the list
+                    nodes.remove(node)
+                    # we can finish the for loop here
+                    break
+                elif not unit_is_stopping and not node:
+                    # This unit is starting, we must assure it shows up in the list
+                    # Let's retry
+                    raise OpenSearchStartError("Node not in the list of nodes")
```
Mehdi-Bendriss, line 1116: Can you explain this? if we call `settle_voting_exclusions` from `post_start_init`, why should we have this check + exception?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1060,6 +1086,80 @@ def _restart_opensearch(self, event: _RestartOpenSearch) -> None:
...
+        # For the sake of predictability, we always sort the cluster managers
+        sorted_cm = sorted(cms)
+        # We always clean the voting.
+        # Then, we deal with the specific cases.
+        self.opensearch_exclusions.delete_voting()
+
+        # Each of the possible cases
+        if len(cms) == 1:
+            # This condition only happens IF the cluster had 2x units and now is scalig down to 1
+            # In this scenario, we should exclude the node that is going away, to force election
```
Mehdi-Bendriss, line 1130: Can you explain why is this needed? if the cluster scales down from 2 to 1 unit, doesn't that mean that an election will happen anyway?

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: I think I overlooked and missed a safety component there, where if we remain with 2 units - and the one being removed is **not** the current elected CM; maybe it makes sense to add the other unit to the voting exclusion. 
This should reduce switchovers and the risk it entails

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1060,6 +1086,80 @@ def _restart_opensearch(self, event: _RestartOpenSearch) -> None:
...
+            # Now, we clean up the sorted_cm list, as we want to be sure the new manager is elected
+            # and different than the excluded units.
+            sorted_cm.pop(0)
+            # We do not exclude the self.unit_name
+        else:
+            # In this case, we either are scaling down to 0 or len(cms) > 2.
+            # There is nothing more to do then cleanup the exclusions
+            return
+
+        # Now, we must be sure a new manager is elected, or there was a failure
```
Mehdi-Bendriss, line 1153: I'm not sure this is needed / will work. If no CM is elected, we will never have online nodes / API to consume.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -44,73 +36,54 @@ def __init__(self, charm):
...
-        if need_voting_cleanup and self._delete_voting():
-            self._charm.peers_data.delete(self._scope, VOTING_TO_DELETE)
-
+    def allocation_cleanup(self) -> None:
+        """Delete alloc exclusions that failed to be deleted."""
         allocations_to_cleanup = self._charm.peers_data.get(
             self._scope, ALLOCS_TO_DELETE, ""
         ).split(",")
-        if allocations_to_cleanup and self._delete_allocations(allocations_to_cleanup):
+        if allocations_to_cleanup and self.delete_allocations_exclusion(allocations_to_cleanup):
```
Mehdi-Bendriss, line 44: nit: adding `_exclusions` at the end is redundant, given that we are already in the "exclusions" module.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -44,73 +36,54 @@ def __init__(self, charm):
...
 
-    def _delete_voting(self) -> bool:
+    def delete_voting(self, alt_hosts: Optional[List[str]] = None) -> bool:
         """Remove all the voting exclusions - cannot target 1 exclusion at a time."""
         # "wait_for_removal" is VERY important, it removes all voting configs immediately
         # and allows any node to return to the voting config in the future
         try:
             self._opensearch.request(
                 "DELETE",
                 "/_cluster/voting_config_exclusions?wait_for_removal=false",
```
Mehdi-Bendriss, line 78: I do not remember why did I set in the past `wait_for_removal=false` instead of `true`. I now don't see a reason why it shouldn't be `true`

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -155,6 +128,19 @@ def _fetch_allocations(self) -> Set[str]:
         finally:
             return allocation_exclusions
 
+    def _fetch_voting_exclusions(self, alt_hosts) -> Set[str]:
```
Mehdi-Bendriss, line 131: nit: `_exclusions` is redundant

tests/integration/ha/continuous_writes.py
Mehdi-Bendriss: Can we rename the method to something more along the lines of `_create_index_with_0_all_replicas`

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1060,6 +1086,80 @@ def _restart_opensearch(self, event: _RestartOpenSearch) -> None:
 
         self._start_opensearch_event.emit()
 
+    def _settle_voting_exclusions(self, unit_is_stopping: bool = False):  # noqa C901
```
Mehdi-Bendriss, line 1089: Could we move this to the `opensearch_nodes_exclusions` module? and rename it as `settle_voting`?

Mehdi-Bendriss (commented):
Thanks Pedro - I have a few questions. There is also the aspect of the missing exception handling of `settle_voting` in various places in the code.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -573,6 +585,9 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
         restart_requested = False
         if self.opensearch_config.update_host_if_needed():
             restart_requested = True
+            # Review voting exclusions as our IP has changed: we may be coming back from a network
+            # outage case.
+            self._settle_voting_exclusions(unit_is_stopping=False)
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss so, `self.alt_hosts` will not return the local host in the list? How can I get it then?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -537,17 +548,26 @@ def _on_update_status(self, event: UpdateStatusEvent):
 
         # if there are exclusions to be removed
         if self.unit.is_leader():
-            self.opensearch_exclusions.cleanup()
```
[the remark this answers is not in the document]
phvalguima: Yes, the health checks below allow anything pass, unless the cluster is really on a bad state (i.e. UNKNOWN). So, moved the check below these first health checks because it makes more sense.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1004,37 +1028,39 @@ def _post_start_init(self, event: _StartOpenSearch):  # noqa: C901
...
         if self.opensearch.is_node_up():
             try:
-                nodes = self._get_nodes(True)
-                # do not add exclusions if it's the last unit to stop
-                # otherwise cluster manager election will be blocked when starting up again
-                # and re-using storage
                 if len(nodes) > 1:
-                    # TODO: we should probably NOT have any exclusion on restart
-                    # https://chat.canonical.com/canonical/pl/bgndmrfxr7fbpgmwpdk3hin93c
-                    # 1. Add current node to the voting + alloc exclusions
```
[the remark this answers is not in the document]
phvalguima: The first description, yes. The second: `TODO: we should probably NOT have any exclusion on restart` is outdated. Now, we only add nodes to the voting exclusions IF we are scaling between 3-1 nodes or vice versa.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: If `self.alt_hosts` returns the right value in any case, do I still need that if and `self.opensearch_peer_cm.is_consumer()` then?

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: First, settle voting logic must be executed synchronously with a (re)start. This is why I am adding these two retry loops (one now, another at its end). Ideally, if we have a problem here, that may cause the entire cluster to get stuck in "metadata error" as we were seeing before.

Because between powering up and actually having the service up takes a while, this for loop assures we wait long enough for the opensearch service to come up.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: If we made to this point, it means any previous voting exclusions have been already removed. So, we have 2x voting units available in this cluster.

The main problem now is, if we have 2x voting units, then the minimum quorum for us to keep working will be:
```
(N=2)  / 2 + 1 => 2 units
```

If I remove one unit, then the other cannot re-elect. So, I apply this exclusion to reduce the voting number down to 1x unit + also moving the cluster manager, if the unit going away was the manager.

phvalguima: It must happen before stopping, so we are sure the metadata will be updated on the unit that is staying.

This is the only way I got on my own tests to safely downscale from 2->1 unit.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: The node is marked as leaving after 30s, max. The election, if needed, will take place right after. Given we are always excluding the unit that is leaving; if that unit was the elected manager, than it will trigger a re-election right away.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
phvalguima: @Mehdi-Bendriss on my own tests, I could not get this DELETE to work with `wait_for_removal=true`... I wanted to move it to True as well, btw!

lib/charms/opensearch/v0/opensearch_base_charm.py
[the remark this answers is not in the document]
phvalguima: Changed the way we call both ClusterTopology's `elected_manager`  and `nodes`. They will now do more checks on self.alt_hosts.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: @Mehdi-Bendriss I would not recommend that. The main reason is because I noticed moving the elected manager between nodes is far faster than juju hooks. We need to be predictable in this specific case, even if it means moving the elected manager.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: This logic has changed following `settle_voting` latest updates.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
phvalguima: @Mehdi-Bendriss, found this in the docs:

> Defaults to true, meaning that all excluded nodes must be removed from the cluster before this API takes any action.

So, wait_for_removal=true expects the node to be gone entirely... That is why it is not suitable for us.

PR #368, opened 2024-07-16
[DPE-4857] improve naming of truststore and keystore secrets

reneradoi:
## Issue
Currently the secrets for storing the truststore and keystore passwords are named `keystore-password-{category}`, which is not ideal considering they are already stored as part of the tls secret for each category.

## Solution
Rename the secrets to:
- `truststore-password` for the ca truststore
- `keystore-password` as the generic name that can be referenced within the respective secrets app-admin, unit-transport and unit-http

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -423,16 +423,21 @@ class OpenSearchTLS(Object):
 
     def _create_keystore_pwd_if_not_exists(self, scope: Scope, cert_type: CertType, alias: str):
         """Create passwords for the key stores if not already created."""
-        keystore_pwd = None
+        store_pwd = None
+        if alias == "ca":
+            store_type = "truststore"
+        else:
+            store_type = "keystore"
+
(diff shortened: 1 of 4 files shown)


lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -423,16 +423,21 @@ def get_unit_certificates(self) -> Dict[CertType, str]:
 
     def _create_keystore_pwd_if_not_exists(self, scope: Scope, cert_type: CertType, alias: str):
         """Create passwords for the key stores if not already created."""
-        keystore_pwd = None
+        store_pwd = None
+        if alias == "ca":
+            store_type = "truststore"
+        else:
+            store_type = "keystore"
```
Mehdi-Bendriss, line 430: nit for conciseness:
```suggestion
        store_type = "truststore" if alias == "ca" else "keystore"
```

Mehdi-Bendriss (dismissed):

phvalguima (dismissed):

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #373, opened 2024-07-19
`publish_host` is a single IP, not a list

phvalguima:
The way `publish_host` is being dealt, it is expected a list of hostnames / IPs. However, publish_host has a single value. Hence, we should not filter it with `set()`.

This fix has been originally applied in this PR: https://github.com/canonical/opensearch-operator/pull/367. It has been also verified with successful passed runs of `test_ha_networking.py` checks, which triggers this logic.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_config.py
+++ b/lib/charms/opensearch/v0/opensearch_config.py
@@ -244,8 +244,8 @@ class OpenSearchConfig:
             ),
             NetworkHost(
                 "network.publish_host",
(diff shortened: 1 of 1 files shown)


reneradoi (approved):
Hey @phvalguima Just for my understanding: Why a separate PR if it's already included in #367?

Edit: Ok nvm, just saw your message in the chat.

Mehdi-Bendriss (approved):

Issue #323, opened 2024-06-06
hook failed: "node-lock-fallback-relation-changed"


juditnovak:
## Steps to reproduce
Shut down and resart the server hosting the Juju model.
Opensearch units will be blocked on `net.ipv4.tcp_retries2 should be 5 `
Destroy a unit and add a new one.
The new one will have this issue (see attached screenshots)

## Expected behavior
Model running smooth.

## Actual behavior

```
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-opensearch-2/charm/./src/charm.py", line 267, in <module>
    main(OpenSearchOperatorCharm)
  File "/var/lib/juju/agents/unit-opensearch-2/charm/venv/ops/main.py", line 544, in main
    manager.run()
  File "/var/lib/juju/agents/unit-opensearch-2/charm/venv/ops/main.py", line 520, in run
    self._emit()
  File "/var/lib/juju/agents/unit-opensearch-2/charm/venv/ops/main.py", line 509, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name)
  File "/var/lib/juju/agents/unit-opensearch-2/charm/venv/ops/main.py", line 143, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-opensearch-2/charm/venv/ops/framework.py", line 350, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-opensearch-2/charm/venv/ops/framework.py", line 849, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-opensearch-2/charm/venv/ops/framework.py", line 939, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-opensearch-2/charm/lib/charms/opensearch/v0/opensearch_locking.py", line 164, in _on_peer_relation_changed
    if self._unit_with_lock and self._unit_requested_lock(
  File "/var/lib/juju/agents/unit-opensearch-2/charm/lib/charms/opensearch/v0/opensearch_locking.py", line 103, in _unit_requested_lock
    value = self._relation.data[unit].get("lock-requested")
  File "/var/lib/juju/agents/unit-opensearch-2/charm/venv/ops/model.py", line 1582, in __getitem__
    return self._data[key]
KeyError: <ops.model.Unit opensearch/0>
```

## Versions

Operating system: 

Juju CLI: 3.1.8-genericlinux-amd64

[...]

Issue #222, opened 2024-04-11
[LOW IMPORTANCE][STABILITY] KeyError: 'node.roles'


juditnovak:
I've seen this error a few times on local runs, so I add this ticket to signify - in case it may occur to others or on pipelines pls add more info to the ticket.

The following exception has been raised a couple of times:

```
unit-opensearch-1: 11:00:28 ERROR unit.opensearch/1.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-opensearch-1/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 408, in current
    nodes = self.request("GET", f"/_nodes/{self.node_id}", alt_hosts=self._charm.alt_hosts)
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/var/lib/juju/agents/unit-opensearch-1/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 375, in node_id
    nodes = self.request("GET", "/_nodes").get("nodes")
  File "/var/lib/juju/agents/unit-opensearch-1/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 297, in request
    resp = call(retries, resp_status_code)
  File "/var/lib/juju/agents/unit-opensearch-1/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 252, in call
    raise OpenSearchHttpError()
charms.opensearch.v0.opensearch_exceptions.OpenSearchHttpError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-opensearch-1/charm/./src/charm.py", line 94, in <module>
    main(OpenSearchOperatorCharm)
  File "/var/lib/juju/agents/unit-opensearch-1/charm/venv/ops/main.py", line 456, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-opensearch-1/charm/venv/ops/main.py", line 144, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-opensearch-1/charm/venv/ops/framework.py", line 351, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-opensearch-1/charm/venv/ops/framework.py", line 853, in _emit
    self._reemit(event_path)
[...]

PR #374, opened 2024-07-19
[DPE-4682] Add fix for user management

### LINKED ISSUE #331: Remove "non-charmed" user deletion at `update-status` and `client-broken`
Currently, we have `OpenSearchUserManager.remove_users_and_roles`, which removes users and roles that are not linked to system users or existing relations. [The removal happens here](https://github.com/canonical/opensearch-operator/blob/5174eed53acc0ba5a90da144e9ca0c16fd8c3da9/lib/charms/opensearch/v0/opensearch_users.py#L263-L264).

This method is called in two different places: [update status and client relation broken events](https://github.com/search?q=repo%3Acanonical/opensearch-operator%20remove_users_and_roles&type=code). Now, removing the relation user and role makes sense, but we should only remove that user / role, not a generic clean-up. We should also remove the call from `update_status`.

phvalguima:
# Issue

Addressing  #331 , namely that currently it's impossible to add users manually to Opensearch, as they are cleaned up on the first `update-status`.

# Solution

Saving every relation ID with the corresponding user's name in a dictionary on the databag.
This allows us to keep track of relation users and the relations they belong to.
This data structure is maintained such as
 - `index-created` : adding new relation - usrer pair as craeting the user in Opensearch
 - `update-status`, `relation-broken`, `relation-departed`: removing (stale or dedicated) relations - user pairs from the databag as removing Opensearch users

## Developers' notes

In order to get involved scopes clear:
- Operations on the new databag structure are strongly coupled togeter. Thus they MUST be in the same module.
  (This principle is also in attempt to decrease the "spagetti".)
- Since removing users is now clearly specific to client relations, this functionality should reside in the Provider code.
  Accompaigned with the suited user creation functionality encapsulating all relation-specific details (role name same as username, databag structures to be maintained, etc.)
- Note: since extra user roles may be shared across multiple users, and re-used, we do NOT clean up those. But normal roles only.
- Note2: We are taking advantage of usernames being equal to their related roles (and don't store roles specifically).

### DIFF
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -3,15 +3,13 @@
 
 """Utility classes and methods for getting cluster info, configuration info and suggestions."""
 import logging
-from typing import TYPE_CHECKING, Dict, List, Optional
+from typing import Dict, List, Optional
 
 from charms.opensearch.v0.constants_charm import GeneratedRoles
 from charms.opensearch.v0.helper_enums import BaseStrEnum
 from charms.opensearch.v0.models import App, Node
+from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
 from tenacity import retry, stop_after_attempt, wait_exponential
-
-if TYPE_CHECKING:
-    from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
 
 # The unique Charmhub library identifier, never change it
 LIBID = "80c3b9eff6df437bb4175b1666b73f91"
@@ -44,7 +42,7 @@ class ClusterTopology:
 
     @staticmethod
     def get_cluster_settings(
-        opensearch: "OpenSearchDistribution",
+        opensearch: OpenSearchDistribution,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
         include_defaults: bool = False,
@@ -144,7 +142,7 @@ class ClusterTopology:
 
     @staticmethod
     def nodes(
-        opensearch: "OpenSearchDistribution",
+        opensearch: OpenSearchDistribution,
         use_localhost: bool,
         hosts: Optional[List[str]] = None,
     ) -> List[Node]:
@@ -183,7 +181,7 @@ class ClusterState:
         reraise=True,
     )
     def shards(
-        opensearch: "OpenSearchDistribution",
+        opensearch: OpenSearchDistribution,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
         verbose: bool = False,
@@ -229,7 +227,7 @@ class ClusterState:
         reraise=True,
     )
     def allocation_explain(
-        opensearch: "OpenSearchDistribution",
+        opensearch: OpenSearchDistribution,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
     ) -> List[Dict[str, str]]:
@@ -248,7 +246,7 @@ class ClusterState:
         reraise=True,
     )
     def indices(
-        opensearch: "OpenSearchDistribution",
+        opensearch: OpenSearchDistribution,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
     ) -> Dict[str, Dict[str, str]]:
@@ -275,7 +273,7 @@ class ClusterState:
 
     @staticmethod
     def shards_by_state(
-        opensearch: "OpenSearchDistribution",
+        opensearch: OpenSearchDistribution,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
     ) -> Dict[str, List[str]]:
@@ -292,7 +290,7 @@ class ClusterState:
 
     @staticmethod
     def busy_shards_by_unit(
-        opensearch: "OpenSearchDistribution",
+        opensearch: OpenSearchDistribution,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
     ) -> Dict[str, List[str]]:
@@ -315,7 +313,7 @@ class ClusterState:
 
     @staticmethod
     def health(
-        opensearch: "OpenSearchDistribution",
+        opensearch: OpenSearchDistribution,
         wait_for_green: bool,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -3,13 +3,15 @@
 
 """Utility classes and methods for getting cluster info, configuration info and suggestions."""
 import logging
-from typing import Dict, List, Optional
+from typing import TYPE_CHECKING, Dict, List, Optional
 
 from charms.opensearch.v0.constants_charm import GeneratedRoles
 from charms.opensearch.v0.helper_enums import BaseStrEnum
 from charms.opensearch.v0.models import App, Node
-from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
 from tenacity import retry, stop_after_attempt, wait_exponential
+
+if TYPE_CHECKING:
+    from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
 
 # The unique Charmhub library identifier, never change it
 LIBID = "80c3b9eff6df437bb4175b1666b73f91"
@@ -42,7 +44,7 @@ class ClusterTopology:
 
     @staticmethod
     def get_cluster_settings(
-        opensearch: OpenSearchDistribution,
+        opensearch: "OpenSearchDistribution",
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
         include_defaults: bool = False,
@@ -142,7 +144,7 @@ class ClusterTopology:
 
     @staticmethod
     def nodes(
-        opensearch: OpenSearchDistribution,
+        opensearch: "OpenSearchDistribution",
         use_localhost: bool,
         hosts: Optional[List[str]] = None,
     ) -> List[Node]:
@@ -181,7 +183,7 @@ class ClusterState:
         reraise=True,
     )
     def shards(
-        opensearch: OpenSearchDistribution,
+        opensearch: "OpenSearchDistribution",
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
         verbose: bool = False,
@@ -227,7 +229,7 @@ class ClusterState:
         reraise=True,
     )
     def allocation_explain(
-        opensearch: OpenSearchDistribution,
+        opensearch: "OpenSearchDistribution",
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
     ) -> List[Dict[str, str]]:
@@ -246,7 +248,7 @@ class ClusterState:
         reraise=True,
     )
     def indices(
-        opensearch: OpenSearchDistribution,
+        opensearch: "OpenSearchDistribution",
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
     ) -> Dict[str, Dict[str, str]]:
@@ -273,7 +275,7 @@ class ClusterState:
 
     @staticmethod
     def shards_by_state(
-        opensearch: OpenSearchDistribution,
+        opensearch: "OpenSearchDistribution",
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
     ) -> Dict[str, List[str]]:
@@ -290,7 +292,7 @@ class ClusterState:
 
     @staticmethod
     def busy_shards_by_unit(
-        opensearch: OpenSearchDistribution,
+        opensearch: "OpenSearchDistribution",
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
     ) -> Dict[str, List[str]]:
@@ -313,7 +315,7 @@ class ClusterState:
 
     @staticmethod
     def health(
-        opensearch: OpenSearchDistribution,
+        opensearch: "OpenSearchDistribution",
         wait_for_green: bool,
         host: Optional[str] = None,
         alt_hosts: Optional[List[str]] = None,
--- a/lib/charms/opensearch/v0/opensearch_relation_provider.py
+++ b/lib/charms/opensearch/v0/opensearch_relation_provider.py
@@ -445,8 +445,7 @@ class OpenSearchProvider(Object):
         and their name is saved in the databag for later reference.
         """
         self.user_manager.create_role(role_name=user, permissions=permissions)
-        if not (users := self.charm.peers_data.get_object(Scope.APP, ClientUsersDict)):
-            users = {}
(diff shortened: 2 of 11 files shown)


reneradoi (approved):
Looks good!

lib/charms/opensearch/v0/helper_cluster.py
```diff
@@ -41,7 +43,7 @@ def generated_roles() -> List[str]:
 
     @staticmethod
     def get_cluster_settings(
-        opensearch: OpenSearchDistribution,
+        opensearch: "OpenSearchDistribution",
```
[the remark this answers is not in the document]
phvalguima: @reneradoi, I've noticed that, if I run `tests/unit/lib/test_helper_cluster.py` unit test separated, it would fail because of a circular dependency. If I run it using the entire unit tests we have, then it works, likely because OpenSearchDistribution was already loaded by some other test.

I changed this so I could remove the dependency on "OpenSearchDistribution", which was causing this loop dependency in the first place!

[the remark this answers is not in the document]
phvalguima: Check it here: https://pastebin.ubuntu.com/p/87F6qYwQDs/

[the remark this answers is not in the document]
reneradoi: Ah I remember that, we briefly chatted about it. Thanks for the reminder!

reneradoi (approved):

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -299,10 +307,24 @@ def create_opensearch_users(
...
+        users = json.loads(
+            self.charm.peers_data.get(Scope.APP, ClientUsersDict, auto_casting=False) or "{}"
+        )
+        if rel_id not in users:
+            users[rel_id] = []
+        roles = json.loads(
+            self.charm.peers_data.get(Scope.APP, ClientRolesDict, auto_casting=False) or "{}"
+        )
+        if rel_id not in roles:
+            roles[rel_id] = []
```
Mehdi-Bendriss, line 326: 1. Why do we store the `roles` - given the role is exactly the username? 
2. If really necessary - Could we merge these 2 into a single dict: `{"rel_id": {"roles": ..., "users": ....}}`

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -311,16 +333,22 @@ def create_opensearch_users(
                 role_name=username,
                 permissions=self.get_extra_user_role_permissions(extra_user_roles, index),
             )
-            roles = [username]
```
Mehdi-Bendriss, line 314: We should keep this variable - I believe this was quite explanatory as to what `[username]` means exactly in the user creation workflow

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -311,16 +333,22 @@ def create_opensearch_users(
...
-            self.user_manager.create_user(username, roles, hashed_pwd)
+            roles[rel_id].extend([username] + extra_user_roles.split(","))
```
Mehdi-Bendriss, line 336: I'm not sure I understand the need for adding `extra_user_roles.split(",")` - what would this mean in the context of OpenSearch. This is mainly used for the charm to know which permissions to grant the role assigned to this user.

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -311,16 +333,22 @@ def create_opensearch_users(
...
+                [{"op": "replace", "path": "/opendistro_security_roles", "value": [username]}],
             )
+            users[rel_id].append(username)
         except OpenSearchUserMgmtError as err:
             logger.error(err)
             raise
 
+        if rel_id:
+            self.charm.peers_data.put(Scope.APP, ClientUsersDict, json.dumps(users))
+            self.charm.peers_data.put(Scope.APP, ClientRolesDict, json.dumps(roles))
```
Mehdi-Bendriss, line 350: same comment about merging these 2 into a single dict - if necessary to store roles at all.

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -251,17 +254,33 @@ def remove_users_and_roles(self, departed_relation_id: Optional[int] = None):
...
-        relations = self.model.relations.get(ClientRelationName, [])
-        relation_users = set(
-            [
-                f"{ClientRelationName}_{relation.id}"
-                for relation in relations
-                if relation.id != departed_relation_id
-            ]
-        )
-        self._remove_lingering_users(relation_users)
-        self._remove_lingering_roles(relation_users)
```
Mehdi-Bendriss, line 264: I believe we can keep it simple and keep this and remove the newly added code, and simply add `rel_id` to the `_remove_lingering_*` methods.

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -251,17 +254,33 @@ def remove_users_and_roles(self, departed_relation_id: Optional[int] = None):
...
+            # Check if this role is not present in any other relation:
+            if any(role in r for r in relation_roles.values()):
+                continue
+            try:
+                self.remove_role(role)
+            except OpenSearchUserMgmtError:
+                logger.error(f"failed to remove role {role}")
+
+        self.charm.peers_data.put(Scope.APP, ClientUsersDict, json.dumps(relation_users))
+        self.charm.peers_data.put(Scope.APP, ClientRolesDict, json.dumps(relation_roles))
```
Mehdi-Bendriss, line 283: let's move this to under the `except OpenSearchUserMgmtError` blocks in the deleted methods below.

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -313,33 +332,3 @@ def put_internal_user(self, user: str, hashed_pwd: str):
                 COSUser,
                 [{"op": "replace", "path": "/opendistro_security_roles", "value": roles}],
             )
-
-    def _remove_lingering_users(self, relation_users: Set[str]):
```
Mehdi-Bendriss, line 317: let's rollback

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -313,33 +332,3 @@ def put_internal_user(self, user: str, hashed_pwd: str):
...
-            logger.error("failed to get users")
-            return
-
-        for username in database_users - app_users:
-            try:
-                self.remove_user(username)
-            except OpenSearchUserMgmtError:
-                logger.error(f"failed to remove user {username}")
-
-    def _remove_lingering_roles(self, roles: Set[str]):
```
Mehdi-Bendriss, line 331: let's rollback

Mehdi-Bendriss (changes requested):
Thanks Pedro - i believe we need to simplify / rollback quite a bit of this PR.

lib/charms/opensearch/v0/helper_cluster.py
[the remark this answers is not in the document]
juditnovak: Guys, PLS let's not groom and grow circular dependencies in our code, as much as avoidable...

This problem can be trivially resolved using an application constant, instead of the hard-coded list of default roles in this module. See https://github.com/canonical/opensearch-operator/commit/d227aa9c48e540edc6b86a492cd15ae9962833e6

juditnovak:
@phvalguima Now that I took it over U're invited on this PR to provide insighst :-) 
(I can't add you as a reviewer)

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Solution fully re-worked: closing this thread.

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Solution fully re-worked: closing this thread.

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Solution fully re-worked: closing this thread. 

Extra user roles are not stored anymore.

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Reworked, we only have a single list per relation now.

lib/charms/opensearch/v0/opensearch_users.py
juditnovak: Fully re-fatctored (into `opensearch_provider` ). Closing thread as it's outdated, new one may raise relating to actual code.

lib/charms/opensearch/v0/opensearch_users.py
juditnovak: All re-worked, closing thread.

lib/charms/opensearch/v0/opensearch_users.py
juditnovak: Reworked, closing thread. New thread may raise alongside the new solution.

lib/charms/opensearch/v0/opensearch_users.py
juditnovak: Reworked, structure changed. We should discuss latest changes.

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -433,3 +435,70 @@ def update_dashboards_password(self):
...
+        self, user: str, permissions: dict[str], hashed_pwd: str, relation_id: int
+    ):
+        """Create a relation user.
+
+        Relation users are registered with a dedicated role which maps to the username,
+        and their name is saved in the databag for later reference.
+        """
+        self.user_manager.create_role(role_name=user, permissions=permissions)
+        if not (users := self.charm.peers_data.get_object(Scope.APP, ClientUsersDict)):
+            users = {}
```
Mehdi-Bendriss, line 449: nit:
```suggestion
        users = self.charm.peers_data.get_object(Scope.APP, ClientUsersDict) or {}
```

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -433,3 +435,70 @@ def update_dashboards_password(self):
...
+
+        Args:
+            departed_relation_id: if a relation is departing, pass in the ID and its user will be
+                deleted.
+        """
+        if not self.opensearch.is_node_up() or not self.unit.is_leader():
+            return
+
+        if not (relation_users := self.charm.peers_data.get_object(Scope.APP, ClientUsersDict)):
+            relation_users = {}
```
Mehdi-Bendriss, line 475: Similar nit as above, should help reduce 1 level of indentation

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -433,3 +435,70 @@ def update_dashboards_password(self):
...
+
+            if departed_relation_id:
+                logging.warning(
+                    "User for relation %d wasn't registered in internal cham workflows.",
+                    departed_relation_id,
+                )
+
+        clearnup_rel_ids = []
```
Mehdi-Bendriss, line 483: typo:
```suggestion
        cleanup_rel_ids = []
```

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -433,3 +435,70 @@ def update_dashboards_password(self):
...
+        if departed_relation_id:
+            clearnup_rel_ids = [str(departed_relation_id)]
+        else:
+            rel_ids = [str(relation.id) for relation in self.opensearch_provides.relations]
+            clearnup_rel_ids = list(set(relation_users.keys()) - set(rel_ids))
```
Mehdi-Bendriss, line 488: I think this logic can run outside of the `else`, so that we always try to cleanup everything

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -433,3 +435,70 @@ def update_dashboards_password(self):
...
+
+        for rel_id in clearnup_rel_ids:
+            if username := relation_users.get(rel_id):
```
Mehdi-Bendriss, line 491: Could you explain when would this be None?
the `del` below suggests otherwise (it will raise a key error if it was ever None)

Mehdi-Bendriss (commented):
Thanks Judit! I have a couple of minor questions

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: With the current code It could be `None` if --for example-- for any "unexpected" reason the function is called with a `departed_relation_id` for a relation that may not have been registered in the databag (anymore).

I'm adding changes so that this will not be an option anymore.

Yet, TBH I find it safer to ensure smooth user experience over non-crtiical errors. We had too many issues (https://github.com/canonical/opensearch-operator/issues/323, https://github.com/canonical/opensearch-operator/issues/222, https://github.com/canonical/opensearch-operator/issues/422) on being confident referencing internal data structures with no security measures.
(I agree that we should signal any inconsistency, but I wonder if it's worth to put the charm in an `error` state. Ideally the internal data structure should recover itself, and the `update-status` usage ensures no lingering users staying in the system.)

You are right that the `del` below is incorrect without a `try...catch`. Changing it so we log and exit in case the key was not found. 

Shame on me not to have had a unittest on the corner case. Adding that, too.

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Following a live question on this matter, my understanding was to only do a full cleanup when called in the `update-status` hook (i.e. w/o a `departed_relation_id`).  

Of course, if preferred, we can do a full cleanup on each run.

Mehdi-Bendriss (approved):

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Done.

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Done.

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Done.

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Done.

lib/charms/opensearch/v0/opensearch_relation_provider.py
juditnovak: Done. (Tests covered the refactored code.)

Mehdi-Bendriss (approved):

PR #376, opened 2024-07-22
[DPE-4900] bugfix truststore secret

reneradoi:
## Issue
This is a follow-up to https://github.com/canonical/opensearch-operator/pull/357. If the `on_certificate_available` hook is run before the `on_relation_created` (for the TLS relation), it may happen that the truststore password is not yet available and is therefore created. But this must only happen on leader-units, because the truststore-secret is of scope `app` and otherwise the secret creation would fail with `ops.model.ModelError: ERROR this unit is not the leader` (as seen in [this CI run](https://github.com/canonical/opensearch-dashboards-operator/actions/runs/10007015106/job/27661171763#step:22:211)).

## Solution
Only create the truststore password secret if the current unit is the juju-leader.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -443,7 +443,8 @@ class OpenSearchTLS(Object):
         keytool = f"sudo {self.jdk_path}/bin/keytool"
 
         admin_secrets = self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
-        self._create_keystore_pwd_if_not_exists(Scope.APP, CertType.APP_ADMIN, "ca")
+        if self.charm.unit.is_leader():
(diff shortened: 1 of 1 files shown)


juditnovak (commented):
Is there a way to test this? I mean the test that was supposed to have failed before the fix is there? :-)

reneradoi:
> Is there a way to test this? I mean the test that was supposed to have failed before the fix is there? :-)

I think this would only be possible with manual effort, e.g. by commenting out [this line](https://github.com/canonical/opensearch-operator/blob/main/lib/charms/opensearch/v0/opensearch_tls.py#L147), then building and deploying while checking the `debug-log`. But not really on CI I guess.

Mehdi-Bendriss (approved):

juditnovak (approved):
A docstring on the test like "Only leader is making an attempt to create truststore password" may be nice to have :-)

PR #380, opened 2024-07-30
[DPE-4792] reload certificates via API

reneradoi:
## Issue
Currently the OpenSearch application is restarted when TLS certificates are renewed. This could be simplified by using the "hot reload" API OpenSearch provides for renewing the TLS certs.

## Solution
- activate the plugins.security.ssl_cert_reload_enabled setting
- when TLS conf is set, query the API instead of restarting OpenSearch

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_config.py
+++ b/lib/charms/opensearch/v0/opensearch_config.py
@@ -186,6 +186,13 @@ class OpenSearchConfig:
             True,
         )
 
+        # enable hot reload of TLS certs (without restarting the node)
+        self._opensearch.config.put(
+            self.CONFIG_YML,
+            "plugins.security.ssl_cert_reload_enabled",
+            True,
+        )
+
     def remove_temporary_data_role(self):
         """Remove the data role that was added temporarily to the first dedicated CM node."""
         conf = self._opensearch.config.load(self.CONFIG_YML)


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -721,15 +715,18 @@ def on_tls_conf_set(
                 truststore_pwd=truststore_pwd,
                 keystore_pwd=keystore_pwd,
             )
-        else:
+
             # write the admin cert conf on all units, in case there is a leader loss + cert renewal
-            self.opensearch_config.set_admin_tls_conf(current_secrets)
+            if not admin_secrets.get("subject"):
```
juditnovak, line 720: Why specific query on the `subject`?  Why not on the cert ifself or so?

juditnovak: (Whatever the answer may be, perhaps it should go in the docstring, if the check is correct and fully intentional :-) )

juditnovak: I'm a bit wondering on a verbose-named little function that should rather go in the `opensearch_tls` module instead... Like `has_valid_admin_cert()` or so...? Instead of the raw check

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -621,3 +618,52 @@ def delete_stored_tls_resources(self):
...
+            response_transport = requests.put(
+                url_transport,
+                cert=(admin_cert, admin_key),
+                verify=f"{self.certs_path}/chain.pem",
+                timeout=5,
+            )
+            if not (
+                response_http.status_code == requests.codes.ok
+                and response_transport.status_code == requests.codes.ok
+            ):
```
juditnovak, line 661: @Mehdi-Bendriss We've been wondering if it could be a good idea to extend the already existing `request()` method of Opensearch to handle this case (i.e. query with cert instead of admin credentials). What do you think? The use case is unique and out of ordinary, however may result in a more consistent usage of API requrests...

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -69,7 +69,7 @@ def set_admin_tls_conf(self, secrets: Dict[str, any]):
         self._opensearch.config.put(
             self.CONFIG_YML,
             "plugins.security.authcz.admin_dn/{}",
-            f"{normalized_tls_subject(secrets['subject'])}",
+            f"{normalized_tls_subject(secrets.get('subject'))}",
```
juditnovak, line 72: What if `secrets.get('subject')` returns `None`? What does `normalized_tls_subject(secrets.get('subject'))` return in that case?

Shouldn't we first check if there's a valid (not `None`, not "", etc.) value, and _then_ `put()` the value to the config?

tests/integration/tls/test_tls.py
```diff
@@ -86,3 +94,63 @@ async def test_cluster_formation_after_tls(ops_test: OpsTest) -> None:
     leader_unit_ip = await get_leader_unit_ip(ops_test)
 
     assert await check_cluster_formation_successful(ops_test, leader_unit_ip, unit_names)
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "large"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_tls_renewal_without_restart(ops_test: OpsTest) -> None:
```
juditnovak, line 102: Since we can't really check if there was a restart or not, I'd call this test `test_tls_renewal` or so.
As in reality, _that's_ what it is testsing

tests/unit/lib/test_opensearch_base_charm.py
```diff
@@ -289,6 +289,30 @@ def test_on_update_status(self, _, cert_expiration_remaining_hours, _stop_opense
             self.charm.on.update_status.emit()
             self.assertTrue(isinstance(self.harness.model.unit.status, BlockedStatus))
 
+    @patch(f"{BASE_LIB_PATH}.opensearch_tls.OpenSearchTLS.store_admin_tls_secrets_if_applies")
+    @patch(f"{BASE_CHARM_CLASS}.is_admin_user_configured")
+    @patch(f"{BASE_LIB_PATH}.opensearch_tls.OpenSearchTLS.is_fully_configured")
+    @patch(f"{BASE_LIB_PATH}.opensearch_tls.OpenSearchTLS.reload_tls_certificates")
+    def test_on_tls_conf_set(
+        self,
```
juditnovak, line 297: This function could have a name like `test_tls_configured_without_restart()` or so. As it _is_ verifying that!

lib/charms/opensearch/v0/opensearch_tls.py
reneradoi: Some context: I decided against extending the existing `request()` method because here is the only case where we need to authenticate against OpenSearch API with cert and key (instead of username/password).

If this should be added to the existing method, I'm fine with it. Would have been my approach as soon as there was another use case for cert/key authentication.

lib/charms/opensearch/v0/opensearch_config.py
reneradoi: This is handled in the base charm module with `if not admin_secrets.get("subject"):` (the one you commented on above).

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: The subject is checked because it's what is required to be put into the config in this case. This check makes sure that the `set_admin_tls_conf` in `opensearch_tls` has valid data to put into the config.

You're right, in general it would make sense to have a function that checks it and is then called multiple times throughout the code. Problem here is, that is changing what can be considered valid and it's also changing what is in the secret, both depending on the workflow. There might be the csr fist, then the cert and the ca and chain, later it also requires the subject and at a certain point also the keystore/truststore password. But all of these are different stages, so it will be pretty hard to make this valid at all times.

tests/integration/tls/test_tls.py
reneradoi: Done

tests/unit/lib/test_opensearch_base_charm.py
reneradoi: Done

Mehdi-Bendriss (commented):
Thanks René - looks good! I have a few questions

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -605,7 +602,7 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
             if self.unit.is_leader():
                 self.status.set(MaintenanceStatus(PluginConfigCheck), app=True)
 
-            if self.plugin_manager.run() and not restart_requested:
```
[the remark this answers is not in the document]
reneradoi: The intention to remove this was that there should not be a second restart anymore, because TLS workflow only reloads the certs via API.

lib/charms/opensearch/v0/opensearch_config.py
reneradoi: Done

lib/charms/opensearch/v0/opensearch_base_charm.py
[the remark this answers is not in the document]
reneradoi: The intention to remove the `restart_requested` variable is that there should not be a restart anymore on TLS certificate renewal (as the new certs are loaded via API), and therefore no need to set this marker.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -721,15 +715,17 @@ def on_tls_conf_set(
                 truststore_pwd=truststore_pwd,
                 keystore_pwd=keystore_pwd,
             )
-        else:
+
             # write the admin cert conf on all units, in case there is a leader loss + cert renewal
-            self.opensearch_config.set_admin_tls_conf(current_secrets)
+            if not admin_secrets.get("subject"):
```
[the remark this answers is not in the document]
reneradoi: Actually in the renewal flow, the subject should not change. It consists of the cluster name and "admin", and none of these should be updated here.

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -327,20 +327,17 @@ def _get_sans(self, cert_type: CertType) -> Dict[str, List[str]]:
...
 
-        dns = {self.charm.unit_name, socket.gethostname(), socket.getfqdn()}
+        # in order to be able to use the API for reloading TLS certificates it is necessary
+        # to only have one dns name and one ip address in the sans
+        # otherwise the following upstream bug will mess the sorting of these fields
+        # https://github.com/opensearch-project/security/issues/4480
+        dns = {socket.getfqdn()}
         ips = {self.charm.unit_ip}
 
-        host_public_ip = get_host_public_ip()
```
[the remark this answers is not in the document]
reneradoi: Done, issue is here: https://github.com/canonical/opensearch-operator/issues/382

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -430,7 +427,7 @@ def _create_keystore_pwd_if_not_exists(self, scope: Scope, cert_type: CertType,
         if secrets:
             store_pwd = secrets.get(f"{store_type}-password")
 
-        if not store_pwd:
+        if not store_pwd and not self.charm.opensearch_peer_cm.is_consumer(of="main"):
```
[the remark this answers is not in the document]
reneradoi: Reason for this: on `CertificateAvailable`, the keystore password could be created if it does not exist yet. If this happens before the relation data is updated into the secret, the secret does not get populated anymore with data from the main orchestrator. I saw this fail in various test runs, and there is actually a test failure every now and then in our main branch, where there seems to be a wrong keystore password. I did not understand why this could happen for quite some time, but tracked this down to this now.

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -621,3 +618,52 @@ def delete_stored_tls_resources(self):
...
+
+    def reload_tls_certificates(self):
+        """Reload transport and HTTP layer communication certificates via REST APIs."""
+        url_http = f"https://{self.charm.unit_ip}:9200/_plugins/_security/api/ssl/http/reloadcerts"
+        url_transport = (
+            f"https://{self.charm.unit_ip}:9200/_plugins/_security/api/ssl/transport/reloadcerts"
+        )
+
+        # using the SSL API requires authentication with app-admin cert and key
+        admin_secret = self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val) or {}
```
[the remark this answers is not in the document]
reneradoi: Done

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -621,3 +618,52 @@ def delete_stored_tls_resources(self):
...
+
+        if not admin_secret:
```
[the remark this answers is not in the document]
reneradoi: You're right, this should not happen.

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -621,3 +618,52 @@ def delete_stored_tls_resources(self):
...
+            )
+            response_transport = requests.put(
+                url_transport,
+                cert=(admin_cert, admin_key),
+                verify=f"{self.certs_path}/chain.pem",
+                timeout=5,
+            )
+            if not (
+                response_http.status_code == requests.codes.ok
+                and response_transport.status_code == requests.codes.ok
```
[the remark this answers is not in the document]
reneradoi: I tried to find a reason why to do so but couldn't that's why I decided to keep it simple and not add complexity here.

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -621,3 +618,52 @@ def delete_stored_tls_resources(self):
...
+
+        if not admin_secret:
+            logger.error("Can not update TLS certs, no app-admin secret found.")
+            return
+
+        admin_cert = f"{self.certs_path}/admin.cert"
+        admin_key = f"{self.certs_path}/admin.key"
+
+        with open(admin_cert, "w") as cert:
+            cert.write(admin_secret["cert"])
```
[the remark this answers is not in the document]
reneradoi: agree

lib/charms/opensearch/v0/opensearch_tls.py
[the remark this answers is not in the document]
reneradoi: > the above condition can be simplified to self.charm.opensearch_peer_cm.is_provider(typ="main")

Is this really the same like `not self.charm.opensearch_peer_cm.is_consumer(of="main")`, also in cases where there is no peer-cluster-relation?

lib/charms/opensearch/v0/opensearch_base_charm.py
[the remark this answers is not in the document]
Mehdi-Bendriss: Exactly - so in a renewal: `admin_secrets.get("subject")` will not be None, and will proceed with the flow as if the "new" admin certificates were already configured - when in fact only the old ones are still stored.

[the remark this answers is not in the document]
reneradoi: But this if-condition only refers to setting the config for `plugins.security.authcz.admin_dn` in opensearch.yml. The admin certificates are stored anyway with `self.tls.store_admin_tls_secrets_if_applies()`, independent of the if-condition for `admin_secrets.get("subject")`.

Or am I getting it wrong what you are referring to?

lib/charms/opensearch/v0/opensearch_tls.py
juditnovak: Yes. The "pro" side was
 - extension on the existing, Opensearch internal `request()` method is not much
 - thus we have a homogenioius usage of `request()` across the Opensearch code.
 
 I'm fine with either solution, it's just worth a thought why we choose the one we do :-)

reneradoi: Mehdi and I agreed on your proposal to extend the existing `request()` method for using certs instead of user/password for authentication.

lib/charms/opensearch/v0/opensearch_base_charm.py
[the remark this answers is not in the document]
reneradoi: The availability of updated certs is checked via `is_fully_configured()`, so this is solved.

lib/charms/opensearch/v0/opensearch_tls.py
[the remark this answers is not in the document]
reneradoi: Added a check for `CertType.APP_ADMIN` to make sure this is done for the `unit-http` and `unit-transport` cases on the consumer-units of large deployments.

juditnovak (approved):
LGTM, thanks for the live sync we had over this code :-)

juditnovak:
Just to confirm: test failures are over unrelated matters? (If not they should be greeen...)

Mehdi-Bendriss (approved):

PR #381, opened 2024-08-02
Fix to support charmcraft 3.1.1

reneradoi:
## Issue
With the latest version 3.1.1 of charmcraft, build and deploying charms fails if there is a prime part in the `charmcraft.yaml` file: `Warning: use of 'prime' in a charm part is deprecated and no longer works, see https://juju.is/docs/sdk/include-extra-files-in-a-charm`.

## Solution
Fix the charmcraft.yaml config file.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -10,6 +10,11 @@ bases:
       - name: "ubuntu"
         channel: "22.04"
 parts:
+  files:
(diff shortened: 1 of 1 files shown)


Mehdi-Bendriss (approved):

juditnovak (approved):
LGTM

PR #387, opened 2024-08-06
[DPE-4931] fix locking with unassigned shards

reneradoi:
## Issue
This PR addresses the issues https://github.com/canonical/opensearch-operator/issues/327 and https://github.com/canonical/opensearch-operator/issues/324.

When Opensearch is in the process of shutting down, the operator currently does not wait for data to be moved away from the stopping unit. This may result in shards not being assigned and could cause loss of data. In cases where the index `.charm_node_lock` is impacted, the operator can no longer acquire the lock to start or stop Opensearch. This will result in `503` errors in the logfile.

The behavior can be seen in this CI run, with some additional logging information added for debugging: https://github.com/canonical/opensearch-operator/actions/runs/10269420444/job/28415611890?pr=387
```
Shards before relocation: [...{'index': '.charm_node_lock', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': '1', 'store': '31.6kb', 'ip': '10.19.29.219', 'node': 'opensearch-0.e42'}, {'index': '.charm_node_lock', 'shard': '0', 'prirep': 'r', 'state': 'STARTED', 'docs': '1', 'store': '8.7kb', 'ip': '10.19.29.239', 'node': 'opensearch-1.e42'}]

Shards after relocation: [... {'index': '.opendistro_security', 'shard': '0', 'prirep': 'p', 'state': 'RELOCATING', 'docs': '10', 'store': '61.5kb', 'ip': '10.19.29.219', 'node': 'opensearch-0.e42 -> 10.19.29.239 yt3jiuSZRTCY8NnoeIni5w opensearch-1.e42'}, {'index': '.charm_node_lock', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': '1', 'store': '31.6kb', 'ip': '10.19.29.219', 'node': 'opensearch-0.e42'}]
```

Shortly after, the error is there:
```
unit-opensearch-0: 16:25:11 DEBUG unit.opensearch/0.juju-log opensearch-peers:1: https://10.19.29.239:9200 "GET /.charm_node_lock/_source/0 HTTP/11" 503 287
unit-opensearch-0: 16:25:11 ERROR unit.opensearch/0.juju-log opensearch-peers:1: Error checking which unit has OpenSearch lock
...
File "/var/lib/juju/agents/unit-opensearch-0/charm/lib/charms/opensearch/v0/opensearch_locking.py", line 263, in acquired
    unit = self._unit_with_lock(host)
  File "/var/lib/juju/agents/unit-opensearch-0/charm/lib/charms/opensearch/v0/opensearch_locking.py", line 225, in _unit_with_lock
    document_data = self._opensearch.request(
  File "/var/lib/juju/agents/unit-opensearch-0/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 306, in request
    raise OpenSearchHttpError(
charms.opensearch.v0.opensearch_exceptions.OpenSearchHttpError: HTTP error self.response_code=503
self.response_body={'error': {'root_cause': [{'type': 'no_shard_available_action_exception', 'reason': 'No shard available for [get [.charm_node_lock][0]: routing [null]]'}], 'type': 'no_shard_available_action_exception', 'reason': 'No shard available for [get [.charm_node_lock][0]: routing [null]]'}, 'status': 503}
unit-opensearch-0: 16:25:11 DEBUG unit.opensearch/0.juju-log opensearch-peers:1: Lock to start opensearch not acquired. Will retry next event
```

## Solution
When stopping Opensearch, the operator should wait for the shards relocation to be completed. This should be happening right after adding the currently stopping unit to the allocation exclusions. The check should be blocking, meaning that Opensearch must not stop until the relocation is finished.

This will look something like this:
```
unit-opensearch-0: 10:07:28 DEBUG unit.opensearch/0.juju-log Shards before relocation: [... {'index': '.plugins-ml-config', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': '1', 'store': '3.9kb', 'ip': '10.18.168.228', 'node': 'opensearch-0.ccc'}, ...]

[...]

unit-opensearch-0: 10:07:32 DEBUG unit.opensearch/0.juju-log Shards after relocation: [... {'index': '.plugins-ml-config', 'shard': '0', 'prirep': 'p', 'state': 'STARTED', 'docs': '1', 'store': '3.9kb', 'ip': '10.18.168.28', 'node': 'opensearch-1.ccc'} ...]
```
To check if there are still some moving shards, the API `_cluster/health` can be queried for `"relocating_shards"`. If these are not `0`, the process of stopping should be halted. Depending on the amount of data, this can take quite some time. A reasonable maximum waiting time of 15 minutes has been added, after that an error will be raised.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_health.py
+++ b/lib/charms/opensearch/v0/opensearch_health.py
@@ -115,7 +115,7 @@ class OpenSearchHealth:
         # we differentiate between a temp yellow (moving shards) and a permanent
         # one (such as: missing replicas)
         if status in [HealthColors.GREEN, HealthColors.YELLOW] and (
-                response["initializing_shards"] > 0 or response["relocating_shards"] > 0
+            response["initializing_shards"] > 0 or response["relocating_shards"] > 0
         ):
             try:
                 logger.debug(
--- a/lib/charms/opensearch/v0/opensearch_health.py
+++ b/lib/charms/opensearch/v0/opensearch_health.py
@@ -112,47 +112,33 @@ class OpenSearchHealth:
             logger.error(e)  # means the status was reported as an int (i.e: 503)
             return HealthColors.UNKNOWN
 
-        if status != HealthColors.YELLOW:
-            return status
-
-        try:
-            logger.debug(
-                f"\n\nHealth: {status} -- Shards: {ClusterState.shards(self._opensearch, host, verbose=True)}\n\n"
-            )
-            logger.debug(
-                f"Allocation explanations: {ClusterState.allocation_explain(self._opensearch, host)}\n\n"
-            )
-        except OpenSearchHttpError:
-            pass
-
         # we differentiate between a temp yellow (moving shards) and a permanent
         # one (such as: missing replicas)
-        if response["initializing_shards"] > 0 or response["relocating_shards"] > 0:
+        if status in [HealthColors.GREEN, HealthColors.YELLOW] and (
+                response["initializing_shards"] > 0 or response["relocating_shards"] > 0
+        ):
+            try:
+                logger.debug(
+                    f"\n\nHealth: {status} -- Shards: {ClusterState.shards(self._opensearch, host, verbose=True)}\n\n"
+                )
+                logger.debug(
+                    f"Allocation explanations: {ClusterState.allocation_explain(self._opensearch, host)}\n\n"
+                )
+            except OpenSearchHttpError:
+                pass
             return HealthColors.YELLOW_TEMP
-        return HealthColors.YELLOW
+
+        return status
 
     @retry(stop=stop_after_attempt(90), wait=wait_fixed(10), reraise=True)
     def wait_for_shards_relocation(self) -> None:
         """Blocking function until the shards relocation completes in the cluster."""
-        endpoint = "/_cluster/health"
+        health = self.get(local_app_only=False)
 
-        try:
-            response = self._opensearch.request(
-                "GET",
-                endpoint,
-                host=self._charm.unit_ip,
-                alt_hosts=self._charm.alt_hosts,
-            )
-        except OpenSearchHttpError:
-            logger.error("Error while waiting for shard relocation to complete")
-            raise
-
-        # we throw an error because various operations should NOT start while data
-        # is being relocated. Examples are: simple stop, unit removal, upgrade
-        if not response["relocating_shards"] == 0:
-            logger.info(
-                f"Shards still moving before stopping Opensearch: relocating: {response['relocating_shards']}"
-            )
+        if health == HealthColors.YELLOW_TEMP:
+            logger.info("Shards still moving before stopping Opensearch.")
+            # we throw an error because various operations should NOT start while data
+            # is being relocated. Examples are: simple stop, unit removal, upgrade
             raise OpenSearchHAError("Shards haven't completed relocating.")
 
     def _apply_for_app(self, status: str) -> None:
--- a/lib/charms/opensearch/v0/opensearch_health.py
+++ b/lib/charms/opensearch/v0/opensearch_health.py
@@ -20,6 +20,7 @@ from charms.opensearch.v0.opensearch_exceptions import (
     OpenSearchHttpError,
 )
 from ops.model import BlockedStatus, MaintenanceStatus, WaitingStatus
+from tenacity import retry, stop_after_attempt, wait_fixed
 
 # The unique Charmhub library identifier, never change it
 LIBID = "93d2c27f38974a59b3bbe39fb27ac98d"
@@ -130,19 +131,20 @@ class OpenSearchHealth:
             return HealthColors.YELLOW_TEMP
         return HealthColors.YELLOW
 
+    @retry(stop=stop_after_attempt(15), wait=wait_fixed(5), reraise=True)
     def wait_for_shards_relocation(self) -> None:
         """Blocking function until the shards relocation completes in the cluster."""
-        endpoint = "/_cluster/health?wait_for_no_relocating_shards=true"
+        endpoint = "/_cluster/health"
         timeout = 10
 
         try:
-            relocation_complete = self._opensearch.request(
+            response = self._opensearch.request(
                 "GET",
                 endpoint,
                 host=self._charm.unit_ip,
                 alt_hosts=self._charm.alt_hosts,
                 timeout=timeout,
-                retries=12,
+                retries=3,
             )
         except OpenSearchHttpError:
             logger.error("Error while waiting for shard relocation to complete")
@@ -150,7 +152,7 @@ class OpenSearchHealth:
 
         # we throw an error because various operations should NOT start while data
         # is being relocated. Examples are: simple stop, unit removal, upgrade
-        if not relocation_complete:
+        if not (response["unassigned_shards"] == 0 and response["relocating_shards"] == 0):
             raise OpenSearchHAError("Shards haven't completed relocating.")
 
     def _apply_for_app(self, status: str) -> None:
--- a/lib/charms/opensearch/v0/opensearch_health.py
+++ b/lib/charms/opensearch/v0/opensearch_health.py
@@ -20,7 +20,6 @@ from charms.opensearch.v0.opensearch_exceptions import (
     OpenSearchHttpError,
 )
 from ops.model import BlockedStatus, MaintenanceStatus, WaitingStatus
-from tenacity import retry, stop_after_attempt, wait_fixed
 
 # The unique Charmhub library identifier, never change it
 LIBID = "93d2c27f38974a59b3bbe39fb27ac98d"
@@ -131,15 +130,28 @@ class OpenSearchHealth:
             return HealthColors.YELLOW_TEMP
         return HealthColors.YELLOW
 
-    @retry(stop=stop_after_attempt(15), wait=wait_fixed(5), reraise=True)
     def wait_for_shards_relocation(self) -> None:
         """Blocking function until the shards relocation completes in the cluster."""
-        if self.get(wait_for_green_first=True) != HealthColors.YELLOW_TEMP:
-            return
+        endpoint = "/_cluster/health?wait_for_no_relocating_shards=true"
+        timeout = 10
+
+        try:
+            relocation_complete = self._opensearch.request(
+                "GET",
+                endpoint,
+                host=self._charm.unit_ip,
+                alt_hosts=self._charm.alt_hosts,
+                timeout=timeout,
+                retries=12,
+            )
+        except OpenSearchHttpError:
+            logger.error("Error while waiting for shard relocation to complete")
+            raise
 
         # we throw an error because various operations should NOT start while data
         # is being relocated. Examples are: simple stop, unit removal, upgrade
-        raise OpenSearchHAError("Shards haven't completed relocating.")
+        if not relocation_complete:
+            raise OpenSearchHAError("Shards haven't completed relocating.")
 
     def _apply_for_app(self, status: str) -> None:
         """Cluster wide / app status."""
--- a/lib/charms/opensearch/v0/opensearch_health.py
+++ b/lib/charms/opensearch/v0/opensearch_health.py
@@ -149,12 +149,9 @@ class OpenSearchHealth:
 
         # we throw an error because various operations should NOT start while data
         # is being relocated. Examples are: simple stop, unit removal, upgrade
-        if not (response["unassigned_shards"] == 0 and response["relocating_shards"] == 0):
+        if not response["relocating_shards"] == 0:
             logger.info(
-                f"""Some shards are still moving before stopping Opensearch: \
-                relocating: {response['relocating_shards']},
-                unassigned: {response['unassigned_shards']}
-                """
+                f"Shards still moving before stopping Opensearch: relocating: {response['relocating_shards']}"
             )
             raise OpenSearchHAError("Shards haven't completed relocating.")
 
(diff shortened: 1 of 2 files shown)


lib/charms/opensearch/v0/opensearch_health.py
```diff
@@ -131,15 +131,29 @@ def get(  # noqa: C901
...
+                alt_hosts=self._charm.alt_hosts,
+            )
+        except OpenSearchHttpError:
+            logger.error("Error while waiting for shard relocation to complete")
+            raise
 
         # we throw an error because various operations should NOT start while data
         # is being relocated. Examples are: simple stop, unit removal, upgrade
-        raise OpenSearchHAError("Shards haven't completed relocating.")
+        if not response["relocating_shards"] == 0:
```
phvalguima, line 152: There are more status that are relevant here, [check the opensearch code](https://github.com/opensearch-project/OpenSearch/blob/a918530397e3e70d5171d879b2238b3005e5c66f/server/src/main/java/org/opensearch/cluster/health/ClusterShardHealth.java#L68).

So, this check is the same as [`self.opensearch.health() != YELLOW_TEMP`](https://github.com/canonical/opensearch-operator/blob/a48f9c2b504db4170c18060a2eec639869c0107e/lib/charms/opensearch/v0/opensearch_health.py#L130), just adding the `Retrying`. And that does not cover all cases. We may have unassigned replicas that WILL move to other nodes (i.e. unassigned > relocating > active) and unassigned that will get stuck as unassigned as there is not enough resources or nodes to move them around.

We also should make this distinction...

First, we are missing a good set of information that comes from [/_cluster/allocation/explain](https://opensearch.org/docs/latest/api-reference/cluster-api/cluster-allocation/). This API provides a json informing how each replica looks like, and if that replica is unassigned, if there are any allocation rules blocking it from actually being assigned.

You can check one example:
* Replicas' status: https://pastebin.ubuntu.com/p/gPcDWvKTj5/
* Explain details: https://pastebin.ubuntu.com/p/vxs6rVndBX/

Also notice there is an "unassigned.reason" field in the replica status, [which comes from here](https://github.com/opensearch-project/OpenSearch/blob/a918530397e3e70d5171d879b2238b3005e5c66f/server/src/main/java/org/opensearch/cluster/routing/UnassignedInfo.java#L93).

It is worthy checking if we can have replicas that are "unassigned" but they are just waiting to be placed. I think we can either explore the `unassigned.reason` or the `explain` API above.

So, in my opinion, we check for health:
* GREEN: just return, we are done
* RED: throw an error, we have shards that are irretrievable
* Any other status, next step

For the last point, we want to look for:
1) Shards that are relocating, we just wait for them
2) Shards that are unassigned, BUT "cluster/explain" returns that that shard has all "YES" in its allocation rules -> we just wait for them
3) Shards that are unassigned, do one of the following at least: (1) "cluster/explain" returns at least one "NO": there is no way this replica will be allocated, we throw an error here; OR (2) check "unassigned.reason" and decide which replicas can be placed where

There is one point I am not sure thou. Shards that have, e.g. `replica: 0-all`, they will have replicas marked as "unassigned" for some time, and then these replicas will go away as a node goes away. I am not sure how they look in `reason` during that time.

phvalguima (commented):
Hey @reneradoi thanks for this PR.
Unfortunately, I don't think we are quite there yet. I left an explanation on how I see we should dig further in the replica allocation problems.

I think we also should move these logic to `opensearch_health` and get more details out of the APIs to decide if we have a cluster in `YELLOW` or `YELLOW_TEMP`.

In a nutshell, if the cluster is in `YELLOW_TEMP`, we can wait. Otherwise, we have a problem and the user must intervene.

lib/charms/opensearch/v0/opensearch_health.py
phvalguima: In summary: I propose to have a simple logic with a Retrying loop and we keep looping while the cluster is in `YELLOW_TEMP` status.

And we should extend the `health` check for YELLOW_TEMP to also validate the reasons of why a given replica is unassigned and decide if that is something we can wait on or not.

Mehdi-Bendriss: @reneradoi I believe what we had was covering the case we were interested in - with the exception of the Retry schedule which should be similar to the one.
Relocating / Initializing shards => Yellow_temp

reneradoi: Hey guys, thank you for the feedback. I saw the logic that was in place already (just not called), the thing here is: it does not work. I've tested with this for a lot of runs, locally and also on CI, and result was always the same: Cluster health is green even though shards are obviously relocating. This can be seen [here for example](https://github.com/canonical/opensearch-operator/actions/runs/10269420444/job/28415611890?pr=387#step:30:17534):

```
unit-opensearch-0: 16:25:07 DEBUG unit.opensearch/0.juju-log opensearch-peers:1: https://10.19.29.219:9200/ "GET /_cluster/health?wait_for_status=green&timeout=1m HTTP/11" 200 450
unit-opensearch-1: 16:25:07 DEBUG unit.opensearch/1.juju-log opensearch-peers:1: Getting secret app:admin-password
unit-opensearch-0: 16:25:07 INFO unit.opensearch/0.juju-log opensearch-peers:1: Health: {'cluster_name': 'opensearch-u6wp', 'status': 'green', 'timed_out': False, 'number_of_nodes': 3, 'number_of_data_nodes': 3, 'discovered_master': True, 'discovered_cluster_manager': True, 'active_primary_shards': 5, 'active_shards': 5, 'relocating_shards': 2, 'initializing_shards': 0, 'unassigned_shards': 0, 'delayed_unassigned_shards': 0, 'number_of_pending_tasks': 0, 'number_of_in_flight_fetch': 0, 'task_max_waiting_in_queue_millis': 0, 'active_shards_percent_as_number': 100.0}

[...]

unit-opensearch-0: 16:25:09 DEBUG unit.opensearch/0.juju-log opensearch-peers:1: https://10.19.29.239:9200/ "GET /.charm_node_lock/_source/0 HTTP/11" 503 287
unit-opensearch-0: 16:25:09 ERROR unit.opensearch/0.juju-log opensearch-peers:1: Request GET to https://10.19.29.239:9200/.charm_node_lock/_source/0 with payload: None failed.(Attempts left: 2)
	Error: 503 Server Error: Service Unavailable for url: https://10.19.29.239:9200/.charm_node_lock/_source/0
```

This behavior has only been the reason why I started looking for other solutions. Another one that did not work was using the parameter `wait_for_no_relocating_shards` of the `_cluster/health` API described [here](https://opensearch.org/docs/latest/api-reference/cluster-api/cluster-health/#query-parameters).

The solution I've proposed here is the only one I found so far that is actually waiting until the shards relocation is completed.

And considering the use case we have here, I think this solution might be pretty simple, yet effective: wait for Opensearch to finish moving all the shards, then shutting down. You can see this very nicely in [this CI run](https://github.com/canonical/opensearch-operator/actions/runs/10285672858/job/28465550386) by searching the log for this message: `Shards still moving before stopping Opensearch`.

I'm not sure if this behavior (health green despite relocating shards) is an upstream bug, but I think we need a solution that works for us, and we need it soon.

phvalguima: That is interesting!

It means green is already considering `relocating_shards` then. I am trying to get the source of health and it seems the path for `YELLOW` is:

[`ClusterStateHealth`](https://github.com/opensearch-project/OpenSearch/blob/348c04e7a32e13ea040a1d2e0459c03da9ec0c2c/server/src/main/java/org/opensearch/cluster/health/ClusterStateHealth.java#L117) > [`ClusterIndexHealth`](https://github.com/opensearch-project/OpenSearch/blob/348c04e7a32e13ea040a1d2e0459c03da9ec0c2c/server/src/main/java/org/opensearch/cluster/health/ClusterIndexHealth.java#L179) > [`ClusterShardHealth`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/health/ClusterShardHealth.java#L135-L143) being `YELLOW`.

Now, shards are yellow if: (1) primary is active and active replica count is equal to total # of replicas, [even if they are relocating](https://github.com/opensearch-project/OpenSearch/blob/348c04e7a32e13ea040a1d2e0459c03da9ec0c2c/server/src/main/java/org/opensearch/cluster/health/ClusterShardHealth.java#L122); or (2) primary is inactive [there is a "NO" in one of the rounting deciders, but no failed previous allocations and there is chance to recover the replica](https://github.com/opensearch-project/OpenSearch/blob/348c04e7a32e13ea040a1d2e0459c03da9ec0c2c/server/src/main/java/org/opensearch/cluster/health/ClusterShardHealth.java#L240-L245).

Indeed, [initializing replicas do not count as "active"](https://github.com/opensearch-project/OpenSearch/blob/348c04e7a32e13ea040a1d2e0459c03da9ec0c2c/server/src/main/java/org/opensearch/cluster/health/ClusterShardHealth.java#L128).

So, that status of `GREEN` with relocating shards is accurate, according to the code.

phvalguima: I think, in this case, it means we should update our `opensearch_health` logic to evaluate for `YELLOW_TEMP` *even* if the cluster is green.

Something like this [here](https://github.com/canonical/opensearch-operator/blob/a48f9c2b504db4170c18060a2eec639869c0107e/lib/charms/opensearch/v0/opensearch_health.py#L77):
```
    def get(  # noqa: C901
        self,
        wait_for_green_first: bool = False,
        use_localhost: bool = True,
        local_app_only: bool = True,
    ) -> str:


    ...


        if status == HealthColors.GREEN and (response["initializing_shards"] > 0 or response["relocating_shards"] > 0):
            return HealthColors.YELLOW_TEMP
        if status != HealthColors.YELLOW:
            return status
```

phvalguima:
BTW: I just caught a weird issue that seems related to all our discussions about cluster health. In a nutshell, [`voting exclusions` PR has this idea of checking for `YELLOW_TEMP` at that very same point, within a Retry loop](https://github.com/canonical/opensearch-operator/blob/c3b71e7fd46984e5838785252284743042141826/lib/charms/opensearch/v0/opensearch_base_charm.py#L1055). And I noticed that occasionally, it fails with an odd error. I managed to reproduce it locally and essentially seems to be the same problem as @reneradoi pointed out.

One of the shards simply get unassigned and unreachable:
```
.plugins-ml-config               0 p STARTED     1  3.9kb 10.115.236.56 opensearch-0.799               
.opensearch-observability        0 p STARTED     0   208b 10.115.236.56 opensearch-0.799               
.opensearch-sap-log-types-config 0 p STARTED              10.115.236.56 opensearch-0.799               
.opendistro_security             0 p STARTED    10 61.3kb 10.115.236.56 opensearch-0.799               
.charm_node_lock                 0 p UNASSIGNED        
```

More details: https://pastebin.ubuntu.com/p/GS7kDgMNjn/, which started right after `node-left` with that replica:
```
Aug 07 18:45:24 juju-b5ca37-1 opensearch.daemon[12200]: [2024-08-07T18:45:24,622][INFO ][o.o.c.s.MasterService    ] [opensearch-0.799] node-left[{opensearch-1.799}{fa75fSUbTbCCAXfLplLieA}{qhO8zkoPSIm-GaWsSyyZRA}{10.115.236.191}{10.115.236.191:9300}{coordinating_onlydimml}{shard_indexing_pressure_enabled=t>

...

Aug 07 18:45:26 juju-b5ca37-1 opensearch.daemon[12200]: org.opensearch.action.NoShardAvailableActionException: No shard available for [get [.charm_node_lock][0]: routing [null]]
Aug 07 18:45:26 juju-b5ca37-1 opensearch.daemon[12200]:         at org.opensearch.action.support.single.shard.TransportSingleShardAction$AsyncSingleAction.perform(TransportSingleShardAction.java:256) [opensearch-2.14.0.jar:2.14.0]
Aug 07 18:45:26 juju-b5ca37-1 opensearch.daemon[12200]:         at org.opensearch.action.support.single.shard.TransportSingleShardAction$AsyncSingleAction.start(TransportSingleShardAction.java:231) [opensearch-2.14.0.jar:2.14.0]
Aug 07 18:45:26 juju-b5ca37-1 opensearch.daemon[12200]:         at org.opensearch.action.support.single.shard.TransportSingleShardAction.doExecute(TransportSingleShardAction.java:125) [opensearch-2.14.0.jar:2.14.0]
Aug 07 18:45:26 juju-b5ca37-1 opensearch.daemon[12200]:         at org.opensearch.action.support.single.shard.TransportSingleShardAction.doExecute(TransportSingleShardAction.java:78) [opensearch-2.14.0.jar:2.14.0]
```

lib/charms/opensearch/v0/opensearch_health.py
reneradoi: I'm not sure if it's a good idea to "overwrite" the cluster health from green to yellow for ourselves, as this might be highly confusing to anyone using the operator. If they see yellow(.temp) in the logs when something does not go as they expected, but see the same thing green in Opensearch, they might consider this a bug (at least I would).

phvalguima: I am suggesting override `GREEN` to (non-existent in opensearch anyways) `YELLOW_TEMP`, which we used to do only for YELLOW before.

phvalguima: On a side note, we could rename `HealthColors` To `HealthStatus` and move  `YELLOW_TEMP` to `RELOCATING_SHARDS`. That would be more clear for anyone else using the operator. In any case, `YELLOW_TEMP` is exclusively our operator creation concept.

phvalguima (changes requested):
I am testing with the changes to `OpenSearchHealth` where `YELLOW_TEMP` is being checked if we have a green or yellow cluster, it did improve my last CI run.

Current work on this PR: https://github.com/canonical/opensearch-operator/pull/367

phvalguima (approved):
LGTM

Really nice digging @reneradoi!

Mehdi-Bendriss (approved):

PR #391, opened 2024-08-08
[DPE-4935] no allocation exclusions on restart

reneradoi:
## Issue
When an Opensearch node is restarted by the operator, currently the shards assigned to this node are first moved away from the node, and after restarting moved back. This is a lot of work that, in case of simple restarts, is not necessary.

To avoid this behavior, no allocation exclusion should be added for the departing node, when a restart operation is performed.

## Solution
Check for the `restart` flag in `_stop_opensearch()` and pass it through to `opensearch_exclusions.add_current()`. If set to `True`, only the voting exclusion should be added for the departing node, but no allocation exclusion.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -1015,10 +1015,8 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
                 # otherwise cluster manager election will be blocked when starting up again
                 # and re-using storage
                 if len(nodes) > 1:
-                    # TODO: we should probably NOT have any exclusion on restart
-                    # https://chat.canonical.com/canonical/pl/bgndmrfxr7fbpgmwpdk3hin93c
                     # 1. Add current node to the voting + alloc exclusions
-                    self.opensearch_exclusions.add_current()
+                    self.opensearch_exclusions.add_current(restart=restart)
             except OpenSearchHttpError:
                 logger.debug("Failed to get online nodes, voting and alloc exclusions not added")
 
@@ -1030,8 +1028,6 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
         self.status.set(WaitingStatus(ServiceStopped))
 
         # 3. Remove the exclusions
-        # TODO: we should probably NOT have any exclusion on restart
-        # https://chat.canonical.com/canonical/pl/bgndmrfxr7fbpgmwpdk3hin93c
         if not restart:
             try:
                 self.opensearch_exclusions.delete_current()
--- a/lib/charms/opensearch/v0/opensearch_health.py
+++ b/lib/charms/opensearch/v0/opensearch_health.py
@@ -3,6 +3,7 @@
 
 """Base class for the OpenSearch Health management."""
 import logging
+import time
 from typing import Dict, Optional
 
 from charms.opensearch.v0.constants_charm import (
@@ -133,6 +134,8 @@ class OpenSearchHealth:
     @retry(stop=stop_after_attempt(90), wait=wait_fixed(10), reraise=True)
     def wait_for_shards_relocation(self) -> None:
         """Blocking function until the shards relocation completes in the cluster."""
+        time.sleep(5)
+
         health = self.get(local_app_only=False)
 
         if health == HealthColors.YELLOW_TEMP:
--- a/lib/charms/opensearch/v0/opensearch_health.py
+++ b/lib/charms/opensearch/v0/opensearch_health.py
@@ -131,7 +131,7 @@ class OpenSearchHealth:
 
         return status
 
-    @retry(stop=stop_after_attempt(90), wait=wait_fixed(10), reraise=True)
+    @retry(stop=stop_after_attempt(90), wait=wait_fixed(5), reraise=True)
     def wait_for_shards_relocation(self) -> None:
         """Blocking function until the shards relocation completes in the cluster."""
         time.sleep(5)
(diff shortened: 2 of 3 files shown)


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1015,10 +1015,8 @@ def _stop_opensearch(self, *, restart=False) -> None:
                 # otherwise cluster manager election will be blocked when starting up again
                 # and re-using storage
                 if len(nodes) > 1:
-                    # TODO: we should probably NOT have any exclusion on restart
-                    # https://chat.canonical.com/canonical/pl/bgndmrfxr7fbpgmwpdk3hin93c
                     # 1. Add current node to the voting + alloc exclusions
-                    self.opensearch_exclusions.add_current()
+                    self.opensearch_exclusions.add_current(restart=restart)
```
Mehdi-Bendriss, line 1019: should we perhaps keep it simple and directly test on `if not restart and len(nodes) > 1`?

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -44,13 +44,14 @@ def __init__(self, charm):
...
 
-    def add_current(self) -> None:
+    def add_current(self, restart=False) -> None:
         """Add Voting and alloc exclusions."""
         if (self._node.is_cm_eligible() or self._node.is_voting_only()) and not self._add_voting():
             logger.error(f"Failed to add voting exclusion: {self._node.name}.")
 
-        if self._node.is_data() and not self._add_allocations():
-            logger.error(f"Failed to add shard allocation exclusion: {self._node.name}.")
+        if not restart:
```
phvalguima, line 52: This can be "dangerous" as we may lose non-replicated shards.

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: Oh, I thought the voting exclusions were supposed to be added still even if it was a restart? That's why I've added the `restart` flag, to be able to distinguish between voting and allocation exclusions in `add_current()`.

phvalguima (commented):
@reneradoi my remark here is that `allocation_exclusions` allows us to move the shards across the nodes *before* stop / starting. That is particularly relevant if we have non-replicated shards (i.e. indices that are either configured with no replicas or, for some reason, had not enough time to replicate across the nodes). In this case, the cluster will be `red` while the given node stops and restarts.

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: you're right! I overlooked that, thanks!

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
Mehdi-Bendriss: You're right, but I believe it may be an acceptable risk in general - from a similar section on the [elastic docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/rolling-upgrades.html#_upgrading_your_cluster):

> When you shut down a data node, the allocation process waits for index.unassigned.node_left.delayed_timeout (by default, one minute) before starting to replicate the shards on that node to other nodes in the cluster, **which can involve a lot of I/O. Since the node is shortly going to be restarted, this I/O is unnecessary.** You can avoid racing the clock by [disabling allocation](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/modules-cluster.html#cluster-routing-allocation-enable) of replicas before shutting down [data nodes](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/modules-node.html#data-node):

lib/charms/opensearch/v0/opensearch_health.py
```diff
@@ -133,6 +134,8 @@ def get(  # noqa: C901
     @retry(stop=stop_after_attempt(90), wait=wait_fixed(10), reraise=True)
     def wait_for_shards_relocation(self) -> None:
         """Blocking function until the shards relocation completes in the cluster."""
+        time.sleep(5)
```
reneradoi, line 137: I have added this change to the `wait_for_shards_relocation` method because I saw error 503 still occurring randomly in integration tests. My assumption is, that checking the shards relocation immediately after adding the allocation exclusion might be to fast for Opensearch, and relocation may not have been started yet. Tbh. five seconds is just a guess, but I think in the case of shutdown, this extra wait is acceptable.

Mehdi-Bendriss: I'm not sure it will help, given we already have the `wait_fixed(10)` in the tenacity decorator.

reneradoi: I'm afraid the `wait` only applies on retries. If the health is green on the first attempt (because it hasn't started relocating yet), it will go through without waiting.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -44,13 +44,14 @@ def __init__(self, charm):
 
         self._scope = Scope.APP if self._charm.unit.is_leader() else Scope.UNIT
 
-    def add_current(self) -> None:
+    def add_current(self, restart=False) -> None:
```
phvalguima, line 47: nit: type checking

phvalguima (changes requested):

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #392, opened 2024-08-08
[DPE-5125] Remove peer relation lock

phvalguima:
This PR separates the lock-related changes from the DPE-4575.

The main goal is to ensure we are not unnecessarily requesting for the lock for a single unit.

## Removing the peer lock

There are two situations that could justify the use of peer lock:

### Use Case 1: moving from 1 unit up while that single unit is powered off

1) 1x unit opensearch (opensearch/0) working as intended
2) A restart is triggered in opensearch/0 -> it sets the .charm-node-lock as taken
3) While opensearch/0 is down, opensearch/{1-X} are added
4) Given opensearch/0 is the main leader, and the elected manager, units {1-X} won't be able to proceed on their process
5) opensearch/0 comes back and cleans its own .charm-node-lock
6) Units {1-X} can start their process and start to start

### Use Case 2: a cluster with X elegible managers, where a subset is powered down already

In this case, we may have two situations: (i) we have a minimal quorum of nodes powered up - hence the other nodes can pick the .charm-node-lock one by one and no need for the peer relation here; or  (ii) we powered down nodes to less than the minimal quorum: in this case, the cluster is essentially stopped and unresponsive.

For case (ii), recovering units will only matter when the minimal quorum of voting seats have been achieved once again. Before that, the cluster will not be able to define shard allocation. On the other hand, if we have to bring several nodes at once up, then we may end up with a lot of recovery traffic happening at once as well. OpenSearch does provide a protection for these flooding, using [`cluster.routing.allocation.node_concurrent_{|outgoing|incoming}_recoveries or node_initial_primaries_recoveries` settings](https://github.com/opensearch-project/OpenSearch/blob/555a56dba55da5be98abbff5ed2fad4d7320f073/server/src/main/java/org/opensearch/cluster/routing/allocation/decider/ThrottlingAllocationDecider.java#L52-L72). That sets a limit of how many concurrent shard relocation tasks may happen at once.

Therefore, a case of (ii) will eventually become a case (i), and can then start to use the charm lock in OpenSearch for restarting. This PR will not address how to decide when we are in case (ii) and what do. This is going to be subject of further investigation.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_locking.py
+++ b/lib/charms/opensearch/v0/opensearch_locking.py
@@ -254,9 +254,27 @@ class OpenSearchNodeLock(ops.Object):
                         self._opensearch, use_localhost=host is not None, hosts=alt_hosts
                     )
                 )
+                try:
+                    logger.debug(
+                        "Current shard allocation status: %s",
+                        self._opensearch.request(
+                            "GET",
+                            "/_cluster/allocation/explain?include_yes_decisions=true&include_disk_info=true",
+                            payload={
+                                "index": self.OPENSEARCH_INDEX,
+                                "shard": 0,
+                                "primary": "true",
+                            },
+                        ),
+                    )
+                except Exception:
+                    logger.debug("Current shard allocation status: error to connect with API")
+                    pass
             except OpenSearchHttpError:
                 logger.exception("Error getting OpenSearch nodes")
-                return False
+                # If we are trying to acquire the lock at application removal, this condition
+                # will be eventually hit
+                return len(self.units) <= 1
             logger.debug(f"[Node lock] Opensearch {online_nodes=}")
             assert online_nodes > 0
             try:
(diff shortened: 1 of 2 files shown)


lib/charms/opensearch/v0/opensearch_locking.py
```diff
@@ -266,10 +284,10 @@ def acquired(self) -> bool:  # noqa: C901
                 # if the node lock cannot be acquired, fall back to peer databag lock
                 # this avoids hitting deadlock situations in cases where
                 # the .charm_node_lock index is not available
-                if online_nodes <= 1:
+                if online_nodes <= 1 and self._charm.app.planned_units() > 1:
                     return self._peer.acquired
                 else:
-                    return False
+                    return self._charm.app.planned_units() > 1
```
reneradoi, line 290: I'm not sure if this is correct: If there's an error when querying which unit has the lock, and there's more than one unit planned, you just want to return true? If several units do this at once, they'd all just go down (or start) at once.

PR #394, opened 2024-08-13
make TLS certs reload more robust

reneradoi:
## Issue
In cases where reloading TLS certs via API fails, Opensearch should be restarted to pick up the new certs. This could happen e.g. after a network cut, if the API is not available and new TLS certs have been requested because of a new IP address.

This can be seen [in the current failures](https://github.com/canonical/opensearch-operator/actions/runs/10361988050/job/28683584556#step:27:5111) of the integration test `test_ha_networking.py`.

## Solution
If an error occurs when reloading the certs, restart Opensearch to pick up the new certificates.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -725,7 +725,11 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
 
         # In case of renewal of the unit transport layer cert - restart opensearch
         if renewal and self.is_admin_user_configured() and self.tls.is_fully_configured():
-            self.tls.reload_tls_certificates()
+            try:
+                self.tls.reload_tls_certificates()
+            except OpenSearchHttpError:
+                logger.error(f"Could not reload TLS certificates via API, will restart.")
+                self._restart_opensearch_event.emit()
 
     def on_tls_relation_broken(self, _: RelationBrokenEvent):
         """As long as all certificates are produced, we don't do anything."""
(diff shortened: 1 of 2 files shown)


Mehdi-Bendriss (dismissed):
Very nice!

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -725,7 +725,11 @@ def on_tls_conf_set(
 
         # In case of renewal of the unit transport layer cert - restart opensearch
         if renewal and self.is_admin_user_configured() and self.tls.is_fully_configured():
-            self.tls.reload_tls_certificates()
+            try:
+                self.tls.reload_tls_certificates()
+            except OpenSearchHttpError:
+                logger.error("Could not reload TLS certificates via API, will restart.")
+                self._restart_opensearch_event.emit()
```
phvalguima, line 732: @reneradoi is this the original TLS flow if the HTTP request fails?

Also, given a restart is expensive, I think it would be worthy here to have a retry.

reneradoi: Yes, in case of error, the original restart is initiated. See also here: https://github.com/canonical/opensearch-operator/pull/380/files#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43L732

I've added the retries.

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #396, opened 2024-08-13
[DPE-4875] Apply sysctl requirements if possible

Mehdi-Bendriss:
## Issue
This PR implements [DPE-4875](https://warthogs.atlassian.net/browse/DPE-4875), namely:
- check sysctl requirements 
- apply them if needed and possible
- check them again for verification

[DPE-4875]: https://warthogs.atlassian.net/browse/DPE-4875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_distro.py
+++ b/lib/charms/opensearch/v0/opensearch_distro.py
@@ -452,24 +452,30 @@ class OpenSearchDistribution(ABC):
 
         return exclusions
 
-    @staticmethod
-    def missing_sys_requirements() -> List[str]:
+    def missing_sys_requirements(self) -> List[str]:
         """Checks the system requirements."""
+
+        def apply(prop: str, value: str) -> bool:
+            """Apply a sysctl value and check if it was set."""
+            try:
+                self._run_cmd(f"sysctl -w {prop}={value}")
+                return self._run_cmd(f"sysctl -n {prop}") == value
+            except OpenSearchCmdError:
+                return False
+
         missing_requirements = []
 
-        max_map_count = int(subprocess.getoutput("sysctl vm.max_map_count").split("=")[-1].strip())
-        if max_map_count < 262144:
-            missing_requirements.append("vm.max_map_count should be at least 262144")
+        prop, val = "vm.max_map_count", "262144"
+        if self._run_cmd(f"sysctl -n {prop}") < val and not apply(prop, val):
+            missing_requirements.append(f"{prop} should be at least {val}")
 
-        swappiness = int(subprocess.getoutput("sysctl vm.swappiness").split("=")[-1].strip())
-        if swappiness > 0:
-            missing_requirements.append("vm.swappiness should be 0")
+        prop, val = "vm.swappiness", "0"
+        if self._run_cmd(f"sysctl -n {prop}") > val and not apply(prop, val):
+            missing_requirements.append(f"{prop} should be {val}")
 
-        tcp_retries = int(
-            subprocess.getoutput("sysctl net.ipv4.tcp_retries2").split("=")[-1].strip()
-        )
-        if tcp_retries > 5:
-            missing_requirements.append("net.ipv4.tcp_retries2 should be 5")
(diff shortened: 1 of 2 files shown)


reneradoi (approved):
nit: iterate over the tuples of prop, val and the operator (>, <) to get rid of the multiple if's. But I'll leave it up to you if you want that or not.

Mehdi-Bendriss:
Thanks @reneradoi ! I thought about it a bit after reading your comment, and given that the missing requirements message varies (depending on the comparison operator) it will end up with as much lines as the current one with slightly less readability

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -59,7 +59,7 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
```
phvalguima, line 62: Is this correct?

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -452,24 +452,30 @@ def normalize_allocation_exclusions(exclusions: Union[List[str], Set[str], str])
...
+            missing_requirements.append(f"{prop} should be {val}")
 
-        tcp_retries = int(
-            subprocess.getoutput("sysctl net.ipv4.tcp_retries2").split("=")[-1].strip()
-        )
-        if tcp_retries > 5:
-            missing_requirements.append("net.ipv4.tcp_retries2 should be 5")
+        prop, val = "net.ipv4.tcp_retries2", "5"
+        if self._run_cmd(f"sysctl -n {prop}") > val and not apply(prop, val):
+            missing_requirements.append(f"{prop} should be at most {val}")
```
phvalguima, line 478: What will we do about `fs.file-max`?

phvalguima (commented):
We should also clean up our docs from cloud-init. We will not need it if we will try setting sysctl via charm. Main reason:
1. If we are running as VMs, then we will be able to set all sysctl parameters
2. If we are running within LXC, then we cannot set them either way: via charm or via cloud-init; we need an external intervention to set it on the host itself

So, these docs are now redundant.

lib/charms/opensearch/v0/opensearch_relation_provider.py
Mehdi-Bendriss: Yes - the Libpatch on charmhub is 1, i'm a bit unsure why we had 3 set in the first place

lib/charms/opensearch/v0/opensearch_distro.py
Mehdi-Bendriss: This is already set in the snap through `ulimits` see [here](https://github.com/canonical/opensearch-snap/blob/main/scripts/wrappers/start.sh#L8) and [here](https://github.com/canonical/opensearch-snap/blob/main/scripts/wrappers/sys/set-sys-config.sh#L29-L32)

Mehdi-Bendriss:
@phvalguima The reason why I chose to keep as is the docs is to because the charm only sets "the minimal" best values - if the user deployed using cloud-init better values our charm logic should/will not change those. 
So I prefer to allow the user some flexibility to apply better values than we do ourselves.

phvalguima:
> @phvalguima The reason why I chose to keep as is the docs is to because the charm only sets "the minimal" best values - if the user deployed using cloud-init better values our charm logic should/will not change those.
So I prefer to allow the user some flexibility to apply better values than we do ourselves.

@Mehdi-Bendriss then in this case we should move the session down in our README, as the cloud init is not mandatory anymore. Maybe adding a `Customizing` session, but def. cloud-init is not part of `Usage` anymore.

phvalguima (approved):
Approving as we will review the sysctl docs and testing on a next PR

PR #405, opened 2024-08-16
[DPE-4575] Manage voting exclusions in more detail

phvalguima:
Currently, our charm is very simplified when it comes to managing voting exclusions. The main issue is that we have a limited number of possible votes to exclude at a time, set by `cluster.max_voting_config_exclusions` (default: `10`). Therefore, we need to be cautious on cleaning up this list periodically.

This PR does the following changes:
* Adds a method to fetch for the current voting exclusions
* Decides which votes should stay, based on a current list of votings to be deleted
* Move the `delete_current()` to be executed at `_start_opensearch` (already there, in `_post_start_init`)
* Add integration test to scale from 3->1->3 and check for the exclusion rules
* Continuous writes now set non balanced indices with "0-all" replicas


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1030,17 +1030,6 @@ def _stop_opensearch(self, *, restart=False) -> None:
         self.peers_data.delete(Scope.UNIT, "started")
         self.status.set(WaitingStatus(ServiceStopped))
 
-        # 3. Remove the exclusions
-        if not restart:
-            try:
-                self.opensearch_exclusions.delete_current()
```
phvalguima, line 1036: We do not remove it here, but rather at "start_openseach" method now

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1175,7 +1164,7 @@ def _remove_data_role_from_dedicated_cm_if_needed(  # noqa: C901
         self.opensearch_config.remove_temporary_data_role()
 
         # wait until data moves out completely
-        self.opensearch_exclusions.add_current()
```
phvalguima, line 1178: We should not mix with the voting exclusions here, as we just want to move the shards allocation away.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: FYI, the `delete_current()` is already called on `_post_start_init`. No changes needed there then

phvalguima:
I can see [ha/test_large_deployments_relation.py](https://github.com/canonical/opensearch-operator/actions/runs/10445469730/job/28921775363) has started to fail. Last time I noticed it successfully executed was [3 days ago](https://github.com/canonical/opensearch-operator/actions/runs/10413056296/job/28840179364)

reneradoi (dismissed):

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -50,17 +47,22 @@ def add_current(self, restart: bool = False) -> None:
...
-            self._charm.peers_data.put(self._scope, VOTING_TO_DELETE, True)
-
-        if self._node.is_data() and not self._delete_allocations():
+        if self._node.is_cm_eligible() or self._node.is_voting_only():
+            if self._delete_voting():
+                self._charm.peers_data.put(
+                    self._scope,
+                    VOTING_TO_DELETE,
+                    ",".join(self._fetch_voting_exclusions()),
+                )
```
Mehdi-Bendriss, line 61: Can you explain why we fetch and store all exclusions? my understanding was that we only keep track on the relations data of the ones that failed to be deleted (similar to the logic of allocations below)

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -70,55 +72,134 @@ def delete_current(self) -> None:
...
 
+    def _removed_units_to_cleanup(self) -> Optional[List[str]]:
+        """Deletes all units that have left the cluster via Juju.
+
+        This method ensures we keep a small list of voting exclusions at all times.
+        """
+        deployment_desc = self._charm.opensearch_peer_cm.deployment_desc()
+        if not deployment_desc or deployment_desc.typ not in [
+            DeploymentType.MAIN_ORCHESTRATOR,
+            DeploymentType.FAILOVER_ORCHESTRATOR,
```
Mehdi-Bendriss, line 83: We should only do this for the `MAIN_ORCHESTRATOR` - as it's the only one supposed to be at the center of the star.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -70,55 +72,134 @@ def delete_current(self) -> None:
...
+            return []
+
+        short_id = self._charm.opensearch_peer_cm.deployment_desc().app.short_id
+        # We do not need to add our own unit as this code is running on it!
+        peers = self._charm.model.get_relation(PeerRelationName)
+        peers = [] if not peers else peers.units
+
+        cms = set(
+            [
+                unit.name + "." + short_id
```
Mehdi-Bendriss, line 94: you should use `format_unit_name(unit)`

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -70,55 +72,134 @@ def delete_current(self) -> None:
...
+                for relation in self._charm.model.relations.get(
+                    PeerClusterOrchestratorRelationName
+                )
+                for unit in relation.units
+            ]
+            + [unit.name + "." + short_id for unit in peers]
+        )
```
Mehdi-Bendriss, line 101: I believe you can replace the above with: 
1. Check if current is main provider `if self.charm.opensearch_peer_cm.is_provider(typ="main")`
    1. fetch all `units = self.charm.peers_data.get(Scope.APP, "cluster_fleet_apps")`
    2. parse it as show [here](https://github.com/canonical/opensearch-operator/blob/main/lib/charms/opensearch/v0/opensearch_peer_clusters.py#L401)
2. else => small deployments
    1. `units = [format_unit_name(u) for u in all_units(self.charm)]`

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -70,55 +72,134 @@ def delete_current(self) -> None:
...
-        need_voting_cleanup = self._charm.peers_data.get(self._scope, VOTING_TO_DELETE, False)
-        if need_voting_cleanup and self._delete_voting():
+        need_voting_cleanup = set(
+            self._charm.peers_data.get(self._scope, VOTING_TO_DELETE, "").split(",")
+            + self._removed_units_to_cleanup()
+        )
+        if (
+            need_voting_cleanup
+            and self._delete_voting(to_remove=need_voting_cleanup)
+            and not self._fetch_voting_exclusions()  # called after the entire deletion, must be empty now
```
Mehdi-Bendriss, line 118: not sure we should be too harsh here, in case the user sets some manually - we should perhaps track only the ones we set from our charm

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -70,55 +72,134 @@ def delete_current(self) -> None:
...
 
-    def _add_voting(self) -> bool:
+    def _add_voting(self, exclusions: Optional[Set[str]] = None) -> bool:
         """Include the current node in the CMs voting exclusions list of nodes."""
         try:
+            to_add = exclusions or set([self._node.name])
             self._opensearch.request(
                 "POST",
-                f"/_cluster/voting_config_exclusions?node_names={self._node.name}&timeout=1m",
+                f"/_cluster/voting_config_exclusions?node_names={','.join(to_add)}&timeout=1m",
```
Mehdi-Bendriss, line 134: nice

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -70,55 +72,134 @@ def delete_current(self) -> None:
...
+            logger.debug("Removed voting: FAILED")
             return False
 
-    def _add_allocations(
+    def _fetch_voting_exclusions(self) -> Set[str]:
+        """Fetch the registered voting exclusions."""
+        try:
+            resp = self._opensearch.request(
+                "GET",
+                "/_cluster/state/metadata/voting_config_exclusions",
```
Mehdi-Bendriss, line 177: TIL, thanks!

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -70,55 +72,134 @@ def delete_current(self) -> None:
...
             self._opensearch.request(
                 "DELETE",
                 "/_cluster/voting_config_exclusions?wait_for_removal=false",
                 alt_hosts=self._charm.alt_hosts,
                 resp_status_code=True,
             )
-            return True
+            logger.debug("Removed voting")
+            remove_set = to_remove or set([self._node.name])
+            nodes = set(remove_set)
```
Mehdi-Bendriss, line 159: We can remove this, right?

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -70,55 +72,134 @@ def delete_current(self) -> None:
...
+
+    def add_allocations(
         self, allocations: Optional[Set[str]] = None, override: bool = False
     ) -> bool:
         """Register new allocation exclusions."""
         try:
             existing = set() if override else self._fetch_allocations()
             all_allocs = existing.union(
-                allocations if allocations is not None else {self._node.name}
+                allocations if allocations is not None else set([self._node.name])
```
Mehdi-Bendriss, line 202: I believe this is a more expensive call, building a list, building a set, as opposed to what we had

Mehdi-Bendriss (commented):
Thanks Pedro! I left some comments

tests/integration/ha/continuous_writes.py
```diff
@@ -56,6 +57,8 @@ async def start(self, repl_on_all_nodes: bool = False, is_bulk: bool = True) ->
         # create index if custom conf needed
         if repl_on_all_nodes:
             await self._create_fully_replicated_index()
+        else:
+            await self._create_index_with_0_all_replicas()
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss I am scaling from 3->1, I need to have the `test_index` with 0 replicas, otherwise I will always fail. Apart from that, we are aligned any other test should always use `repl_on_all_nodes`

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
phvalguima: Voting exclusions sits on the critical path of the 3<->2<->1 nodes scaling. It does not apply to all the cases.

Now, on these cases, we **must** apply them.

However, in this specific case, I **must** apply the voting exclusion before, say, scaling 3->2, or clean it up before 1->2 scaling. The reason is because once we reach 2x nodes, then we cannot guarantee the API will work forever, as it may eventually try another election and we lose the known leader. 

So, indeed, keeping track of the exclusions in a peer relation is not needed. We can fetch the results anyways. I kept that only to have some similarity with allocations.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
phvalguima: Fair enough. The only concern I can see is that we have 10x votes by default... If we consider the user may try using it, then we may end up not having enough votes!

We can raise that as an exception, ofc.

tests/integration/ha/continuous_writes.py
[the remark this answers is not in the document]
phvalguima: Btw, I can leave a comment of "Avoid this 0-all option at all costs!" on the code...

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
phvalguima: Giving it some extra thought, I will do the following: use it to keep track of our own voting exclusions only instead of a `fetch`.

phvalguima: @Mehdi-Bendriss I created another logic:
* `CURRENT_EXCLUSIONS`: a peer databag list that tracks the exclusions defined by the charm
* At `_add_voting`: we add one unit to `CURRENT_EXCLUSIONS` and a new voting exclusion
* At `_cleanup_voting`: (1) **potential nodes to remove**: check the list of nodes present in the full voting exclusions and not present in our deployment; (2) iterate over `CURRENT_EXCLUSIONS`: if a node is present in this list AND in the **potential nodes to remove**, then we should take this node out of the voting exclusions as it is occupying a vote and it is actually gone
* At `_delete_voting`: (1) run the _cleanup_voting; (2) remove the node itself from that list as well; and (3) remove all votes and readd the remaining ones to the list

This way, we make sure we can preserve any user-defined changes but still keep the list clean.

Some questions:
1) We should use the peer-cluster relation as well to publish CURRENT_EXCLUSIONS, right? Otherwise, if we have a failover, then the list is now present in the peer databag of the `failover cluster`
2) We have a hard limit of voting count (10 in general), there is a risk we will hit this limit if we let the user also set voting and we respect these values, what should we do in this case?

tests/integration/ha/continuous_writes.py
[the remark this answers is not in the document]
Mehdi-Bendriss: @phvalguima Not all tests rely on an `1-all` (full) replication factor - since many have the replication factor set on first document ingestion. And for those tests, they will essentially have the `0-all` logic you added - which would be wrong. 

I suggest you replace the arg `repl_on_all_nodes: bool = False` by: `replication_mode: Literal["0-all",  "1-all", "default"] = "default"` and replace the `else` by an `elif replication_mode == "0-all"`

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
Mehdi-Bendriss: I think we could potentially simplify it and build on top of what we already have by only making use of `ALLOCS_TO_DELETE` and `VOTING_TO_DELETE` which we update whenever we have a failed deletion, and recurrent cleanup routine attempts to wipe the content of these no?

phvalguima:
Also, has to clean-up: https://github.com/canonical/opensearch-operator/pull/407#discussion_r1758630831

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -503,6 +497,18 @@ def _on_peer_relation_departed(self, event: RelationDepartedEvent):
...
+            return
+
+        # Now, we register in the leader application the presence of departing unit's name
+        # We need to save them as we have a count limit
+        for attempt in Retrying(stop=stop_after_attempt(6), wait=wait_fixed(10)):
+            with attempt:
+                if not self.alt_hosts:
+                    # No neighbors found, it means this is the last unit to go away.
+                    break
+                self.opensearch_exclusions.add_exclusions(unit_name=event.departing_unit)
```
phvalguima, line 510: As the `event.departing_unit` is going away, we need to register it in the app-level peer relation. Otherwise, we may not be able to clean it up.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -538,6 +544,9 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
                 pass
         try:
             self._stop_opensearch()
+            if self.alt_hosts:
+                # There is enough peers available for us to try removing the unit
+                self.opensearch_exclusions.delete_current()
```
phvalguima, line 549: We try to clean this unit out of the voting/allocation exclusions. The challenge: if we fail, its own `Scope.UNIT` entry that registers this exclusion will also go away. Therefore, we use the peer-departed to also register this departing unit to the app-level databag.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -41,75 +42,190 @@ class OpenSearchExclusions:
...
+            if response < 200 or response >= 300:
+                logger.debug("Failed to remove voting exclusions, response %s", response)
+                return False
+
+            logger.debug("Removed voting for: %s", to_stay)
+            return to_stay and self._add_voting(to_stay)
         except OpenSearchHttpError:
-            return False
+            pass
+        return False
```
Mehdi-Bendriss, line 205: what we had was cleaner I believe

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -41,75 +42,190 @@ class OpenSearchExclusions:
...
         try:
-            self._opensearch.request(
+            response = self._opensearch.request(
                 "DELETE",
                 "/_cluster/voting_config_exclusions?wait_for_removal=false",
                 alt_hosts=self._charm.alt_hosts,
                 resp_status_code=True,
             )
-            return True
+            if response < 200 or response >= 300:
```
Mehdi-Bendriss, line 197:
```suggestion
            if response >= 400:
```

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -41,75 +42,190 @@ class OpenSearchExclusions:
...
                 alt_hosts=self._charm.alt_hosts,
                 resp_status_code=True,
-                retries=3,
             )
-            return True
+            logger.debug("Added voting, response: %s", response)
+
+            self._charm.peers_data.put(self._scope, VOTING_TO_DELETE, ",".join(to_add))
+            # The voting excl. API returns a status only
+            return response >= 200 and response < 300
```
Mehdi-Bendriss, line 168:
```suggestion
            return response < 400
```

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -41,75 +42,190 @@ class OpenSearchExclusions:
...
         try:
-            self._opensearch.request(
+            to_add = exclusions or {self._node.name}
+            response = self._opensearch.request(
                 "POST",
-                f"/_cluster/voting_config_exclusions?node_names={self._node.name}&timeout=1m",
+                f"/_cluster/voting_config_exclusions?node_names={','.join(to_add)}&timeout=1m",
                 alt_hosts=self._charm.alt_hosts,
                 resp_status_code=True,
-                retries=3,
```
Mehdi-Bendriss, line 93: why remove the retry?

Mehdi-Bendriss (commented):
Thanks Pedro! I have a few questions

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
phvalguima: Ack, it is just a personal view: don't like a method without final return... Anyways, returned to the original setting!

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -503,6 +497,18 @@ def _on_peer_relation_departed(self, event: RelationDepartedEvent):
         else:
             event.defer()
 
+        if not self.unit.is_leader():
+            return
+
+        # Now, we register in the leader application the presence of departing unit's name
+        # We need to save them as we have a count limit
+        for attempt in Retrying(stop=stop_after_attempt(6), wait=wait_fixed(10)):
```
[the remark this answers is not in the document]
phvalguima: Indeed, changing the `add` to be a boolean and raise an `Exception` in case of `False`.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1059,7 +1068,7 @@ def _stop_opensearch(self, *, restart=False) -> None:
                 # and re-using storage
                 if len(nodes) > 1:
                     # 1. Add current node to the voting + alloc exclusions
-                    self.opensearch_exclusions.add_current(restart=restart)
+                    self.opensearch_exclusions.add_current()
```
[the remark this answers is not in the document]
phvalguima: @Mehdi-Bendriss agreed for allocation, but not for exclusion. I will update it, but I'd rather have the `restart` boolean dealt on the `_stop` / `_start` methods, e.g.:
```
self.opensearch_exclusions.add_current(voting=True, allocation=not restart)
```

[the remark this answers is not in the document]
phvalguima: My main concern is: what if we were resetting a single node of 3, and we fail to bring it back online... Then, we may end up with a cluster error because the 3->2 scaling only works if we carefully exclude the unit from the voting pool before stopping.

If we fail bringing up or even take longer than the default wait time ( `30s`), then we will break the cluster in this case.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
```diff
@@ -41,76 +42,196 @@ class OpenSearchExclusions:
     def __init__(self, charm):
         self._charm = charm
         self._opensearch = self._charm.opensearch
-
         self._scope = Scope.APP if self._charm.unit.is_leader() else Scope.UNIT
 
-    def add_current(self, restart: bool = False) -> None:
+    def add(self, unit_name: str) -> None:
```
juditnovak, line 47: I'm adding the question here though it should go rather on line 45 -- impacting the whole module.

I gather that we do not save anything in the databag on the leader unit, as leader only manipulates the `app` databag.

Now, if a leader is restarted, and leadership may be granted to another unit, then the ex-leader comes back with an empty unit databag (in terms of voting exclusions).

Why is that not an issue...?

tests/integration/ha/test_scale_to_one_and_back.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
phvalguima: @juditnovak these moments are purposefully short... we'd have to explicitly simulate an issue at restart.

lib/charms/opensearch/v0/opensearch_nodes_exclusions.py
phvalguima: So, the `VOTING_TO_DELETE` only matters at `cleanup()`. At that moment, we will do the following:
1) Load all the `VOTING_TO_DELETE` entries from our unit/app databag (depending if we are the leader or not)
2) For each entry in that set, we only will clean it from "voting exclusion" IF that unit is gone from juju
3) We apply the final cleanup

At restart, we call: `add_current()` at `_stop_opensearch`, i.e. we will forcely add our own unit to the exclusions. We then stop and start the service. At `_start_opensearch`, we call `delete_current()`, which will remove this same unit from the exclusion, forcely, and then update its own databag.

Therefore, my changes only really take effect on: (1) `-departed`, the app leader registers the new unit to be removed; and (2) at cleanup: we remove the unit from the voting exclusion if the unit is not available on Juju anymore.

tests/integration/ha/test_scale_to_one_and_back.py
```diff
@@ -0,0 +1,191 @@
...
+        await ops_test.model.applications[app].destroy_unit(f"{app}/{elected_cm_unit_id}")
+
+        await wait_until(
+            ops_test,
+            apps=[app],
+            units_statuses=["active"],
+            wait_for_exact_units=init_count - 1,
+            idle_period=IDLE_PERIOD,
+        )
+        assert len(voting_exclusions) == 0
```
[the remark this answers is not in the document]
phvalguima: Indeed! Thanks

tests/integration/ha/test_scale_to_one_and_back.py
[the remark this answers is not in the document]
juditnovak: Grrrrrr..... then could we have pls-pls-pls a 1-2 unittests...?

My problem is that... The current tests could (eventually) execute all well also if the voting exclusion databag fields are empty all way throught...

tests/unit/lib/test_opensearch_exclusions.py
```diff
@@ -0,0 +1,221 @@
...
+    def setUp(self) -> None:
+        self.harness = Harness(OpenSearchOperatorCharm)
+        self.addCleanup(self.harness.cleanup)
+        self.harness.set_leader(True)
+
+        self.harness.begin()
+
+        self.charm = self.harness.charm
+        self.peer_rel_id = self.harness.add_relation(PeerRelationName, self.charm.app.name)
+        self.lock_rel_id = self.harness.add_relation(NodeLockRelationName, self.charm.app.name)
```
juditnovak, line 30: U may B able 2 get rid of this line.

tests/unit/lib/test_opensearch_exclusions.py
```diff
@@ -0,0 +1,221 @@
...
+                "unit1",
+                "unit2",
+                "unit3",
+            ]
+
+            # Unit loses leadership, the unit stops checking the Scope.APP
+            self.harness.set_leader(False)
+            # Override the self._scope:
+            self.charm.opensearch_exclusions._scope = Scope.UNIT
+            # Simulating the unit start, hence calling delete_current()
```
juditnovak, line 129: I guess this is to be swapped with the line below, no? :-)

tests/integration/ha/test_scale_to_one_and_back.py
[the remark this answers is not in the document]
juditnovak: Thank VERY-VERY much for the particularly verbose unittests.

They surely serve as documentation as much as a security net.

juditnovak (dismissed):
Thank you soooooo much for the elaborate testing!!!!!!!!!!! :-) :-) :-)

reneradoi (commented):
Hey @phvalguima the PR looks good! Logic wise I'm fine with it, there are only a few formalities I'd like to change.

At some point it would probably be good to have an integration test to see the scaling also work in large deployments, but that can also be added later.

tests/unit/lib/test_opensearch_exclusions.py
phvalguima: Not really, the thing is, we never recreate the `self.charm.opensearch_exclusions` object, hence we never update its `_scope`. In theory, I do not even need to do `self.harness.set_leader(False)` to get the same result here.

tests/unit/lib/test_opensearch_exclusions.py
phvalguima: Only the last line. The peer rel is needed so I can set the deployment-description content later down the road.

juditnovak (approved):

reneradoi (approved):

PR #407, opened 2024-08-17
[DPE-4588] Hook start checks if service was previously running

phvalguima:
In this PR, we target to automatically process the service restart post full cluster reboot. Currently, neither charm nor systemd will restart the service back once the node has restarted.

### Current Issue in the Charm

Once the node is restarted, it will throw the `start` hook once again.
Each node will detect the new hook and run the routine outside of the `if self.is_node_up()`. Each unit will issue a `restart-event` and try to acquire the lock.

The first unit that acquires the lock will manage to get the peer-lock. That means only that unit has it and it can restart safely.

Once that unit is on, then all the remaining units will have to acquire the node-lock. However, the single service will not be fully online, as it will be an 1/2,3,X... nodes. Given it still has its own metadata informing of neighbours, then it will be blocked waiting for these neighbours.

Therefore, the entire cluster will get stuck waiting for the lock.

### Proposal

We will focus only on nodes that have "cluster_manager" as one of the roles. All the other nodes should work with the lock requests in any case.

Akin to systemd's "enabled" concept: once a service is active, it will be started automatically after a reboot; the charm will now have a logic that detects a "start" hook (hence, a reboot may have happened) and:
1) Checks first if the unit is up (i.e. `is_node_up() == True`)
2) If not, check if the peer relation data stated this unit as "started" == "True". If yes, then we may have one of either: (i) deferred "start" that is wrapping its last task, or (ii)  we have a reboot. The (i) case will be a exception (i.e. a "start" hook happened with service down and "started==True")
3) As we do not need to differ from (i) or (ii), then we execute the same clean-up logic as `is_node_up()==True` condition
4) Restart the systemd service
5) Finally, we restart the systemd service


phvalguima:
I can see [ha/test_large_deployments_relation.py](https://github.com/canonical/opensearch-operator/actions/runs/10445469730/job/28921775363) has started to fail. Last time I noticed it successfully executed was [3 days ago](https://github.com/canonical/opensearch-operator/actions/runs/10413056296/job/28840179364)

reneradoi (dismissed):

phvalguima:
After some further investigation, it seems the community will abandon most of `gateway.*` option on the next major release. Discussion has been taken upstream on: https://github.com/opensearch-project/OpenSearch/issues/15599

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -136,6 +136,31 @@ def format_unit_name(unit: Union[Unit, str], app: App) -> str:
     return f"{unit.replace('/', '-')}.{app.short_id}"
 
 
+def all_units_names(charm: "OpenSearchBaseCharm") -> List[Unit]:
```
phvalguima, line 139: Keeping this logic, although not used anymore in this code.

We need a way to recover all units' names, disregard if this is a small, large, etc deployment.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -312,8 +313,28 @@ def _on_start(self, event: StartEvent):
             if self.peers_data.get(Scope.UNIT, "bootstrap_contributor"):
                 self._cleanup_bootstrap_conf_if_applies()
 
+        if self.opensearch.is_node_up():
+            __on_start_post_cleanup()
             return
 
+        elif (
+            self.peers_data.get(Scope.UNIT, "started")
```
phvalguima, line 321: `self.peers_data[Scope.UNIT]["started"]` is a boolean

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -64,6 +64,15 @@ def set_client_auth(self):
             True,
         )
 
+    def set_recover_after_nodes(self, recover_after_nodes: Optional[int] = None):
```
phvalguima, line 67: Not used anymore, but keeping it as an option.

reneradoi: Maybe add a comment here to refer the deprecation notice and - once [your issue](https://github.com/opensearch-project/OpenSearch/issues/15599) has been replied to - remove completely if it is no longer supported.

phvalguima: @reneradoi I think even better is to only focus on data nodes tbh. These will stay moving forward.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -300,9 +300,10 @@ def _on_leader_elected(self, event: LeaderElectedEvent):
 
         self.status.clear(AdminUserInitProgress)
 
-    def _on_start(self, event: StartEvent):
+    def _on_start(self, event: StartEvent):  # noqa C901
         """Triggered when on start. Set the right node role."""
-        if self.opensearch.is_node_up():
+
+        def __on_start_post_cleanup():
```
Mehdi-Bendriss, line 306: nit - given its within a method, we can get away with a simple name, e.g: `cleanup`

src/opensearch.py
```diff
@@ -70,6 +72,50 @@ def install(self):
...
+
+        if not service_running("snap.opensearch.daemon.service"):
+            return False
+
+        # Now, we must dig deeper into the actual status of systemd and the JVM process.
+        # First, we want to make sure the process is not stopped, dead or zombie.
+        # From: https://github.com/torvalds/linux/blob/master/fs/proc/array.c#L126-L140
+        # "Parked" state is ignored as it applies to threads.
+        try:
+            pid = subprocess.check_output(["lsof", "-ti:9200"], text=True).rstrip()
```
Mehdi-Bendriss, line 92: could you use `run_cmd` ?

src/opensearch.py
```diff
@@ -70,6 +72,50 @@ def install(self):
...
+            if not pid or not os.path.exists(f"/proc/{pid}/stat"):
+                return False
+            with open(f"/proc/{pid}/stat") as f:
+                stat = f.read()
+        except subprocess.CalledProcessError:
+            return False
+
+        if stat[2] == "T" and paused:
```
Mehdi-Bendriss, line 100: could you add a comment above this line and explain what `stat[2] == "T"` refers to?

src/opensearch.py
```diff
@@ -70,6 +72,50 @@ def install(self):
...
+            return True
+
+        # We do not check reachability of the service
+        # If that is needed, then use the `is_started` method.
+        return stat[2] not in ["Z", "T", "X"]
```
Mehdi-Bendriss, line 105: same here for future readers

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -312,8 +313,28 @@ def _on_start(self, event: StartEvent):
...
+            )
+
+            # We had a reboot in this node.
+            # We execute the same logic as above:
+            __on_start_post_cleanup()
+
+            # Now, reissue a restart: we should not have stopped in the first place
+            # as "started" flag is still set to True.
+            # We do not wait for the 200 return, as maybe more than one unit is coming back
+            self.opensearch.start_service_only()
```
Mehdi-Bendriss, line 336: I think we need to catch `OpenSearchStartError` here

Mehdi-Bendriss (commented):
Thanks pedro, looks good! I only have minor comments

reneradoi (approved):

src/opensearch.py
```diff
@@ -86,17 +87,28 @@ def is_service_started(self, paused: Optional[bool] = False) -> bool:
 
         # Now, we must dig deeper into the actual status of systemd and the JVM process.
         # First, we want to make sure the process is not stopped, dead or zombie.
-        # From: https://github.com/torvalds/linux/blob/master/fs/proc/array.c#L126-L140
-        # "Parked" state is ignored as it applies to threads.
         try:
-            pid = subprocess.check_output(["lsof", "-ti:9200"], text=True).rstrip()
+            pid = run_cmd("lsof", args="-ti:9200").rstrip()
```
Mehdi-Bendriss, line 91: rstrip is not necessary - as it is already cleaned in `run_cmd`

Mehdi-Bendriss (approved):

src/opensearch.py
phvalguima: @Mehdi-Bendriss I will clean up on a next PR

PR #408, opened 2024-08-20
[DPE-4653] certificate expiration integration test

reneradoi:
This PR provides integration test coverage for the workflow of expiring TLS certificates and their renewal. 

It also includes a bugfix for renewing the expiring certificates, or better creating the certificate signing requests: When initially creating them, they do not include `unique-id` in the subject. But on renewal this is included, which causes the `subject` of the renewed certificate to be different from the initial one.

In addition to that, there's a config change to `TLSCertificatesRequiresV3` in order to get notification for expiring certificates with a reasonable lead time.

Please also note the documentation added for the TLS certificates workflow here: https://github.com/canonical/opensearch-operator/wiki/TLS-certificates-workflow

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -73,7 +73,7 @@ class OpenSearchTLS(Object):
         self.jdk_path = jdk_path
         self.certs_path = certs_path
         self.keytool = self.jdk_path + "/bin/keytool"
-        self.certs = TLSCertificatesRequiresV3(charm, TLS_RELATION, expiry_notification_time=168)
+        self.certs = TLSCertificatesRequiresV3(charm, TLS_RELATION, expiry_notification_time=23)
 
         self.framework.observe(
             self.charm.on.set_tls_private_key_action, self._on_set_tls_private_key
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -302,6 +302,7 @@ class OpenSearchTLS(Object):
         subject = self._get_subject(cert_type)
         organization = self.charm.opensearch_peer_cm.deployment_desc().config.cluster_name
         new_csr = generate_csr(
+            add_unique_id_to_subject_name=False,
             private_key=key,
             private_key_password=(None if key_password is None else key_password.encode("utf-8")),
             subject=subject,
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -73,7 +73,7 @@ class OpenSearchTLS(Object):
         self.jdk_path = jdk_path
         self.certs_path = certs_path
         self.keytool = self.jdk_path + "/bin/keytool"
-        self.certs = TLSCertificatesRequiresV3(charm, TLS_RELATION)
+        self.certs = TLSCertificatesRequiresV3(charm, TLS_RELATION, expiry_notification_time=168)
 
         self.framework.observe(
             self.charm.on.set_tls_private_key_action, self._on_set_tls_private_key
--- a/tests/integration/tls/test_tls.py
+++ b/tests/integration/tls/test_tls.py
@@ -5,8 +5,10 @@
 import logging
 import subprocess
 import time
+from datetime import datetime, timedelta
 
 import pytest
+from charms.tls_certificates_interface.v3.tls_certificates import LIBID as TLS_LIBID
 from pytest_operator.plugin import OpsTest
 
 from ..helpers import (
@@ -190,22 +192,18 @@ async def test_tls_expiration(ops_test: OpsTest) -> None:
         wait_for_exact_units=1,
     )
 
-    # Now apply a hack to make the certificate secrets expire in short time
-    # set the secret expiry to a fixed timedelta of 180 seconds to give time to start initially.
-    # This happens on the tls_certificates lib and we apply the patch via sed-command.
-    # The patch accesses the method `_get_next_secret_expiry_time` in the tls_certificate libs.
-    # If the test should fail on updates to this lib, check if the method was renamed or removed.
-    search_expression = "expire=self._get_next_secret_expiry_time\\(certificate\\)"
-    replace_expression = "expire=timedelta\\(seconds=180\\)"
+    # Relate OpenSearch to TLS and wait until all is settled
+    await ops_test.model.integrate(APP_NAME, TLS_CERTIFICATES_APP_NAME)
+
+    await wait_until(
+        ops_test,
+        apps=[APP_NAME],
+        units_statuses=["active"],
+        wait_for_exact_units=1,
+    )
 
     unit_id = get_application_unit_ids(ops_test, APP_NAME)[0]
     unit_ip = await get_leader_unit_ip(ops_test)
-    lib_file = f"/var/lib/juju/agents/unit-opensearch-{unit_id}/charm/lib/charms/tls_certificates_interface/v3/tls_certificates.py"
-    cmd = f"juju ssh {APP_NAME}/{unit_id} sudo sed -i 's/{search_expression}/{replace_expression}/g' {lib_file}"
-    subprocess.run(cmd, shell=True)
-
-    # Relate OpenSearch to TLS and wait until all is settled
-    await ops_test.model.integrate(APP_NAME, TLS_CERTIFICATES_APP_NAME)
 
     # wait for the unit to be ready and API's available
     cluster_health_resp = await cluster_health(ops_test, unit_ip, wait_for_green_first=True)
@@ -214,6 +212,19 @@ async def test_tls_expiration(ops_test: OpsTest) -> None:
     # now start with the actual test
     # first get the currently used certs
     current_certs = await get_loaded_tls_certificates(ops_test, unit_ip)
+
+    # Now apply a hack to make the certificate secrets expire in short time
+    # set the secret expiry to a fixed timedelta of 180 seconds to give time to start initially.
+    expire = datetime.now() + timedelta(seconds=180)
+
+    juju_secrets = await ops_test.model.list_secrets()
+    for secret in juju_secrets:
+        if secret.label.startswith(TLS_LIBID):
+            cmd = f"""juju exec -u {APP_NAME}/{unit_id} --  \
+             /var/lib/juju/tools/unit-{APP_NAME}-{unit_id}/secret-set \
+            {secret.uri.split(":")[1]} --expire {expire.isoformat()}
+            """
+            subprocess.run(cmd, shell=True)
 
     # now wait for the expiration period to pass by (and a bit longer for things to settle)
     # we can't use `wait_until` here because the unit might not be idle in the meantime
--- a/tests/integration/tls/test_tls.py
+++ b/tests/integration/tls/test_tls.py
@@ -191,8 +191,10 @@ async def test_tls_expiration(ops_test: OpsTest) -> None:
     )
 
     # Now apply a hack to make the certificate secrets expire in short time
-    # set the secret expiry to a fixed timedelta of 180 seconds to give time to start initially
-    # this happens on the tls_certificates lib and we apply the patch via sed-command
+    # set the secret expiry to a fixed timedelta of 180 seconds to give time to start initially.
+    # This happens on the tls_certificates lib and we apply the patch via sed-command.
+    # The patch accesses the method `_get_next_secret_expiry_time` in the tls_certificate libs.
+    # If the test should fail on updates to this lib, check if the method was renamed or removed.
     search_expression = "expire=self._get_next_secret_expiry_time\\(certificate\\)"
     replace_expression = "expire=timedelta\\(seconds=180\\)"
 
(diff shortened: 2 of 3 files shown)


lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -73,7 +73,7 @@ def __init__(
         self.jdk_path = jdk_path
         self.certs_path = certs_path
         self.keytool = self.jdk_path + "/bin/keytool"
-        self.certs = TLSCertificatesRequiresV3(charm, TLS_RELATION)
+        self.certs = TLSCertificatesRequiresV3(charm, TLS_RELATION, expiry_notification_time=168)
```
juditnovak, line 76: Tiny nitpick, can B ignored.

May it be useful to add a comment saying 
```
We want to be notified a week in advance
```
or so (highlighting the uing of 168 :-) )

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -302,6 +302,7 @@ def _request_certificate_renewal(
         subject = self._get_subject(cert_type)
         organization = self.charm.opensearch_peer_cm.deployment_desc().config.cluster_name
         new_csr = generate_csr(
+            add_unique_id_to_subject_name=False,
```
juditnovak, line 305: Wow, nice catch!!!!!

(Perhaps an explanatory comment above as well, why this is essential?)

tests/integration/tls/test_tls.py
```diff
@@ -154,3 +157,72 @@ async def test_tls_renewal(ops_test: OpsTest) -> None:
...
+
+    await wait_until(
+        ops_test,
+        apps=[APP_NAME],
+        units_statuses=["blocked"],
+        wait_for_exact_units=1,
+    )
+
+    # Now apply a hack to make the certificate secrets expire in short time
+    # set the secret expiry to a fixed timedelta of 300 seconds to give time to start initially
```
juditnovak, line 193: Though technically this is true, I'm not sure if we should go into this level of detail.

From our perspective: we hack the library to create certs with an early expiry.
(Now it's the lib's implementation detail that it goes on to secrtets, etc. but at this level we don't need to know about that.) All we wanted was a cert that expires in 3 mins insteatd the usual, long ahead expiry time.

tests/integration/tls/test_tls.py
```diff
@@ -154,3 +157,72 @@ async def test_tls_renewal(ops_test: OpsTest) -> None:
...
+    cmd = f"juju ssh {APP_NAME}/{unit_id} sudo sed -i 's/{search_expression}/{replace_expression}/g' {lib_file}"
+    subprocess.run(cmd, shell=True)
+
+    # Relate OpenSearch to TLS and wait until all is settled
+    await ops_test.model.integrate(APP_NAME, TLS_CERTIFICATES_APP_NAME)
+    assert await is_up(ops_test, unit_ip), "OpenSearch service hasn't started."
+    logger.info("Waiting for certificates to expire.")
+
+    # wait for the certs API to be ready
+    time.sleep(30)
```
juditnovak, line 210: `sleep()` generally introduces the risk of flakyness...

Couldn't we rather `wait_for` the service, or add `tenacity` retries to the upcoming calls...?

tests/integration/tls/test_tls.py
```diff
@@ -154,3 +157,72 @@ async def test_tls_renewal(ops_test: OpsTest) -> None:
...
+
+    # now start with the actual test
+    # first get the currently used certs
+    current_certs = await get_loaded_tls_certificates(ops_test, unit_ip)
+
+    # now wait for the expiration period to pass by (and a bit longer for things to settle)
+    # we can't use `wait_until` here because the unit might not be idle in the meantime
+    time.sleep(240)
```
juditnovak, line 218: Can't we wait until the unit becomes idle?
Or for any other condition that has to unavoidably occur?

Again it would be resulting in more stable pipelines...
We had a LOT of trouble in the past with unstable Opensearch pipelines...
I think we should be more strict on preserving current reliable quality and avoid any shadow of flakyness...

tests/integration/tls/test_tls.py
```diff
@@ -154,3 +157,72 @@ async def test_tls_renewal(ops_test: OpsTest) -> None:
...
+    # set the secret expiry to a fixed timedelta of 300 seconds to give time to start initially
+    # this happens on the tls_certificates lib and we apply the patch via sed-command
+    search_expression = "expire=self._get_next_secret_expiry_time\\(certificate\\)"
+    replace_expression = "expire=timedelta\\(seconds=180\\)"
+
+    unit_id = get_application_unit_ids(ops_test, APP_NAME)[0]
+    unit_ip = await get_leader_unit_ip(ops_test)
+    lib_file = f"/var/lib/juju/agents/unit-opensearch-{unit_id}/charm/lib/charms/tls_certificates_interface/v3/tls_certificates.py"
+    cmd = f"juju ssh {APP_NAME}/{unit_id} sudo sed -i 's/{search_expression}/{replace_expression}/g' {lib_file}"
+    subprocess.run(cmd, shell=True)
```
juditnovak, line 202: Huh though it's a pretty cool hack, I'm afraid it strongly wouldn't stand the test of time.

The TLS lib's owners clearly should have the liberty to change the name of their helper functions. Howerver now if that happens, we have a broken pipeline.
(Which is a highly unrewarding experience for the Developer working on a totally unrelated task to debug :-) )

I'm wondering if 
 - there may be a long-term reliable way to add a hack (like monkeypatching a public function like `generate_certificates` or so)

However here I would stop for a moment.

The primary thing we wanted to check was that the notification was raised, no?
I.e. the new behavior introduce in the Opensearch charm code.

Now _that_ could be verified in a reliable way ASSUMING that Juju secret expiry is to be used as a mean of the notification workflow from TLS team. (This we could easily check with them if the secret expiry is an "interface-level" part of their workflow, that should be documented and preserved on their side -- or rather an implementation detail they may change in the future.)

ASSUMING it's part of their public workflow, we could change the secret expiry by resetting the secret info -- which is a safer hack than applying `sed` on someone else's code :-) 

At this point we assume that the rest of the TLS workflow (as shipped by the TLS operator and library) and it's not our role to test if their workflow is safe. Of course if we want to --for the sake of the security of our charm-- then we must find a safe way that could preserve the behavior. I'm actually thinking to ask the TLS libs team to provide a way we could set shorter expiry perhaps than 1d...?

juditnovak (changes requested):
1. I may have overlooed: we are not testing for the notification to be raised? But for the cert expiry/renewal (only)?

2. Could we change the title of the PR? It's not only introducing tests but a behavior change for the Opensearch charm (namely to get notifications ahead of a cert expiry), which is a "product-level" change :-)

reneradoi:
Hey Judit, thanks for your review!

> 1. I may have overlooed: we are not testing for the notification to be raised? But for the cert expiry/renewal (only)?
No the goal is not to test for the notification to be raised (aka the secret to expire). This is part of what the `tls-operator` and `juju` do for us, and we only use these mechanisms. What we really want to test is that our reaction to that even is as expected, and the expiring certificates are renewed by our operator, with Opensearch being still functional in the meantime. That's why this test is focussed on the certs to be updated, when this event occurs.

> 2. Could we change the title of the PR? It's not only introducing tests but a behavior change for the Opensearch charm (namely to get notifications ahead of a cert expiry), which is a "product-level" change :-)

The notification is actually not a new feature that we are inventing. It is already there, and we would already get notified. I only added the config to make this notification a useful one by adjusting this time to a week. This will result in the secret expiry to be set to one week in advance of the certificate expiry instead of just some hours. I don't consider this a big change tbh.

tests/integration/tls/test_tls.py
reneradoi: I understand your point here, and yes, if they would change that function name, we would have to do work on our side. But given that this in not in their charm but in the libs, the failing test would be the direct impact of a lib update and would thus not cause flaky tests but instead only the test to fail on that PR updating this lib.

I'd like to add that this hack/patch is very helpful for this test to be a "quick" feedback loop. I have to admit that I don't consider waiting for one hour to be acceptable for an integration test, except if there is absolutely no other way around.

tests/integration/tls/test_tls.py
reneradoi: Same remark as above: No, unfortunately not. As the certificates will be reloaded during that waiting period, the unit would not be active/idle anymore and the wait time would start again. Instead we have a fixed time after that we expect the certs to be renewed, that's why I chose `sleep` here.

tests/integration/tls/test_tls.py
reneradoi: No, unfortunately not. As the certificates will be reloaded during that waiting period, the unit would not be active/idle anymore and the wait time would start again. Instead we have a fixed time after that we expect the certs to be renewed, that's why I chose `sleep` here.

reneradoi: This `sleep` could be removed by instead waiting for the cluster health to be `green`.

tests/integration/tls/test_tls.py
reneradoi: In this case, a constant was added to signal this sleep is not just some arbitrary time.

tests/integration/tls/test_tls.py
reneradoi: Several other approaches have been discussed, but none was working. Even directly executing the hook tools of juju, in this case for `secret-set` didn't work out. Therefore I will add a more detailed comment, but leave it as it is for now.

juditnovak: All right, after a small research on the topic, if I'm correct this should work:

```
juju exec -u opensearch/4 -- /var/lib/juju/tools/unit-opensearch-4/secret-set cr3ib0gqjdij98lt3h00 --expire <I'm_not_sure_of_the_format>
```
Reverse engineering it from the `ops` library (https://ops.readthedocs.io/en/latest/_modules/ops/model.html -> serach for "def secret_set" )  the format should be `datetime.isoformat()`

Mehdi-Bendriss (commented):
Thanks René! I have 1 comment for added sanity check

tests/integration/tls/test_tls.py
```diff
@@ -154,3 +158,74 @@ async def test_tls_renewal(ops_test: OpsTest) -> None:
...
+    assert cluster_health_resp["status"] == "green"
+
+    # now start with the actual test
+    # first get the currently used certs
+    current_certs = await get_loaded_tls_certificates(ops_test, unit_ip)
+
+    # now wait for the expiration period to pass by (and a bit longer for things to settle)
+    # we can't use `wait_until` here because the unit might not be idle in the meantime
+    logger.info("Waiting for certificates to expire.")
+    time.sleep(SECRET_EXPIRY_WAIT_TIME)
```
Mehdi-Bendriss, line 221: we could however have a `wait_until` with a longer `IDLE_PERIOD` set, e.g: 120sec, no?

tests/integration/tls/test_tls.py
```diff
@@ -154,3 +158,74 @@ async def test_tls_renewal(ops_test: OpsTest) -> None:
...
+
+    # now compare the current certificates against the earlier ones and see if they were updated
+    updated_certs = await get_loaded_tls_certificates(ops_test, unit_ip)
+    logger.info(f"Certificates before expiry: {current_certs}")
+    logger.info(f"Certificates after expiry: {updated_certs}")
+
+    assert (
+        updated_certs["transport_certificates_list"][0]["not_before"]
+        > current_certs["transport_certificates_list"][0]["not_before"]
+    )
```
Mehdi-Bendriss, line 231: Could we add 2 additional checks here: 
- if all the nodes are online 
- if the new certificates are effectively loaded / used by all the units

tests/integration/tls/test_tls.py
reneradoi: I think this won't work, as during this period, the secrets expire and the certs will be renewed. This causes the units to not be active, and the `wait_until` would start the `IDLE_PERIOD` from 0. This would end in a timeout, as I've seen in the early stages of testing this.

tests/integration/tls/test_tls.py
reneradoi: Was added.

tests/integration/tls/test_tls.py
reneradoi: After running multiple tests with `juju exec ... secret-set` it turned out that this solution will also not be feasible, thus I defaulted back to accepting a one-hour duration of the test-run and adjusted the test for this.

lib/charms/opensearch/v0/opensearch_tls.py
juditnovak: Double-checking: why did this conversation get closed? 

Decision on no comment or perhaps by accident?
(I think it may be nice for the next developer in 6 months or so to have a quick reminder why this boolean was critical. Regardless, I leave it to your judgement :-) )

juditnovak (approved):

lib/charms/opensearch/v0/opensearch_tls.py
reneradoi: Hey Judit, yeah I decided not to comment, as the change is just to use the exactly same parameters as in the other `generate_csr` call in the code. To me this was just a bug I fixed along the way, nothing worth to comment.

Mehdi-Bendriss (approved):

PR #409, opened 2024-08-22
[DPE-5202] Test TLS through Manual TLS Certificates

skourta:
Add a test for an OpenSearch deployment with TLS certificates issued by the `manual-tls-certificates` operator.

### DIFF
--- a/tests/integration/tls/helpers_manual_tls.py
+++ b/tests/integration/tls/helpers_manual_tls.py
@@ -7,7 +7,6 @@ import asyncio
 import base64
 import json
 import logging
-from collections import deque
 from typing import TYPE_CHECKING, NamedTuple
 
 from charms.tls_certificates_interface.v3.tls_certificates import (
@@ -80,7 +79,7 @@ class ManualTLSAgent:
         self.tls_unit = tls_unit
         self.ca_key = generate_private_key()
         self.ca = generate_ca(self.ca_key, "CN_CA")
-        self.csr_queue = deque()
+        self.csr_queue: list[CSR] = []
 
     async def get_outstanding_certificate_requests(self) -> None:
         """Get the outstanding certificate requests from the TLS operator.
@@ -101,7 +100,7 @@ class ManualTLSAgent:
             )
             raise GetOutstandingCertificateRequestsError(message)
         csrs = json.loads(action.results["result"])
-        self.csr_queue = deque([CSR.from_dict(csr) for csr in csrs])
+        self.csr_queue = [CSR.from_dict(csr) for csr in csrs]
 
     @retry(
         wait=wait_exponential(multiplier=1, min=5, max=20),
@@ -168,13 +167,12 @@ class ManualTLSAgent:
     async def process_queue(self) -> None:
         """Process the certificate signing requests in the queue."""
         while self.csr_queue:
-            csr = self.csr_queue.popleft()
+            csr = self.csr_queue.pop()
             await self.process_csr(csr)
 
 
 async def main() -> None:
     """Run the ManualTLSAgent."""
-    logging.basicConfig(level=logging.INFO)
     logging.info("Starting ManualTLSAgent")
     model = Model()
     await model.connect()
--- a/tests/integration/tls/helpers_manual_tls.py
+++ b/tests/integration/tls/helpers_manual_tls.py
@@ -1,0 +1,186 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import base64
+import json
+import logging
+from collections import deque
+from typing import TYPE_CHECKING, NamedTuple
+
+from charms.tls_certificates_interface.v3.tls_certificates import (
+    generate_ca,
+    generate_certificate,
+    generate_private_key,
+)
+from juju.unit import Unit
+from tenacity import retry, stop_after_attempt, wait_exponential
+
+if TYPE_CHECKING:
+    from juju.action import Action
+
+logger = logging.getLogger(__name__)
+
+MANUAL_TLS_CERTIFICATES_APP_NAME = "manual-tls-certificates"
+
+
+class GettingOutstandingCertificateRequestsFailedError(Exception):
+    """Exception raised when getting outstanding certificate requests fails."""
+
+    def __init__(self, message: str) -> None:
+        """Initialise the exception."""
+        super().__init__(message)
+
+
+class CSRsMissingError(Exception):
+    """Exception raised when the number of CSRs in the queue is less than the expected number."""
+
+    def __init__(self, message: str) -> None:
+        """Initialise the exception."""
+        super().__init__(message)
+
+
+class ProvidingCertificateFailedError(Exception):
+    """Exception raised when providing a certificate fails."""
+
+    def __init__(self, message: str) -> None:
+        """Initialise the exception."""
+        super().__init__(message)
+
+
+class CSR(NamedTuple):
+    """CSR represents the information about a certificate signing request."""
+
+    relation_id: str
+    application_name: str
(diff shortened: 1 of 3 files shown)


tests/integration/tls/test_manual_tls.py
```diff
@@ -0,0 +1,91 @@
...
+        my_charm,
+        num_units=len(UNIT_IDS),
+        series=SERIES,
+        application_name=APP_NAME,
+    )
+
+    # Deploy TLS Certificates operator.
+    tls_app: Application = await ops_test.model.deploy(
+        MANUAL_TLS_CERTIFICATES_APP_NAME,
+        channel="beta",
```
reneradoi, line 37: What's the reason why you chose to test against the `beta` channel? Most other tests use `stable` for the channel of the `self-signed-certificates`, I think we should stick to this if no specific reason for `beta`.

reneradoi (commented):
Hey @skourta really nice work! I like especially how you organized the `manual-tls-operator` in the helper and also the logging of the integration test itself.

There's only one small question that came up. Once that is clarified, I'm fine with merging this.

Thank you for contributing this!

tests/integration/tls/test_manual_tls.py
skourta: TBH I just followed the [readme](https://github.com/canonical/manual-tls-certificates-operator) they had. There is no specific reason. I will test with the stable version, there shouldn't be any issue.

skourta: They have both `stable` and `beta` as the same revision. I changed the channel to `stable`. I'm closing the conversation.

reneradoi (approved):

tests/integration/tls/helpers_manual_tls.py
```diff
@@ -0,0 +1,209 @@
...
+        if action.status != "completed":
+            message = action.safe_data.get(
+                "message",
+                "Failed to get outstanding certificate requests",
+            )
+            if "No outstanding certificate requests" in message:
+                logging.info(message)
+            if "No certificates relation has been created yet" in message:
+                logging.info(message)
+            logging.error(message)
```
Mehdi-Bendriss, line 117: are these all needed? given we raise an exception with this message

tests/integration/tls/helpers_manual_tls.py
```diff
@@ -0,0 +1,209 @@
...
+            raise GettingOutstandingCertificateRequestsFailedError(message)
+        csrs = json.loads(action.results["result"])
+        self.csr_queue = deque([CSR.from_charm_csr(csr) for csr in csrs])
```
Mehdi-Bendriss, line 120: question: why opt for a `deque` instead of a simple list? and use `pop(0)` instead of `popleft`

tests/integration/tls/helpers_manual_tls.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @retry(
+        wait=wait_exponential(multiplier=1, min=5, max=20),
+        stop=stop_after_attempt(100),
+    )
+    async def wait_for_csrs_in_queue(self, nbr_csrs: int = 1) -> None:
```
Mehdi-Bendriss, line 126:
```suggestion
    async def wait_for_csrs_in_queue(self, csrs_count: int = 1) -> None:
```

tests/integration/tls/helpers_manual_tls.py
```diff
@@ -0,0 +1,209 @@
...
+    async def process_queue(self) -> None:
+        """Process the certificate signing requests in the queue."""
+        while self.csr_queue:
+            csr = self.csr_queue.popleft()
+            await self.process_csr(csr)
+
+
+async def main() -> None:
+    """Run the ManualTLSAgent."""
+    logging.basicConfig(level=logging.INFO)
```
Mehdi-Bendriss, line 193: safe to delete

tests/integration/tls/helpers_manual_tls.py
```diff
@@ -0,0 +1,209 @@
...
+                "certificate": base64.b64encode(certificate).decode(),
+                "ca-certificate": base64.b64encode(self.ca).decode(),
+                "certificate-signing-request": base64.b64encode(
+                    csr.csr,
+                ).decode(),
+            },
+        )
+        action = await action.wait()
+        if action.status != "completed":
+            message = "Failed to provide certificate for %s", csr.unit_name
```
Mehdi-Bendriss, line 179: This creates a tuple
```suggestion
            message = f"Failed to provide certificate for {csr.unit_name}"
```

tests/integration/tls/test_manual_tls.py
```diff
@@ -0,0 +1,91 @@
...
+        units_statuses=["active"],
+        wait_for_exact_units=len(UNIT_IDS),
+        timeout=2000,
+    )
+    assert len(ops_test.model.applications[APP_NAME].units) == len(UNIT_IDS)
+
+    # Scale up the application by adding a new unit
+    logger.info("Scaling up the application by adding a new unit")
+    await os_app.add_unit(1)
+
```
Mehdi-Bendriss, line 74: We should perhaps add a `wait_until` for consistency with other tests

Mehdi-Bendriss (commented):
Thanks Smail! I have a few comments

tests/integration/tls/helpers_manual_tls.py
```diff
@@ -0,0 +1,209 @@
...
+
+class ProvidingCertificateFailedError(Exception):
+    """Exception raised when providing a certificate fails."""
+
+    def __init__(self, message: str) -> None:
+        """Initialise the exception."""
+        super().__init__(message)
+
+
+class CSR(NamedTuple):
```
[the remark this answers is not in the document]
skourta: Can you please explain? I used `NamedTuple` because I'm using CSRs as immutable. Based on my knowledge `NamedTuple` is more adequate and efficient than `Dataclass` for this use.

tests/integration/tls/helpers_manual_tls.py
skourta: A Python `list` is implemented as a dynamic array in C. If we use a list and pop(0) all subsequent items in the list must be shifted which renders the operation very inefficient. `deque` (short for double-ended queue) has O(1) performance for both append and pop on either side of the queue. [Resource](https://docs.python.org/3/library/collections.html#collections.deque)

tests/integration/tls/helpers_manual_tls.py
skourta: Do you mean the config setting or all the main function? I'm using the module manually as well with the Manual TLS Operator to do tests.

tests/integration/tls/helpers_manual_tls.py
Mehdi-Bendriss: In general - in **tests** and with this limited scale, we optimize for readability more than we do for performances

tests/integration/tls/helpers_manual_tls.py
Mehdi-Bendriss: I meant the logging level setting

tests/integration/tls/helpers_manual_tls.py
skourta: I understand. Since we do not care about the order of the CSRs (they can arrive in any order.) I will change it to a `list` and use `pop` to treat it as a stack rather than a queue. This way we get readability and still be efficient. WDYT?

Mehdi-Bendriss: sounds good!

Mehdi-Bendriss (approved):

reneradoi (approved):

PR #412, opened 2024-08-23
[DPE-4049][BUGFIX] Exception when offline unit is missing `node.roles` from static config

juditnovak:
## Issue

https://github.com/canonical/opensearch-operator/issues/222

Context (as highlighted in the [issue comment](https://github.com/canonical/opensearch-operator/issues/222#issuecomment-2306491781)):

"When a unit is online, unit-specific service information is taken from the Opensearch API (`/_nodes`).

However if the unit is offline, we fall back to static configuration.
This is the part where we had the issue, in case `node.roles` was missing from the configuration."

## Solution

We shouldn't directly refer to a field that may be missing from the config.

NOTE: Since the `opensearch_distro` had no unittests, this PR is also adding a new unttest module dedicated to `opensearch_distro`. For the time being only the concerned `current()` method is being tested (both for happy and broken path -- the latter was the target of the fix.) The test module should be extended in the future for additional test coverage for the `opensearch_distro` module.

NOTE2: The PR is including a global fix for Opensearch Unit Tests, as a global mock was applied (in a hidden corner) imacting ALL tests.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_distro.py
+++ b/lib/charms/opensearch/v0/opensearch_distro.py
@@ -17,8 +17,12 @@ from typing import Dict, List, Optional, Set, Tuple, Union
 
 import requests
 import urllib3.exceptions
-from charms.opensearch.v0.helper_charm import mask_sensitive_information
-from charms.opensearch.v0.helper_cluster import ClusterTopology, Node
+from charms.opensearch.v0.constants_charm import GeneratedRoles
+from charms.opensearch.v0.helper_charm import (
+    format_unit_name,
+    mask_sensitive_information,
+)
+from charms.opensearch.v0.helper_cluster import Node
 from charms.opensearch.v0.helper_conf_setter import YamlConfigSetter
 from charms.opensearch.v0.helper_http import error_http_retry_log
 from charms.opensearch.v0.helper_networking import get_host_ip, is_reachable
@@ -30,6 +34,7 @@ from charms.opensearch.v0.opensearch_exceptions import (
     OpenSearchStartTimeoutError,
 )
 from charms.opensearch.v0.opensearch_internal_data import Scope
+from pydantic.error_wrappers import ValidationError
 from tenacity import (
     Retrying,
     retry,
@@ -417,7 +422,7 @@ class OpenSearchDistribution(ABC):
         """Return Port of OpenSearch."""
         return 9200
 
-    def current(self) -> Node:
+    def current(self) -> Node:  # noqa: C901
         """Returns current Node."""
         try:
             nodes = self.request("GET", f"/_nodes/{self.node_id}", alt_hosts=self._charm.alt_hosts)
@@ -431,30 +436,56 @@ class OpenSearchDistribution(ABC):
                 unit_number=self._charm.unit_id,
                 temperature=current_node.get("attributes", {}).get("temp"),
             )
+
         except OpenSearchHttpError:
+
             # we try to get the most accurate description of the node from the static config
             conf = self.config.load("opensearch.yml")
+
+            # also, if possible we rely on the Deployment Description (databag)
             deployment_desc = self._charm.opensearch_peer_cm.deployment_desc()
+
+            # Application Priority: Deployment Description
+            # Reason: No reason to re-construct the App object
+            #  - it's available 99% of scenarios
+            #  - it's the same object as a re-constructed one (i.e. no dynamic changes on App)
+            if deployment_desc is None:
+                try:
+                    app = App(id=conf.get("node.attr.app_id"))
+                except ValidationError:
+                    raise OpenSearchError("Can not determine app details.")
+            else:
+                app = deployment_desc.app
+
+            # Roles (Temperature) Priority: local config
+            # Reason:
+            #  - Deployment Description is holding "expected state" (that may not be applied)
+            #  - Static config holds the currently applied settings
             try:
                 roles = conf["node.roles"]
             except KeyError:
-                if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES:
-                    roles = deployment_desc.config.roles
+                if deployment_desc:
+                    if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES:
+                        roles = deployment_desc.config.roles
+                    else:
+                        roles = GeneratedRoles
                 else:
-                    roles = ClusterTopology.generated_roles()
+                    raise OpenSearchError("Can not determine roles.")
 
+            temperature = None
             try:
                 temperature = conf["node.attr.temp"]
             except KeyError:
-                temperature = deployment_desc.config.data_temperature
-            if not roles:
-                raise OpenSearchError("Can not determine app name and/or roles.")
+                if deployment_desc:
+                    temperature = deployment_desc.config.data_temperature
 
             return Node(
-                name=self._charm.unit_name,
+                # NOTE: We are NOT using self._charm.unit_name, as it refers to deployment_desc()
+                # that is not to be assumed to be always available at this point
+                name=format_unit_name(self._charm.unit, app=app),
                 roles=roles,
                 ip=self._charm.unit_ip,
-                app=deployment_desc.app,
+                app=app,
                 unit_number=self._charm.unit_id,
                 temperature=temperature,
             )
--- a/lib/charms/opensearch/v0/opensearch_distro.py
+++ b/lib/charms/opensearch/v0/opensearch_distro.py
@@ -18,11 +18,11 @@ from typing import Dict, List, Optional, Set, Tuple, Union
 import requests
 import urllib3.exceptions
 from charms.opensearch.v0.helper_charm import mask_sensitive_information
-from charms.opensearch.v0.helper_cluster import Node
+from charms.opensearch.v0.helper_cluster import ClusterTopology, Node
 from charms.opensearch.v0.helper_conf_setter import YamlConfigSetter
 from charms.opensearch.v0.helper_http import error_http_retry_log
 from charms.opensearch.v0.helper_networking import get_host_ip, is_reachable
-from charms.opensearch.v0.models import App
+from charms.opensearch.v0.models import App, StartMode
 from charms.opensearch.v0.opensearch_exceptions import (
     OpenSearchCmdError,
     OpenSearchError,
@@ -434,30 +434,27 @@ class OpenSearchDistribution(ABC):
         except OpenSearchHttpError:
             # we try to get the most accurate description of the node from the static config
             conf = self.config.load("opensearch.yml")
+            deployment_desc = self._charm.opensearch_peer_cm.deployment_desc()
             try:
                 roles = conf["node.roles"]
             except KeyError:
-                roles = self._charm.opensearch_peer_cm.deployment_desc().config.roles
-
-            try:
-                app = App(id=conf["node.attr.app_id"])
-            except KeyError:
-                app = App(id=self._charm.opensearch_peer_cm.deployment_desc().app.id)
+                if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES:
+                    roles = deployment_desc.config.roles
+                else:
+                    roles = ClusterTopology.generated_roles()
 
             try:
                 temperature = conf["node.attr.temp"]
             except KeyError:
-                temperature = (
-                    self._charm.opensearch_peer_cm.deployment_desc().config.data_temperature
-                )
-            if not (app and roles):
+                temperature = deployment_desc.config.data_temperature
+            if not roles:
                 raise OpenSearchError("Can not determine app name and/or roles.")
 
             return Node(
                 name=self._charm.unit_name,
                 roles=roles,
                 ip=self._charm.unit_ip,
-                app=app,
+                app=deployment_desc.app,
                 unit_number=self._charm.unit_id,
                 temperature=temperature,
             )
--- a/lib/charms/opensearch/v0/opensearch_distro.py
+++ b/lib/charms/opensearch/v0/opensearch_distro.py
@@ -421,6 +421,7 @@ class OpenSearchDistribution(ABC):
         """Returns current Node."""
         try:
             nodes = self.request("GET", f"/_nodes/{self.node_id}", alt_hosts=self._charm.alt_hosts)
+
             current_node = nodes["nodes"][self.node_id]
             return Node(
                 name=current_node["name"],
@@ -431,15 +432,34 @@ class OpenSearchDistribution(ABC):
                 temperature=current_node.get("attributes", {}).get("temp"),
             )
         except OpenSearchHttpError:
-            # we try to get the most accurate description of the node
-            conf_on_disk = self.config.load("opensearch.yml")
+            # we try to get the most accurate description of the node from the static config
+            conf = self.config.load("opensearch.yml")
+            try:
+                roles = conf["node.roles"]
+            except KeyError:
+                roles = self._charm.opensearch_peer_cm.deployment_desc().config.roles
+
+            try:
+                app = App(id=conf["node.attr.app_id"])
+            except KeyError:
+                app = App(id=self._charm.opensearch_peer_cm.deployment_desc().app.id)
+
+            try:
+                temperature = conf["node.attr.temp"]
+            except KeyError:
+                temperature = (
+                    self._charm.opensearch_peer_cm.deployment_desc().config.data_temperature
+                )
+            if not (app and roles):
+                raise OpenSearchError("Can not determine app name and/or roles.")
+
             return Node(
                 name=self._charm.unit_name,
-                roles=conf_on_disk["node.roles"],
+                roles=roles,
                 ip=self._charm.unit_ip,
-                app=App(id=conf_on_disk.get("node.attr.app_id")),
+                app=app,
                 unit_number=self._charm.unit_id,
-                temperature=conf_on_disk.get("node.attr.temp"),
+                temperature=temperature,
             )
 
     @staticmethod
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -98,6 +98,7 @@ COSUser = "monitor"
 COSRelationName = "cos-agent"
 COSRole = "readall_and_monitor"
 COSPort = "9200"
+GeneratedRoles = ["data", "ingest", "ml", "cluster_manager"]
 
 
 # Opensearch Users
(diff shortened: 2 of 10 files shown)


lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -433,9 +434,10 @@ def current(self) -> Node:
         except OpenSearchHttpError:
             # we try to get the most accurate description of the node
             conf_on_disk = self.config.load("opensearch.yml")
+
             return Node(
                 name=self._charm.unit_name,
-                roles=conf_on_disk["node.roles"],
+                roles=conf_on_disk.get("node.roles") or [],
```
Mehdi-Bendriss, line 440: Roles cannot be null / empty, as we rely on those elsewhere in the code - this will result in unpredictability in the code. 

These are the sources of getting the roles from, by order of priority and availability:
1. check API (as done in line 423)
2. if not available - check Disk as done in this line 
3. if not available - check `opensearch_peer_cm.deployment_desc().config.roles`
4. if not available - check where roles are queried too early and fix it

Mehdi-Bendriss (changes requested):
Thanks Judit, I left a comment.

lib/charms/opensearch/v0/opensearch_distro.py
juditnovak: Sure.

In case NO roles found at all: raising an exception?

Out of curiousity. In the use-case where we experience the [issue](https://github.com/canonical/opensearch-operator/issues/222) (departing unit) where should the role come from?
If it is the config, I would adjust the config in the test space to include roles, simulating closer to reality.

Thanks.

juditnovak: Question #1:

> 4. if not available - check where roles are queried too early and fix it

I'm confused how this may apply (i.e. "too early") given that we are getting the error as a node is departing from the cluster. 

As the [issue](https://github.com/canonical/opensearch-operator/issues/222) trace indicates, the current error is triggered in attempt to delete the current node (`opensearch_node_exclusions` -> `delete_current`)

Could you please confirm roles MUST be available until the last moment the unit exists, including the pattern above? I.e. after quitting the `peer` relation, etc. 

(In case not: how should we act then?)

Question #2:

We have 2 other attributes (`node.attr.app_id`, `node.attr.temp`) similarly used in this workflow. Are they supposed to be defined the same way (and sources) as `node.roles`? If yes, we probably want to apply the same checks for all.

juditnovak: > Well the current code has already 1 and 2 above. We need to add the 3rd check. If it still fails, it means that the call happens too early - in which case, the call needs to be made later in the code.

That part is clear :-)  My question was targeting the [issue](https://github.com/canonical/opensearch-operator/issues/222) specifically. I try to make it more clear.

In this particular case (i.e. departing node already offline) where do we expect to find the roles? In other words, typically for example do we ALWAYS have roles configured statically (in `opensearch.yaml`)

I would like to adjust test environment accordingly, i.e. providing coverage for the bug fix according to the expected behavior of the product.

juditnovak: Summarizing the offline conversation:

Two worklows may be impacted: initialization and shutdown process. 
A node should be receiving `roles` information in the following order from the following sources.

Initialization:
 1. `storage-attached` (i.e. Openserach gets its disk space)
 2. roles information is written to static config
 3. roles information is populated on the `peer` relation describing the `Deployment` configuration
 4. once the service is up, the Opensearch API is used for roles information

Shutdown (2-3 order may be reverse):
 1. Opensearch API may not be available anymore
     - either as the unit may get offline 
     - or there may be no Opensearch online unit to contact anymore
 2. `storage-detached` (no more config file or disk space)
 3. peer relation departed (no more `Deployment` configuration

The bug is raised around 2-3. In addition we should check if before 2) the static config may not get wiped out.

juditnovak:
I'm doing my best to reproduce the issue for deeper understanding of the original cause, but I'm struggling to get the failure again.

 - Within approx 7 tries `main` doesn't trigger it anymore
 - tried rev 123 but there another known error is perhaps shadowing this issue
 - currently experimenting with rev130, but I'm strating to wonder if we shouldn't do all this investigating in case the issue would be observed again...
    - NOTE: regarding our assumption that the config may be wiped out -- shouldn't we have a bunch of other errors/issues in that case? If that happened shouldn't we have a way more severe failures (from multiple "angles") than this one...? 
 
image

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -431,15 +432,25 @@ def current(self) -> Node:
                 temperature=current_node.get("attributes", {}).get("temp"),
             )
         except OpenSearchHttpError:
-            # we try to get the most accurate description of the node
-            conf_on_disk = self.config.load("opensearch.yml")
+            # we try to get the most accurate description of the node from the static config
+            if conf := self.config.load("opensearch.yml"):
+                roles = conf.get("node.roles")
+                app = App(id=conf.get("node.attr.app_id"))
```
Mehdi-Bendriss, line 438: this will result in setting `None` values to those 2 attributes if the `opensearch.yml` exists but no value is set in those entries. (it's okay for the temperature).
  
A safer approach would be to use `conf[node.roles]` and catch the exception and use the `deployment_desc.config..` in the catch block or simply check on the existence of those 2 keys.

Mehdi-Bendriss (changes requested):
Thanks Judit - I have a comment

lib/charms/opensearch/v0/opensearch_distro.py
juditnovak: Sure.
My assupmption was that --in case there IS a config-- these values MUST be set there.
(I'm still confused about what we should assume at which stage of the lifetime of the charm on these levels.)

Changing as requested.

juditnovak: Done.

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -431,15 +432,34 @@ def current(self) -> Node:
                 temperature=current_node.get("attributes", {}).get("temp"),
             )
         except OpenSearchHttpError:
-            # we try to get the most accurate description of the node
-            conf_on_disk = self.config.load("opensearch.yml")
+            # we try to get the most accurate description of the node from the static config
+            conf = self.config.load("opensearch.yml")
```
Mehdi-Bendriss, line 436:
```suggestion
            conf = self.config.load("opensearch.yml")
            deployment_desc = self._charm.opensearch_peer_cm.deployment_desc()
```

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -431,15 +432,34 @@ def current(self) -> Node:
...
+            try:
+                roles = conf["node.roles"]
+            except KeyError:
+                roles = self._charm.opensearch_peer_cm.deployment_desc().config.roles
+
+            try:
+                app = App(id=conf["node.attr.app_id"])
+            except KeyError:
+                app = App(id=self._charm.opensearch_peer_cm.deployment_desc().app.id)
```
Mehdi-Bendriss, line 445: The app should only be fetched from the deployment_desc - no need to rebuild it 
```suggestion
```

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -431,15 +432,34 @@ def current(self) -> Node:
...
+
+            try:
+                temperature = conf["node.attr.temp"]
+            except KeyError:
+                temperature = (
+                    self._charm.opensearch_peer_cm.deployment_desc().config.data_temperature
+                )
```
Mehdi-Bendriss, line 452:
```suggestion
                temperature = deployment_desc.config.data_temperature
```

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -431,15 +432,34 @@ def current(self) -> Node:
...
             )
         except OpenSearchHttpError:
-            # we try to get the most accurate description of the node
-            conf_on_disk = self.config.load("opensearch.yml")
+            # we try to get the most accurate description of the node from the static config
+            conf = self.config.load("opensearch.yml")
+            try:
+                roles = conf["node.roles"]
+            except KeyError:
+                roles = self._charm.opensearch_peer_cm.deployment_desc().config.roles
```
Mehdi-Bendriss, line 440: Apologies, I missed this in the previous review
```suggestion
                if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES:
                    roles = deployment_desc.config.roles
                else:
                    roles = ClusterTopology.generated_roles()
```

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -431,15 +432,34 @@ def current(self) -> Node:
...
+            except KeyError:
+                app = App(id=self._charm.opensearch_peer_cm.deployment_desc().app.id)
+
+            try:
+                temperature = conf["node.attr.temp"]
+            except KeyError:
+                temperature = (
+                    self._charm.opensearch_peer_cm.deployment_desc().config.data_temperature
+                )
+            if not (app and roles):
```
Mehdi-Bendriss, line 453: The app should never be None
```suggestion
            if not roles:
```

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -431,15 +432,34 @@ def current(self) -> Node:
...
+                raise OpenSearchError("Can not determine app name and/or roles.")
+
             return Node(
                 name=self._charm.unit_name,
-                roles=conf_on_disk["node.roles"],
+                roles=roles,
                 ip=self._charm.unit_ip,
-                app=App(id=conf_on_disk.get("node.attr.app_id")),
+                app=app,
```
Mehdi-Bendriss, line 460:
```suggestion
                app=deployment_desc.app,
```

Mehdi-Bendriss (commented):
Thanks Judit - I have a few comments

juditnovak:
@Mehdi-Bendriss your suggestions (overall a new version of the original proposal) have been all committed.

I'm confused why you've proposed this code: I do not undestand underlying workflows (further than the instuctions I've been following from earlier comment https://github.com/canonical/opensearch-operator/pull/412#discussion_r1729248248).
I'm interested to learn about them (so next time my contribution could be more useful than it seems to be here :-( )

Your proposal introduces a circular import, as `ClusterTopology` itself internally uses `OpenSearchDistributon` (the class we are modifying).

image

Do you have another proposal? Would you like to give highlights of underlying logic, so I could have one?

juditnovak:
Summarizing offline discussion and consequences:

@Mehdi-Bendriss suggested to use keep the circular import.
Implemented in commit https://github.com/canonical/opensearch-operator/pull/412/commits/f282ef4aeec756f2cfa4107df7423e89d83f0b97
Integration tests show no issues, on a superficial check the disabled config unittest may.

I believe that we should avoid the circular imports and keep clear integrity of the `ClusterTopology` object relying on the `OpenSearchDistribution` as suggested by the original design.

By moving the hard-coded list of default roles into an application-wide constant, this is simple to do. Note that this solution had NO interference with any tests (which should be a signal of a healthy internal refactor).

This proposed solution is added in commit https://github.com/canonical/opensearch-operator/pull/412/commits/d227aa9c48e540edc6b86a492cd15ae9962833e6. (In case circular import is preferred, we just need to roll back this commit.)

Question (@Mehdi-Bendriss, @reneradoi , @phvalguima ):
The current solution is ignoring the Deployment Description stored in the databag if the the application was NOT started with `WITH_PROVIDED_ROLES`. I understand that this is how the application is normally configured.
Question: is it possible that throughout runtime of the application the Deployment Description roles (stored in the databag) may change?
If this scenario is impossible, I'm changing the corresponding new unittest. Thank you.

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -431,15 +433,31 @@ def current(self) -> Node:
...
+            if not roles:
+                raise OpenSearchError("Can not determine app name and/or roles.")
+
             return Node(
                 name=self._charm.unit_name,
-                roles=conf_on_disk["node.roles"],
+                roles=roles,
                 ip=self._charm.unit_ip,
-                app=App(id=conf_on_disk.get("node.attr.app_id")),
+                app=deployment_desc.app,
```
juditnovak, line 458: Questoin @Mehdi-Bendriss @reneradoi @phvalguima 

Normally `OpenSearchPeerClustersManager.deployment_desc()` returns a data structure with values from the databag -- or `None` if databag information couldn't be retrieved.

I.e. the `deployment_desc` variable techinically can be `None`, resulting in the above snippet failing at muliple levels.

Now, the original bug https://github.com/canonical/opensearch-operator/issues/222 was raised at unit shutdown.

Is it possible that `OpenSearchDistribution.current()` may be executed at a point when peer data may not be available anymore? (Why?)

juditnovak: In other words, are we sure that this solution is not duplicating the already existing new issue https://github.com/canonical/opensearch-operator/issues/422 (I just opened about a week ago), throwing precisely the error that I foresee will be also happening if we merge the code above:

```
  File "/var/lib/juju/agents/unit-opensearch-25/charm/lib/charms/opensearch/v0/opensearch_base_charm.py", line 1573, in unit_name
    return format_unit_name(self.unit, app=self.opensearch_peer_cm.deployment_desc().app)
AttributeError: 'NoneType' object has no attribute 'app'
```

...?

juditnovak: @Mehdi-Bendriss  earlier [suggested](https://github.com/canonical/opensearch-operator/pull/412#discussion_r1729248248) to follow this strategy for `roles`: disk, peer relation, error if neither.

Now, the `app_id` is stored in the local config (see screenshot below), and we _did_ check it (only) there with the original code. 

I'd like to understand why do we replace the original strategy, instead of extending it with an additional check (as done for `roles`)?

Also I'm expressing that I'm not sure if the current proposal is solid but I may be mistaken.

image

juditnovak: Summarizing live conversation: reasons to use different sources for `app` vs. `roles` got clarified (see upcoming comments around related code).

It also seemed reasonable to fall back to the original (local config) for `app` in case the Peer Relation may not be available (as in corner cases like the error above).

juditnovak:
Summarizing offline discussion: Deployment Description is holding the "desired state", not the current state.
The current state is available in the static config.
(This also explains why it's safer to back to Deployment Description defaults rather than actual values, in case static config may not be available.)

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -430,16 +435,55 @@ def current(self) -> Node:
...
+            #  - Static config holds the currently applied settings
+            try:
+                roles = conf["node.roles"]
+            except KeyError:
+                if deployment_desc and deployment_desc.start == StartMode.WITH_PROVIDED_ROLES:
+                    roles = deployment_desc.config.roles
+                else:
+                    roles = GeneratedRoles
+
+            if not roles:
```
Mehdi-Bendriss, line 468: This will never be true, given the previous `else`

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -430,16 +435,55 @@ def current(self) -> Node:
...
+            # Reason:
+            #  - Deployment Description is holding "expected state" (that may not be applied)
+            #  - Static config holds the currently applied settings
+            try:
+                roles = conf["node.roles"]
+            except KeyError:
+                if deployment_desc and deployment_desc.start == StartMode.WITH_PROVIDED_ROLES:
+                    roles = deployment_desc.config.roles
+                else:
+                    roles = GeneratedRoles
```
Mehdi-Bendriss, line 466: We Have conflicting conditions here and with `if not roles`.

we can either keep `if not roles:` later, or apply the following: 
```suggestion
                if deployment_desc: 
                    if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES:
                        roles = deployment_desc.config.roles
                    else:
                        roles = GeneratedRoles
```

Or directly access `deployment_desc.config.roles` and let it crash if `not deployment_desc`. Not both.

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -430,16 +435,55 @@ def current(self) -> Node:
...
+
+            if not roles:
+                raise OpenSearchError("Can not determine app name and/or roles.")
+
+            temperature = None
+            try:
+                temperature = conf["node.attr.temp"]
+            except KeyError:
+                if deployment_desc:
```
Mehdi-Bendriss, line 475: I believe this is not needed - as we'd have been into `if not roles` and raised an exception.

Mehdi-Bendriss (commented):
Thanks Judit, I left minor comments.

lib/charms/opensearch/v0/opensearch_distro.py
juditnovak: Indeed my mistake on the `if` condition.

Good call, thanks, I believe that the above was the intention, mistake on my side. Fixing 2morr.

lib/charms/opensearch/v0/opensearch_distro.py
juditnovak: Redundant indeed. I'll remove it.

lib/charms/opensearch/v0/opensearch_distro.py
juditnovak: Valid call, sry, I overlooked (moe details on https://github.com/canonical/opensearch-operator/pull/412/files#r1752624651). Closing this one as solution referred in the other thread applies here.

lib/charms/opensearch/v0/opensearch_distro.py
juditnovak: Keeping this check after modifications above. Currenty it's not redundant.

It _is_ (technically) possible that we may have a config file, but there would no `node.attr.temp` (while it _did_ include `node.roles`, thus no exceptions above were triggered). 
Now it's accepted **not** to have a Deployment Description at this point. Critical data (`app`, `roles`) was retrieved from config, we're all good -- yet (optional/missable) temperature is missing from the config. See new unittests (clearly corner-cases).

Temperature value detection should _only_ fall back to Deployment Description in case the latter exists.

lib/charms/opensearch/v0/opensearch_distro.py
juditnovak: Done.

reneradoi (approved):

PR #423, opened 2024-08-31
[DPE-5339][UNITTESTS] Unittests for "Implement ca replacement routine unittests" task

juditnovak:
This PR is an extension to https://github.com/canonical/opensearch-operator/pull/353 adding Unit Tests.

Sadly the tests provide hardly more than "happy path" :-( (Given the volume I may still just stop here for now.)
Value delivered:
 - test for all deployments
 - detailed documentation of the workflow
 - clear demonstration of each step by 
   - unit role (leader/non-leader)
   - action taken together with involved data flow

The proposed change adds +7% test coverage to the original PR (that has 55%)
image

Additional changes:
  A bug was found along the way:
  A field such as `ca` will never exist in a secret as it is invlaid to define 2-character long fields. Thus the corresponding function should have always returned `False`.


juditnovak:
Tests added to target PR https://github.com/canonical/opensearch-operator/pull/353

PR #424, opened 2024-09-03
Enable cluster manager only nodes

reneradoi:
## Issue
Currently we always add the `data` role to a node if it is `cluster-manager`. This is required because otherwise the security index could not be initialized directly after startup of the first node.

## Solution
This PR drafts a solution for enabling "cluster-manager-only" nodes in large deployments. The workaround for adding the `data` role by default is removed.

The solution is implemented according this workflow:
- start `cluster-manager` -> no initialization of security index -> defer `start` event, set status to `blocked`
- start  `data` node -> initialize security index -> connect to `cluster-manager` node
- `cluster-manager` processes deferred `start` -> processes `post_start_init` -> cluster is formed

### DIFF
--- a/lib/charms/opensearch/v0/models.py
+++ b/lib/charms/opensearch/v0/models.py
@@ -275,8 +275,8 @@ class PeerClusterRelDataCredentials(Model):
     admin_password_hash: str
     kibana_password: str
     kibana_password_hash: str
-    monitor_password: str
-    admin_tls: Dict[str, Optional[str]]
+    monitor_password: Optional[str]
+    admin_tls: Optional[Dict[str, Optional[str]]]
     s3: Optional[S3RelDataCredentials]
 
 
@@ -286,6 +286,7 @@ class PeerClusterApp(Model):
     app: App
     planned_units: int
     units: List[str]
+    leader_host: str
     roles: List[str]
 
 
--- a/lib/charms/opensearch/v0/models.py
+++ b/lib/charms/opensearch/v0/models.py
@@ -286,6 +286,7 @@ class PeerClusterApp(Model):
     app: App
     planned_units: int
     units: List[str]
+    roles: List[str]
 
 
 class PeerClusterFleetApps(Model):
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -316,6 +316,8 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
 
         # apply the directives computed and emitted by the peer cluster manager
         if not self._apply_peer_cm_directives_and_check_if_can_start():
+            # todo: remove logging
+            logger.debug("Cannot start because of peer cluster manager directives.")
             event.defer()
             return
 
@@ -348,7 +350,18 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
 
         # request the start of OpenSearch
         self.status.set(WaitingStatus(RequestUnitServiceOps.format("start")))
-        self._start_opensearch_event.emit()
+
+        deployment_desc = self.opensearch_peer_cm.deployment_desc()
+        ignore_lock = (
+                    "data" not in deployment_desc.config.roles
+                    and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+                    or "data" in deployment_desc.config.roles
+                    and self.unit.is_leader()
+                    and deployment_desc.typ == DeploymentType.OTHER
+                    and not self.peers_data.get(Scope.APP, "security_index_initialised", False))
+#        ignore_lock = self.unit.is_leader() and "data" in self.opensearch_peer_cm.deployment_desc().config.roles and not self.peers_data.get(
+#                Scope.APP, "security_index_initialised", False)
+        self._start_opensearch_event.emit(ignore_lock=ignore_lock)
 
     def _apply_peer_cm_directives_and_check_if_can_start(self) -> bool:
         """Apply the directives computed by the opensearch peer cluster manager."""
@@ -356,8 +369,12 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             # the deployment description hasn't finished being computed by the leader
             return False
 
+        # todo: remove logging
+        logger.debug(f"deployment desc: {deployment_desc}")
         # check possibility to start
         if self.opensearch_peer_cm.can_start(deployment_desc):
+            # todo: remove logging
+            logger.debug("can start after checking deployment desc")
             try:
                 nodes = self._get_nodes(False)
                 self.opensearch_peer_cm.validate_roles(nodes, on_new_unit=True)
@@ -851,24 +868,6 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             # Retrieve the nodes of the cluster, needed to configure this node
             nodes = self._get_nodes(False)
 
-            cluster_apps = self.peers_data.get_object(Scope.APP, "cluster_fleet_apps")
-            # todo: remove logging
-            logger.debug(f"cluster apps: {cluster_apps}")
-
-            data_node_available = False
-            for app in cluster_apps.values():
-                cluster_app = PeerClusterApp.from_dict(app)
-                # todo: remove logging
-                logger.debug(f"app roles: {cluster_app.roles}")
-                if "data" in cluster_app.roles:
-                    data_node_available = True
-                    break
-
-            if not data_node_available:
-                logger.info("No data role available, deferring event.")
-                event.defer()
-                return
-
             # validate the roles prior to starting
             self.opensearch_peer_cm.validate_roles(nodes, on_new_unit=True)
 
@@ -886,13 +885,25 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             return
 
         try:
+            deployment_desc = self.opensearch_peer_cm.deployment_desc()
             self.opensearch.start(
                 wait_until_http_200=(
                     not self.unit.is_leader()
                     or self.peers_data.get(Scope.APP, "security_index_initialised", False)
                 )
             )
-            self._post_start_init(event)
+            if (
+                self.peers_data.get(Scope.APP, "security_index_initialised", False)
+                or self.unit.is_leader()
+                and "data" in deployment_desc.config.roles
+                and deployment_desc.typ == DeploymentType.OTHER
+                ):
+                self._post_start_init(event)
+            else:
+                logger.info("No data role available, deferring event post_start_init.")
+                self.node_lock.release()
+                event.defer()
+                return
         except (OpenSearchHttpError, OpenSearchStartTimeoutError, OpenSearchNotFullyReadyError):
             event.defer()
         except (OpenSearchStartError, OpenSearchUserMgmtError) as e:
@@ -900,12 +911,16 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             self.node_lock.release()
             self.status.set(BlockedStatus(ServiceStartError))
             event.defer()
+#
 
     def _post_start_init(self, event: _StartOpenSearch):  # noqa: C901
         """Initialization post OpenSearch start."""
         # initialize the security index if needed (and certs written on disk etc.)
-        if self.unit.is_leader() and not self.peers_data.get(
-            Scope.APP, "security_index_initialised"
+        # this happens only on the first data node to join the cluster
+        if (
+            self.unit.is_leader()
+            and "data" in self.opensearch_peer_cm.deployment_desc().config.roles
+            and not self.peers_data.get(Scope.APP, "security_index_initialised")
         ):
             admin_secrets = self.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
             self._initialize_security_index(admin_secrets)
@@ -1148,7 +1163,11 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             or not self.alt_hosts
         ):
             return (
-                self.unit.is_leader() and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+                self.unit.is_leader() and (
+                    deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+                    or deployment_desc.typ == DeploymentType.OTHER
+                    and "data" in deployment_desc.config.roles
+                )
             )
 
         # When a new unit joins, replica shards are automatically added to it. In order to prevent
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -1330,6 +1330,10 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             ):
                 computed_roles.append("data")
                 self.peers_data.put(Scope.UNIT, "remove-data-role", True)
+                nodes_config = self.peers_data.get_object(Scope.APP, "nodes_config")
+                for node in nodes_config.values():
+                    cluster_node = Node.from_dict(node)
+                    logger.info(f"Roles: {cluster_node.roles}")
         else:
             computed_roles = ClusterTopology.generated_roles()
 
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -47,7 +47,7 @@ from charms.opensearch.v0.helper_security import (
     generate_hashed_password,
     generate_password,
 )
-from charms.opensearch.v0.models import DeploymentDescription, DeploymentType
+from charms.opensearch.v0.models import DeploymentDescription, DeploymentType, PeerClusterApp
 from charms.opensearch.v0.opensearch_backups import backup
 from charms.opensearch.v0.opensearch_config import OpenSearchConfig
 from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
@@ -851,6 +851,24 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             # Retrieve the nodes of the cluster, needed to configure this node
             nodes = self._get_nodes(False)
 
+            cluster_apps = self.peers_data.get_object(Scope.APP, "cluster_fleet_apps")
+            # todo: remove logging
+            logger.debug(f"cluster apps: {cluster_apps}")
+
+            data_node_available = False
+            for app in cluster_apps.values():
+                cluster_app = PeerClusterApp.from_dict(app)
+                # todo: remove logging
+                logger.debug(f"app roles: {cluster_app.roles}")
+                if "data" in cluster_app.roles:
+                    data_node_available = True
+                    break
+
+            if not data_node_available:
+                logger.info("No data role available, deferring event.")
+                event.defer()
+                return
+
             # validate the roles prior to starting
             self.opensearch_peer_cm.validate_roles(nodes, on_new_unit=True)
 
@@ -1322,18 +1340,15 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             # "data" role in the provided roles, we need to add the role to be able to create
             # and store the security index
             # todo: rework: delay sec index init until 1st data node / handle red health
-            if (
-                self.unit.is_leader()
-                and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
-                and "data" not in computed_roles
-                and not self.peers_data.get(Scope.APP, "security_index_initialised", False)
-            ):
-                computed_roles.append("data")
-                self.peers_data.put(Scope.UNIT, "remove-data-role", True)
-                nodes_config = self.peers_data.get_object(Scope.APP, "nodes_config")
-                for node in nodes_config.values():
-                    cluster_node = Node.from_dict(node)
-                    logger.info(f"Roles: {cluster_node.roles}")
+            # if (
+                # self.unit.is_leader()
+                # and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+                # and "data" not in computed_roles
+                # and not self.peers_data.get(Scope.APP, "security_index_initialised", False)
+            # ):
+                # todo: remove below lines
+                # computed_roles.append("data")
+                # self.peers_data.put(Scope.UNIT, "remove-data-role", True)
         else:
             computed_roles = ClusterTopology.generated_roles()
 
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -23,6 +23,7 @@ from charms.opensearch.v0.constants_charm import (
     COSUser,
     OpenSearchSystemUsers,
     OpenSearchUsers,
+    PClusterNoDataNode,
     PeerClusterRelationName,
     PeerRelationName,
     PluginConfigChangeError,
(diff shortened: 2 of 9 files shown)


lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -203,3 +203,14 @@ def mask_sensitive_information(cmd: str) -> str:
     pattern = re.compile(r"(-tspass\s+|-kspass\s+|-storepass\s+|-new\s+|pass:)(\S+)")
 
     return re.sub(pattern, r"\1" + "xxx", cmd)
+
+
+def data_role_in_cluster_fleet_apps(charm: "OpenSearchBaseCharm") -> bool:
```
Mehdi-Bendriss, line 208: Could you move this one to `helper_cluster.ClusterTopology`?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -876,6 +884,12 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
             )
             self._post_start_init(event)
         except (OpenSearchHttpError, OpenSearchStartTimeoutError, OpenSearchNotFullyReadyError):
+            if "data" not in deployment_desc.config.roles:
```
Mehdi-Bendriss, line 887: does this mean that we always start the opensearch daemon no matter the existence of a data node? 
Why can't we do it conditionally to the content of the roles of the `peer-cluster-fleet` in the peer-cluster-relation-joined event of the providers?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -876,6 +884,12 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
...
+                self.status.set(BlockedStatus(PClusterNoDataNode))
+                self.node_lock.release()
+            if self.opensearch_peer_cm.is_provider():
```
Mehdi-Bendriss, line 890: I believe this should be `is_provider(typ="main")`

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1129,8 +1152,10 @@ def _can_service_start(self) -> bool:
             not self.peers_data.get(Scope.APP, "security_index_initialised", False)
             or not self.alt_hosts
         ):
-            return (
-                self.unit.is_leader() and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+            return self.unit.is_leader() and (
+                deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+                or deployment_desc.typ == DeploymentType.OTHER
+                and "data" in deployment_desc.config.roles
```
Mehdi-Bendriss, line 1158: I believe this check is too broad, as it also impacts subsequent data units / clusters way past the sec index initialisation

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -206,9 +206,12 @@ def remove_temporary_data_role(self):
     def add_seed_hosts(self, cm_ips: List[str]):
         """Add CM nodes ips / host names to the seed host list of this unit."""
         cm_ips_set = set(cm_ips)
-        with open(self._opensearch.paths.seed_hosts, "w+") as f:
-            lines = "\n".join([entry for entry in cm_ips_set if entry.strip()])
-            f.write(f"{lines}\n")
+
+        # only update the file if there is data to update
+        if cm_ips_set:
```
Mehdi-Bendriss, line 211: when can there ever be no seed_hosts?

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -154,7 +151,21 @@ def run_with_relation_data(
...
+
+        # create alternative list for seed_hosts cluster-manager-only-nodes
+        hosts = []
+        if all_apps := self._charm.peers_data.get_object(Scope.APP, "cluster_fleet_apps"):
+            for app in all_apps.values():
+                p_cluster_app = PeerClusterApp.from_dict(app)
+                if "cluster_manager" not in p_cluster_app.roles:
+                    continue
+
+                hosts.append(p_cluster_app.leader_host)
```
Mehdi-Bendriss, line 163: related to the previous question

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -433,24 +443,25 @@ def _rel_err_data(  # noqa: C901
...
-        elif not self.charm.secrets.get(Scope.APP, self.charm.secrets.password_key(COSUser)):
-            blocked_msg = f"'{COSUser}' user not created yet."
-        else:
-            try:
-                if not self._fetch_local_cm_nodes(deployment_desc):
-                    blocked_msg = f"No 'cluster_manager' eligible nodes found {message_suffix}"
-            except OpenSearchHttpError as e:
-                logger.error(e)
-                blocked_msg = f"Could not fetch nodes {message_suffix}"
+        elif "data" in deployment_desc.config.roles:
```
Mehdi-Bendriss, line 446: I don't think we should have this. This would prevent any dedicated CM cluster to ever run the below logic
I believe we should check on the existence of "data" in the peer-cluster-apps object given we have all visibility over the fleet

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -433,24 +443,25 @@ def _rel_err_data(  # noqa: C901
...
+            if not self.charm.is_admin_user_configured():
```
Mehdi-Bendriss, line 447: this is configurable without data nodes (simply file on disk)

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -433,24 +443,25 @@ def _rel_err_data(  # noqa: C901
...
+                blocked_msg = f"Admin user not fully configured {message_suffix}."
+            elif not self.charm.tls.is_fully_configured_in_cluster():
```
Mehdi-Bendriss, line 449: This is configurable without needing data nodes

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -348,7 +350,17 @@ def _on_start(self, event: StartEvent):
...
+
+        deployment_desc = self.opensearch_peer_cm.deployment_desc()
+        ignore_lock = (
+            "data" not in deployment_desc.config.roles
+            and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+            or "data" in deployment_desc.config.roles
+            and self.unit.is_leader()
+            and deployment_desc.typ == DeploymentType.OTHER
+            and not self.peers_data.get(Scope.APP, "security_index_initialised", False)
+        )
```
Mehdi-Bendriss, line 362: I think we should block / defer the start until after the first data node joins (we'll get notified through the large deployment relations)

Mehdi-Bendriss (commented):
Thanks René. I have mainly questions about synchronisation of the start sequence of all apps / nodes.

Can you confirm the following flow when there is no data node in the cluster:
1. when no data node in the application:
    1. no opensearch starts
    2. the start hook gets deferred endlessly  
3. when a data node joins - through large deployment relations:
    1. we become aware of it through the `fleet_apps` object
    2. we then start the leader unit of the main orchestrator 
    3. which then notifies the large deployment relations
    4. the leader unit of the data cluster starts too 
    5. start flow of the rest of the fleet resumes as usual

reneradoi:
Hey @Mehdi-Bendriss thank you for the feedback! This is very valuable, as the current design is different than your expections given below. 

Currently, when no data node is there, the `cluster-manager` does get started, but without running the `post_start_init`. That means it is there, but not initialized. The start hook will get deferred until the first data node is up and initialized (or endlessly, if that doesn't happen).

When a data node joins, it will currently start independently, initialize the security index and on that start, join the cluster (by contacting the already started cluster-manager). On the next re-emitted start hook, the cluster-manager will then fully come up.

I've chosen this design because I found the current code has a lot of situations on startup where, if the node is not fully up, the start event gets deferred and the update to the large deployment relation didn't happen. That's why I wanted to let them start independently and let the data node find the cluster once up.

I will try to rework that according to your given specification below.

Thanks!

> Thanks René. I have mainly questions about synchronisation of the start sequence of all apps / nodes.
> 
> Can you confirm the following flow when there is no data node in the cluster:
> 
> 1. when no data node in the application:
>    
>    1. no opensearch starts
>    2. the start hook gets deferred endlessly
> 2. when a data node joins - through large deployment relations:
>    
>    1. we become aware of it through the `fleet_apps` object
>    2. we then start the leader unit of the main orchestrator
>    3. which then notifies the large deployment relations
>    4. the leader unit of the data cluster starts too
>    5. start flow of the rest of the fleet resumes as usual

lib/charms/opensearch/v0/models.py
```diff
@@ -286,6 +286,8 @@ class PeerClusterApp(Model):
     app: App
     planned_units: int
     units: List[str]
+    leader_host: str
```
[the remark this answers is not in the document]
reneradoi: has been removed.

lib/charms/opensearch/v0/helper_charm.py
reneradoi: done.

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: has been updated.

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: done.

lib/charms/opensearch/v0/opensearch_config.py
reneradoi: has been adjusted so that `cm_ips` is always available.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
reneradoi: Has been removed, `cm_nodes` is now provided from the deployment desc and unit data if the nodes API is not available.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
reneradoi: adjusted.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
reneradoi: adjusted.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
reneradoi: adjusted.

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: adjusted.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -348,7 +355,17 @@ def _on_start(self, event: StartEvent):
...
         self.status.set(WaitingStatus(RequestUnitServiceOps.format("start")))
-        self._start_opensearch_event.emit()
+
+        deployment_desc = self.opensearch_peer_cm.deployment_desc()
+        ignore_lock = (
+            "data" not in deployment_desc.config.roles
+            and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+            or "data" in deployment_desc.config.roles
+            and self.unit.is_leader()
+            and deployment_desc.typ == DeploymentType.OTHER
```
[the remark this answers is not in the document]
reneradoi: This is only happening if the security index is not initialized and the unit is the leader, so is only valid for the first data node to start.

juditnovak (commented):
Guys, this is another VERY complex workflow we are adding here.

Could we provide at least a base description of the logic? I.e. documentation within the code as done before (CA rotation).

Also, for such a complexity, could we please have more than "happy path" testing...? (I'm glad to have it either as unit or as integration tests.)

If at least we could hold on to the earlier objective, that each PR would add 2 unittest to the testsuite...
(I volunteer to add them if no other way.)

reneradoi:
Hey @juditnovak! Thanks for the comment. I agree, these are not just "minor tweaks" or code adjustments. Sometimes this can get lost when being deeply invested into a topic.

> Could we provide at least a base description of the logic? I.e. documentation within the code as done before (CA rotation).
> 
> [...]
> 
> If at least we could hold on to the earlier objective, that each PR would add 2 unittest to the testsuite... (I volunteer to add them if no other way.)

I've added a comment explaining the workflow at where it starts, and also added a few unit tests to document the changed behaviour. Hope this is fine for you.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -346,9 +348,34 @@ def _on_start(self, event: StartEvent):
...
+            self.status.set(BlockedStatus(PClusterNoDataNode))
+            event.defer()
+            return
+
         # request the start of OpenSearch
         self.status.set(WaitingStatus(RequestUnitServiceOps.format("start")))
-        self._start_opensearch_event.emit()
+
+        # if this is the first data node to join, start without getting the lock
+        ignore_lock = (
```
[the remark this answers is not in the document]
reneradoi: Hey Pedro, thanks for your comment. I've been thinking about it a bit, and I don't think that the 2nd case (node lock not available) is an issue. If this happens, because no data node is online, the API-calls will not be successful, and in that case, the operator should fall back to the peer database lock (see [here](https://github.com/canonical/opensearch-operator/blob/main/lib/charms/opensearch/v0/opensearch_locking.py#L279-L287)).

skourta:
Nice work @reneradoi .
I tested the workflow and it follows exactly what @Mehdi-Bendriss described.
I have a couple of notes:
1. Once the main cluster manager node is up and running we add a failover node. The failover node status is set to "waiting" with the requesting lock message. This is misleading as it is actually blocked waiting for data nodes to join and the main cluster manager to be initialized.
2. When you deploy the data nodes they go in an active state/idle state with no message. This is also misleading as they are waiting to be integrated with the cluster manager. I think we should change the state and add a message clarifying what is happening.

phvalguima (changes requested):
~~check my next review~~

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -346,9 +348,34 @@ def _on_start(self, event: StartEvent):
...
+            "data" in deployment_desc.config.roles
+            and self.unit.is_leader()
+            and deployment_desc.typ == DeploymentType.OTHER
+            and not self.peers_data.get(Scope.APP, "security_index_initialised", False)
```
phvalguima, line 376:
```suggestion
            and (
                not self.peers_data.get(Scope.APP, "security_index_initialised", False)
                or (
                    self.peers_data.get(Scope.UNIT, "STARTED")
                    and not self.opensearch.is_service_started())
                )
            )
```

phvalguima (changes requested):
I am trying to get this PR aligned with our full cluster restart work on https://github.com/canonical/opensearch-operator/pull/407

This PR seems complementary: you are exclusively worried with the 1st data node coming up, whereas I am concerned that every restart, we will bring all CMs up.

For https://github.com/canonical/opensearch-operator/pull/407 to work with your proposal, then we need to extend a similar logic here, i.e. the 1st data node should be restart ignoring the lock IF:

we are still bootstrapping the cluster (i.e. security is not yet initialized)
OR if this is a node restart: (i) we have a "STARTED" tag on peer data and (ii) the service is powered down

lib/charms/opensearch/v0/opensearch_base_charm.py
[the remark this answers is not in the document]
phvalguima: @reneradoi well, the snippet you showed means we can have up to 2x nodes running (1st booting is the CM unit leader, 2nd is the next unit that tries to get the lock and sees a single CM up).

However, I cannot think of a scenario where this may be a problem right now... Only case that concerns me is if we have the entire data application powered down. But only case I can think of is a restart of the app itself. I will close this topic and I think we should focus instead on: https://github.com/canonical/opensearch-operator/pull/424#pullrequestreview-2302418430

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: Offline conversation: In order not to mix up PR's with different scopes and goals, this topic will be postponed and addressed in a separate task. Once both this PR and #407 are merged, this will be followed-up.

It will be tracked with https://warthogs.atlassian.net/browse/DPE-5460

phvalguima (dismissed):
LGTM

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1129,8 +1167,11 @@ def _can_service_start(self) -> bool:
             not self.peers_data.get(Scope.APP, "security_index_initialised", False)
             or not self.alt_hosts
         ):
-            return (
-                self.unit.is_leader() and deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+            return self.unit.is_leader() and (
+                deployment_desc.start == StartMode.WITH_GENERATED_ROLES
+                or deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+                or deployment_desc.typ == DeploymentType.OTHER
```
Mehdi-Bendriss, line 1173: I believe we can remove this line. A failover orchestrator can also have a "data" role

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1303,8 +1285,10 @@ def _initialize_security_index(self, admin_secrets: Dict[str, any]) -> None:
     def _get_nodes(self, use_localhost: bool) -> List[Node]:
         """Fetch the list of nodes of the cluster, depending on the requester."""
         # This means it's the first unit on the cluster.
-        if self.unit.is_leader() and not self.peers_data.get(
-            Scope.APP, "security_index_initialised", False
+        if (
+            self.unit.is_leader()
+            and "data" in self.opensearch_peer_cm.deployment_desc().config.roles
```
Mehdi-Bendriss, line 1290: should the first unit of the main orchestrator (when dedicated `cluster_manager`) also receive `[]` from this method?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -477,6 +495,19 @@ def _fetch_local_cm_nodes(self, deployment_desc: DeploymentDescription) -> List[
             use_localhost=self._opensearch.is_node_up(),
             hosts=self.charm.alt_hosts,
         )
+
+        if not nodes:
+            # create a node from the deployment desc and unit data only
+            return [
+                Node(
+                    name=self.charm.unit_name,
+                    roles=deployment_desc.config.roles,
```
Mehdi-Bendriss, line 504: We need to handle the case where the roles are autogenerated, this list will be None

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: I think that would result in a break of the workflow. Currently we try to get the online nodes, if not possible, we defer (see [here](https://github.com/canonical/opensearch-operator/blob/main/lib/charms/opensearch/v0/opensearch_base_charm.py#L940-L951), logic was not changed with this PR).

If we change that and return an empty list, we would instead raise a `OpenSearchNotFullyReadyError` and not defer / come back. Then the cluster-manager would never fully start up, and the later-starting data node could not join the cluster after initializing the security index.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -876,6 +898,9 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
             )
             self._post_start_init(event)
         except (OpenSearchHttpError, OpenSearchStartTimeoutError, OpenSearchNotFullyReadyError):
+            self.node_lock.release()
+            if self.opensearch_peer_cm.is_provider(typ="main"):
```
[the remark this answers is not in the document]
reneradoi: done

lib/charms/opensearch/v0/opensearch_base_charm.py

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
reneradoi: done

Mehdi-Bendriss (approved):

skourta (approved):

lib/charms/opensearch/v0/opensearch_base_charm.py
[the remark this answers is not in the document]
phvalguima: @reneradoi @Mehdi-Bendriss what happens if we have:
* MAIN_ORCHESTRATOR: CM only
* FAILOVER_ORCHESTRATOR: CM + single data cluster

I think, in this case, we'd never initialize the data cluster?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -173,6 +174,14 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):
             trigger_rel_id=event.relation.id,
         )
 
+        if (
+            deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
```
phvalguima, line 178: Future: for ticket [DPE-5460](https://warthogs.atlassian.net/browse/DPE-5460), we may want to just resolve this condition at `_on_start`.

phvalguima (commented):
I left one comment that we should discuss further. Otherwise, LGTM.

lib/charms/opensearch/v0/opensearch_base_charm.py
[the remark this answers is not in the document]
phvalguima:
```suggestion
            and deployment_desc.typ in [DeploymentType.OTHER, DeploymentType.FAILOVER]
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
phvalguima: PS.: this is for awareness only, I copied the link of this comment to the mentioned ticket.

lib/charms/opensearch/v0/opensearch_base_charm.py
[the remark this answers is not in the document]
reneradoi: > @reneradoi @Mehdi-Bendriss what happens if we have:
> 
> * MAIN_ORCHESTRATOR: CM only
> * FAILOVER_ORCHESTRATOR: CM + single data cluster
> 
> I think, in this case, we'd never initialize the data cluster?

Thanks for this comment @phvalguima it's a really good question! Thinking about the code again, I came to the conclusion that the start and initialization of this failover-node (cluster manager + data roles) will not be prevented. The added logic does not impact the existing capabilities to start such nodes, so the main-orchestrator that was started first should wait with starting for this node, but once the failover has been added, they should just form the cluster and that's it.

The logic here is only to allow non-cm-nodes to start without acquiring the lock, this will not prevent starting "regular" cm+data nodes.

We should add test coverage for that, but I can only do this next week because I will travel back now.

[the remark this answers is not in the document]
juditnovak: Could we just have a test for this case.... _Please_...?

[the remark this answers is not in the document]
phvalguima: @reneradoi ack, I will attach this discussion to our sync-up ticket.

phvalguima (approved):

PR #429, opened 2024-09-08
[DPE-5402] Fix wrong roles setting of "coordinating" node

Mehdi-Bendriss:
## Issue
This PR implements [DPE-5402](https://warthogs.atlassian.net/browse/DPE-5402) - namely this PR fixes:
- setting a wrong role `coordinating_only`
- not setting `node.roles: []` when only the `coordinating` role was set
- This PR also prevents the TLS the workflow to start on non-main orchestrator when the current cluster was **not a consumer** (which would create a mismatch in the APP_ADMIN tls values). 

[DPE-5402]: https://warthogs.atlassian.net/browse/DPE-5402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/charms/opensearch/v0/helper_cluster.py
+++ b/lib/charms/opensearch/v0/helper_cluster.py
@@ -37,7 +37,7 @@ class ClusterTopology:
     @staticmethod
     def generated_roles() -> List[str]:
         """Get generated roles for a Node."""
-        return ["data", "ingest", "ml", "coordinating_only", "cluster_manager"]
+        return ["data", "ingest", "ml", "cluster_manager"]
 
     @staticmethod
     def get_cluster_settings(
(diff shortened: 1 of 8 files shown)


phvalguima (approved):
LGTM

For the reference: OpenSearch always set the coordinating task to the node. If we leave the `node.roles: [}`, then it will only read cluster metadata and do request routing.

reneradoi (approved):
For context: The adjustments on TLS are a workaround as long as https://github.com/canonical/opensearch-operator/blob/feature/TLS-CA-rotation/lib/charms/opensearch/v0/opensearch_tls.py#L147-L160 is not merged into main yet.

Issue #244, opened 2024-04-19
Remove reference to proxy requests in backups


Mehdi-Bendriss:
The backups module makes use of a a proxy request method which handles empty, None, error responses in a special way.
We need to revisit this part and make the module converge with using exclusively the main request method.

phvalguima:
Hi @Mehdi-Bendriss do you mean actual proxy setup? Or a change on `s3-integrator`.

Mehdi-Bendriss:
@phvalguima i mean that the `_request` method in the backups module is a proxy function the `request` method of the `opensearch_distro` module

PR #435, opened 2024-09-10
[DPE-4196] Plugin Management Refactor

### LINKED ISSUE #252: [RFE] Extend plugin manager

phvalguima:
The main goal is to: (1) separate API call (/_cluster) between plugin-specific calls and cluster settings; (2) curb the requirements for restart for any plugin changes; (3) get a faster response time using cached entries wherever possible; (4) add new use-case with secret handling within the plugin logic itself; and (5) define models so we can standardize the plugin data exchanged between the objects and via relation.

### The config-only plugin

These are plugin configured via config options. In this case, it is only needed to add a new OpenSearchPlugin -child class that manages the config options to be added or removed from the cluster.

For example, opensearch-knn receives the config from the charm and returns the options to be set in the opensearch.yml.

## New Plugin Manager Infra

Now, the plugin manager is able to manage plugins that depend on config options, API calls and secrets. Whenever adding a new plugin, we should consider:

opensearch_plugins.py: this plugin should have a representation that is consumable by plugin_manager; it should be composed of all the configurations and keys to be added or removed to the cluster's main configuration
opensearch_plugin_manager.py: add the new plugin to the plugin dict; the manager must be able to instantiate this new plugin
opensearch_{plugin-name}.py: if the plugin is managed by a given relation, this lib will implement the relation manager and interface with OpenSearch's plugin-specific APIs
models.py: add any relation data model to this lib

## Secret Management

Each plugin that handles the specific secrets must implement the secret management logic in its operation. The goal is to avoid filling the opensearch_secrets.py methods with ifs for each plugin case and separating / isolating each plugin code.

## Adds Azure Backup Support

This PR builds on top of #531 and provides all the azure backup support and integration with the Blob store.

## Remove unneeded restarts and add caching

We ensure that any configuration changes that come from plugin management are applied via API before being persisted on config files. If the API responds with a 200 status, then we should only write the new value to the configuration and finish without a need for restart.

In case the service is down and API is not available, we can assume we will eventually start the service back up. In this case, it suffices to write the config entries to the files and leave to the next start to pick them up.

This task is going to be divided into 3x parts:

Addresses low ranging fruits where we reduce the number of restarts and add caching support
Three main actions: (i) Merge {add,delete}_plugin together and its equivalents in OpenSearchPluginConfig class; (ii) we receive one big dictionary where a key: None means we want to simply delete that entry; and (iii) the main OpenSearchKeystore must observe secret changes and update its values accordingly
Returns unit tests: this is going to be commented out whilst Parts 1 and 2 happen, given this part of the code was covered with extensive testing
The current implementation of plugin_manager.run waits for the cluster to be started before processing its config changed. We relax this demand and open to the option where the cluster is not yet ready, so we can modify the configuration without issuing a restart request.

https://github.com/canonical/opensearch-operator/issues/252 is closed with OpenSearchPluginRelationsHandler interface. It allows plugins to define how they will handle its relation(s). opensearch_backup module extends this interface and defines a checker to process either small or large deployments details.

Other relevant changes:

* Renaming method check_plugin_manager_ready to check_plugin_manager_ready_for_api
* Any plugin that needs to manage things via API call should check the health of the cluster using check_plugin_manager_ready_for_api
* Moving opensearch_distro.version to load the workload_version file we have present instead of an API call: this is two fold, 1. removes the dependency to a cluster to be ready and 2. makes this method in-sync with recent changes for upgrades logic
* Waive the need of loading the default settings if this particular unit is powered down: which makes sense, in this moment we can do any config changes as we will eventually powered it back up later
* If /_cluster/settings is available: apply the configs via API and do not add a restart request
* On config-changed handler, the upgrade_in_progress check gets precedence and will continuously defer config-changed event until upgrade is finished before calling the plugin manager
* Create a OpenSearchKeyStoreNotReadyYetError: responsible to identify the keystore has not been initialized yet across the cluster and hence, we cannot manage any plugins that use it; however, we always apply the opensearch.yml changes from that plugin
* Add cached_property whenever it makes sense, also adds logic to clean the cache if there was any relevant changes to its content.
* That still frees the config_changed to just call plugin_manager.run() before everything is set, as the run() method changes hard configuration only.

Closes https://github.com/canonical/opensearch-operator/issues/252, https://github.com/canonical/opensearch-operator/issues/280, #244


phvalguima:
Based on original reviews from: https://github.com/canonical/opensearch-operator/pull/282

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -387,94 +419,70 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+    @override
     def _on_s3_relation_broken(self, event: EventBase) -> None:
         """Processes the non-orchestrator cluster events."""
         self.charm.status.clear(S3RelMissing)
         if self.charm.unit.is_leader():
             self.charm.status.clear(S3RelShouldNotExist, app=True)
         logger.info("Non-orchestrator cluster, abandon s3 relation event")
-        return
-
-    def _on_s3_relation_action(self, event: EventBase) -> None:
```
phvalguima, line 450: Moved to the base class

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -825,36 +843,31 @@ def _on_s3_credentials_changed(self, event: EventBase) -> None:  # noqa: C901
...
+            logger.error(e)
+            return
         except OpenSearchError as e:
-            if isinstance(e, OpenSearchNotFullyReadyError):
-                logger.warning("s3-changed: cluster not ready yet")
-            else:
-                self.charm.status.set(BlockedStatus(PluginConfigError))
-                # There was an unexpected error, log it and block the unit
-                logger.error(e)
+            self.charm.status.set(BlockedStatus(PluginConfigError))
```
phvalguima, line 868: No reason to separate the exceptions here

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -486,10 +485,8 @@ def version(self) -> str:
...
-        try:
-            return self.request("GET", "/").get("version").get("number")
-        except OpenSearchHttpError:
-            logger.error(
-                "failed to get root endpoint, implying that this node is offline. Retry once node is online."
-            )
-            raise OpenSearchError()
+        # Will have a format similar to:
+        # Version: 2.14.0, Build: tar/.../2024-05-27T21:17:37.476666822Z, JVM: 21.0.2
+        output = self.run_bin("opensearch", "--version 2>/dev/null")
```
phvalguima, line 490: avoid running through the API call early in the plugin setup routine

tests/integration/plugins/test_plugins.py
```diff
@@ -522,11 +581,12 @@ async def test_knn_training_search(ops_test: OpsTest, deploy_type: str) -> None:
         await _wait_for_units(ops_test, deploy_type)
 
         # Now use it to compare with the restart
-        assert await is_each_unit_restarted(ops_test, APP_NAME, ts)
+        assert not await is_each_unit_restarted(ops_test, APP_NAME, ts)
```
phvalguima, line 584: Here we are now making sure the plugin manager did not trigger a restart

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -544,11 +546,13 @@ def _on_update_status(self, event: UpdateStatusEvent):
         if self.unit.is_leader():
             self.opensearch_exclusions.cleanup()
 
-            if (health := self.health.apply(wait_for_green_first=True)) not in [
+            if (health := self.health.apply(wait_for_green_first=False)) not in [
                 HealthColors.GREEN,
                 HealthColors.IGNORE,
             ]:
-                event.defer()
```
phvalguima, line 551: This defer brings no real benefit, as another `update-status` will forcefully happen.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -220,48 +226,50 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+    def _on_s3_relation_broken(self, event: EventBase) -> None:
+        """Defers the s3 relation broken events."""
+        raise NotImplementedError
+
     def _on_s3_relation_action(self, event: EventBase) -> None:
         """No deployment description yet, fail any actions."""
         logger.info("Deployment description not yet available, failing actions.")
         event.fail("Failed: deployment description not yet available")
 
-    def _request(self, *args, **kwargs) -> dict[str, Any] | None:
```
phvalguima, line 233: Closes #244: remove proxy method

tests/helpers.py
```diff
@@ -7,12 +7,25 @@
...
 from unittest.mock import patch
 
+import tenacity
 from cryptography import x509
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
 from cryptography.x509.oid import NameOID
 
 
+def patch_wait_fixed() -> Callable:
```
phvalguima, line 17: Speeds up tenacity by replacing the wait for a smaller range one

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -889,12 +923,12 @@ def apply_api_config_if_needed(self) -> None:
...
         if state != BackupServiceState.SUCCESS:
             logger.error(f"Failed to setup backup service with state {state}")
-            if self.charm.unit.is_leader():
-                self.charm.status.set(BlockedStatus(BackupSetupFailed), app=True)
             self.charm.status.clear(BackupConfigureStart)
-            return
-        if self.charm.unit.is_leader():
-            self.charm.status.clear(BackupSetupFailed, app=True)
+            self.charm.status.set(BlockedStatus(BackupSetupFailed))
+            self.charm.status.set(BlockedStatus(BackupSetupFailed), app=True)
```
phvalguima, line 928: Following conversation on: https://chat.canonical.com/canonical/pl/oqwsfdejufgepettcqzobnw7xh

It is neither guaranteed nor wanted by the entire team to support a: all units in a given status -> app status updated automatically

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -34,46 +34,25 @@ class OpenSearchKeystoreError(OpenSearchError):
     """Exception thrown when an opensearch keystore is invalid."""
 
 
+class OpenSearchKeystoreNotReadyYetError(OpenSearchKeystoreError):
+    """Exception thrown when the keystore is not ready yet."""
```
phvalguima, line 38: This error is thrown when we try to reach out to the to reload the keys via API and it fails.

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -412,38 +439,62 @@ def config(self) -> OpenSearchPluginConfig:
...
         pass
 
 
+class OpenSearchPluginDataProvider:
+    """Implements the data provider for any charm-related data access.
+
+    Plugins may have one or more relations tied to them. This abstract class
+    enables different modules to implement a class that can specify which
+    relations should plugin manager listen to.
+    """
```
zmraul, line 464: **question:** Does this class include large deployment relation?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -452,66 +503,141 @@ def name(self) -> str:
         return "opensearch-knn"
 
 
+class OpenSearchPluginBackupDataProvider(OpenSearchPluginDataProvider):
```
zmraul: **nit**: I would personally remove most of the `OpenSearch` prefix on these classes, since they are somewhat redundant and create noise when parsing or searching the file.

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -133,23 +112,51 @@ def __init__(self, charm):
...
+        if not os.path.exists(self.keystore):
+            raise OpenSearchKeystoreNotReadyYetError()
+
         if not entries:
             return  # no key/value to remove, no need to request reload of keystore either
         for key in entries:
             self._delete(key)
 
-    def list(self, alias: str = None) -> List[str]:
+    def update(self, entries: Dict[str, Any]) -> None:
```
zmraul, line 89: **nit**: Isn't there a lot of redundancy with these methods? I feel like a general `add|remove` or just `update` for all of them, instead of having 3 public + 2 internal methods.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -387,94 +408,70 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
             self.charm.on[S3_RELATION].relation_broken, self._on_s3_relation_broken
         )
 
+        for event in [
+            charm.on[PeerClusterRelationName].relation_joined,
+            charm.on[PeerClusterRelationName].relation_changed,
+            charm.on[PeerClusterRelationName].relation_departed,
```
zmraul: **question** Is `relation_departed` needed?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -220,48 +226,50 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
-            kwargs["timeout"] = 10
-        # We are interested to see the entire response
-        kwargs["resp_status_code"] = False
-        try:
-            result = self.charm.opensearch.request(*args, **kwargs)
-        except OpenSearchHttpError as e:
-            return e.response_body if e.response_body else None
-        return result if isinstance(result, dict) else None
-
     def _is_restore_in_progress(self) -> bool:
```
zmraul: **nit**: should this one be public as well? same as `is_backup_in_progress`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1082,5 +1092,8 @@ def backup(charm: "OpenSearchBaseCharm") -> OpenSearchBackupBase:
         # of cluster are we. Once we have that defined, then we will process.
         return OpenSearchBackupBase(charm)
     elif charm.opensearch_peer_cm.deployment_desc().typ == DeploymentType.MAIN_ORCHESTRATOR:
+        # Using the deployment_desc() method instead of is_provider()
+        # In both cases: (1) small deployments or (2) large deployments where this cluster is the
+        # main orchestrator, we want to instantiate the OpenSearchBackup class.
         return OpenSearchBackup(charm)
```
zmraul: **question**: shouldn't this one also return `OpenSearchBackup` when type is `DeploymentType.FAILOVER_ORCHESTRATOR`?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -452,66 +503,141 @@ def name(self) -> str:
...
-        )
-        """
-        if not self._extra_config.get("access-key") or not self._extra_config.get("secret-key"):
+        """Returns OpenSearchPluginConfig composed of configs used at plugin configuration."""
+        conf = self.data.credentials.dict()
+        # First, let's check if credentials are set
+        if any([val is None for val in conf.values()]):
             raise OpenSearchPluginMissingConfigError(
-                "Plugin {} missing: {}".format(
+                "Plugin {} missing credentials".format(
```
zmraul: This message is not informative enough. When integrating with s3, you can have both `access_key` and `secret_key` set on s3 integrator and still getting this message. After setting those, and adding eg `bucket` you get the more informative message from below that shows the missing fields.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -121,46 +112,16 @@ def get_plugin_status(self, plugin_class: Type[OpenSearchPlugin]) -> PluginState
...
-                    node
-                    for node in self._charm._get_nodes(True)
-                    if node.app.id == deployment_desc.app.id
-                ]
-            )
-            == self._charm.app.planned_units()
-            and self._charm.health.get()
-            in [HealthColors.GREEN, HealthColors.YELLOW, HealthColors.IGNORE]
-        )
+    def is_ready_for_api(self) -> bool:
```
zmraul, line 120: More of a design opinion here, but these checks should probably be centralized at some point. They are invoked from outside this file, which means that it is a relevant check for other components.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -399,33 +430,30 @@ def _needs_upgrade(self, plugin: OpenSearchPlugin) -> bool:
...
+            self._clean_cache_if_needed()
+
         except OpenSearchCmdError as e:
             if "not found" in str(e):
                 logger.info(f"Plugin {plugin.name} to be deleted, not found. Continuing...")
                 return False
             raise OpenSearchPluginRemoveError(plugin.name)
         return True
 
+    def _clean_cache_if_needed(self):
```
zmraul: **nit**: Not sure why this is needed. Clearing the cache on `self.plugins` will lead to `ConfigExposedPlugins` being read again, which is static, so no benefit to clearing. `self.plugins` is being used as a `read_only` property as far as I can see. 

For `_installed_plugins` I would make that a `@property` instead, and just evaluate every time.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -121,46 +112,16 @@ def get_plugin_status(self, plugin_class: Type[OpenSearchPlugin]) -> PluginState
...
+    def is_ready_for_api(self) -> bool:
+        """Checks if the plugin manager is ready to run API calls."""
+        return self._charm.health.get() not in [HealthColors.RED, HealthColors.UNKNOWN]
 
     def run(self) -> bool:
         """Runs a check on each plugin: install, execute config changes or remove.
 
         This method should be called at config-changed event. Returns if needed restart.
         """
+        is_manager_ready = True
```
zmraul: **question**: Shouldn't this method be gated by `is_ready_for_api`? It is calling API commands on `apply` and it's called from base-charm.

zmraul (commented):
Huge amount of work on this PR. Nice!

I've been running this branch locally and didn't find issues so far. Left some comments.

lib/charms/opensearch/v0/models.py
```diff
@@ -258,15 +260,98 @@ def set_promotion_time(cls, values):  # noqa: N805
 class S3RelDataCredentials(Model):
     """Model class for credentials passed on the PCluster relation."""
 
-    access_key: str = Field(alias="access-key")
-    secret_key: str = Field(alias="secret-key")
+    access_key: str = Field(alias="access-key", default=None)
```
Mehdi-Bendriss, line 283: when would this ever be None?

lib/charms/opensearch/v0/models.py
```diff
@@ -258,15 +260,98 @@ def set_promotion_time(cls, values):  # noqa: N805
...
+    """
+
+    bucket: str = Field(default="")
+    endpoint: str = Field(default="")
+    region: Optional[str] = None
+    base_path: Optional[str] = Field(alias="path", default=S3_REPO_BASE_PATH)
+    protocol: Optional[str] = None
+    storage_class: Optional[str] = Field(alias="storage-class")
+    tls_ca_chain: Optional[str] = Field(alias="tls-ca-chain")
+    credentials: S3RelDataCredentials = Field(alias=S3_CREDENTIALS, default=S3RelDataCredentials())
```
Mehdi-Bendriss, line 305: can you excplain the `default` here?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -220,48 +226,50 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
         ]:
             self.framework.observe(event, self._on_s3_relation_action)
 
+    def _on_secret_changed(self, event: EventBase) -> None:
```
Mehdi-Bendriss: should be marked as abstract

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -220,48 +226,50 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+
     def _on_s3_relation_event(self, event: EventBase) -> None:
         """Defers the s3 relation events."""
         logger.info("Deployment description not yet available, deferring s3 relation event")
         event.defer()
 
+    @abstractmethod
+    def _on_s3_relation_broken(self, event: EventBase) -> None:
+        """Defers the s3 relation broken events."""
+        raise NotImplementedError
```
Mehdi-Bendriss: a `pass` or `...` would be better here

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -220,48 +226,50 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+                or {}
+            )
+        except OpenSearchHttpError:
+            # Defaults to True if we have a failure, to avoid any actions due to
+            # intermittent connection issues.
+            logger.warning(
+                "_is_restore_in_progress: failed to get indices status"
+                " - assuming restore is in progress"
+            )
+            return True
```
Mehdi-Bendriss: I'm not sure I understand this path, why return true and assume a restore is in progress when it may not be, instead of let it crash?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -220,48 +226,50 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
          2) check for each index shard: for all type=SNAPSHOT and stage=DONE, return False.
         """
-        indices_status = self._request("GET", "/_recovery?human") or {}
+        try:
+            indices_status = (
+                self.charm.opensearch.request(
+                    "GET",
+                    "/_recovery?human",
+                    retries=6,
+                    timeout=10,
```
Mehdi-Bendriss, line 233: the timeout is too high, any specific reason why a request could take up to 10 sec to return?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -290,7 +298,12 @@ def _query_backup_status(self, backup_id: Optional[str] = None) -> BackupService
                 with attempt:
                     target = f"_snapshot/{S3_REPOSITORY}/"
                     target += f"{backup_id.lower()}" if backup_id else "_all"
-                    output = self._request("GET", target)
+                    output = self.charm.opensearch.request(
+                        "GET",
+                        target,
+                        retries=6,
+                        timeout=10,
```
Mehdi-Bendriss: similar comment on timeout

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -342,6 +359,9 @@ def get_service_status(  # noqa: C901
             return BackupServiceState.SNAPSHOT_RESTORE_ERROR_INDEX_NOT_CLOSED
         if type == "snapshot_restore_exception":
             return BackupServiceState.SNAPSHOT_RESTORE_ERROR
+        if type:
```
Mehdi-Bendriss: when do we expect to have a type not listed above? 
note: pattern matching is better suited for this count of `if`s.

lib/charms/opensearch/v0/models.py
```diff
@@ -258,15 +260,98 @@ def set_promotion_time(cls, values):  # noqa: N805
...
+    """Model class for the S3 relation data.
+
+    This model should receive the data directly from the relation and map it to a model.
+    """
+
+    bucket: str = Field(default="")
+    endpoint: str = Field(default="")
+    region: Optional[str] = None
+    base_path: Optional[str] = Field(alias="path", default=S3_REPO_BASE_PATH)
+    protocol: Optional[str] = None
```
Mehdi-Bendriss, line 302: can you explain why so many optionals and default=""?

lib/charms/opensearch/v0/models.py
```diff
@@ -258,15 +260,98 @@ def set_promotion_time(cls, values):  # noqa: N805
...
+            and not s3_creds.access_key
+            and not s3_creds.secret_key
+        ):
+            raise ValueError("Missing fields: access_key, secret_key")
+
+        # Both bucket and endpoint must be set, or not.
+        if values.get("bucket") and not values.get("endpoint"):
+            raise ValueError("Missing field: endpoint")
+        if values.get("endpoint") and not values.get("bucket"):
+            raise ValueError("Missing field: bucket")
```
Mehdi-Bendriss: these can be avoided by not making them optional

lib/charms/opensearch/v0/models.py
```diff
@@ -258,15 +260,98 @@ def set_promotion_time(cls, values):  # noqa: N805
...
+        if not conf:
+            return None
+
+        data = conf
+        if isinstance(conf, dict):
+            # We are
+            data = S3RelDataCredentials.from_dict(conf)
+
+        for value in data.dict().values():
+            if value.startswith("secret://"):
```
Mehdi-Bendriss, line 345: why is this needed? we can assume we receive well formatted secrets

lib/charms/opensearch/v0/models.py
```diff
@@ -258,15 +260,98 @@ def set_promotion_time(cls, values):  # noqa: N805
...
+    def get_endpoint_protocol(endpoint: str) -> str:
+        """Returns the protocol based on the endpoint."""
+        if not endpoint:
+            return "http"
+
+        if endpoint.startswith("http://"):
+            return "http"
+        if endpoint.startswith("https://"):
+            return "https"
+        return "https"
```
```suggestion

        return "https"
```

Can you explain why the default `https` when if no endpoint passed is default to `http`?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -366,18 +386,19 @@ def is_idle_or_not_set(self) -> bool:
         "idle": configured but there are no backups nor restores in progress.
         "not_set": set by the children classes
         """
-        return not (self.is_backup_in_progress() or self._is_restore_in_progress())
-
-    def manual_update(self, event: EventBase) -> None:
-        """Manually update the backup plugin config (credentials)"""
-        pass
+        return not (
+            self.is_set() or self.is_backup_in_progress() or self._is_restore_in_progress()
```
Mehdi-Bendriss: We need to test this path. What does `is_set()` represent? I do not see it anywhere in the code.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -366,18 +386,19 @@ def is_idle_or_not_set(self) -> bool:
...
 
 
 class OpenSearchNonOrchestratorClusterBackup(OpenSearchBackupBase):
     """Simpler implementation of backup relation for non-orchestrator clusters.
 
     In a nutshell, non-orchestrator clusters should receive the backup information via
     peer-cluster relation instead; and must fail any action or major s3-relation events.
+
+    This class means we are sure this juju app is a non-orchestrator. In this case, we must
+    manage the update status correctly if the user ever tries to relate the s3-credentials.
```
Mehdi-Bendriss: what should we do in `update-status`?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -387,94 +408,70 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+
+            # I've discussed this offline with @wallyworld and the main idea
+            # is that the unit is already aware of the secret change, why triggering
+            # a new hook in this case?
+            # Now, opensearch_backups.py was originally using opensearch_secrets.py to
+            # update its the s3-credentials secret and inform this class via "manual_update"
+
+            # Listening to the peer cluster relation is another alternative:
+            # Effectively it will call the common method that both _on_secret_changed and
+            # _on_peer_cluster_relation_event uses to update the keystore.
```
Mehdi-Bendriss: please summarize comments and keep them focused / take out usernames etc.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -387,94 +408,70 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+            self.framework.observe(event, self._on_peer_cluster_relation_event)
```
Mehdi-Bendriss, line 576: This is a complex event to listen to, if you see the business logic / validation that is involved in this hook, it is difficult to make safe assumptions as to the fact you're receiving the data you need. This was the reason we had secret changed invoke manual update. Otherwise we'll duplicate all the validation of this hook here.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -387,94 +408,70 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
-        except OpenSearchError as e:
-            logger.warning(
-                f"s3-changed: failed disabling with {str(e)}\n"
-                "repository-s3 maybe it was not enabled yet"
-            )
-        # It must be able to enable the plugin
+    def _on_peer_cluster_relation_event(self, event: EventBase) -> None:
+        """Processes the peer-cluster relation events."""
+        if not self.charm.secrets.get_object(Scope.APP, S3_CREDENTIALS):
+            logger.warning(f"Secret {S3_CREDENTIALS} found but missing s3-credentials set.")
```
Mehdi-Bendriss: This will not evaluate to true the first invocation of this hook. Given that this secret is only effectively stored at the end of this hook by the leader and will only effectively be true after a secrets changed flow.
In which case, why listen to it altogether?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -220,48 +226,50 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
          1) no restore requested: return False
          2) check for each index shard: for all type=SNAPSHOT and stage=DONE, return False.
         """
-        indices_status = self._request("GET", "/_recovery?human") or {}
+        try:
+            indices_status = (
+                self.charm.opensearch.request(
+                    "GET",
+                    "/_recovery?human",
+                    retries=6,
```
Mehdi-Bendriss, line 232: why `6` specifically? :D

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -387,94 +408,70 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+            # _on_peer_cluster_relation_event uses to update the keystore.
+            self.framework.observe(event, self._on_peer_cluster_relation_event)
+
     @override
-    def manual_update(self, event: EventBase) -> None:
-        """Processes the non-orchestrator cluster events."""
-        if not self.charm.plugin_manager.check_plugin_manager_ready():
-            logger.warning("s3-changed: cluster not ready yet")
+    def _on_secret_changed(self, event: EventBase) -> None:
+        """Clean secret from the plugin cache."""
```
Mehdi-Bendriss: wrong docstring ?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -387,94 +408,70 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+        secret = event.secret
+
```
Mehdi-Bendriss:
```suggestion
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -387,94 +408,70 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+            # Effectively it will call the common method that both _on_secret_changed and
+            # _on_peer_cluster_relation_event uses to update the keystore.
+            self.framework.observe(event, self._on_peer_cluster_relation_event)
+
     @override
-    def manual_update(self, event: EventBase) -> None:
-        """Processes the non-orchestrator cluster events."""
-        if not self.charm.plugin_manager.check_plugin_manager_ready():
-            logger.warning("s3-changed: cluster not ready yet")
+    def _on_secret_changed(self, event: EventBase) -> None:
```
Mehdi-Bendriss: Wrong type hint. Please review all type hints in this module. e.g: `_on_s3_relation_action`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -503,6 +509,7 @@ def _on_s3_relation_event(self, event: EventBase) -> None:
         if self.charm.opensearch_peer_cm.is_provider(typ="main"):
             self.charm.peer_cluster_provider.refresh_relation_data(event)
 
+    @override
     def _on_s3_relation_action(self, event: EventBase) -> None:
         """Just overloads the base method, as we process each action in this class."""
         pass
```
Mehdi-Bendriss: can we drop?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -484,17 +481,26 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = S3_RELATIO
...
             self.s3_client.on.credentials_changed, self._on_s3_credentials_changed
         )
         self.framework.observe(self.charm.on.create_backup_action, self._on_create_backup_action)
         self.framework.observe(self.charm.on.list_backups_action, self._on_list_backups_action)
         self.framework.observe(self.charm.on.restore_action, self._on_restore_backup_action)
 
+    @override
+    def _on_secret_changed(self, event: EventBase) -> None:
+        # This method is not needed anymore, as we already listen to credentials_changed event.
+        pass
```
Mehdi-Bendriss: can we drop?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -650,7 +661,15 @@ def is_idle_or_not_set(self) -> bool:
...
-        output = self._request("GET", f"_snapshot/{S3_REPOSITORY}")
+        try:
+            output = self.charm.opensearch.request(
+                "GET",
+                f"_snapshot/{S3_REPOSITORY}",
+                retries=6,
+                timeout=10,
+            )
+        except OpenSearchHttpError as e:
+            return e.response_body if e.response_body else None
```
Mehdi-Bendriss: this violates the signature of the method, this is not a boolean nor an exception. What should this output be used for?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -889,12 +922,12 @@ def apply_api_config_if_needed(self) -> None:
...
         if state != BackupServiceState.SUCCESS:
             logger.error(f"Failed to setup backup service with state {state}")
-            if self.charm.unit.is_leader():
-                self.charm.status.set(BlockedStatus(BackupSetupFailed), app=True)
             self.charm.status.clear(BackupConfigureStart)
-            return
-        if self.charm.unit.is_leader():
-            self.charm.status.clear(BackupSetupFailed, app=True)
+            self.charm.status.set(BlockedStatus(BackupSetupFailed))
+            self.charm.status.set(BlockedStatus(BackupSetupFailed), app=True)
```
Mehdi-Bendriss: this will fail if not leader unit

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -889,12 +922,12 @@ def apply_api_config_if_needed(self) -> None:
...
+            raise OpenSearchBackupError()
+        self.charm.status.clear(BackupSetupFailed)
+        self.charm.status.clear(BackupSetupFailed, app=True)
```
Mehdi-Bendriss: this will fail if not leader unit.

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: what if the next event is a non update status related event?

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -133,23 +112,51 @@ def __init__(self, charm):
...
+
         if not entries:
             return  # no key/value to add, no need to request reload of keystore either
         for key, value in entries.items():
             self._add(key, value)
 
     def delete(self, entries: List[str]) -> None:
         """Removes a given key from "opensearch" keystore."""
+        if not os.path.exists(self.keystore):
+            raise OpenSearchKeystoreNotReadyYetError()
```
Mehdi-Bendriss: why not return?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -52,16 +56,18 @@
...
     },
     "repository-s3": {
         "class": OpenSearchBackupPlugin,
         "config": None,
-        "relation": "s3-credentials",
     },
 }
 
 
+class OpenSearchPluginManagerNotReadyYetError(OpenSearchPluginError):
```
Mehdi-Bendriss, line 72:
```suggestion
class OpenSearchPluginManagerNotReadyError(OpenSearchPluginError):
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -190,9 +151,27 @@ def run(self) -> bool:
...
+            except OpenSearchKeystoreNotReadyYetError:
+                # Plugin manager must wait until the keystore is to finish its setup.
+                # This separated exception allows to separate this error and process
+                # it differently, once we have inserted all plugins' configs.
+
+                # Store the error and continue
+                # We want to apply all configuration changes to the cluster and then
+                # inform the caller this method needs to be reran later to update keystore.
+                # The keystore does not demand a restart, so we can process it later.
+                is_manager_ready = False
```
Mehdi-Bendriss: why is the keystore the only decisive factor about the manager being ready or not?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -190,9 +151,27 @@ def run(self) -> bool:
...
+                logger.debug(f"Finished Plugin {plugin.name} waiting for keystore")
+            else:
+                logger.debug(f"Finished Plugin {plugin.name} status: {self.status(plugin)}")
         logger.info(f"Plugin check finished, restart needed: {restart_needed}")
 
+        if not is_manager_ready:
+            # Next run, configurations above will not change, as they have been applied, and
+            # only the missing keystore will be set.
+            raise OpenSearchKeystoreNotReadyYetError()
```
Mehdi-Bendriss: same question as above

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -261,13 +241,7 @@ def _configure_if_needed(self, plugin: OpenSearchPlugin) -> bool:
...
         """If disabled, removes plugin configuration or sets it to other values."""
         try:
-            if self._user_requested_to_enable(plugin) or self.status(plugin) not in [
-                PluginState.ENABLED,
-                PluginState.WAITING_FOR_UPGRADE,
-            ]:
-                # Only considering "INSTALLED" or "WAITING FOR UPGRADE" status as it
-                # represents a plugin that has been installed but either not yet configured
-                # or user explicitly disabled.
+            if self.status(plugin) != PluginState.DISABLING_NEEDED:
```
Mehdi-Bendriss, line 249: perhaps more clear `DISABLING_REQUESTED` ?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -277,14 +251,16 @@ def _compute_settings(
         self, config: OpenSearchPluginConfig
     ) -> Tuple[Dict[str, str], Dict[str, str]]:
         """Returns the current and the new configuration."""
+        if not self._charm.opensearch.is_node_up() or not self.is_ready_for_api():
```
Mehdi-Bendriss, line 259: is node_up can be part of `is_read_for_api`, no?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -618,7 +627,7 @@ def _close_indices_if_needed(self, backup_id: int) -> Set[str]:
     def _restore(self, backup_id: int) -> Dict[str, Any]:
         """Runs the restore and processes the response."""
         backup_indices = self._list_backups().get(backup_id, {}).get("indices", {})
-        output = self._request(
+        output = self.charm.opensearch.request(
             "POST",
             f"_snapshot/{S3_REPOSITORY}/{backup_id.lower()}/_restore?wait_for_completion=true",
```
Mehdi-Bendriss, line 759: backup_id is an integer.

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -27,9 +27,13 @@ class and implementing the necessary extra logic.
 
 The plugin lifecycle runs through the following steps:
 
+
 MISSING (not installed yet) > INSTALLED (plugin installed, but not configured yet) >
-ENABLED (configuration has been applied) > WAITING_FOR_UPGRADE (if an upgrade is needed)
-> ENABLED (back to enabled state once upgrade has been applied)
+ENABLING_NEEDED (the user requested to be enabled, but not configured yet) >
```
Mehdi-Bendriss, line 32:
```suggestion
ENABLING_REQUESTED (the user requested to be enabled, but not configured yet) >
```

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -412,38 +439,62 @@ def config(self) -> OpenSearchPluginConfig:
...
+        pass
+
+    @abstractmethod
+    def get_data(self) -> Dict[str, Any]:
+        """Returns the data from the relation databag.
+
+        Exceptions:
+            ValueError: if the data is not valid
+        """
+        raise NotImplementedError
```
Mehdi-Bendriss, line 482: not needed

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -452,66 +503,141 @@ def name(self) -> str:
...
+    """
+
+    def __init__(self, charm):
+        """Creates the OpenSearchPluginBackupDataProvider object."""
+        super().__init__(charm)
+        self._relation = None
+        if not self._charm.opensearch_peer_cm.deployment_desc():
+            # Temporary condition: we are waiting for CM to show up and define which type
+            # of cluster are we. Once we have that defined, then we will process.
+            raise OpenSearchPluginMissingConfigError("Missing deployment description in peer CM")
```
Mehdi-Bendriss, line 526: how is this exception caught?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -452,66 +503,141 @@ def name(self) -> str:
...
+    def get_data(self) -> Dict[str, Any]:
+        """Returns the data from the relation databag.
+
+        Exceptions:
+            ValueError: if the data is not valid
+        """
+        if not self.get_relation():
+            return {}
+        result = dict(self.get_relation().data[self._relation.app]) or {}
+        if not self.is_main_orchestrator:
```
Mehdi-Bendriss, line 549: no proxy method please.

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -452,66 +503,141 @@ def name(self) -> str:
...
+            # Peer relations exchange secrets via peer-cluster secret
+            result |= self._charm.secrets.get_object(Scope.APP, S3_CREDENTIALS) or {}
```
Mehdi-Bendriss, line 551: we read the same secret on the opensearch_backups module. Looks like we have duplication we need to remove

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -452,66 +503,141 @@ def name(self) -> str:
...
 
+    MODEL = S3RelData
+    MANDATORY_CONFS = [
+        "bucket",
+        "endpoint",
+        "region",
+        "base_path",
+        "protocol",
+        "credentials",
+    ]
```
Mehdi-Bendriss: similar comment on duplication, pydantic should already perform the required validation.

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -452,66 +503,141 @@ def name(self) -> str:
...
+    DATA_PROVIDER = OpenSearchPluginBackupDataProvider
```
Mehdi-Bendriss: why a class variable?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -452,66 +503,141 @@ def name(self) -> str:
...
     """Manage backup configurations.
 
     This class must load the opensearch plugin: repository-s3 and configure it.
+
+    The plugin is responsible for managing the backup configuration, which includes relation
+    databag or only the secrets' content, as backup changes behavior depending on the juju app
+    role within the cluster.
     """
 
+    MODEL = S3RelData
```
Mehdi-Bendriss, line 565: why a class variable?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -647,7 +648,16 @@ def _set_security_conf(self, data: PeerClusterRelData) -> None:
...
+            self.charm.secrets.put_object(
+                Scope.APP, S3_CREDENTIALS, s3_creds.to_dict(by_alias=True)
+            )
+        else:
+            # Set the S3 credentials to empty
+            self.charm.secrets.put_object(
+                Scope.APP,
+                S3_CREDENTIALS,
+                S3RelDataCredentials().to_dict(by_alias=True),
+            )
```
Mehdi-Bendriss: can you explain why?

lib/charms/opensearch/v0/opensearch_plugins.py
```diff
@@ -452,66 +503,141 @@ def name(self) -> str:
...
                     self.name,
-                    [
-                        conf
-                        for conf in ["access-key", "secret-key"]
-                        if not self._extra_config.get(conf)
-                    ],
                 )
             )
 
+        if self.dp.is_main_orchestrator:
```
Mehdi-Bendriss, line 605: this is too smart for the plugins - which is supposed to be dumb. The heavy lifting should have been on the data provider and only inject to the plugin what it needs.

Mehdi-Bendriss (commented):
Thanks Pedro. I left some comments. 
It seems to me that we are making the plugin components too smart as opposed to the previous implementation, which I believe was clearer and had better separation of concerns.
The large deployments workflow needs to be carefully analyzed

lib/charms/opensearch/v0/models.py
phvalguima: If the user did not set that yet via action, this model will return an empty content.

lib/charms/opensearch/v0/models.py
phvalguima: Missing configs from the s3-integrator

lib/charms/opensearch/v0/models.py
phvalguima: Missing configs from the s3-integrator

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Well, there are a few answers here:
1) Update status [can be set to arbitrarily long values or even disabled](https://juju.is/docs/juju/list-of-model-configuration-keys#heading--update-status-hook-interval), hence, we should never be doing here anything other than keeping the status
2) On an one year span, update-status will outpace any other type of event, if we start deferring them we may have multiple update-status happening at once
3) The update-status sets a time threshhold, if our hooks take too long, then we accumulate update-status. If our update-status is taking too long (e.g. because of deferred previous update-status) then we will have an endless loop

What is the benefit this deferral brings?

lib/charms/opensearch/v0/opensearch_keystore.py
phvalguima: I recall we needed this signal to be dealt on upper classes, hence the exception vs. return allowed me to "bypass" all the way to these classes.

phvalguima: Maybe we can just return a bool, worth checking

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
phvalguima: This is the case we are missing s3 information, then we just create an empty object to signal that.

lib/charms/opensearch/v0/opensearch_keystore.py
phvalguima: I have checked and indeed we need an exception. We raise this exception all the way down here but we only really catch it on the base charm itself,  on `self.plugin_manager.run()` call. Also, this method would have three possible answers: "yes, removed", "not removed, key not found", or "error, we are not ready yet as keystore does not exist".

lib/charms/opensearch/v0/opensearch_keystore.py
phvalguima: yes, indeed the `add` and `delete` are redundant. We can get away with update only.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: @zmraul can you give more precise references? AFAIU other places are asking if the opensearch is healthy, I am asking if it is responsive.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: If no keystore, then some plugins cannot work. This check allows us to collect this case, run all the other plugins who can continue executing and then report back there was at least one plugin with a problem.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: Maybe this is more a naming issue. The idea here was more `is_keystore_ready`.

lib/charms/opensearch/v0/models.py
phvalguima: Same as the above.

lib/charms/opensearch/v0/models.py
phvalguima: It is a safety check.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: My lucky number? Lol, no idea why I put it, I guess it was just a typo of "I want 5, pressed 6"

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: But indeed, I agree this name is bad... I am changing it to `is_keystore_ready`.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: Yeah, I agree. But the thing here is that I need some "factory" method that differs between MAIN or OTHERS...

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: Not directly, but that is the idea. This class exists to abstract the access to the databags from basic plugins. The plugin should be "dumb", i.e. just v basic dataclasses. This class provides the plugins with any relation info.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: **Yes**. Reason: if you have s3 keys in the opensearch-keystore but you do not have backup configured, then the application breaks. Therefore, we need to know when the relation is gone.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: Changed the name, check if that makes more sense.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Block and state a message.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Sure, but my goal here is exactly to break the big peer-cluster handler out. That method is massive and I cannot really grasp what it does on each case. There is no real risk of *listening* to a complex event if we do not defer, as we are not necessarily making changes to its databag (hence, we are not *triggering* peer-cluster-changed).

Now, to your point, I noticed we defer.

So, I will first try by removing that defer.

The other option is to listen to this event, then we raise a custom event that is focused only on processing peer-cluster relations changes *for backups*. This way, we can defer this small event, internal, backup event instead of the peer cluster.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: > This will not evaluate to true the first invocation of this hook.

Why not? Can you provide more details why this `if` will not be true for every `peer-cluster-changed` call?

> Given that this secret is only effectively stored at the end of this hook by the leader

Why?

> In which case, why listen to it altogether?

Simply put, we are only interested in the `secret-changed`. The way we manage secrets is by receiving these secrets via peer-cluster relation and then setting them in the app-level. Now, that triggers secret-changed or secret-removed in some units.

As my comment previously says: unit-level secrets do not trigger secret-changed, nor app-level secrets trigger secret-changed in the app leader. Therefore, the only real way for us to load the secrets coming from the orchestrator *on every unit* is to listen to the peer cluster.

The secret-changed truly only exists to call the `secret.get_content(refresh=True)` and get the secret content updated.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: All calls of `apply_api_config_if_needed` are gated previously: if we are not the leader, we already returned. Only the leader manages the APIs.

phvalguima: I will add an explicit `if` there.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: Changed the names, check it again.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: Yes, I think this particular change predates our fix for the API calls where we iterate over the multiple addresses, instead of always sending the same API request to the same unit on every retry... I think it is safe to just have the is_ready_for_api().

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: So, that is explicitly tested in our CI. It is a string [as stated here](https://github.com/canonical/opensearch-operator/blob/2/edge/actions.yaml#L61), as I understand it [from upstream](https://github.com/opensearch-project/OpenSearch/blob/98dbc4a31cce52e2553523f07ee25c04a71a62b3/plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/async/UploadRequest.java#L23). The annotation is wrong, I will fix it.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: It is caught on the main touch points of the `openseach_plugins`: the `self.plugin_manager.run()` and in backups, we catch it on `self.charm.plugin_manager.apply_config()`.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: They are very tightly coupled. The idea is to have each subclass defining its own provider type. To simplify it, I am stating a class and then the upstream classes creating the plugin do not need to worry to create 2x objects instead of one.

lib/charms/opensearch/v0/models.py
phvalguima: I am just following our principle of TLS first. Here, if the user is not explicitly they want to expose their data via HTTP, we enforce the HTTPS.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: I remember now, it was 6x 10s, which gives an entire minute. And that is the time out period we use over our health checks on the cluster.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: I remember now, it was 6x 10s, which gives an entire minute. And that is the time out period we use over our health checks on the cluster.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: It was 6x 10s, which gives an entire minute. And that is the time out period we use over our health checks on the cluster.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Not really, we actually call this method if we do not know yet what deployment-description we will get but we did get a secret changed. In this case, we call and pass right away.

This is an empty method which is called in certain special cases, not an abstract.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Actually, this is not supposed to be an abstract. We use the Base class.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: I kept the exception handling so we can add an error log here. I will reraise this exception however.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Well, if we have already a match pattern, then I always like to add the final "default" check. In this case, the final default check grabs returns unknown.

I am removing the `if type` reference and leaving as a default catch in this match.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: So, update-status makes sense actually. What we could have is an observer for `update-status` that checks if wrong relations exist. If yes, then we update the status.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Thinking this through, indeed relation_departed is not needed in this case.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Sure, but we should adopt pyright here

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Not really, otherwise the behavior of base will dictate what to do with the action. And hence, we will get event fail every time.

We are overloading so we can just dismiss this handler now that we know we are the MAIN orchestrator.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: True, this method is linked to the upgrade. if we cannot define we have an idle backup system, then we should be conservative and return False.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: We do not really care in this case, they all receive instructions via peer-cluster. If we do have a failove,r then that cluster will run another hook later down the road which will call the backup system and this time be the "OpenSearchBackup" class.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: So, that message continues there, check the if right after:
```
        if self.dp.is_main_orchestrator:
```

This IF ensures we only do this more thorough check if we are the main orchestrator. And that makes sense, as the main orchestrator is the one that first interacts with the S3. The remaining will only receive the data IF you have passed via peer relation, which will only happen if you had all the contents in the first place.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: I guess you mean the `ready_for_api` right? This is a proxy method to check the health actually. So, if we received an answer and the cluster is minimally healthy, then we can proceed.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
phvalguima: True, makes sense. I will remove that method and move the _installed_plugin to be managed as a property. The real cache we need to watch for is the cluster_config, which is deleted somewhere else.

Removed this method.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
zmraul: The ready for API one, yes.

zmraul (dismissed):
Nice! After some discussions around this PR, my opinion is that it successfully enables integrating plugins into the codebase keeping the following items in a centralized development environment:
- Plugins that are configured via relations
- Secrets that need to be created for storing credentials (related to plugins specifically)
- Large deployment data abstraction (including passing secrets around)
-----------------
#### Follow up work

There are some leftover tasks that there should be looked at once this is merged. I'm writing them down here, but I'll create issues for these once merged:
- Look at default settings on `<Plugin|Relation>RelData(Model)` classes. These defaults are not adding value, since the validator should be the main point where checks happen. But just removing defaults interacts with Plugin `data` and `config` methods, so I would rather have specific work done on this to make sure validation is well done.
- Some extra `abstraction | constant` is needed around s3 (and azure) secrets, since there is too many literal strings being used around, which makes following the flow of data hard within the code.

lib/charms/opensearch/v0/models.py
phvalguima: for the `S3RelData`, we are keeping the defaults. For `S3RelDataCredentials` and also `AzureRelDataCredentials`, default is `None`.

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: This is gone now

lib/charms/opensearch/v0/opensearch_plugins.py
phvalguima: This is gone later

Mehdi-Bendriss (approved):
We will merge this one - as we wait for https://github.com/canonical/opensearch-operator/pull/541 to follow and address the remaining feedback points.

zmraul (approved):
Re-approving.

Issue #417, opened 2024-08-27
Long delay on restart of units after each cert renewal following a CA renewal


Mehdi-Bendriss:
After a successful CA renewal, 2 issues occur:
1. After certificate renewal - for +4 minutes the following error occurs: 
 ```
unit-main-0: 15:14:59 ERROR unit.main/0.juju-log opensearch-peers:1: Cannot connect to the OpenSearch server...
unit-main-0: 15:15:00 ERROR unit.main/0.juju-log opensearch-peers:1: [Errno 111] Connection refused
unit-main-0: 15:15:03 ERROR unit.main/0.juju-log opensearch-peers:1: [Errno 111] Connection refused
unit-main-0: 15:15:06 ERROR unit.main/0.juju-log opensearch-peers:1: [Errno 111] Connection refused
unit-main-0: 15:15:09 ERROR unit.main/0.juju-log opensearch-peers:1: [Errno 111] Connection refused
unit-main-0: 15:15:12 ERROR unit.main/0.juju-log opensearch-peers:1: [Errno 111] Connection refused
unit-main-0: 15:15:15 DEBUG unit.main/0.juju-log opensearch-peers:1: Getting secret app:admin-password
unit-main-0: 15:15:15 DEBUG unit.main/0.juju-log opensearch-peers:1: Starting new HTTPS connection (1): 10.122.32.198:9200
unit-main-0: 15:15:15 DEBUG unit.main/0.juju-log opensearch-peers:1: Error when checking if host 10.122.32.198 is up: HTTP error self.response_code=None
self.response_text="HTTPSConnectionPool(host='10.122.32.198', port=9200): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1007)')))"
unit-main-0: 15:15:18 DEBUG unit.main/0.juju-log opensearch-peers:1: Getting secret app:admin-password
unit-main-0: 15:15:18 DEBUG unit.main/0.juju-log opensearch-peers:1: Starting new HTTPS connection (1): 10.122.32.198:9200
unit-main-0: 15:15:18 DEBUG unit.main/0.juju-log opensearch-peers:1: Error when checking if host 10.122.32.198 is up: HTTP error self.response_code=None
[...]

PR #436, opened 2024-09-11
[DPE-5283] stabilization for CA rotation and integration test for large deployments

skourta:
This PR addresses the stabilization of the CA rotation workflow.

The following issues are fixed:
1) Error with `No cert in -in file '/tmp/tmp1uvvhlyo.cert' matches private key [xxx] certificate routines:X509_check_private_key:key values mismatch:../crypto/x509/x509_cmp.c:405:`: The private key was regenerated on every admin certificate request. When a CA rotation is restarting the nodes, the certificate that was being loaded was generated using an old private key and can no longer be validated with the stored key (see https://github.com/canonical/opensearch-operator/issues/417).

2) A long delay when restarting after renewing the TLS certificates on CA rotation (see https://github.com/canonical/opensearch-operator/issues/417).

3) An issue with the CA truststore file being updated/cleaned too early (before all nodes have been restarted in rolling fashion), which caused the respective node not to be able to join the cluster anymore after restarting.

4) An issue with updating the pem-file too early. This file is used for API requests by the operator to the node on which it is stored. If this file is updated too early, the operator can not communicate with the node anymore. The same if it is updated too late in the workflow.

5) An issue with deferring `CertificateAvailableEvents` on node initialization which caused the deployment in setups with `manual tls-operator` to end in endless deferrals and never to start correctly.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -245,15 +245,19 @@ class OpenSearchTLS(Object):
             cert_type, self.charm.secrets.get_object(scope, cert_type.val)
         )
 
-        # store the admin certificates in non-leader units
-        if not self.charm.unit.is_leader():
-            if self.all_certificates_available():
-                self.store_admin_tls_secrets_if_applies()
-
-        # apply the chain.pem file for API requests only if the CA cert has not been updated
-        admin_secrets = self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)
+        # apply the chain.pem file for API requests, only if the CA cert has not been updated
+        admin_secrets = self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val) or {}
         if admin_secrets.get("chain") and not self._read_stored_ca(alias="old-ca"):
             self.update_request_ca_bundle()
+
+        # store the admin certificates in non-leader units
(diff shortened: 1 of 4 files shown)


lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -152,8 +157,8 @@ def _on_tls_relation_created(self, event: RelationCreatedEvent) -> None:
                 Scope.APP, CertType.APP_ADMIN, CertType.APP_ADMIN.val
             )
 
-            if not admin_cert:
-                self._request_certificate(Scope.APP, CertType.APP_ADMIN)
+            # if not admin_cert:
```
reneradoi, line 160: I think you can remove this line.

reneradoi (commented):
Hey Smail, I've seen that the unit tests didn't pass. This was due to the test data, as we were just having a random string as `key` in the unit test, which then caused the error with `incorrect padding` when being parsed.

I've added this as a commit directly to this PR, as this test as just recently been added and fixing it would maybe have been a bit to much to ask from you :)

Now let's see if the integration tests also pass.

skourta:
I'm also still testing with the manual tls operator. I think there's still a bug there I need to work out.

reneradoi (approved):

Mehdi-Bendriss (approved):

PR #442, opened 2024-09-15
[DPE-4880] TLS CA rotation in large deployments

reneradoi:
This PR adds integration testing for TLS CA rotation in large deployments as well as the functionality to enable it, by setting the flags for ongoing TLS CA rotation not only to the `opensearch-peers` relation databag, but also to the relations `peer-cluster` and `peer-cluster-orchestrator`. By that, all related units in a large deployment broadcast their current TLS CA rotation status to the relation databags.


phvalguima (changes requested):
Nice! I left a comment in the test coverage, besides checking the CA, I'd recommend trying to get the `ContinuousWrites` class to load the new CA as well and explicitly checking it.

Mehdi-Bendriss (commented):
Thanks René! I noticed a bit of an odd behavior in the CI run logs, that indicates that the a unit stayed in `Waiting for TLS to be configured...` for over 4min (20:32:28 - 20:36:39)
Any idea what that might be due to?

reneradoi:
> Thanks René! I noticed a bit of an odd behavior in the CI run logs, that indicates that the a unit stayed in `Waiting for TLS to be configured...` for over 4min (20:32:28 - 20:36:39) Any idea what that might be due to?

Looks like this some odd log output from the `wait_until` method, it only logs the first application it can find, not all. According to the logs, at this time the `opensearch-data-0` unit was updated.

tests/integration/tls/test_ca_rotation.py
```diff
@@ -99,3 +112,113 @@ async def test_rollout_new_ca(ops_test: OpsTest) -> None:
...
+            FAILOVER_APP: {"active": []},
+        },
+        units_statuses=["active"],
+        wait_for_exact_units={app: units for app, units in APP_UNITS.items()},
+        timeout=2400,
+        idle_period=IDLE_PERIOD,
+    )
+
+    more_writes = await c_writes.count()
+    await c_writes.stop()
```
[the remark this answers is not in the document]
reneradoi: added to the test

skourta (approved):

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1142,6 +1142,10 @@ def _post_start_init(self, event: _StartOpenSearch):  # noqa: C901
         if self.opensearch_peer_cm.is_provider():
             self.peer_cluster_provider.refresh_relation_data(event, can_defer=False)
 
+        # before resetting the CA rotation state, we remove the old ca from the truststore
+        if self.peers_data.get(Scope.UNIT, "tls_ca_renewed", False):
+            self.tls.remove_old_ca()
```
Mehdi-Bendriss, line 1147: This should only happen if all nodes / all applications have finished rotating the certificates after a CA rotation. I think we need an additional flag on the peer-cluster relation data if it's a large deployment, no?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1142,6 +1142,10 @@ def _post_start_init(self, event: _StartOpenSearch):  # noqa: C901
         if self.opensearch_peer_cm.is_provider():
             self.peer_cluster_provider.refresh_relation_data(event, can_defer=False)
 
+        # before resetting the CA rotation state, we remove the old ca from the truststore
+        if self.peers_data.get(Scope.UNIT, "tls_ca_renewed", False):
```
Mehdi-Bendriss, line 1146: I think we should instead check on `ca_rotation_complete_in_cluster` to ensure that all nodes / all applications have finished rotating the certificates after a CA rotation, no?

Mehdi-Bendriss (approved):
Thanks René - approved with a question

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: Yes, currently the `old-ca` would be removed from the truststore even before all cluster-members have updated the new certs. But the truststore file would only be loaded on another restart, so it does not get active immediately.

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: But that would never happen at the same time, or we would need to defer the `_StartOpenSearch` event here.

PR #445, opened 2024-09-17
[DPE-5483] Shuffle URLs we try to connect to

phvalguima:
Currently, the "retrial" mechanism of the `request()` method always target the very same unit, even if we specified more than one option. 

This PR applies the retrial to the different targets of the list.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_distro.py
+++ b/lib/charms/opensearch/v0/opensearch_distro.py
@@ -25,6 +25,7 @@ from charms.opensearch.v0.helper_charm import (
 )
 from charms.opensearch.v0.helper_cluster import Node
 from charms.opensearch.v0.helper_conf_setter import YamlConfigSetter
+from charms.opensearch.v0.helper_http import error_http_retry_log
 from charms.opensearch.v0.helper_networking import get_host_ip, is_reachable
 from charms.opensearch.v0.models import App, StartMode
 from charms.opensearch.v0.opensearch_exceptions import (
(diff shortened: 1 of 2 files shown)


lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -247,17 +247,18 @@ def request(  # noqa
...
 
-        def call(url: str) -> requests.Response:
+        def call(urls: List[str]) -> requests.Response:
             """Performs an HTTP request."""
             for attempt in Retrying(
                 retry=retry_if_exception_type(requests.RequestException)
                 | retry_if_exception_type(urllib3.exceptions.HTTPError),
                 stop=stop_after_attempt(retries),
                 wait=wait_fixed(1),
-                before_sleep=error_http_retry_log(logger, retries, method, url, payload),
```
phvalguima, line 257: WIP: removed for now, I think we can replace the `url` for `urls` here and log all the options available...

phvalguima: PS.: this should be a debug log level: if we got to the point of asking ourselves which retry is failing, then something is going wrong...

The actual, final error, should then be a `error` level as it does prevent the functioning of the charm.

skourta (approved):

juditnovak (approved):
In my opinion we should have a `property` on `OpenSearchDistro` that's referring to the available endpoints. Instead of the snippet in `request` to retrieve them.

But that's just an opinion, and doesn't have to be included in this PR :-)

Mehdi-Bendriss (approved):

PR #447, opened 2024-09-18
[DPE-4359] TLS ca/certificate rotation without downtime

skourta:
## Issue
We currently make use of PEM files for the:

- CA
- Certificates (admin, unit-http/unit-transport)
- private keys

This is fine in all cases but the case where the CA is rotated, in which case downtime will be incurred as soon as the first unit gets a new CA and restarts.

We need to switch from individual certificate resources to:

- java keystores: for private keys, admin/unit-http-transport certificates 
- java trust stores: for CA

We need to implement the whole 2-stages rolling restart routine for this case.  
### Implementation:

Ensure that:

1. TLS certificates rotation on expiration works as expected without downtime

2. TLS CA certificates rotation on expiration works as expected without downtime

3. TLS certificates rotation works as expected after a CA rotation, without downtime

4. For both simple and large deployments 

### Testing:

- unit tests 

- integration tests coverage for the 4 previous cases

### Documentation:
Document on charmhub the whole TLS rotation workflow for both deployment modes:
- simple 
- large


phvalguima (approved):
This is a feature branch that has been worked and tested on for some time. LGTM

Mehdi-Bendriss (approved):
Pending a rebase.

phvalguima (approved):

juditnovak (approved):
Finally.... +5% unittest coverage :-)

reneradoi (approved):

PR #452, opened 2024-09-20
[DPE-5496] Release 2.17

phvalguima:
Update opensearch to 2.17. Also fixes a race condition in our upgrade tests.


tests/integration/upgrades/test_small_deployment_upgrades.py
```diff
@@ -276,7 +276,7 @@ async def test_upgrade_rollback_from_local(
         await wait_until(
             ops_test,
             apps=[app],
-            apps_statuses=["blocked"],
+            apps_statuses=["active"],
```
phvalguima, line 279: This `blocked` allows an early escape from the `wait_until`, before the rollback is completed. Juju will fail the refresh command then if there are any units still executing.

juditnovak (approved):
Latest changes (upgrade tests) look good now.

paulomach (approved):
impressive

PR #455, opened 2024-09-22
[DPE-5536][test_upgrades.py] CI - Gracefully fail refresh cmd error

### LINKED ISSUE #454: [CI] Intermittent failures in upgrade tests
Now and then, our upgrade tests fail with the following message:
```
ERROR "opensearch" deploy incomplete, please try refresh again in a little bit.
```

That comes from juju directly, [as it is defined in its cmd](https://github.com/juju/juju/blob/4584c53183a1b128ce0ffbe5f4e3c4b6235290bc/cmd/juju/application/refresh.go#L371). In this case, we should add a `Retry` loop to any of the refresh calls we have in these tests.

Latest CI failing: https://github.com/canonical/opensearch-operator/actions/runs/10980795554/job/30487099742?pr=453

phvalguima:
Currently, our CI is failing intermittently whenever we call `juju refresh`, [as the controller is still busy downloading the charm](https://github.com/juju/juju/blob/4584c53183a1b128ce0ffbe5f4e3c4b6235290bc/cmd/juju/application/refresh.go#L371)

To avoid failing due to this intermittent error, we add a retry loop at refresh call.

Closes #454

### DIFF
--- a/tests/integration/upgrades/test_small_deployment_upgrades.py
+++ b/tests/integration/upgrades/test_small_deployment_upgrades.py
@@ -3,8 +3,6 @@
 # See LICENSE file for licensing details.
 
 import logging
-import subprocess
-from typing import Optional
 
 import pytest
 from pytest_operator.plugin import OpsTest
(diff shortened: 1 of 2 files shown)


tests/integration/upgrades/test_small_deployment_upgrades.py
```diff
@@ -61,6 +63,39 @@
 #  Auxiliary functions
 #
 #######################################################################
+async def _refresh(
```
Mehdi-Bendriss, line 66: maybe externalise it in a helper since all other upgrade related tests will benefit from it

Mehdi-Bendriss (commented):
looks good, i think it needs to be applied to all upgrade tests

skourta (approved):

PR #456, opened 2024-09-24
[DPE-5299] Check deployment_desc() available before security index init

### LINKED ISSUE #419: Initialize security index at `relation-broken` -- `AttributeError: 'NoneType' object has no attribute 'config'`
## Steps to reproduce

1. Had a cluster instllation at rev140
2. After reboot the cluster was blocked on lock (see screenshot on MM chat https://chat.canonical.com/canonical/pl/sk3tqsiudfby5n5grhpm7x3zuw)
3. I issued `juju remove-application opensearch`

What I observe is a single unit left standing out of a 3-units cluster with the error as on the screenshot below.

## Actual behavior

image

image

## Versions

image

phvalguima:
Currently, if we have a single node cluster, we may end up on a removal whilst still starting the cluster. Due to LP#2076599, that will cause a failure due to empty `deployment_desc()` response.

Closes #419

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -1042,6 +1042,10 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
                 logger.debug(f"Error when initializing the security index: {e.out}")
                 event.defer()
                 return
+            except OpenSearchStartError:
+                logger.debug("Cluster not ready yet: missing deployment-description")
+                event.defer()
+                return
 
         # it sometimes takes a few seconds before the node is fully "up" otherwise a 503 error
         # may be thrown when calling a node - we want to ensure this node is perfectly ready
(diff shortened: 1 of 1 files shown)


skourta (approved):

reneradoi (changes requested):
Hey @phvalguima, it's a good idea to make sure the `deployment_desc` is there before initializing the security index. But I think it would be better to check this earlier. Before calling the `_initialize_security_index()` method there is a condition in `_post_start_init()` [with checks on the deployment_desc](https://github.com/canonical/opensearch-operator/blob/5104c6d0be887a37958138286b20f50b135d3e0e/lib/charms/opensearch/v0/opensearch_base_charm.py#L1032-L1034). This is where it should already fail, if no `deployment_desc` is available. And even earlier, in `_start_opensearch()`, there is [a check for missing](https://github.com/canonical/opensearch-operator/blob/5104c6d0be887a37958138286b20f50b135d3e0e/lib/charms/opensearch/v0/opensearch_base_charm.py#L915) `deployment_desc`, referring to the same bug. Maybe we should make sure we don't run into this bug there?

phvalguima:
@reneradoi I am being overzealous, indeed :)

Meaning, that IF may never really be necessary. Let's do the following, I will close this PR for now, and close the bug. If we have another occurrence, then we revive this proposal.

PR #458, opened 2024-09-26
[DPE-5558] Break CA rotation into integration test groups

phvalguima:
Currently, we are having a lot of time outs in CA rotation testing. Breaking between small and large deployments and parallelized runners will help with that overall duration. 

Pointing to `main` as switching back to `2/edge` also brings the changes from: https://github.com/canonical/opensearch-operator/pull/461


tests/unit/lib/test_opensearch_tls.py
```diff
@@ -235,6 +235,11 @@ def test_on_relation_created_non_admin(self, _, __, _request_certificate, deploy
             state=DeploymentState(value=State.ACTIVE),
         )
         event_mock = MagicMock()
+        self.secret_store.put_object(
```
phvalguima: this was not directly added by me... But may have come when switching target branches...

skourta (approved):
LGTM

reneradoi (approved):
Thanks Pedro for splitting it up!

PR #460, opened 2024-09-26
[DPE-5460] restart data only nodes without lock

reneradoi:
## Issue
In Large Deployments with cluster-manager-only nodes and data-only nodes, a deadlock might appear if the whole data-application has been powered down and needs to restart. In this case, starting the data node should be possible without acquiring the lock first, because the lock index is not available.

## Solution
Start the leader of the data-nodes without acquiring the lock if it was previously started but is not running anymore.

The PR also adds unit test coverage for the case where a cluster-manager-only has not initialized the security index and a failover-orchestrator with data-role joins and wants to start (see [here](https://github.com/canonical/opensearch-operator/pull/424#discussion_r1768132249) for context).


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -412,7 +412,15 @@ def cleanup():
...
-            and not self.peers_data.get(Scope.APP, "security_index_initialised", False)
+            and (
+                not self.peers_data.get(Scope.APP, "security_index_initialised", False)
+                or (
+                    # in case all data-nodes are powered down after being previously started
+                    # ignore the lock to get a data-node started, as it holds security index
+                    self.peers_data.get(Scope.UNIT, "STARTED")
+                    and not self.opensearch.is_service_started()
+                )
+            )
```
phvalguima: The big question I have here is: should we run the `_post_start_init` or not, after a reboot?

My immediate thought is: we should not. First, because if the unit is running and marked as "STARTED", `_post_start_init` had already been called successfully. Second, because that routine has a multiple checks against OpenSearch itself (e.g. voting exclusions API calls to erase its own exclusion). These calls are not guaranteed to work if we had a full cluster restart.

The voting exclusions API is particularly interesting: you can manage it as long as you have a quorum of CMs, given voting exclusions are set on cluster-level metadata. Our lock is an index, hence we *also* need at least a data node online. We cannot guarantee neither will be available right after a full cluster restart.

Therefore, I would recommend not to call `_post_start_init`, which means not calling `self._start_opensearch_event.emit`, even if we have `ignore_lock=True`. If we are okay with that, then we should move this `or` logic above to the same `elif` in line 324.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -412,7 +412,15 @@ def cleanup():
             "data" in deployment_desc.config.roles
             and self.unit.is_leader()
             and deployment_desc.typ == DeploymentType.OTHER
-            and not self.peers_data.get(Scope.APP, "security_index_initialised", False)
+            and (
```
phvalguima, line 415: This is unrelated to this line, but just caught it. We are missing a check on line 390. Currently, this logic leaves us vulnerable to [LP#2076599](https://launchpad.net/bugs/2076599). Can we have instead:
```
if not (deployment_desc := self.opensearch_peer_cm.deployment_desc()):
    event.defer()
    return
```

reneradoi: No need to check this here, is already covered in line 359 with `_apply_peer_cm_directives_and_check_if_can_start()`, which checks if the `deployment_desc` is available.

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: I agree with your assumption. So the only change here would be to remove the condition `and "cluster_manager" in self.opensearch.roles` in line 326, which was added by your PR recently. What was the reason for restricting this `opensearch.start_service_only()` to cluster-managers?

phvalguima: > I agree with your assumption. So the only change here would be to remove the condition and "cluster_manager" in self.opensearch.roles in line 326

I would recommend instead:
```
and any([role in deployment_desc.config.roles for role in ["cluster_manager", "voting_only", "data"]])
```
That also covers the "voting_only" which I was missing originally.

>  which was added by your PR recently. What was the reason for restricting this opensearch.start_service_only() to cluster-managers?

So, my reasoning was: we need to do something similar to `systemctl enable ...`. Systemd will always power your service ON if the node is rebooted. However, we do not want it to be powered ON IF we rebooted the machine BUT it was not yet "STARTED".

The logic for `start_service_only()` is: if we have all the conditions, then "just start", equivalent to what `systemctl enable` would do, just start the actual service even if we are missing other peers. We had a node reboot and things will eventually settle to their original state.

That logic will also leave the `start` event right after starting. Hence, we are not really hoping everything is settled right now but we know whatever *was* started will eventually get back online.

reneradoi:
Assigned to @phvalguima to follow up on the discussion if the leader of the data-only-nodes should be running `_post_start_init` when rebooting or not.

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: After offline conversation with @Mehdi-Bendriss, the preferred way to go is to run the full `_post_start_init` on the restarting data-node (after the crash/when rebooting), if possible, to make sure to have a fully-functional cluster again after restarting.

So I've conducted a test with the PR as it is now, with the following steps:
- create a large deployment on an AWS EC2 (lxd cloud) with:
  - 1 cluster manager (`--config roles=cluster_manager`)
  - 3 data nodes (`--config init_hold=True --config roles=data`)
  - 1 failover (`--config init_hold=True --config roles=cluster_manager`)
- reboot the EC2 instance (`sudo reboot`)
- wait for the cluster to come back online

Result is: it worked fine. The cluster manager and failover nodes came back immediately (starting only the service), and the data nodes came back with a full `_post_start_init`. They formed a fully-functional cluster and all nodes where online.

So I suggest we continue with the implementation as it currently is in this PR.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -412,7 +412,15 @@ def cleanup():
...
+                not self.peers_data.get(Scope.APP, "security_index_initialised", False)
+                or (
+                    # in case all data-nodes are powered down after being previously started
+                    # ignore the lock to get a data-node started, as it holds security index
+                    self.peers_data.get(Scope.UNIT, "STARTED")
```
skourta: This should be lowercase

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: As we discussed @reneradoi if you force reboot the machines of the data nodes then they will be blocked and the ignore mechanism will not work. Here are the steps I used:
1. Deploy a large deployment:
```
juju deploy self-signed-certificates
juju deploy ./opensearch_ubuntu-22.04-amd64.charm --series jammy --show-log --verbose --config cluster_name="app" --config roles="cluster_manager"
juju deploy ./opensearch_ubuntu-22.04-amd64.charm opensearch-failover --series jammy --show-log --verbose --config cluster_name="app" --config roles="cluster_manager" --config init_hold="true"
juju deploy ./opensearch_ubuntu-22.04-amd64.charm opensearch-data -n 2 --series jammy --show-log --verbose --config cluster_name="app" --config roles="data" --config init_hold="true"
```
2. Integrate everything:
```
juju integrate opensearch self-signed-certificates                                  
juju integrate opensearch-failover self-signed-certificates
juju integrate opensearch-data self-signed-certificates
```
```
juju integrate opensearch-failover:peer-cluster opensearch:peer-cluster-orchestrator
juju integrate opensearch-data:peer-cluster opensearch:peer-cluster-orchestrator
juju integrate opensearch-data:peer-cluster opensearch-failover:peer-cluster-orchestrator
```
3. On idle force reboot the machines containing the data nodes:
```
juju ssh opensearch-data/1 "sudo reboot" ; juju ssh opensearch-data/0 "sudo reboot"
```

Upon initial investigation the `started` flag seems not to be set for the unit which causes the condition to be falsy.

skourta (changes requested):

skourta (dismissed):

skourta (approved):

Mehdi-Bendriss (approved):

PR #463, opened 2024-09-27
[DPE-3654] Peer cluster data stored in a secret

skourta:
## Issue
The data of the peer cluster relation was stored in the app data bag in text format including passwords and sensitive data.

## Solution
Move the data of the peer cluster relation to a juju secret


Mehdi-Bendriss (changes requested):
Thanks Smail. 

Having the rel data as a whole as part of the secrets makes us lose valuable information from the relation.

I would rather: 
- we keep the relation flow as it was and just replace the plaintext values of the passwords etc.. by their secret labels in the relation data.
- grant access to all these secrets to the consumers
- consumer is able to fetch the content from these labels 
- when a secret changes, we'll have to trigger a rel changed event for the consumers to see it (or do we if the secrets are already granted?)

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -64,7 +65,7 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
```
reneradoi: Updating is not required.

lib/charms/opensearch/v0/opensearch_secrets.py
```diff
@@ -40,7 +40,7 @@
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
```
reneradoi: Updating is not required.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -519,6 +537,55 @@ def _fetch_local_cm_nodes(self, deployment_desc: DeploymentDescription) -> List[
             if node.is_cm_eligible() and node.app.id == deployment_desc.app.id
         ]
 
+    def _rel_data_redacted_dict(self, rel_data: PeerClusterRelData) -> dict[str, Any]:
+        """Convert the secret data to a dict for storage in a secret."""
+        # If we use the values of the secrets then the size of the secret
+        # would be too large for juju we store the secret ids instead and
+        # fetch the secrets when needed in the requirer side
+        secrets = self.charm.secrets
```
reneradoi: Should be moved to the initialization of the class.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -960,3 +1026,72 @@ def _clear_errors(self, *error_labels: str):
...
+
+
+def rel_data_from_redacted_str(
+    charm: "OpenSearchBaseCharm", redacted_dict_str: str
+) -> PeerClusterRelData | None:
+    """Construct the peer cluster rel data from the secret data."""
+    content = json.loads(redacted_dict_str)
+    credentials = content["credentials"]
+
+    secrets = charm.secrets
```
reneradoi: Should be moved to the initialization of the class.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -806,11 +872,11 @@ def _cm_nodes(self, orchestrators: PeerClusterOrchestrators) -> List[Node]:
...
                 continue
 
-            data = self.get_obj_from_rel(key="data", rel_id=rel_id)
-            if not data:  # not ready yet
+            redacted_dict_str = self.get_from_rel(key="data", rel_id=rel_id, remote_app=True)
+            if not redacted_dict_str:  # not ready yet
                 continue
 
-            data = PeerClusterRelData.from_dict(data)
+            data = rel_data_from_redacted_str(self.charm, redacted_dict_str)
```
reneradoi: You could use the PeerClusterRelData.from_dict() method here to create the structure. Therefore return the updated data from `rel_data_from_redacted_str()` to here and load it via the PeerClusterRelData.from_dict() method.

reneradoi (changes requested):
Hey Smail, thanks for tackling this topic! I think it needs some adjustments as we are currently mixing using the `opensearch_secrets.py` class and accessing secrets directly through the model, especially in the `_grant_rel_data_secrets()` method. This should be avoided, if possible, and methods missing the in `opensearch_secrets` class should be added there. This is to make sure we have a clean interface with the secrets, and don't have to adjust multiple places in the code of the ops-implementation of secrets changes.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: This is outside the class because it is also used in the `opensearch_peer_clusters` module

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -519,6 +545,59 @@ def _fetch_local_cm_nodes(self, deployment_desc: DeploymentDescription) -> List[
...
+        credentials = rel_data_secret_content["credentials"]
+        for key, secret_id in credentials.items():
+            # s3 and admin-username are not secrets
+            if key == "admin_username":
+                continue
+
+            secret = self.model.get_secret(id=secret_id)
+            for rel_id in all_rel_ids:
+                if relation := self.get_rel(rel_id=rel_id):
+                    secret.grant(relation)
```
reneradoi: The secrets are accessed directly via the ops module. I think it would be possible to have a method in `opensearch_secrets.py` to get the secret from a given id and then use grant a secret to a relation there.

That would help to have all secret related methods in that class and use this as a clean interface.

reneradoi (dismissed):
Thanks for applying all the changes I asked for!

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -376,18 +395,25 @@ def _s3_credentials(
                 return None
 
             # As the main orchestrator, this application must set the S3 information.
+            access_key = self.charm.backup.s3_client.get_s3_connection_info().get("access-key")
+            secret_key = self.charm.backup.s3_client.get_s3_connection_info().get("secret-key")
+
+            # set the secrets in the charm
+            self.charm.secrets.put(Scope.APP, "s3-access-key", access_key)
+            self.charm.secrets.put(Scope.APP, "s3-secret-key", secret_key)
```
phvalguima, line 415: We should revisit and reformat this code once s3 relation starts supporting relation secrets. Can you add a `# TODO` so we can come back and remove this secret?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -519,6 +552,72 @@ def _fetch_local_cm_nodes(self, deployment_desc: DeploymentDescription) -> List[
...
+            redacted_dict["credentials"]["admin_tls"] = admin_tls
+
+        if (
+            rel_data.credentials.s3
+            and rel_data.credentials.s3.access_key
+            and rel_data.credentials.s3.secret_key
+        ):
+            redacted_dict["credentials"]["s3"] = {
+                "access-key": self.secrets.get_secret_id(Scope.APP, "s3-access-key"),
+                "secret-key": self.secrets.get_secret_id(Scope.APP, "s3-secret-key"),
```
phvalguima, line 609: Likewise, this secret should be caught straight from the s3 relation, once it supports granting secrets around.

phvalguima (dismissed):
I am approving it, but if we have to do any changes on this PR still, @skourta, can you add the `TODO` notes as described? Also, we should make sure the backup tests are passing in our CI, as they depend on the secret exchanges.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -525,10 +529,10 @@ def rel_data(self) -> Optional[PeerClusterRelData]:
         rel = self._charm.model.get_relation(
             PeerClusterOrchestratorRelationName, orchestrators.main_rel_id
         )
-        if not (data := rel.data[rel.app].get("data")):
+        if not (redacted_dict_str := rel.data[rel.app].get("data")):
```
Mehdi-Bendriss: let's keep it simple, and just name it `data` - given that the new data format is with "redacted" secrets

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -590,3 +594,61 @@ def _deployment_type(config: PeerClusterConfig, start_mode: StartMode) -> Deploy
             if not config.init_hold
             else DeploymentType.FAILOVER_ORCHESTRATOR
         )
+
+    def rel_data_from_redacted_str(self, redacted_dict_str: str) -> PeerClusterRelData | None:
```
Mehdi-Bendriss: i don't think this method can return `None`, a renaming may also make sense here:
```suggestion
    def rel_data_from_str(self, redacted_dict_str: str) -> PeerClusterRelData:
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -260,6 +262,13 @@ def refresh_relation_data(
         # compute the data that needs to be broadcast to all related clusters (success or error)
```
Mehdi-Bendriss, line 271:
```suggestion
        # compute the data that needs to be broadcast to all related clusters (success or error)
        # if rel_data is an error, prepare to broadcast it to all related clusters
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -260,6 +262,13 @@ def refresh_relation_data(
...
         rel_data = self._rel_data(deployment_desc, orchestrators)
 
+        # if rel_data is an error, prepare to broadcast it to all related clusters
```
```suggestion
        # if rel_data is NOT an error, we will replace the plaintext credentials in 
        # the object, with their corresponding secret IDs
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -519,6 +552,72 @@ def _fetch_local_cm_nodes(self, deployment_desc: DeploymentDescription) -> List[
             if node.is_cm_eligible() and node.app.id == deployment_desc.app.id
         ]
 
+    def _rel_data_redacted_dict(self, rel_data: PeerClusterRelData) -> dict[str, Any]:
+        """Convert the secret data to a dict for storage in a secret."""
```
```suggestion
        """Replace the secrets' plain text content in the rel data by their IDs."""
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -519,6 +552,72 @@ def _fetch_local_cm_nodes(self, deployment_desc: DeploymentDescription) -> List[
...
+        # If we use the values of the secrets then the size of the secret
+        # would be too large for juju we store the secret ids instead and
+        # fetch the secrets when needed in the requirer side
```
Mehdi-Bendriss: That may be true, but this is the not the reason we're doing this. We are doing this in order to not transmit plaintext secret data in the relations.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -519,6 +552,72 @@ def _fetch_local_cm_nodes(self, deployment_desc: DeploymentDescription) -> List[
...
+        redacted_dict = rel_data.to_dict()
```
Mehdi-Bendriss, line 576: why is needed to serialize to a dict, why not work with the object and its attributes directly? I would rather we work with the former.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -260,6 +262,13 @@ def refresh_relation_data(
         # compute the data that needs to be broadcast to all related clusters (success or error)
         rel_data = self._rel_data(deployment_desc, orchestrators)
 
+        # if rel_data is an error, prepare to broadcast it to all related clusters
+        if isinstance(rel_data, PeerClusterRelData):
+            rel_data_redacted_dict = self._rel_data_redacted_dict(rel_data)
```
Mehdi-Bendriss, line 278: why is it needed to serialize into a dict?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
Mehdi-Bendriss: and it should be only one `s3` secret, not 2 like we currently do

Mehdi-Bendriss (commented):
Thanks Smail. Good job. This looks good overall. I left some comments.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: We do not want to mutate the object directly. We can either use the dict or copy the object and update the content of the secret to their respective ids.

skourta: @Mehdi-Bendriss We cannot directly use the `PeerClusterRelData` object directly because of the `admin_tls` field in the credentials which is [`dict`](https://github.com/canonical/opensearch-operator/blob/1fa589cda715971fba071a447169fb9271a20224/lib/charms/opensearch/v0/models.py#L295). The secret IDs are `str`. I can either create a new model for the redacted peer cluster relation data or stick with the `dict`. WDYT?

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (dismissed):

reneradoi (dismissed):
Re-approval after this PR had already been reviewed from my side previously.

reneradoi (approved):

phvalguima (approved):

PR #466, opened 2024-10-01
[DPE-4115] Performance Profile Support

phvalguima:
This PR extends the current charm to support performance profiles following spec DA031 and add supports for the following profiles:
* testing:
  * focused in the integration tests and our CI -> 1G RAM dedicated to the Heap and no automation
* staging:
  * HA capabilities must be available: for that, we will enforce index template that encompasses all the indices and sets replica: 1-all
  * Extends heap to: `max(1G, 10% of RAM)`
  * `indices.memory.index_buffer_size` extends to `25%`
  * Adds three component templates, that will be described later
* production:
  * Same features as the staging, but heap is set instead to: `max(1G, 50% of RAM)`

The options above are set based on the following documents:
https://opensearch.org/docs/latest/tuning-your-cluster/performance/
https://opensearch.org/docs/latest/search-plugins/knn/performance-tuning/

The user can switch between the three options above, and depending on the selected value, the templates are created or destroyed.

---------------------------------------------------------------------------------------------------------------------

UPDATE

One important question about this PR is if index templates with '*' will apply to system indices. So, the first part of this answer is: index templates are only applied at index creation, [as shown here](https://github.com/opensearch-project/OpenSearch/blob/23d1c7a55a63250b962c1fad4e6fb962fdd156cc/server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java#L447). We can delete index templates after indices were created based on that template.

Manual configuration (e.g. setting `0-all`) will always take precedence.

There is an exception for hidden (not necessarily system) indices: if we have templates that are "catch-all", then they are not applied [to hidden indices](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/metadata/MetadataIndexTemplateService.java#L1170).


lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -272,14 +272,13 @@ def replace(
...
         path = f"{self.base_path}{config_file}"
-
         if not exists(path):
             raise FileNotFoundError(f"{path} not found.")
 
         with open(path, "r+") as f:
             data = f.read()
 
-            if regex and old_val and re.compile(old_val).match(data):
+            if regex and old_val and re.compile(old_val, re.MULTILINE).findall(data):
```
phvalguima: Extended the regex, as it was failing to match: `-Xmx1g` in a file with content: `-Xms1g\n-Xmx1g`. Setting the multiline and forcing it to be more greedy and search for any combinations resolves this problem.

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +356,246 @@ def promote_failover(self) -> None:
...
+    B = 1  # noqa: N815
+    kB = 1024  # noqa: N815
+    mB = 1024 * kB  # noqa: N815
+    gB = 1024 * mB  # noqa: N815
+
+    @staticmethod
+    def get(name: str) -> int:
+        """Convert the value to the required unit."""
+        val = name.lower()
+        if val == "kb" or val == "k":
```
phvalguima: Java uses the model of [kmgKMG], no "B" char is added to the heap parameters.

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +356,246 @@ def promote_failover(self) -> None:
         self.main_app = self.failover_app
         self.main_rel_id = self.failover_rel_id
         self.delete("failover")
+
+
+class ByteUnit(Enum):
```
phvalguima: @Mehdi-Bendriss in chat with Big data team, they also have interest in these functionalities. Let's stick with this PR for now and, once you are okay with this format, we move the `ByteUnit` and `JavaHeapSize` class to @MiaAltieri helper libs.

tests/unit/lib/test_opensearch_tls.py
```diff
@@ -1578,7 +1578,11 @@ def test_on_certificate_available_rotation_ongoing_on_this_unit(
         # No action taken, no change on status or certificates
         assert run_cmd.call_count == 0
         assert self.harness.model.unit.status == original_status
-        self.charm.on.certificate_available.defer.called_once()
+        if leader:
```
phvalguima: Seems we are mixing `leader=True` and `leader=False`, and defer will be called only for leader units. We should not limit deferrals to leaders only. That will cause issues if we lose the leader unit later.

Mehdi-Bendriss (commented):
Thanks @phvalguima.

I have a few points:

### 1. Replication factor:
We should **never** set the replication factor to `1-all` - this is unnatural for non system indices and may cause the disk of all units to overflow quickly.

Replication factors are set by the user. 

If we want to be safer, we should look for indices whose `number_of_replicas < 1` and set it to 1. Not more. 

### 2. Codec used:
Why did you choose `zstd_no_dict` instead of `zst` or instead `qat_lz4` (in supported systems) or others? which one to choose? How? 

It may also be needed to set a [compression algorithm](https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/index-settings/#static-index-level-index-settings) with each? which one to choose? how? 

### 3. Heap size:

The heap size should, in production, be set to 50% but **lower than 32Gb** as per the official opensearch recommendations. [here](https://opster.com/guides/opensearch/opensearch-basics/opensearch-heap-size-usage-and-jvm-garbage-collection/) and [here](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/auto-tune.html).

### 4. Units conversion:
 
The main complexity of this PR revolves around unit conversions, 2 things to note:
  1. `/proc/meminfo` **always** returns units in `Kb`. 
    1. any reason to not use `psutil` ? this should save you the parsing of `/meminfo`
  2. Jvm XMS and XMX properties accept either `g|G -- m|M -- k|K`, meaning the smallest unit is `k`.

With this in mind - it makes sense to normalize the values to the smallest unit supported by both, which is the `Kb` and work with it exclusively (when reading, calculating or writing to file).
Something along the lines of:

  ```
  def jvm_size_in_kb(input: str) -> int:
      """Normalize the size values set in the jvm.options from supported units to Kb."""
      match = re.match(r"(\d+)([gmk])$", input.lower())

      value = float(match.group(1))
      jvm_formatted_unit = match.group(2)

      factor = 1
      if jvm_formatted_unit == "m":
          factor = 1024
      elif jvm_formatted_unit == "g":
          factor = 1024 * 1024

      # the loss is minimal since we're working with KBs
      return int(value * factor)
  ```

With this, I believe we do not need `ByteUnit` and `JavaByteSize` classes. 

Similar to the the percentage method, which can simply be calculated as `int(0.25 * val)` ==> again, the loss of rounding to the floor is minimal because we are dealing with Kilobytes.

phvalguima:
Hi @Mehdi-Bendriss, I will go over your points one by one.

---------------------------------------------------------------------------------

> 1. Replication factor:
We should never set the replication factor to 1-all - this is unnatural for non system indices and may cause the disk of all units to overflow quickly.
...
If we want to be safer, we should look for indices whose number_of_replicas < 1 and set it to 1. Not more.

Indeed, I corrected that right after our conversation earlier this week in [this commit](https://github.com/canonical/opensearch-operator/pull/466/commits/7c47f207bf69440b78d0f339cdbd6e1842e5a4a2). 

Although I agree with the removal of `-all`, thinking of real world deployments, we always go with at least 3x nodes. Other operators like SQL DBs will always have 1x main + 2x replicas in-sync when deployed by field. We remove the `-all` but we could have a similar experience and set it to `2`, as we are in the **HA** type of deployment, hence at least 3 nodes will be present.

> Replication factors are set by the user.

True, but on our specification: `DA031 - Profile config option`, it is stated that both `staging` and `production`, we will be *providing a highly-available and scalable service to be used in production.*. Now, that is a bit open for interpretation in my view: "highly-available" in terms of OpenSearch services (i.e. only the service indices should be set for HA by our charm) OR in terms of *using* OpenSearch as well (i.e. even indices created by the user). 

That is why I am trying to put together a minimum "index template", that at least assures an user that indices will be replicated, unless this user explicitly states otherwise.

---------------------------------------------------------------------------------


> 2. Codec used:

TBH, I am not entirely set on each of the values we will discuss below. That is one of the reasons I have added them as component templates, so an user will build their own index templates on top.

I'd rather benchmark these for comparison on top. Maybe getting that landed on this PR is too much. TL;DR I think we can play it safer as follows:

1) break the "component template" setup on a separate branch, for now
2) we run performance tests until we are okay with which values to suggest
3) if we are okay with the results, we can report it and update first our documentation (an user can then follow and create their own indices / index template)
4) Eventually, come back to the branch from step (1) and merge it. 

WDYT?

I will start with a later question:

> which one to choose?

The codec selection came from [these results](https://opensearch.org/docs/latest/im-plugin/index-codecs/#benchmarking).

> Why did you choose zstd_no_dict instead of zst 

In the case of `zstd_no_dict`, from the results you will see above, we would sacrifice 5% compression for +7% (net) p90 latency and +7% throughput (net) when compared with `zstd`.

> or instead qat_lz4 (in supported systems) or others?

Happy to set QAT if we get the logic to detect and enable it. We are not yet there.

> How?

First, we should keep in mind some types of indices cannot simply work with all the codecs (e.g. the vector indices).

As you also noticed, I am not really onboard with these results, as there are other parameters to be set. In this case, the **How??** will have to be done the same way we've done the AVX testing: by running performance tests and comparing results.

> It may also be needed to set a [compression algorithm](https://opensearch.org/docs/latest/install-and-configure/configuring-opensearch/index-settings/#static-index-level-index-settings) with each? which one to choose? how?

Yes, I also have this concern. What I think we should do here is run these component templates against benchmarks and document their results.


---------------------------------------------------------------------------------


> 3. Heap size:
The heap size should, in production, be set to 50% but lower than 32Gb as per the official opensearch recommendations. [here](https://opster.com/guides/opensearch/opensearch-basics/opensearch-heap-size-usage-and-jvm-garbage-collection/) and [here](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/auto-tune.html).

Thanks, that is a really good point. I will set a hard limit of up to 32G. Indeed having the GC going over 100s of Gigs does not sound good :)

---------------------------------------------------------------------------------


> 4. Units conversion:

So, some thoughts here: (1) indeed sticking with Kb, as long as the JVM can accept the "32G" in Kb format, is a good idea and would simplify a lot; and (2) I was discussing with Big Data team could benefit from this logic here. The idea is to eventually move to the `data_platform_helpers`. From what I gathered, they are currently setting the JVM heap on a hard limited number rn.

> The main complexity of this PR revolves around unit conversions, 2 things to note:
1. /proc/meminfo always returns units in Kb.

That is true [as the kernel code shows](https://github.com/torvalds/linux/blob/2a130b7e1fcdd83633c4aa70998c314d7c38b476/fs/proc/meminfo.c#L31).

> 1. any reason to not use psutil ? this should save you the parsing of /meminfo

Yes, two reasons: (1) the parsing goes from L589 to L598... I prefer 10 LoCs that process a file whose format is set in stone by the kernel than adding a new dependency (that actually is not shipped with stock Ubuntu anyways); and (2) it gives access to hugepages info as well, which we can potentially explore later, and would be "just there".


> Jvm XMS and XMX properties accept either g|G -- m|M -- k|K, meaning the smallest unit is k.
With this in mind - it makes sense to normalize the values to the smallest unit supported by both, which is the Kb and work with it exclusively (when reading, calculating or writing to file).

Yes, I was going down to the bytes and coming back but indeed makes sense to stop on Kb instead and we handle all this in Kb.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -676,22 +689,42 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
         if not self.plugin_manager.check_plugin_manager_ready():
             return
 
+        if self.upgrade_in_progress:
```
phvalguima, line 702: Moved this upwards as, from here down to the end of `_config_changed`, changes should not be accepted during upgrade.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -355,6 +361,13 @@ def cleanup():
                 logger.error("Service previously started but now misses the snap.")
                 return
 
+        # Store the current perf. profile we are applying
+        self.peers_data.put(
+            Scope.UNIT,
```
phvalguima: It is hard to make sense of the current performance profile just by pooling the raw heap size and other sources. Better to just store it in the peer data.

With that said, may be worthy to save it on the app level.

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -398,6 +406,32 @@ def _build_paths(self) -> Paths:
...
+                "/_index_template/charmed-index-tpl",
+            ]:
+                try:
+                    self.request("DELETE", endpoint)
+                except OpenSearchHttpError:
+                    pass
+            # Nothing to do anymore
+            return
+
+        for idx, template in self.perf_profile.charmed_index_template.items():
```
phvalguima: TODO: this can only happen on MAIN_ORCHESTRATOR

tests/unit/lib/test_ml_plugins.py
```diff
@@ -120,7 +120,6 @@ def test_disable_via_config_change(
 
         self.harness.update_config({"plugin_opensearch_knn": False})
         self.charm.plugin_manager.check_plugin_manager_ready.assert_called()
-        self.charm._restart_opensearch_event.emit.assert_called_once()
```
phvalguima, line 123: This line will be changed anyways on [DPE-4196 PR](https://github.com/canonical/opensearch-operator/pull/435/files#diff-edbec226d3bd45d1b8bb2a562d8c71e006c439ea5bf75d0a5e9cf490c7a41f0eR136)

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -166,12 +172,21 @@ def __init__(self, handle, *, ignore_lock=False):
         super().__init__(handle, ignore_lock=ignore_lock)
 
 
+class _ApplyProfileTemplatesOpenSearch(EventBase):
```
phvalguima: This event is needed as we have to add profile templates and the cluster may not be ready. This custom events allows us to execute these tasks when the cluster is set.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -676,22 +731,47 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
...
+                self.opensearch.perf_profile = OpenSearchPerfProfile.from_str(
+                    self.config.get(PERFORMANCE_PROFILE)
+                )
+                # If we have a running service, and our profile changed
+                # then we need a restart to apply the new profile
+                self.opensearch_config.apply_performance_profile(self.opensearch.perf_profile)
+                self._apply_profile_templates_event.emit()
+                if self.unit.is_leader():
+                    self.peers_data.put(
+                        Scope.APP,
```
phvalguima: After discussing with @reneradoi we need to have it as Scope.UNIT. Otherwise we may end up in a race condition where the leader is the first one to execute the change and all the other units see that new peer value == new config performance profile.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -249,6 +268,32 @@ def __init__(self, *args, distro: Type[OpenSearchDistribution] = None):
...
+            # we can safely abandon this event as we already had a previous call on the same hook
+            return
+        self._apply_profile_templates_has_been_called = True
+
+        if not self.opensearch_peer_cm.deployment_desc() or not self.opensearch.is_node_up():
+            logger.info("Applying profile templates but cluster not ready yet.")
+            event.defer()
+            return
+
+        if self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR:
```
reneradoi: To summarize offline discussion about it, there are a few questions about how to handle the profile configuration in large deployments:

1. Should the user be able to run different Opensearch applications within a large deployment with different profile settings?
2. If so, how should be determined which performance template to apply to the indices?

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +358,76 @@ def promote_failover(self) -> None:
...
         self.delete("failover")
+
+
+class OpenSearchPerfProfile(Model):
+    """Generates an immutable description of the performance profile."""
+
+    class Config:
+        """Pydantic config for this model."""
+
+        arbitrary_types_allowed = True
```
Mehdi-Bendriss: I don't think this is needed. Any specific reason to introduce more flexibility?

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +358,76 @@ def promote_failover(self) -> None:
...
+
+    typ: PerformanceType
+    heap_size_in_kb: int = MIN_HEAP_SIZE
+    opensearch_yml: Dict[str, str] = {}
+    charmed_index_template: Dict[str, str] = {}
+    charmed_component_templates: Dict[str, str] = {}
+
+    @classmethod
+    def from_str(cls, input_str: str):
```
Mehdi-Bendriss: I don't think this is needed, the base method `from_str` handles it, as long as your string representation is from a well dumped object.

This works fine.
```
OpenSearchPerfProfile.from_str(json.dumps({"typ": "production"})
```

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +358,76 @@ def promote_failover(self) -> None:
...
+        """Create a new instance of this class from a stringified json/dict repr."""
+        return cls(typ=PerformanceType(input_str))
+
+    @root_validator
+    def set_options(cls, values):  # noqa: N805
+        """Generate the attributes depending on the input."""
+        val = values["typ"]
+        if isinstance(val, str):
+            val = PerformanceType(val)
```
Mehdi-Bendriss: why is this check needed? `values["typ"]` is already of type `PerformanceType`

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +358,76 @@ def promote_failover(self) -> None:
...
+        if val == PerformanceType.STAGING:
+            values["heap_size_in_kb"] = min(
+                int(0.25 * OpenSearchPerfProfile.meminfo()["MemTotal"]),
+                MAX_HEAP_SIZE,
+            )
+
+        if val == PerformanceType.TESTING:
+            values["heap_size_in_kb"] = MIN_HEAP_SIZE
+
+        if val != PerformanceType.TESTING:
```
Mehdi-Bendriss: nit:
```suggestion
        if val == PerformanceType.TESTING:
            values["heap_size_in_kb"] = MIN_HEAP_SIZE
            return values 
            
        mem_total = OpenSearchPerfProfile.meminfo()["MemTotal"]
        mem_percent = 0.50 if val == PerformanceType.PRODUCTION else 0.25
        
        values["heap_size_in_kb"] = min(
            int(mem_percent * mem_total), MAX_HEAP_SIZE
        )
```

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +358,76 @@ def promote_failover(self) -> None:
...
+            values["opensearch_yml"] = {"indices.memory.index_buffer_size": "25%"}
+
+            values["charmed_index_template"] = {
+                "charmed-index-tpl": {
+                    "index_patterns": ["*"],
```
Mehdi-Bendriss, line 398: can we exclude built-in security indices? the ones essentially starting with `.`

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +358,76 @@ def promote_failover(self) -> None:
...
+                    "template": {
+                        "settings": {
+                            "number_of_replicas": "2",
```
```suggestion
                            "number_of_replicas": "1",
```

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +358,76 @@ def promote_failover(self) -> None:
...
+    @staticmethod
+    def meminfo() -> dict[str, float]:
+        """Read the /proc/meminfo file and return the values.
+
+        According to the kernel source code, the values are always in kB:
+            https://github.com/torvalds/linux/blob/
+                2a130b7e1fcdd83633c4aa70998c314d7c38b476/fs/proc/meminfo.c#L31
+        """
+        with open("/proc/meminfo") as f:
+            meminfo = f.read()
```
Mehdi-Bendriss: nit - maybe for clarity as the next lines are a bit difficult to read:
```suggestion
            meminfo = f.read().split("\n")
            meminfo = [line.split() for line in meminfo if line.strip()]
```

lib/charms/opensearch/v0/models.py
```diff
@@ -346,3 +358,76 @@ def promote_failover(self) -> None:
...
+
+        return {
+            line.split()[0][:-1]: float(line.split()[1]) for line in meminfo.split("\n") if line
+        }
```
```suggestion
        return {
            line[0][:-1]: float(line[1]) for line in meminfo
        }
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -355,6 +400,16 @@ def cleanup():
                 logger.error("Service previously started but now misses the snap.")
                 return
 
+        # Store the current perf. profile we are applying
+        self.opensearch_config.apply_performance_profile(self.opensearch.perf_profile)
+        self._apply_profile_templates_event.emit()
```
Mehdi-Bendriss: i'm not convinced on why we need this as an event when we can directly apply it above as well as follow the normal route used in plugins later in _config_changed.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -676,22 +731,47 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
...
-            if self.plugin_manager.run():
-                if self.upgrade_in_progress:
-                    logger.warning(
-                        "Changing config during an upgrade is not supported. The charm may be in a broken, "
-                        "unrecoverable state"
+            restart_requested = self.plugin_manager.run()
+            if (
+                self.peers_data.get(Scope.APP, PERFORMANCE_PROFILE)
+                and PerformanceType(self.peers_data.get(Scope.APP, PERFORMANCE_PROFILE))
+                != self.opensearch.perf_profile.typ
```
Mehdi-Bendriss: I would prefer we rely on `deployment_description` - and add a field there "profile" and compare against the config option, and in what comes next

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: As discussed, this shouldn't be the case. Can be handled in a next PR.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -355,6 +400,16 @@ def cleanup():
                 logger.error("Service previously started but now misses the snap.")
                 return
 
+        # Store the current perf. profile we are applying
+        self.opensearch_config.apply_performance_profile(self.opensearch.perf_profile)
+        self._apply_profile_templates_event.emit()
+        if self.unit.is_leader():
+            self.peers_data.put(
```
Mehdi-Bendriss: I prefer we enrich the `deployment-description` object and follow its flow by adding a field `profile: str` in the said object.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -676,22 +731,47 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
...
+                and PerformanceType(self.peers_data.get(Scope.APP, PERFORMANCE_PROFILE))
+                != self.opensearch.perf_profile.typ
+            ):
+                self.opensearch.perf_profile = OpenSearchPerfProfile.from_str(
+                    self.config.get(PERFORMANCE_PROFILE)
+                )
+                # If we have a running service, and our profile changed
+                # then we need a restart to apply the new profile
+                self.opensearch_config.apply_performance_profile(self.opensearch.perf_profile)
+                self._apply_profile_templates_event.emit()
```
Mehdi-Bendriss: I don't see the need of this

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: We can use the deployment-description object

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -98,6 +103,9 @@ def __init__(self, charm, peer_relation_name: str):
         self.config = YamlConfigSetter(base_path=self.paths.conf)
         self._charm = charm
         self._peer_relation_name = peer_relation_name
+        self.perf_profile = OpenSearchPerfProfile.from_str(
```
Mehdi-Bendriss: I don't think we need this - we can call `apply_perf_templates_if_needed` directly with the new profile

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -398,6 +406,43 @@ def _build_paths(self) -> Paths:
         """Build the Paths object."""
         pass
 
+    def apply_perf_templates_if_needed(self) -> bool:  # noqa: C901
```
Mehdi-Bendriss: I would create a new module `performance_profile.py` where the comparison between current deployment-description profile and new config issued profile happens. 
Similar to the plugin manager

src/opensearch.py
```diff
@@ -88,7 +88,7 @@ def is_service_started(self, paused: Optional[bool] = False) -> bool:
         # Now, we must dig deeper into the actual status of systemd and the JVM process.
         # First, we want to make sure the process is not stopped, dead or zombie.
         try:
-            pid = run_cmd("lsof", args="-ti:9200").rstrip()
+            pid = run_cmd("lsof", args="-ti:9200").out.rstrip()
```
Mehdi-Bendriss, line 91: Can you explain this? `run_cmd` returns `output.stdout.strip()` - there is no `.out` and no need for `.rstrip()`

Mehdi-Bendriss (changes requested):
Thanks Pedro. Could you rely on the current `deployment-description` data structure for setting the current profile (from config)? 
Maybe we can simply follow the plugin manager single point of entry logic an get rid of the custom event.

lib/charms/opensearch/v0/models.py
phvalguima: @Mehdi-Bendriss that is  a good question. I have to look into it.

src/opensearch.py
phvalguima: @Mehdi-Bendriss I think we are confusing `OpenSearchDistribution.{run_bin,run_script}` and `from charms.opensearch.v0.helper_charm import run_cmd`. Indeed, the former just provide a `str`. However, they are not generic cmd calls, they are based on `OPENSEARCH_BIN` and `OPENSEARCH_HOME` folders. We want something in the `PATH` (`lsof`). The latter gives this option and returns a `SimpleNamespace`.

I did not notice myself the slight distinction, as we do have `OpenSearchDistribution._run_cmd` as well. I think the `run_cmd` was added by @reneradoi. I will set a quick chat between us tomorrow to go over which is the most appropriate.

lib/charms/opensearch/v0/models.py
phvalguima: Removed

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: @Mehdi-Bendriss there is no guarantees that, at very first `start` and `config-changed`, of the leader of `MAIN_ORCHESTRATOR`, the cluster will be ready to have that setting applied.

src/opensearch.py
phvalguima: @Mehdi-Bendriss I will summarize our offline discussion with @reneradoi:
1) Continue with this implementation
2) Eventually, we will make `OpenSearchDistribution._run_cmd` obsolete

lib/charms/opensearch/v0/models.py
phvalguima: To clarify, I am adding a checker, as `set_options` overloads the `root_validator`. Therefore, it is acceptable to receive `values` without `typ`.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Check my comment here: https://github.com/canonical/opensearch-operator/pull/466/files/51d92e7074c0b37409ea279f5f328151e244218c#r1800783632

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: @Mehdi-Bendriss I think it is not clear what this entry means vs. the `deployment-description`.

DeploymentDescription carries the directive: everyone must be on profile="YY".

This setting is used to tell us which profile this unit we have right now. If:

```
self.peers_data.get(Scope.UNIT, PERFORMANCE_PROFILE) != DeploymentDescription.get(PERFORMANCE_PROFILE)
```

Then we need to change the settings and trigger a restart.

Reason being: it is far easier to keep notes of our current profile state on the peer databag than figuring it out from the current deployment.

config.yaml
```diff
@@ -36,3 +36,13 @@ options:
     default: true
     type: boolean
     description: Enable opensearch-knn
+
+  profile:
+    type: string
+    default: "production"
```
phvalguima, line 42: @Mehdi-Bendriss default should be "production"?

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Check comment on: https://github.com/canonical/opensearch-operator/pull/466/files/51d92e7074c0b37409ea279f5f328151e244218c#r1800815821

lib/charms/opensearch/v0/models.py
phvalguima: @Mehdi-Bendriss added a response to the PR description.

config.yaml
Mehdi-Bendriss: seems like it from [the spec](https://docs.google.com/document/d/1bq7KyUAPIIcUWGgifW05p40v7lGXs3PhMznM2ECaMKY/edit?tab=t.0)

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -196,6 +197,7 @@ def _new_cluster_setup(self, config: PeerClusterConfig) -> DeploymentDescription
                 pending_directives=directives,
                 typ=self._deployment_type(config, start_mode),
                 state=deployment_state,
+                profile=self._charm.performance_profile.current.typ.value,
```
reneradoi: If the config is changed by the user (e.g. from `testing` to `staging`), this will still return the old value and not update the profile itself, as it always returns the current value from the databag. Have verified that by testing an update to the config: it doesn't reflect in the `deployment description` to be updated and the heap is not increased.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -227,6 +229,7 @@ def _new_cluster_setup(self, config: PeerClusterConfig) -> DeploymentDescription
             pending_directives=directives,
             typ=self._deployment_type(config, start_mode),
             state=deployment_state,
+            profile=self._charm.performance_profile.current.typ.value,
```
reneradoi: Same as above.

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,216 @@
...
+                "/_index_template/charmed-index-tpl",
+            ]:
+                try:
+                    self.charm.opensearch.request("DELETE", endpoint)
+                except OpenSearchHttpError as e:
+                    if e.response_code == 404:
+                        # Nothing to do, this error means the template does not exist
+                        pass
+                    logger.warning(f"Failed to delete index template: {e}")
+                    return False
```
reneradoi: If the `404` doesn't matter here, should we maybe not return False in that case? Otherwise the event will be deferred forever if this doesn't exist in testing (which would not be nice).

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -398,6 +408,43 @@ def _build_paths(self) -> Paths:
...
```
reneradoi: As the implementation is now in `opensearch_performance_profile.py`, shoudn't it be removed from here?

reneradoi (commented):
Hey Pedro, the PR is in a good state already from my point of view. I've found some things that might need to be updated or discussed, but once these are clarified, I think we can move forward with this.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
phvalguima: @reneradoi, in terms of documentation and to leave it here what we've discussed in the chat: I have a ticket to go over our docs and make sure that specially smaller setups set the profile correctly.

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: You are not storing the flag in the unit peer data, you are storing it in the app data bag: `Scope.APP,`

phvalguima: Yes, that has been fixed within `opensearch_performance_profile` module.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
phvalguima: @reneradoi thanks for the pointers. Indeed there was a misplaced config set at `_on_config_changed`. I fixed it. Now, there are two other things:
1) there is one extra issue with `YamlConfSetter.replace`. I've separated it from this PR and merged the fix back here, original fix: [here](https://github.com/canonical/opensearch-operator/pull/484)
2) The `DeploymentDescription` will not get updated, *unless* there is a peer-cluster-orchestrator requirer. The reason is because I have a gateway to the `refresh_relation_data`.

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: Indeed, fixed!

lib/charms/opensearch/v0/opensearch_distro.py
phvalguima: I will move it to the dedicated module.

phvalguima: I copied its content to the new module, but forgot to remove this :)

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -703,11 +712,21 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
...
+        else:
+            if self.unit.is_leader():
+                self.status.clear(PluginConfigCheck, app=True)
+                self.status.clear(PluginConfigChangeError, app=True)
 
-        if self.unit.is_leader():
-            self.status.clear(PluginConfigCheck, app=True)
-            self.status.clear(PluginConfigChangeError, app=True)
+        if deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR:
+            # Only the main orchestrator can change the performance profile
```
```suggestion
            # TODO extend later. For now, only the main orchestrator can change the perf. profile 
```

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -98,6 +98,11 @@ def __init__(self, charm, peer_relation_name: str):
         self.config = YamlConfigSetter(base_path=self.paths.conf)
         self._charm = charm
         self._peer_relation_name = peer_relation_name
+        self.perf_profile = OpenSearchPerfProfile.from_dict(
```
Mehdi-Bendriss: I don't think we need this, right? we already have it as part of the relation data. This class symbolizes the "current" opensearch instance - the config values symbolize: current or future opensearch instance

lib/charms/opensearch/v0/opensearch_peer_clusters.py
Mehdi-Bendriss: Good catch! 

- René's comment: This can be solved by adding it as part of `def _user_config(self):` then we may be able to remove the need for having `performance_profile.current`. And the config change / diff will always be computed in `config_change`.
- For number 2: not quite, it will get updated as long as there is a diff in the user config. as highlighted in the above point. 

TL;DR: sorry, I should have been more careful -- `profile` should have been as part of `DeploymentDescription.config` not `DeploymentDescription` 

If too involved. We can perhaps change the field in the deployment_desc, and do further structural changes in a future PR.

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,214 @@
...
+
+    The main reason to have a separate event, is to be able to wait for the cluster.
+    It defers otherwise and only defers the execution of this particular task.
+    """
+
+
+class OpenSearchPerformance(ops.Object):
+    """Base class for OpenSearch charms."""
+
+    _apply_profile_templates_event = EventSource(_ApplyProfileTemplatesOpenSearch)
```
Mehdi-Bendriss, line 54: Is there a way we defer config-changed instead of this event?

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,214 @@
...
+        self.framework.observe(
+            charm.on[PeerClusterRelationName].relation_changed,
+            self._on_peer_cluster_relation_changed,
+        )
+        self.framework.observe(
+            self._apply_profile_templates_event, self._on_apply_profile_templates
+        )
+        self._apply_profile_templates_has_been_called = False
+
+    def _on_install(self, event: EventBase):
```
Mehdi-Bendriss: I don't think we need this hook, config_change hook is invoked on both following cases:
1. as part of start 
2. as part of config change

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -703,11 +712,21 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
...
+            perf_profile_needs_restart = self.performance_profile.apply(
+                self.config.get(PERFORMANCE_PROFILE)
+            )
+            if self.opensearch_peer_cm.is_provider(typ="main") and perf_profile_needs_restart:
```
phvalguima: @Mehdi-Bendriss do you think we should gate the `refresh_relation_data` with `self.opensearch_peer_cm.is_provider(typ="main")`. Effectively, that means the `DeploymentDescription` only gets updated at when the first orch. requirer shows up.

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,214 @@
...
+            # Deployment description not available yet
+            # Nothing to do
+            return
+
+        if self.apply(deployment_desc.profile):
+            # We need to restart
+            self.charm.restart_event.emit()
+
+    @property
+    def current(self) -> OpenSearchPerfProfile | None:
```
Mehdi-Bendriss, line 66: Similarly, we may not need this - if we update the deployment_desc.onfig as above

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,214 @@
...
+            return
+
+        if (
+            self.charm.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR
+            and not self.charm.unit.is_leader()
+        ):
+            return
+
+        # Configure templates if needed
+        if not self.charm.opensearch.apply_perf_templates_if_needed():
```
Mehdi-Bendriss:
```suggestion
        if not self.apply_perf_templates_if_needed():
```

Mehdi-Bendriss (commented):
Thanks Pedro - I believe one final round, with points to clarify

lib/charms/opensearch/v0/opensearch_peer_clusters.py
phvalguima: @Mehdi-Bendriss I'd rather do it on a follow-up PR

lib/charms/opensearch/v0/opensearch_distro.py
phvalguima: Yes, that officially became dead code :)

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: @Mehdi-Bendriss I don't want to have these big events being deferred... It is increasing the complexity a lot in my view. This is also an experiment: I have a "component": capable of managing some relations and events and has some internal, custom events.

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: 1. I need to make sure we have that set before the very first config-changed is called, from Juju docs: [install -> config-changed -> start](https://juju.is/docs/sdk/config-changed-event)
2. So, I put it on install because: (1) it goes back to encapsulating the internals of this component to itself; and (2) I don't want to have yet another option if you ask `opensearch_performance_profile.current`: I'd rather have it 
setting as PerformanceType only, not PerformanceType|None.

Now, that said, we can find another place given the constraints above!

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: The `deployment_desc` does not get updated if we have small deployments. Reason is because I am setting the config as the following:
```
        if deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR:
            # Only the main orchestrator can change the performance profile
            perf_profile_needs_restart = self.performance_profile.apply(
                self.config.get(PERFORMANCE_PROFILE)
            )
            if self.opensearch_peer_cm.is_provider(typ="main") and perf_profile_needs_restart:
                # Update the peer-cluster-orchestrator as needed
                self.peer_cluster_provider.refresh_relation_data(event) ## <<< happens if we have at least one peer-cluster-orch. requirer
```

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: Yes! I noticed that on my tests as well :D Thanks for the catch btw

lib/charms/opensearch/v0/opensearch_peer_clusters.py
phvalguima: @Mehdi-Bendriss @reneradoi if I understand it correctly, it feels like a gap in our config design for large deployments. Maybe we should have a "configurator/proxy" charm that manages all the config options?

In any case: as we go into the plugins, this is going to become a more often problem. We should stop and do a spec of how to proceed, before anything. WDYT?

Mehdi-Bendriss: > it feels like a gap in our config design for large deployments

Not really. 
- The deployment description is a description of an app with all its config options. It is unrelated to large deployments. 
- Config design in **both** Simple and Large deployments is handled centrally in the `opensearch_peer_clusters.run` method as described above. 
- it manages 3 config options already, not exclsuively related to large deployments (Refer to [the data model here](https://github.com/canonical/opensearch-operator/blob/2/edge/lib/charms/opensearch/v0/models.py#L201-L207))  
- The logic is reactive to the changes in the `deployment_desc.config` where it compares the previous with the current. 
- Based on which, it decides if the path to follow is: `new_cluster_setup`  (deploy time config options) or `existing_cluster_setup` (config options changes afterwards) 
- Those are exclusively triggered and ran based on a single point of entry `opensearch_peer_clusters.run` - exclsuively ran from `charm._on_config_change`. 

In any way. We can talk about improving the overall config management later.

lib/charms/opensearch/v0/opensearch_performance_profile.py
Mehdi-Bendriss: As explained above. The Deployment description object is unrelated to large deployments. and it does get updated in small deployments when the config changes :)
refer to [this comment](https://github.com/canonical/opensearch-operator/pull/466#discussion_r1803825307)

What is preventing us from moving the field from `deployment_desc.profile` to `deployment_desc.config.profile`?

Doing so will update the deployment_desc automatically on config changes.

reneradoi:
Hey @phvalguima I've tested the PR for large deployments. It does work when creating the deployment and also when updating the profile for the main-orchestrator. As commented in the code, updating the profile for non-main-orchestrators does not work currently.

Is this restriction supposed to stay? If so, I think we should display it to the user (e.g. with an error-message in the log) when applying the config-change on non-main-orchestrators, otherwise we would have a config mismatch and no feedback why.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
phvalguima: Yes, it is an interesting topic to discuss next week.

src/opensearch.py
phvalguima: Reported as a bug: https://github.com/canonical/opensearch-operator/issues/489

reneradoi (dismissed):
Testing with both large and small deployments worked now. Test was config change `testing -> staging -> testing`. All units have been restarted, changes were applied.

tests/unit/lib/test_opensearch_tls.py
```diff
@@ -89,6 +89,9 @@ def setUp(self, _) -> None:
         self.harness.add_network("1.1.1.1", endpoint=TLS_RELATION)
         self.harness.begin()
         self.charm = self.harness.charm
+        self.charm.tls._remove_ca_from_request_bundle = MagicMock()
+        self.charm.tls._add_ca_to_request_bundle = MagicMock()
```
phvalguima: @skourta I cannot get the TLS tests passings without mocking those two methods. Is it a problem to mock these values?

lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -289,10 +288,9 @@ def replace(
         if output_type in [OutputType.console, OutputType.all]:
             logger.info(data)
 
-        if output_file is None:
-            output_file = config_file
```
Mehdi-Bendriss: this is deleted without any substitute - we should put back

lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -289,10 +288,9 @@ def replace(
...
-
-        with open(output_file, "w") as f:
+        if output_file and output_file != config_file:
```
Mehdi-Bendriss:
```suggestion
        if output_file != config_file:
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -671,7 +679,20 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
...
+        perf_profile_needs_restart = None
+        if (
+            (previous_deployment_desc := self.opensearch_peer_cm.deployment_desc())
+            and previous_deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+            and not self.upgrade_in_progress
+        ):
+            perf_profile_needs_restart = self.performance_profile.apply(
+                self.config.get(PERFORMANCE_PROFILE)
+            )
+        else:
```
Mehdi-Bendriss: can you explain why we always defer if the current app is not `MAIN_ORCHESTRATOR` ?

lib/charms/opensearch/v0/opensearch_performance_profile.py
Mehdi-Bendriss: I understand the lifecycle, but we are mixing concerns here - install hook is about installing. 
Given you're now relying on the DeploymentDesc, there is no difference between this flow and the flow of setting roles at deploy time -- they're both taken into account in `new_cluster_setup`.

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,240 @@
...
+        on the beginning of the charm lifecycle. Therefore, we check if the unit is down.
+        """
+        if not os.path.exists(
+            os.path.join(
+                self.charm.opensearch.paths.conf,
+                self.charm.opensearch_config.JVM_OPTIONS,
+            )
+        ):
+            event.defer()
+            return
```
Mehdi-Bendriss: no need for this. This comes for free when installing, by the time you're in config_changed you are good.

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,240 @@
...
+
+        if (
+            deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR
+            and (rel_data := self.charm.opensearch_peer_cm.rel_data())
+            and rel_data.deployment_desc
+        ):
+            # Not the main orchestrator: set the profile from the peer-cluster
+            if self.apply(rel_data.deployment_desc.config.profile):
+                # We need to restart
+                self.charm.trigger_restart()
```
Mehdi-Bendriss: why do we set the profile from the main orchestrator? every application should have its own profile.

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,240 @@
...
+            # We need to restart
+            self.charm.trigger_restart()
+
+    @property
+    def current(self) -> OpenSearchPerfProfile | None:
+        """Return the current performance profile.
+
+        The profile is saved as a string in the charm peer databag.
+        """
+        if not self.peers_data.get(Scope.UNIT, PERFORMANCE_PROFILE) or not (
```
Mehdi-Bendriss: 1. why is this stored in the unit?
2. why is are we not just returning whatever's in the deployment_desc

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,240 @@
...
+            )
+
+        if deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR:
+            return OpenSearchPerfProfile.from_dict(
+                {"typ": self.peers_data.get(Scope.UNIT, PERFORMANCE_PROFILE)}
+            )
+        return OpenSearchPerfProfile.from_dict({"typ": deployment_desc.config.profile})
+
+    @current.setter
+    def current(self, value: OpenSearchPerfProfile | str):
```
Mehdi-Bendriss, line 78: similar as above

Mehdi-Bendriss (commented):
Thanks Pedro. I have some questions, mainly on the large deployment flow.

lib/charms/opensearch/v0/helper_conf_setter.py
phvalguima: @Mehdi-Bendriss path has been already defined. So the question is: are we writing to a different file, or the same?

Check L274

lib/charms/opensearch/v0/helper_conf_setter.py
phvalguima: Yes, check my comment before

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Indeed. I did not like that code to be at the beginning of the method either. I will roll that back to using `refresh_relation_data` instead.

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: The profile also sets global, cluster-wide settings. I don't think we should have all the apps enforcing a given index template just because "data.cold" is being set to profile=production.

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: self.peers_data.get(Scope.UNIT, PERFORMANCE_PROFILE) -> what do I have configured right now? If I do not save it in the peer databag, then I have to figure it out using the meminfo + JVM heap + any template indices available. That is far more complex.

This is literally a "note for myself".

I cannot trust the `self.charm.config` either. As I may be right now changing my profile and the `self.charm.config` will represent the "new state" I have to be.

I cannot trust the deployment description: that is set by the leader unit. What happens if I am the last unit to be set and the deployment desc. has been updated?

phvalguima: This setting is there to make it easier to find what is my real state, right now.

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: That is something the main orch. must mandate instead.

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: Ack, it is also causing problems with upgrades. So, I am assigning it explicitly at the beginning of config-changed.

lib/charms/opensearch/v0/opensearch_performance_profile.py
phvalguima: To clarify: I moved the profile entry to the `deployment_desc.config.profile`.

Now, that is not enough for this logic. As I have two answer basic one question:
`the state I am == the state I must be` >> this will trigger the profile change or not

Now, the question:
`the state I am` is defined by my **peer** relation databag. I store there the current status every time I update it

The question:
`the state I must be` is defined by either a config change (if this is a MAIN app) or by the `PeerClusterRelData.DeploymentDescription.config.profile`

So, I need the current method to know the profile.

Mehdi-Bendriss: the state I am == can be based on the content of the opensearch.yml heap no?

phvalguima: no, based on the `jvm.options`, the current jvm args (meaning, the process args for Xms and Xmx) and the index templates

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -680,27 +699,27 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
...
+            # The following changes in _on_config_changed are not supported during an upgrade
+            # Therefore, we leave now
+            logger.warning(
+                "Changing config during an upgrade is not supported. The charm may be in a broken, "
+                "unrecoverable state"
+            )
+            event.defer()
             return
 
+        perf_profile_needs_restart = False
```
reneradoi, line 712: This can be removed, it will be overwritten with `True` or `False` in line 738.

lib/charms/opensearch/v0/opensearch_performance_profile.py
```diff
@@ -0,0 +1,178 @@
...
+    def current(self) -> OpenSearchPerfProfile | None:
+        """Return the current performance profile.
+
+        The profile is saved as a string in the charm peer databag.
+        """
+        if not self.peers_data.get(Scope.UNIT, PERFORMANCE_PROFILE):
+            return None
+        return OpenSearchPerfProfile.from_dict(
+            {"typ": self.peers_data.get(Scope.UNIT, PERFORMANCE_PROFILE)}
+        )
```
Mehdi-Bendriss, line 75: nit for conciseness:
```suggestion
        if not (profile := self.peers_data.get(Scope.UNIT, PERFORMANCE_PROFILE)):
            return None
        return OpenSearchPerfProfile.from_dict({"typ": profile})
```

Mehdi-Bendriss (approved):
Thanks Pedro!

reneradoi (approved):
Looks good! Only one minor thing that could potentially be changed, depending on your preference.

PR #474, opened 2024-10-11
Increase logging when index creation fails

### LINKED ISSUE #473: Race condition between `leader-elected` and `opensearch-client-created` events.
The OpenSearch deployment gets stuck with the following error message:
```
opensearch                         blocked      3  opensearch                2/edge         171  no       failed to create datahub_index index - deferring index-requested event...
```

That blocked only happens if the index already exists, [as the code shows](https://github.com/canonical/opensearch-operator/blob/2/edge/lib/charms/opensearch/v0/opensearch_relation_provider.py#L223).

After doing some investigation on data-hub's CI failure: https://github.com/canonical/datahub-k8s-operator/actions/runs/11261394134/job/31314738969?pr=5#step:13:1786

I've noticed that we may have the following sequence of events happening:
1) unit 1 is the leader, executes the `index-requested` event and creates the event
2) leader ship changes -> unit 2 
3) unit 2 executes the `index-requested` event and notices the index already exists

Now, the cluster will be stuck on "failed to create {index-name} index".

It is not possible to validate the flow above as the real cause of this CI failure. Nonetheless, it is a possibility and therefore, we should avoid it.

phvalguima:
This PR will add some extra logging so we can follow-up with any failures due to index creation.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_relation_provider.py
+++ b/lib/charms/opensearch/v0/opensearch_relation_provider.py
@@ -218,12 +218,10 @@ class OpenSearchProvider(Object):
             self.opensearch.request("PUT", f"/{event.index}")
         except OpenSearchHttpError as e:
             if (
-                (
-                    e.response_code == 400
-                    and e.response_body.get("error", {}).get("type")
-                    != "resource_already_exists_exception"
-                ) or e.response_code != 400
-            ):
+                e.response_code == 400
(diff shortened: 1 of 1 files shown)


Mehdi-Bendriss (approved):
Please open a following PR - with a rollback to this temporary change so we don't forget about it.

skourta (approved):

PR #479, opened 2024-10-15
[DPE-5667] Move away from ops_test's wait_for_idle to wait_until

### LINKED ISSUE #476: [CI][test_storage.py] Some units are stuck waiting for idle that never happens
Most of the tests on `test_storage.py` are using `ops_test.model.wait_for_idle`, [e.g. here](https://github.com/canonical/opensearch-operator/blob/2cc457c9094fba28e055e81c9eb5de48ce563707/tests/integration/ha/test_storage.py#L77).

On the status, one can see the deployment seems to be in idle:
```
Model  Controller           Cloud/Region         Version  SLA          Timestamp
test   localhost-localhost  localhost/localhost  3.[5](https://github.com/canonical/opensearch-operator/actions/runs/11337557141/job/31529500083#step:37:6).3    unsupported  03:58:25Z

App                       Version  Status  Scale  Charm                     Channel        Rev  Exposed  Message
opensearch                         active      1  opensearch                                 0  no       
self-signed-certificates           active      1  self-signed-certificates  latest/stable  155  no       

Unit                         Workload  Agent  Machine  Public address  Ports     Message
opensearch/0*                active    idle   1        10.108.5.105    9200/tcp  
self-signed-certificates/0*  active    idle   0        10.108.5.19               
```

We should consider moving these tests to `wait_until`.

phvalguima:
Moves the CI run for `test_storage.py` to use `wait_until`.

Closes #476

### DIFF
--- a/tests/integration/ha/test_storage.py
+++ b/tests/integration/ha/test_storage.py
@@ -178,7 +178,7 @@ async def test_storage_reuse_after_scale_to_zero(
     await wait_until(
         ops_test,
         apps=[app],
-        apps_statuses=["active"],
+        apps_statuses=["active", "blocked"],
         timeout=1000,
         idle_period=IDLE_PERIOD,
         wait_for_exact_units={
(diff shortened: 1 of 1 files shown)


reneradoi (dismissed):
Thanks Pedro! Feels kind of unsatisfying to having to move away from the "standard" method provided by the framework because it does not really work...

Mehdi-Bendriss (approved):

lucasgameiroborges (approved):

PR #494, opened 2024-10-25
[DPE-5710] Enable `update_certs` to throw KeyError

### LINKED ISSUE #493: `update_certs` is called too early in the user mgmt lifecycle
The following deployment:
```
$ juju status --relations
Model       Controller        Cloud/Region      Version  SLA          Timestamp
opensearch  azure-westeurope  azure/westeurope  3.4.4    unsupported  08:00:05Z

App                       Version  Status   Scale  Charm                     Channel        Rev  Exposed  Message
opensearch                         active       3  opensearch                2/edge         185  no       
opensearch-dashboards              blocked      1  opensearch-dashboards     2/stable        22  no       Incompatible Opensearch and Dashboards versions
self-signed-certificates           active       1  self-signed-certificates  latest/stable  155  no       
sysconfig                          active       3  sysconfig                 latest/stable   33  no       ready

Unit                         Workload     Agent      Machine  Public address  Ports     Message
opensearch-dashboards/0*     blocked      idle       1        172.18.0.14     5601/tcp  Incompatible Opensearch and Dashboards versions
opensearch/0                 active       idle       2        172.18.0.12     9200/tcp  
  sysconfig/1                active       idle                172.18.0.12               ready
opensearch/1*                maintenance  executing  3        172.18.0.11               new index .opensearch-dashboards requested
  sysconfig/0*               active       idle                172.18.0.11               ready
opensearch/2                 active       idle       4        172.18.0.13     9200/tcp  
  sysconfig/2                active       idle                172.18.0.13               ready
self-signed-certificates/0*  active       idle       0        172.18.0.15               

Machine  State    Address      Inst id        Base          AZ  Message
0        started  172.18.0.15  juju-a49dc1-0  ubuntu@22.04      
1        started  172.18.0.14  juju-a49dc1-1  ubuntu@22.04      
2        started  172.18.0.12  juju-a49dc1-2  ubuntu@22.04      
3        started  172.18.0.11  juju-a49dc1-3  ubuntu@22.04      
4        started  172.18.0.13  juju-a49dc1-4  ubuntu@22.04      

Integration provider                   Requirer                                 Interface           Type         Message
opensearch-dashboards:dashboard_peers  opensearch-dashboards:dashboard_peers    dashboard_peers     peer         
opensearch-dashboards:restart          opensearch-dashboards:restart            rolling_op          peer         
opensearch-dashboards:upgrade          opensearch-dashboards:upgrade            upgrade             peer         
opensearch:juju-info                   sysconfig:juju-info                      juju-info           subordinate  
opensearch:node-lock-fallback          opensearch:node-lock-fallback            node_lock_fallback  peer         
opensearch:opensearch-client           opensearch-dashboards:opensearch-client  opensearch_client   regular      
opensearch:opensearch-peers            opensearch:opensearch-peers              opensearch_peers    peer         
opensearch:upgrade-version-a           opensearch:upgrade-version-a             upgrade             peer         
self-signed-certificates:certificates  opensearch-dashboards:certificates       tls-certificates    regular      
self-signed-certificates:certificates  opensearch:certificates                  tls-certificates    regular    
```

And it fails when trying to recover the CA chain information at user setup: https://pastebin.ubuntu.com/p/P9w6YDCDxc/

However, it seems the CA chain info is not yet there:
* opensearch/1 show-unit: https://pastebin.canonical.com/p/qCXMcQKqv3/
* self-signed-cert show-unit: https://pastebin.canonical.com/p/jMpZ9PT6BF/

phvalguima:
There is a potential race condition within ` opensearch_relation_provider.py`, where we may ask for an user to be set (and hence, eventually call `update_certs`) before the certificate is ready. That is more often when we deploy a complete bundle at once.

This PR fixes the issue by allowing `update_certs` to detect the missing `chain` config in the TLS secrets and throws a `KeyError`.

Closes: #493


lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -359,20 +364,27 @@ def update_certs(self, relation_id, ca_chain=None):
...
+            # If neither ca_chain nor the secret exists, we'll raise a KeyError.
+            _ch_chain = (
+                ca_chain
+                or self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val)["chain"]
+            )
         except AttributeError:
             # cert doesn't exist - presumably we don't yet have a TLS relation.
             logger.warning("unable to get ca_chain")
             return
+        except TypeError:
```
reneradoi, line 386: I'm confused, shouldn't this be a KeyError if the `["chain"]` doesn't exist (as per [Built-in Exceptions](https://docs.python.org/3/library/exceptions.html#KeyError))?

Edit: Test says KeyError:
```
>>> my_dict = {1: "a", 2: "b"}
>>> print(my_dict[1])
a
>>> print(my_dict[3])
  File "<stdin>", line 1, in <module>
KeyError: 3
```

phvalguima: Yes, and in this case, we let it be raised and we catch it on the calling methods.

Now, if you try instead `None["chain"]`, then you have a type (`None`) being wrongly used as another type (`dict`). That is a TypeError and it is also a possibility here if nothing is set in the secret.

To simplify, I just reraise it as a KeyError so we handle a single exception.

reneradoi (approved):

lib/charms/opensearch/v0/opensearch_relation_provider.py
skourta: In my opinion, I think it would be cleaner to raise a custom exception rather than raising the TypeError as a KeyError

skourta (approved):
I left a minor comment as a suggestion.

phvalguima:
@reneradoi @skourta I agree that TypeError > KeyError is confusing. I will do the following instead: I will set the `get_object` in secrets to accept a default value to return (which will be `None` for everyone else). Here, I will set that default as `{}`. Then, we do not have two errors anymore.

lib/charms/opensearch/v0/opensearch_relation_provider.py
Mehdi-Bendriss: +1 with @skourta - we should not propagate a `KeyError`. There should be a TLS related exception raised / caught in the consumer side. 

Why are we also handling `AttributeError`? 

We can I believe simplify this part as:
```
admin_secrets = self.charm.secrets.get_object(Scope.APP, CertType.APP_ADMIN.val) or {}
if not (_ch_chain := ca_chain or admin_secrets.get("chain")):
    raise TLSCaChainMissingError()
```

phvalguima: @Mehdi-Bendriss on the Error, okay, I will add another error class here (in another PR and ref your comment).

About `AttributeError`, I think the reason there was because original code was:

```
... .get("chain")
```

Which may result in a `None.get("chain")`. I looked into the code underneath the secret calls but it handles None carefully, so there is no way I could see an `AttributeError` bubbling up. TL;DR it is safe to remove that 2nd exception.

PR #495, opened 2024-10-28
[DPE-5830] Bump to 24.04

### LINKED ISSUE #1093: Parsing of storage constraints fails
### Description

Hi Team!

After upgrading our libs to 3.5.2 one of our integration test fails with this error:
```
[...]
File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/juju/constraints.py", line 131, in parse_storage_constraint
    for m in STORAGE.finditer(constraint):
TypeError: expected string or bytes-like object
```
This is presumably due to https://github.com/juju/python-libjuju/pull/1053, which added parsing the storage constraints by applying `re.finditer()` on the given constraints. But as some of these (e.g. size) are integers and no strings, this fails.

We define the storage constraints like this:  `storage = {"opensearch-data": {"pool": "opensearch-pool", "size": 2048}}` This cannot be changed to string as per https://github.com/juju/python-libjuju/blob/main/juju/client/_definitions.py#L6531-L6532.

### Urgency

Annoying bug in our test suite

### Python-libjuju version

3.5.2.0

### Juju version

3.4.3, 3.5.3

### Reproduce / Test

```python
storage = {"opensearch-data": {"pool": "opensearch-pool", "size": 2048}}
ops_test.model.deploy(my_charm, num_units=1, series=SERIES, storage=storage)
```

reneradoi:
Update CI pipeline to enable 24.04 support.

Change list:
- `charmcraft.yaml` uses charmcraftst124 syntax to enable building on multiple bases
- Charm now uses bin commands exposed though the snap: `opensearch.<bin-command>`
- `tmpfiles` are now created inside the snap folders so the snap commands are able to access them
- small fixes and additions on unit tests
- update `juju` version that fixes [storage](https://github.com/juju/python-libjuju/issues/1093) and [websockets](https://github.com/juju/python-libjuju/issues/1184) issues

------------
TODO:
After this PR, some tasks will need addressing:
- [Migration of CI](https://warthogs.atlassian.net/browse/DPE-6251) to use charmcraft 3 support
- [Provider tests bug with the app charm](https://warthogs.atlassian.net/browse/DPE-6252)


lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -79,7 +79,7 @@ def __init__(
         self.peer_relation = peer_relation
         self.jdk_path = jdk_path
         self.certs_path = certs_path
-        self.keytool = self.jdk_path + "/bin/keytool"
+        self.keytool = "opensearch.keytool"
```
reneradoi, line 82: I like that change and that this is now provided by the snap!

.github/workflows/ci.yaml
```diff
@@ -114,10 +114,11 @@ jobs:
         path:
         - .
         - ./tests/integration/relations/opensearch_provider/application-charm/
-    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@v22.0.0
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm.yaml@beta-charmcraftst124
```
reneradoi, line 117: Maybe add some `# todo` comments here?

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -549,13 +547,15 @@ def store_new_ca(self, secrets: Dict[str, Any]) -> bool:  # noqa: C901
             ):
                 raise
 
-        with tempfile.NamedTemporaryFile(mode="w+t") as ca_tmp_file:
+        with tempfile.NamedTemporaryFile(
+            mode="w+t", dir=self.charm.opensearch.paths.conf
+        ) as ca_tmp_file:
```
reneradoi, line 552: :+1: good change. Just for consideration: there is also a `self.charm.opensearch.paths.certs`.

pyproject.toml
```diff
@@ -71,22 +71,23 @@ responses = "^0.25.3"
...
-pytest-operator-groups = {git = "https://github.com/canonical/data-platform-workflows", tag = "v22.0.0", subdirectory = "python/pytest_plugins/pytest_operator_groups"}
-pytest-microceph = {git = "https://github.com/canonical/data-platform-workflows", tag = "v22.0.0", subdirectory = "python/pytest_plugins/microceph"}
+pytest-operator-cache = {git = "https://github.com/canonical/data-platform-workflows", branch = "beta-charmcraftst124", subdirectory = "python/pytest_plugins/pytest_operator_cache"}
+pytest-operator-groups = {git = "https://github.com/canonical/data-platform-workflows", tag = "v23.1.1", subdirectory = "python/pytest_plugins/pytest_operator_groups"}
+pytest-microceph = {git = "https://github.com/canonical/data-platform-workflows", tag = "v23.1.1", subdirectory = "python/pytest_plugins/microceph"}
 # should not be updated unless https://github.com/juju/python-libjuju/issues/1093 is fixed
-juju = "==3.5.0"
+# websockets issue solved from 3.2.5.1 https://github.com/juju/python-libjuju/issues/1184 # TODO: restore to 3.5.0
```
reneradoi, line 81: Nice to see this fixed.

.github/workflows/ci.yaml
zmraul: The big "TODO" is on `charmcraft.yaml` explaining why all of this is needed. Once we generally update how charmcraft is done, all of this other places will be forced to change by necessity.

lib/charms/opensearch/v0/opensearch_tls.py
zmraul: Yes, I was between that one or `conf`. I think either way is fine since the files disappear once `close()` is called.

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -197,23 +197,23 @@ def is_node_up(self, host: Optional[str] = None) -> bool:
...
+        return self._run_cmd(opensearch_command, args, stdin=stdin)
 
     def run_script(self, script_name: str, args: str = None):
         """Run script provided by Opensearch in another directory, relative to OPENSEARCH_HOME."""
         script_path = f"{self.paths.home}/{script_name}"
         if not os.access(script_path, os.X_OK):
             self._run_cmd(f"chmod a+x {script_path}")
 
-        self._run_cmd(f"{script_path}", args)
+        self._run_cmd(f"snap run --shell opensearch.daemon -- {script_path}", args)
```
phvalguima, line 216: nice, makes a lot of sense to run these scripts under the snap context. TBH we only use this logic [here](https://github.com/canonical/opensearch-operator/blob/44162927148aeb3032f9d6334f3a42731eb8d188/lib/charms/opensearch/v0/opensearch_base_charm.py#L1448).

phvalguima (commented):
I think, in terms of changes, it seems okay. Covers the distro module and fixes the calls to our plugins or keytool. What I expect to see before given a final +1:
* CI passes on all tests -> only exceptions I would consider here are `test_storage`, which is a bit flaky; and lib check
* A ticket explicitly highlighting we need to go back and revisit the TODOs of this PR. Assign this ticket to yourself for this pulse. Keep it in "blocked" and list what we need to have so we can unblock

phvalguima (approved):
After discussions with @zmraul, there is an ongoing issue with charmcraft and the test application charm: https://github.com/canonical/charmcraft/issues/2055

Both moving the CI to 24.04 and updating opensearch CI once the charmcraft issue above has been fixed have been reported in the following respective tickets:
* https://warthogs.atlassian.net/browse/DPE-6251
* https://warthogs.atlassian.net/browse/DPE-6252

We can move forward with this PR, aware the `relations/test_opensearch_provider.py` will not be passing for the time being.

zmraul (approved):

PR #497, opened 2024-10-30
Do not apply performance profile if deployment_desc does not exist

reneradoi:
## Issue
If the Opensearch operator receives a config change before the deployment description has been initialized, applying the performance profile will fail and the charm will be in a failed state.

This happens repeatedly on the integration test runs for Opensearch in the data-integrator CI, e.g. [here](https://github.com/canonical/data-integrator/actions/runs/11585189263/job/32253824740).

## Solution
If the deployment description does not exist yet, the performance profile should not be applied and the received event should be deferred.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -743,9 +743,14 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
                 self.status.clear(PluginConfigCheck, app=True)
                 self.status.clear(PluginConfigChangeError, app=True)
 
-        perf_profile_needs_restart = self.performance_profile.apply(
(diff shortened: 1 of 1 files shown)


phvalguima (approved):
Thanks for that @reneradoi!

The original thought here was that `DeploymentDescription` would be already available at `config-changed` but indeed, for non-leader units, that will not necessarily hold true.

It is more a *nit*, but we could move it up to line 720, as there is no point of doing neither perf profile nor plugin configuration if deployment desc. is not set (although plugin_manager handles that case already by itself).

phvalguima:
BTW, on a side-note, I also prepared a PR to data-integrator to decrease its default opensearch profile to "testing": https://github.com/canonical/data-integrator/pull/86

Mehdi-Bendriss (approved):

PR #501, opened 2024-11-11
[DPE-5866] Add Charm Terraform Module

phvalguima:
This PR extends our current charm to add a terraform module for the charm. It also adds github workflow to test + a basic test setup.


terraform/main.tf
```diff
@@ -0,0 +1,12 @@
+resource "juju_application" "opensearch" {
+  charm {
+    name     = "opensearch"
+    channel  = var.channel
+    revision = var.revision
+  }
```
marcoppenheimer: **todo:** As far as I can see, you aren't using `base` here. Revision is *probably* tighter, but was this an omission?

terraform/main.tf
```diff
@@ -0,0 +1,12 @@
...
+  config      = var.config
+  model       = var.model_name
+  name        = var.app_name
+  units       = var.units
+  constraints = var.constraints
```
marcoppenheimer, line 16: **todo:** From what I remember about OS, it's likely that we'll have multiple machines with different internal specs for different roles. High memory for `ML`, high disk for something else etc.

I think you might want to add an optional `machines` var here, that takes a list of `juju_machine` resources for placement.
I'm not 100% sure of the provider syntax for that, and it's changed a lot recently there and in the spec, but hopefully you get my gist.

marcoppenheimer: Same with storages?

terraform/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "vpc" {
+  description = "VPC object copied from variable to the next stage"
+  value       = juju_application.opensearch.placement
+}
```
marcoppenheimer: **todo:** re. [the spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?tab=t.0#heading=h.sothcshk3dl1), I would suggest adding `app_name` and `requires`/`provides` outputs. I believe you mentioned this in the README.

terraform/README.md
marcoppenheimer, line 1: **praise:** This is an **excellent** README. Great job, I'll be stealing this, and I hope most of the team does too.

terraform/main.tf
phvalguima: Yes, I forgot it :)

Thanks!

terraform/main.tf
phvalguima: @marcoppenheimer, TL;DR: this is the "charm terraform". So, we are modeling a single "juju app" here. What you are referring to is the "product terraform". That may be composed by multiple types of juju apps, dashboard, etc. Indeed, for the later, we have to care about that.

Now, back to your questions, I agree with should have both as options: `machines` and `storage`.

terraform/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+variable "config" {
+  description = "Map of charm configuration options"
+  type        = map(string)
+  default     = {}
+}
+
+variable "model_name" {
+  description = "Model name"
+  type        = string
+}
```
ghislainbourgeois: The spec was changed to use the name `model` for the input variable to better communicate the fact that it should a tied to a resource, and not just any string.

Also, I have heard that there will be changes to the provider moving towards UUIDs, so using `model` here would also better fit with that change in the future.

ghislainbourgeois (changes requested):
This looks great, nice job! On top of the comments that @marcoppenheimer already left, I added a comment regarding the variable `model_name`.

terraform/main.tf
phvalguima: @marcoppenheimer @ghislainbourgeois `placement` is broken in our provider. If I deploy *without* using placement, I still get a failure with:
```
╷                                                                                                                                  
│ Error: Provider produced inconsistent result after apply                                                                         
│                                                                                                                                  
│ When applying changes to module.opensearch.juju_application.opensearch, provider "provider[\"registry.terraform.io/juju/juju\"]" 
│ produced an unexpected new value: .placement: was cty.StringVal(""), but now cty.StringVal("1").                                 
│                                                                                                                                  
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.                                     
╵                                                                                                                                  
```

There are bugs already reported:
* For placement specifically: https://github.com/juju/terraform-provider-juju/issues/443, https://github.com/juju/terraform-provider-juju/issues/182
* and I've also seen that for constraints, as reported here: https://github.com/juju/terraform-provider-juju/issues/344, https://github.com/juju/terraform-provider-juju/issues/536

https://github.com/juju/terraform-provider-juju/issues/182 proposes a workaround with:
```
lifecycle {
        ignore_changes = [ placement, ]
    }
```

We should not skip changes in placement, nor constraints in my view.

Therefore, I am keeping placement out of this module for now (it will be commented only).

phvalguima:
@ghislainbourgeois @marcoppenheimer I have closed the comments that seem to be addressed. There is one big question for me about placement. Can you check that out?

terraform/main.tf
phvalguima: BTW, I tried the ignore_changes and dis not work on my test. I am not sure if I missed something.

ghislainbourgeois: I think keeping placement out for now is the right course of action until we find the right approach.

marcoppenheimer (dismissed):

ghislainbourgeois (dismissed):

.github/workflows/ci.yaml
```diff
@@ -32,6 +32,59 @@ jobs:
       - name: Run tests
         run: tox run -e unit
 
+  terraform-test:
+    name: Terraform - Lint and Simple Deployment
```
gabrielcocenza, line 36: **non-blocker**:

TIL: I didn't know that terraform was already installed by default on `ubuntu-latest`

It seems that during transitions like ubuntu 22.04 to 24.04 terraform was not included during some time. See https://github.com/actions/runner-images/issues/10796

IMO making it explicit can make it a little bit more robust for the future, but not big deal

terraform/tests/simple_deployment.tf
```diff
@@ -0,0 +1,39 @@
+module "opensearch" {
+  source     = "../"
+  app_name   = var.app_name
+  model = var.model_name
+  units      = var.simple_opensearch_units
+  config = {
+    profile = "testing"
+  }
+
+  channel = "2/edge"
```
gabrielcocenza, line 10: Wouldn't make more sense in the CI to use the local build charm instead of using from the `2/edge` channel?

I understand that for someone using the terraform from this repo to deploy OpenSearch will want to choose a channel, but running for every PR to test what is on `2/edge` than I think this could be a cronjob that runs with a certain frequency during the day.

gabrielcocenza: In any case, I guess it's nice to have also the option to deploy OpenSearch from a local build, but this can be targeted in the future

phvalguima: I added a comment on our spec for terraform, but currently we are not defining the `path` there.

.github/workflows/ci.yaml
phvalguima: Yes, I discovered it as well a few months ago :)

terraform/variables.tf
```diff
@@ -0,0 +1,67 @@
...
+variable "app_name" {
+  description = "Application name"
+  type        = string
+  default     = "opensearch"
+}
+
+variable "channel" {
+  description = "Charm channel"
+  type        = string
+  default     = null
```
Mehdi-Bendriss: perhaps:
```suggestion
  default     = "2/stable"
```

terraform/variables.tf
```diff
@@ -0,0 +1,67 @@
...
+variable "revision" {
+  description = "Charm revision"
+  type        = number
+  default     = null
+}
+
+variable "units" {
+  description = "Charm units"
+  type        = number
+  default     = 1
```
Mehdi-Bendriss: I would set here:
```suggestion
  default     = 3
```

terraform/variables.tf
```diff
@@ -0,0 +1,67 @@
...
+  description = "Map of storage used by the application"
+  type        = map(string)
+  default     = {}
+}
+
+variable "endpoint_bindings" {
+  description = "Map of endpoint bindings"
+  type        = map(string)
+  default     = {}
+}
```
Mehdi-Bendriss:
```suggestion
}

```

.gitignore
```diff
@@ -16,3 +16,50 @@ cloudinit-userdata.yaml
...
+
+# Include override files you do wish to add to version control using negated pattern
+# !example_override.tf
+
+# Include tfplan files to ignore the plan output of command: terraform plan -out=tfplan
+# example: *tfplan*
+
+# Ignore CLI configuration files
+.terraformrc
+terraform.rc
```
Mehdi-Bendriss:
```suggestion
terraform.rc

```

terraform/tests/preamble.tf
```diff
@@ -0,0 +1,18 @@
...
+}
+
+resource "juju_application" "self-signed-certificates" {
+  charm {
+    name    = "self-signed-certificates"
+    channel = "latest/stable"
+  }
+  model      = var.model_name
+  depends_on = [null_resource.preamble]
+}
```
Mehdi-Bendriss:
```suggestion
}

```

terraform/tests/providers.tf
```diff
@@ -0,0 +1,16 @@
...
+    http = {
+      source  = "hashicorp/http"
+      version = "~> 3.4.5"
+    }
+    external = {
+      source  = "hashicorp/external"
+      version = "~> 2.3.4"
+    }
+  }
+}
```
Mehdi-Bendriss:
```suggestion
}

```

terraform/main.tf
```diff
@@ -0,0 +1,41 @@
...
+    for k, v in var.endpoint_bindings : {
+      endpoint = k, space = v
+    }
+  ]
+
+  storage_directives = var.storage
+
+  lifecycle {
+    precondition {
+      condition     = length(var.machines) == 0 || (length(var.machines) > 0 && length(var.machines) == var.units)
```
Mehdi-Bendriss:
```suggestion
      condition     = length(var.machines) == 0 || length(var.machines) == var.units
```

terraform/main.tf
```diff
@@ -0,0 +1,41 @@
...
+      error_message = "Machine count does not match unit count"
+    }
+    precondition {
+      condition     = length(var.storage) == 0 || (length(var.storage) > 0 && (lookup(var.storage, "count", 0) <= 1) && (lookup(var.storage, "count", 0) > 0))
```
Mehdi-Bendriss: 1. `length(var.storage) > 0` can be dropped.
2. Can you describe the data structure of `var.storage`?

Mehdi-Bendriss (commented):
Thanks Pedro, it looks good. I have small questions

terraform/main.tf
phvalguima: On CC006, it is a `map(string)`. In Juju provider, it [follows this schema](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#nested-schema-for-storage)

terraform/main.tf
phvalguima: Sorry about that, I also found pretty lame but I was getting a bunch of errors because `var.machines` can be null and just `length(var.machines) == var.units` would break...

Anyways, I ended up being hyper(unnecessarily)-conservative in these conditions

terraform/main.tf
phvalguima: That means, eventually count will become a `number`, which can be anything (e.g. `int` or even `float`) as [described here](https://developer.hashicorp.com/terraform/language/expressions/types).

Mehdi-Bendriss (approved):

marcoppenheimer (approved):

PR #502, opened 2024-11-11
[DPE-5932] Remove cert-available check for CA rotation

### LINKED ISSUE #500: OpenSearch's CA rotation gets stuck in large deployments
Our CI runs for CA rotation, specially the large deployments ones, are continuously timing out, e.g.: https://github.com/canonical/opensearch-operator/actions/runs/11599401070/job/32297803436

The test: `test_rollout_new_ca` will eventually time out.

I can manually reproduce it and see. Juju status:
```
Model  Controller           Cloud/Region         Version  SLA          Timestamp
test   localhost-localhost  localhost/localhost  3.5.4    unsupported  10:40:53Z

App                       Version  Status   Scale  Charm                     Channel        Rev  Exposed  Message
opensearch-data                    waiting      1  opensearch                                 0  no       TLS not fully configured in related 'main-orchestrator'.
opensearch-failover                waiting      1  opensearch                                 1  no       TLS not fully configured in related 'main-orchestrator'.
opensearch-main                    active       3  opensearch                                 2  no       
self-signed-certificates           active       1  self-signed-certificates  latest/stable  155  no       

Unit                         Workload     Agent      Machine  Public address  Ports     Message
opensearch-data/0*           maintenance  idle       1        10.72.181.79    9200/tcp  Waiting for TLS to be fully configured...
opensearch-failover/0*       maintenance  idle       2        10.72.181.174   9200/tcp  Waiting for TLS to be fully configured...
opensearch-main/3            maintenance  executing  3        10.72.181.181   9200/tcp  (get-password) Waiting for TLS to be fully configured...
opensearch-main/4*           active       idle       4        10.72.181.111   9200/tcp  
opensearch-main/5            maintenance  idle       5        10.72.181.162   9200/tcp  Waiting for TLS to be fully configured...
self-signed-certificates/0*  active       idle       0        10.72.181.203             

Machine  State    Address        Inst id        Base          AZ  Message
0        started  10.72.181.203  juju-f60bff-0  ubuntu@22.04      Running
1        started  10.72.181.79   juju-f60bff-1  ubuntu@22.04      Running
2        started  10.72.181.174  juju-f60bff-2  ubuntu@22.04      Running
3        started  10.72.181.181  juju-f60bff-3  ubuntu@22.04      Running
4        started  10.72.181.111  juju-f60bff-4  ubuntu@22.04      Running
5        started  10.72.181.162  juju-f60bff-5  ubuntu@22.04      Running
```

juju show-unit for each of the main units:
* opensearch-main/3: https://pastebin.ubuntu.com/p/hVjktcjC2f/
* opensearch-main/4: https://pastebin.ubuntu.com/p/n9vvXfYPrZ/
* opensearch-main/5: https://pastebin.ubuntu.com/p/fqm9G9vQdC/

The reason is because each unit perceives its own neighbors as in-rotating status. I can see it by entering one of the nodes with `juju debug-hooks opensearch-main/3`, and executing the `dispatch` script gives:

```
root@juju-f60bff-3:/var/lib/juju/agents/unit-opensearch-main-3/charm# ./dispatch 
2024-11-11 10:39:05,637 DEBUG    ops 2.16.1 up and running.
2024-11-11 10:39:05,702 DEBUG    Re-emitting deferred event <CertificateAvailableEvent via OpenSearchOperatorCharm/TLSCertificates
RequiresV3[certificates]/on/certificate_available[485]>.
2024-11-11 10:39:05,735 DEBUG    unit.unit-http TLS certificate available.
2024-11-11 10:39:06,573 DEBUG    TLS CA rotation ongoing in unit <ops.model.Unit opensearch-main/4>, will not update tls certifica
tes.
2024-11-11 10:39:06,598 DEBUG    Deferring <CertificateAvailableEvent via OpenSearchOperatorCharm/TLSCertificatesRequiresV3[certif
icates]/on/certificate_available[485]>.
2024-11-11 10:39:06,602 DEBUG    Re-emitting deferred event <CertificateAvailableEvent via OpenSearchOperatorCharm/TLSCertificates
RequiresV3[certificates]/on/certificate_available[486]>.
2024-11-11 10:39:06,605 DEBUG    unit.unit-transport TLS certificate available.
2024-11-11 10:39:07,363 DEBUG    TLS CA rotation ongoing in unit <ops.model.Unit opensearch-main/4>, will not update tls certifica
tes.
2024-11-11 10:39:07,367 DEBUG    Deferring <CertificateAvailableEvent via OpenSearchOperatorCharm/TLSCertificatesRequiresV3[certif
icates]/on/certificate_available[486]>.
2024-11-11 10:39:07,371 DEBUG    Re-emitting deferred event <CertificateAvailableEvent via OpenSearchOperatorCharm/TLSCertificates
RequiresV3[certificates]/on/certificate_available[499]>.
2024-11-11 10:39:07,374 DEBUG    unit.unit-http TLS certificate available.
2024-11-11 10:39:08,056 DEBUG    TLS CA rotation ongoing in unit <ops.model.Unit opensearch-main/4>, will not update tls certifica
tes.
2024-11-11 10:39:08,061 DEBUG    Deferring <CertificateAvailableEvent via OpenSearchOperatorCharm/TLSCertificatesRequiresV3[certif
icates]/on/certificate_available[499]>.
2024-11-11 10:39:08,065 DEBUG    Re-emitting deferred event <CertificateAvailableEvent via OpenSearchOperatorCharm/TLSCertificates
RequiresV3[certificates]/on/certificate_available[500]>.
2024-11-11 10:39:08,068 DEBUG    unit.unit-transport TLS certificate available.
2024-11-11 10:39:08,703 DEBUG    TLS CA rotation ongoing in unit <ops.model.Unit opensearch-main/4>, will not update tls certificates.
2024-11-11 10:39:08,707 DEBUG    Deferring <CertificateAvailableEvent via OpenSearchOperatorCharm/TLSCertificatesRequiresV3[certif
icates]/on/certificate_available[500]>.
2024-11-11 10:39:08,711 DEBUG    Emitting Juju event get_password_action.
2024-11-11 10:39:08,719 DEBUG    Executing command: openssl pkcs12 -in /var/snap/opensearch/current/etc/opensearch/certificates/ca
.p12 -passin pass:xxx
2024-11-11 10:39:08,740 DEBUG    Executing command: openssl x509 -in /tmp/tmp3x85tfgt -noout -issuer
2024-11-11 10:39:08,751 DEBUG    Executing command: openssl pkcs12 -in /var/snap/opensearch/current/etc/opensearch/certificates/un
it-transport.p12 -nodes -passin pass:xxx | openssl x509 -noout -issuer
```

## Reproducer
```
sudo apt install -y python3-pip
sudo pip3 install tox poetry charmcraftcache

sudo snap install charmcraft --classic
sudo snap install juju --classic

sudo lxd init ## just removed the IPv6 and set storage to `dir`

juju bootstrap localhost
juju add-model test

git clone https://github.com/canonical/opensearch-operator
cd opensearch-operator
charmcraftcache pack

tox run -e integration -- 'tests/integration/tls/test_ca_rotation.py' --group='large' -m '' --model test 
```

## Initial Conclusions

The run on debug-hooks above shows that `opensearch-main/3` is considering its leader unit as still executing the CA rotation:

> 2024-11-11 10:39:08,703 DEBUG    TLS CA rotation ongoing in unit <ops.model.Unit opensearch-main/4>, will not update tls certificates.

However, we can see in the show-unit that the `tls_{renewing,renewed}` marks are gone in that unit. Therefore, the main problem here is the fact that opensearch-main/4 is executing [its reset ca rotation state routine](https://github.com/canonical/opensearch-operator/blob/405725f7d87acc2f37e3f91027116b18c54d1153/lib/charms/opensearch/v0/opensearch_tls.py#L845) too early. That fact will make the check [here](https://github.com/canonical/opensearch-operator/blob/405725f7d87acc2f37e3f91027116b18c54d1153/lib/charms/opensearch/v0/opensearch_tls.py#L891) to return as "unit XX is still doing its rotation", although in fact that unit finished its rotation entirely.

phvalguima:
This PR removes one of the filters that was causing a race condition at CA rotation + large deployments. It fixes the CA rotation for large deployments in our CI.

Closes #500

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -859,10 +859,6 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             if self.tls.is_fully_configured():
                 try:
                     self.tls.reload_tls_certificates()
-                except OpenSearchHttpError:
-                    logger.error("Could not reload TLS certificates via API, will restart.")
-                    self._restart_opensearch_event.emit()
-                else:
                     self.status.clear(TLSNotFullyConfigured)
                     self.tls.reset_ca_rotation_state()
                     # if all certs are stored and CA rotation is complete in the cluster
@@ -873,6 +869,10 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
                     ):
                         logger.info("on_tls_conf_set: Detected CA rotation complete in cluster")
                         self.tls.on_ca_certs_rotation_complete()
+
+                except OpenSearchHttpError:
+                    logger.error("Could not reload TLS certificates via API, will restart.")
+                    self._restart_opensearch_event.emit()
 
             else:
                 event.defer()
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -859,6 +859,10 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             if self.tls.is_fully_configured():
                 try:
                     self.tls.reload_tls_certificates()
+                except OpenSearchHttpError:
+                    logger.error("Could not reload TLS certificates via API, will restart.")
+                    self._restart_opensearch_event.emit()
+                else:
                     self.status.clear(TLSNotFullyConfigured)
                     self.tls.reset_ca_rotation_state()
                     # if all certs are stored and CA rotation is complete in the cluster
@@ -869,10 +873,6 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
                     ):
                         logger.info("on_tls_conf_set: Detected CA rotation complete in cluster")
                         self.tls.on_ca_certs_rotation_complete()
-
-                except OpenSearchHttpError:
-                    logger.error("Could not reload TLS certificates via API, will restart.")
-                    self._restart_opensearch_event.emit()
 
             else:
                 event.defer()
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -67,10 +67,6 @@ LIBPATCH = 1
 logger = logging.getLogger(__name__)
 
 
-class OpenSearchTlsRotationWaitNextHookError(OpenSearchError):
-    """We wait for the next hook to finish the TLS rotation."""
-
-
 class OpenSearchTLS(Object):
     """Class that Manages OpenSearch relation with TLS Certificates Operator."""
 
@@ -102,18 +98,6 @@ class OpenSearchTLS(Object):
         self.framework.observe(
             self.certs.on.certificate_invalidated, self._on_certificate_invalidated
         )
-        # We want to avoid a corner case where the leader is the first unit to receive a certificate-available
-        # at CA rotation. It takes the lock (as the leader, it can execute the lock itself rightaway)
-        # which then moves from tls_ca_renewing=True, to `delete tls_ca_renewing` and set tls_ca_renewed=True
-        # That would happen on the call within _on_certificate_available's self.charm.on_tls_ca_rotation()
-        # Once that happens, the leader can still continue to run within the same hook and proceed with
-        # its _on_certificate_available call, reaching: self.charm.on_tls_conf_set(event, scope, cert_type, renewal)
-        # there it could execute the reload successfully and re-execute: reset_ca_rotation_state
-        # Finally, that removes the `tls_ca_renewed` from the databag.
-
-        # Therefore, this boolean exists: we will force the reset_ca_rotation_state to wait a whole hook
-        # between the two calls of reset_ca_rotation_state.
-        self.moved_to_tls_ca_renewed = False
 
     def _on_set_tls_private_key(self, event: ActionEvent) -> None:
         """Set the TLS private key, which will be used for requesting the certificate."""
@@ -268,11 +252,6 @@ class OpenSearchTLS(Object):
                 )
                 self.charm.on_tls_ca_rotation()
                 return
-
-#        elif not self.ca_rotation_complete_in_cluster():
-#            # We have same current_stored_ca and event.ca, but we are executing the rotation
-#            event.defer()
-#            return
 
         # store the certificates and keys in a key store
         self.store_new_tls_resources(
@@ -887,14 +866,11 @@ class OpenSearchTLS(Object):
             )
             self.moved_to_tls_ca_renewed = True
         else:
-#        elif not self.moved_to_tls_ca_renewed:
             # this means only the CA rotation completed, still need to create certificates
             self.charm.peers_data.put(Scope.UNIT, "tls_ca_renewed", True)
             self.update_ca_rotation_flag_to_peer_cluster_relation(
                 flag="tls_ca_renewed", operation="add"
             )
-#        else:
-#            raise OpenSearchTlsRotationWaitNextHookError()
 
     def ca_rotation_complete_in_cluster(self) -> bool:
         """Check whether the CA rotation completed in all units."""
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -222,7 +222,8 @@ class OpenSearchTLS(Object):
 
         if secret != {"chain": ca_chain, "cert": event.certificate, "ca-cert": event.ca}:
             # Juju is not able to check if secrets' content changed between revisions
-            # this IF is intended to reduce a storm of secret-removed/-changed events for the same content
+            # this IF is intended to reduce a storm of secret-removed/-changed events
+            # for the same content
             self.charm.secrets.put_object(
                 scope,
                 cert_type.val,
@@ -864,7 +865,6 @@ class OpenSearchTLS(Object):
             self.update_ca_rotation_flag_to_peer_cluster_relation(
                 flag="tls_ca_renewed", operation="remove"
             )
-            self.moved_to_tls_ca_renewed = True
         else:
             # this means only the CA rotation completed, still need to create certificates
             self.charm.peers_data.put(Scope.UNIT, "tls_ca_renewed", True)
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -67,6 +67,10 @@ LIBPATCH = 1
 logger = logging.getLogger(__name__)
 
 
+class OpenSearchTlsRotationWaitNextHookError(OpenSearchError):
+    """We wait for the next hook to finish the TLS rotation."""
+
+
 class OpenSearchTLS(Object):
     """Class that Manages OpenSearch relation with TLS Certificates Operator."""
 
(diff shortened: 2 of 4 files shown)


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -859,10 +859,6 @@ def on_tls_conf_set(
             if self.tls.is_fully_configured():
                 try:
                     self.tls.reload_tls_certificates()
-                except OpenSearchHttpError:
-                    logger.error("Could not reload TLS certificates via API, will restart.")
-                    self._restart_opensearch_event.emit()
-                else:
```
phvalguima: I am not sure I see the value of the `else` here. Moved the code while testing but I kept it as I think it makes everything easier to read

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -210,23 +210,30 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
         if not self.charm.unit.is_leader() and scope == Scope.APP:
             return
 
-        if not self.ca_rotation_complete_in_cluster():
```
phvalguima, line 213: This is the main change. This check here filters out any `cert-available` events if at least one single unit in the entire cluster started the rotation.

That means we have a race. If we start a rotation, some units will cross this line and eventually get into the rotation part of this function. Once they get there and finish their own flow, then this check will start to return `false` for all the other units that did not yet get a `cert-available` event.

I do not see the value of this code, so I just removed it.

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -210,23 +210,30 @@ def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
...
-            merge=True,
-        )
+        current_secret_obj = self.charm.secrets.get_object(scope, cert_type.val) or {}
+        secret = {
+            "chain": current_secret_obj.get("chain"),
+            "cert": current_secret_obj.get("cert"),
+            "ca-cert": current_secret_obj.get("ca-cert"),
+        }
+
+        if secret != {"chain": ca_chain, "cert": event.certificate, "ca-cert": event.ca}:
```
phvalguima, line 223: Avoiding a wave of `secret-changed` + `secret-removed`, remember juju does not filter it

tests/unit/lib/test_opensearch_tls.py
```diff
@@ -83,6 +83,8 @@ class TestOpenSearchTLS(unittest.TestCase):
 
     @patch("charm.OpenSearchOperatorCharm._put_or_update_internal_user_leader")
     def setUp(self, _) -> None:
+        self.skipTest("TODO: remove this skip")
```
phvalguima: Once we are aligned in the logic, I will fix the unit tests.

@reneradoi keep in mind, some of these checks are failing and I am unsure if my changes are breaking your CA rotation flow. Can you remove this skip, run the unit tests and check if the behavior is intended? If yes, I will update this logic.

lib/charms/opensearch/v0/opensearch_base_charm.py
reneradoi: The `else` part is to make sure and explicit that this code must only run if the `reload_tls_certificates()` was successful. If not, these actions will be done after the restart (in `post_start_init`).

lib/charms/opensearch/v0/opensearch_tls.py
reneradoi: The removed condition is a safeguard to avoid updating certs during a CA rotation. 

Certs should only be updated after the CA has been rotated. If certs with the new CA are stored before all units have the new CA, and this unit/opensearch on this unit restarts (for whatever reason), it will no longer be able to communicate with the other cluster members. This could lead to the cluster to become unavailable.

There is another safeguard for this in `store_new_tls_resources` (see [here](https://github.com/canonical/opensearch-operator/blob/2/edge/lib/charms/opensearch/v0/opensearch_tls.py#L645-L648)). As long as this second safeguard in `store_new_tls_resources` stays in place, I believe we can remove the one her in this line, if it causes a race condition.

lib/charms/opensearch/v0/opensearch_tls.py
reneradoi: Good find!

tests/unit/lib/test_opensearch_tls.py
reneradoi: The unit test failures indicate a changed behaviour. The `run_cmd` they are checking is the one for looking up the current CA and for writing the cert files. If the condition above in `_on_certificate_available` gets removed, then these checks should be changed to assert for 3 runs instead of 0 runs of this command. These 3 runs are happening in `read_stored_ca`. If it is more than 3, then be aware the certs might have been stored.

reneradoi (dismissed):
Hey Pedro, I think this change makes sense. We had a couple of changes on the whole process in the last weeks, therefore the condition `if not self.ca_rotation_complete_in_cluster()` in `_on_certificate_available` is now probably obsolete. The original need for it was to make sure nothing gets updated during the rollout of the new CA: no certs, no secrets, no ca-bundle.

Your assumption of a race condition seems plausible, at least the integration tests look good after the change. Important tough: Make sure the condition `if not self.ca_rotation_complete_in_cluster()` stays in `store_new_tls_resources()`, otherwise the certs might get updated during the rollout of the new CA.

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: I will keep the `else`. I still think it would only matter if there are any other step down in this function that may throw `OpenSearchHttpError`, which is not the case, thou. Otherwise having `else`  or not is irrelevant.

Given this is more a style in this case, I agree leaving things explicit makes more sense.

lib/charms/opensearch/v0/opensearch_tls.py
phvalguima: So, that is interesting. It means the onl real place we have where `store_new_tls_resources`  **really** gets applied then [is here](https://github.com/canonical/opensearch-operator/blob/405725f7d87acc2f37e3f91027116b18c54d1153/lib/charms/opensearch/v0/opensearch_tls.py#L645-L648), would that be correct?

tests/unit/lib/test_opensearch_tls.py
phvalguima: Ack, but does that mean the changes in this PR causes these `run_cmd`  or did we miss a mock in our tests (e.g. we should "tell" the unit test that now `on_ca_rotation_complete` returns False), makes sense?

I am just worried I may accidentally add a problem with these changes...

lib/charms/opensearch/v0/opensearch_tls.py
reneradoi: Yes, this is where the certs are actually applied.

tests/unit/lib/test_opensearch_tls.py
reneradoi: The changes in the PR cause exactly three `run_cmd` commands to be executed: checking the current CA for the admin cert, the unit_http cert and the unit_transport cert. The unit tests assert the amount of times this `run_cmd` command is executed, that's why we need to adjust it.

reneradoi (dismissed):
Hey @phvalguima the unit tests look fine now. The only thing that needs to be adjusted is the comments in line 1581 and 1690 of `test_opensearch_tls.py` saying `In case any stage of a CA cert rotation is being processed, further 'certificate-available' events are deferred.`

Other than that, the change can be merged from my point of view.

reneradoi (approved):

Mehdi-Bendriss (approved):

PR #504, opened 2024-11-14
Change duplicated rules names

gabrielcocenza:
## Issue
Some rules has duplicated names that can be a little bit confusing. 

## Solution
This PR separate into different names to be more clear with the alert rule expression and adds spaces between alerts to increase readability.


phvalguima (approved):
LGTM

Mehdi-Bendriss (approved):
Nice catch, thank you!

PR #505, opened 2024-11-14
Add CI and unit tests for prometheus rules using promtool

gabrielcocenza:
- It's good to have unit tests for the prometheus alerts and also check if the prometheus_alerts.yaml is a valid file.

- using promtool is possible to check the rules and also run unit tests.

## How to test this PR

1. Install the prometheus snap:
```shell
sudo snap install prometheus
```

2. Run to check if the rules file is valid:
```shell
promtool check rules src/alert_rules/prometheus/*.yaml
```

3. Run the unit tests:
```
promtool test rules tests/unit/test_alert_rules/*.yaml
```

### DIFF
--- a/.github/workflows/test_prometheus_rules.yaml
+++ b/.github/workflows/test_prometheus_rules.yaml
@@ -3,12 +3,6 @@ name: Test prometheus rules
 on:
   workflow_call:
   workflow_dispatch:
-  pull_request:
-    types: [opened, synchronize, reopened]
-    branches: [main]
-    paths-ignore:
-      - "**.md"
-      - "**.rst"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
--- a/.github/workflows/test_prometheus_rules.yaml
+++ b/.github/workflows/test_prometheus_rules.yaml
@@ -1,0 +1,34 @@
+name: Test prometheus rules
+
+on:
+  workflow_call:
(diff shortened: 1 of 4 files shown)


.github/workflows/test_prometheus_rules.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - "**.md"
+      - "**.rst"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  promtool:
+    runs-on: ubuntu-latest
```
phvalguima, line 13: Is it possible to move to self-hosted CI?

gabrielcocenza: I think it is. Is there a specific reason to change to self-hosted?

zmraul (changes requested):
See [comment on opensearch-dashboards](https://github.com/canonical/opensearch-dashboards-operator/pull/142#pullrequestreview-2500122619)

.github/workflows/test_prometheus_rules.yaml
phvalguima: Not really, just more runners. But as you removed from the pull request path, as @zmraul suggested, I am fine keeping it here.

phvalguima (approved):

zmraul (approved):

PR #506, opened 2024-11-21
Separate alerts when cluster is in yellow state

gabrielcocenza:
When a cluster is in yellow state, it's because some shards are not active. This can be reached under heavy load, however if there are shards unassigned and the cluster is yellow, that means that some replicas are not allocated and it might be necessary to add new nodes in order to host all shards.

### DIFF
--- a/src/alert_rules/prometheus/prometheus_alerts.yaml
+++ b/src/alert_rules/prometheus/prometheus_alerts.yaml
@@ -11,13 +11,23 @@
     "labels":
       "severity": "critical"
 
+  - "alert": "OpenSearchClusterYellowTemp"
+    "annotations":
+      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Cluster might be under heavy load."
+      "summary": "Cluster health status is temporarily YELLOW"
+    "expr": |
+      sum by (cluster, instance) (opensearch_cluster_shards_number{type=~"relocating|initializing"}) > 0 and on(cluster, instance) opensearch_cluster_status == 1
+    "for": "20m"
+    "labels":
+      "severity": "warning"
+
   - "alert": "OpenSearchClusterYellow"
     "annotations":
-      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Some cluster replicas shards are not allocated."
-      "summary": "Cluster health status is YELLOW"
+      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW with some replica shards unassigned."
+      "summary": "Number of nodes in the cluster might be too low. Consider scaling the application to ensure that it has enough nodes to host all shards."
     "expr": |
-      sum by (cluster) (opensearch_cluster_status == 1)
-    "for": "20m"
+      sum by (cluster, instance) (opensearch_cluster_shards_number{type="unassigned"}) > 0 and on(cluster, instance) opensearch_cluster_status == 1
+    "for": "10m"
     "labels":
       "severity": "warning"
 
--- a/src/alert_rules/prometheus/prometheus_alerts.yaml
+++ b/src/alert_rules/prometheus/prometheus_alerts.yaml
@@ -13,20 +13,20 @@
 
   - "alert": "OpenSearchClusterYellowTemp"
     "annotations":
-      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Cluster might be under heavy load."
+      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Shards are still still relocating or initializing. The cluster might be under heavy load."
       "summary": "Cluster health status is temporarily YELLOW"
     "expr": |
-      sum by (cluster, instance) (opensearch_cluster_shards_number{type=~"relocating|initializing"}) > 0 and on(cluster, instance) opensearch_cluster_status == 1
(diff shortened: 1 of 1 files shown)


src/alert_rules/prometheus/prometheus_alerts.yaml
```diff
@@ -11,16 +11,26 @@
...
-  - "alert": "OpenSearchClusterYellow"
+  - "alert": "OpenSearchClusterYellowTemp"
     "annotations":
-      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Some cluster replicas shards are not allocated."
-      "summary": "Cluster health status is YELLOW"
+      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Cluster might be under heavy load."
+      "summary": "Cluster health status is temporarily YELLOW"
     "expr": |
-      sum by (cluster) (opensearch_cluster_status == 1)
+      sum by (cluster, instance) (opensearch_cluster_shards_number{type=~"relocating|initializing"}) > 0 and on(cluster, instance) opensearch_cluster_status == 1
```
phvalguima: @gabrielcocenza breaking per instance here?

phvalguima: *note* The timing seems perfect: I agree we should wait for a longer time before raising the "YELLOW_TEMP" vs. "YELLOW", as yellow is a more serious condition and yellow temp may just be consequence of a busy cluster,

gabrielcocenza: I thought about breaking per instance to try to show which "node" it's not able to host the shard. In a theoretic scenario nodes can have different capacity.

However I think it would be ok just break for cluster. What do you think?

phvalguima: makes sense, but it will be X alerts at once, or very closely I guess?

phvalguima (dismissed):
LGTM, I don't have a strong opinion on per cluster or per instance, I think per instance makes sense as the user will receive multiple alerts but they will give you an itemized view of the servers in trouble.

src/alert_rules/prometheus/prometheus_alerts.yaml
```diff
@@ -11,16 +11,26 @@
     "labels":
       "severity": "critical"
 
-  - "alert": "OpenSearchClusterYellow"
+  - "alert": "OpenSearchClusterYellowTemp"
     "annotations":
-      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Some cluster replicas shards are not allocated."
-      "summary": "Cluster health status is YELLOW"
+      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Cluster might be under heavy load."
```
Mehdi-Bendriss:
```suggestion
      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Shards are still still relocating or initializing. The cluster might be under heavy load."
```

src/alert_rules/prometheus/prometheus_alerts.yaml
Mehdi-Bendriss: We should not have an alert at the instance level. We should alerts on the cluster level and with a full view over the instances.

src/alert_rules/prometheus/prometheus_alerts.yaml
```diff
@@ -11,16 +11,26 @@
...
     "expr": |
-      sum by (cluster) (opensearch_cluster_status == 1)
+      sum by (cluster, instance) (opensearch_cluster_shards_number{type=~"relocating|initializing"}) > 0 and on(cluster, instance) opensearch_cluster_status == 1
     "for": "20m"
     "labels":
       "severity": "warning"
 
+  - "alert": "OpenSearchClusterYellow"
+    "annotations":
+      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW with some replica shards unassigned."
```
Mehdi-Bendriss:
```suggestion
      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW. Some replica shards are unassigned."
```

Mehdi-Bendriss (changes requested):
Thanks Gabriel. I would rather rather have alerting on a cluster level to not spam with alerts instead of at the instance level.

zmraul (dismissed):

src/alert_rules/prometheus/prometheus_alerts.yaml
```diff
@@ -22,16 +22,26 @@
     "labels":
       "severity": "critical"
 
-  - "alert": "OpenSearchClusterYellow"
+  - "alert": "OpenSearchClusterYellowTemp"
     "annotations":
-      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Some cluster replicas shards are not allocated."
-      "summary": "Cluster health status is YELLOW"
+      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Shards are still still relocating or initializing. The cluster might be under heavy load."
```
Mehdi-Bendriss: Sorry, typo on my previous suggestion: 
```suggestion
      "message": "Cluster {{ $labels.cluster }} health status has been YELLOW for at least 20m. Shards are still relocating or initializing. The cluster might be under heavy load."
```

Mehdi-Bendriss (approved):

zmraul (approved):

PR #509, opened 2024-11-28
Add new alert rules for throttling

gabrielcocenza:
- If OpenSearch is throttling, this is an alert that optimizations are necessary like scaling the number of nodes or changing queries and indexing patterns

### DIFF
--- a/src/alert_rules/prometheus/prometheus_alerts.yaml
+++ b/src/alert_rules/prometheus/prometheus_alerts.yaml
@@ -129,16 +129,6 @@
     "labels":
       "severity": "alert"
 
-  - "alert": "OpenSearchFewNodes"
-    "annotations":
-      "message": "Cluster {{ $labels.cluster }} has less nodes than expected. It's recommend to have at least 3 nodes for high availability."
-      "summary": "Few OpenSearch nodes on cluster"
-    "expr": |
-      sum by (cluster) (opensearch_cluster_nodes_number) / count by (cluster) (opensearch_cluster_nodes_number) < 3
-    "for": "5m"
-    "labels":
-      "severity": "warning"
-
   - "alert": "OpenSearchThrottling"
     "annotations":
       "message": "Cluster {{ $labels.cluster }} is throttling. Please optimize queries and indexing patterns or consider scale the application."
--- a/src/alert_rules/prometheus/prometheus_alerts.yaml
+++ b/src/alert_rules/prometheus/prometheus_alerts.yaml
@@ -105,3 +105,23 @@
     "for": "1m"
     "labels":
       "severity": "alert"
+
+  - "alert": "OpenSearchFewNodes"
+    "annotations":
+      "message": "Cluster {{ $labels.cluster }} has less nodes than expected. It's recommend to have at least 3 nodes for high availability."
+      "summary": "Few OpenSearch nodes on cluster"
+    "expr": |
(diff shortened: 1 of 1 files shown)


src/alert_rules/prometheus/prometheus_alerts.yaml
```diff
@@ -105,3 +105,23 @@
     "for": "1m"
     "labels":
       "severity": "alert"
+
+  - "alert": "OpenSearchFewNodes"
```
zmraul: In my opinion this alert is not too relevant. High availability requirements should be explained on docs, or deployment requisites, not trigger once integrated with observability.

zmraul (dismissed):

src/alert_rules/prometheus/prometheus_alerts.yaml
gabrielcocenza: I kind agree. However, the generated metrics cannot ensure that a node is lost apart from checking if the `up`metric is working or not. I created this as a warning just to trigger something not critical to operators know that the current cluster is not operating as expected.

src/alert_rules/prometheus/prometheus_alerts.yaml
```diff
@@ -105,3 +105,23 @@
...
+    "annotations":
+      "message": "Cluster {{ $labels.cluster }} has less nodes than expected. It's recommend to have at least 3 nodes for high availability."
+      "summary": "Few OpenSearch nodes on cluster"
+    "expr": |
+      sum by (cluster) (opensearch_cluster_nodes_number) / count by (cluster) (opensearch_cluster_nodes_number) < 3
```
phvalguima: We are supporting from 1+ nodes, so this alert would make sense if we have *at least 3x units* and then we have   one or more failing.

src/alert_rules/prometheus/prometheus_alerts.yaml
```diff
@@ -105,3 +105,23 @@
...
+    "for": "5m"
+    "labels":
+      "severity": "warning"
+
+  - "alert": "OpenSearchThrottling"
+    "annotations":
+      "message": "Cluster {{ $labels.cluster }} is throttling. Please optimize queries and indexing patterns or consider scale the application."
```
phvalguima, line 134: That is a good one!

phvalguima (commented):
Hey @gabrielcocenza we cannot unfortunately limit ourselves to 3+ nodes. The scenario of 1-node and 2-node cluster is also supposed to work. On the other side, the throttling alert is a really good one!

src/alert_rules/prometheus/prometheus_alerts.yaml
gabrielcocenza: I had the assumption that for production OpenSearch we would have at least 3 nodes.
I'll remove this rule and just let the throttling one

src/alert_rules/prometheus/prometheus_alerts.yaml
gabrielcocenza: I've removed the node count alert. We can add in the future if necessary

src/alert_rules/prometheus/prometheus_alerts.yaml
gabrielcocenza: Rule removed

phvalguima (approved):

zmraul (approved):

PR #510, opened 2024-12-02
Add revision to the dashboard title of the Grafana

gabrielcocenza:
With this change, two different OpenSearch applications in different revisions, will have separate dashboards when related to COS.

- data-platform-helpers was added to use the get_charm_revision
- titles of dashboards are changed during the charm upgrade hook
- config_changed was added on refresh_events of COSAgentProvider to re-render the dashboards after a charm upgrade

## How to test it

1. Deploy more than one opensearch application
2. Relate both apps with grafana-agent
3. In one of the applications use juju refresh and apply a local build of this patch
4. Go to the Grafana dashboards an you will see that two dashboards are created:
  - one with no revision (from charmhub)
  - one with Rev 0 (meaning that is a local build)
  
  
 
image


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -244,6 +244,7 @@ def __init__(self, *args, distro: Type[OpenSearchDistribution] = None):
             metrics_endpoints=[],
             scrape_configs=self._scrape_config,
             refresh_events=[
+                self.on.config_changed,
```
zmraul, line 247: **question**: Why is this one needed?

gabrielcocenza: The `COSAgentProvider` has [self.on.config_changed](https://github.com/canonical/grafana-agent-operator/blob/main/lib/charms/grafana_agent/v0/cos_agent.py#L655C81-L656C1) by default if no custom events  are passed.

We are passing custom events, so there is no refresh on COS for config-changed.
When you upgrade a charm, after that the config-change hook is triggered.

In summary, I'm re adding the default behavior and at the same time cover my need of refreshing COS relation after charm upgrade

gabrielcocenza: Another thing...
I didn't add `self.on_charm_upgrade` because I was seeing a race condition. Talking with the [COS team](https://matrix.to/#/!nHXpRkcSNJHlHdUGbQ:ubuntu.com/$ravfUK4XN39YUeqybKqMiDBzuPo0fzh9Y_WJp3sVhbs?via=ubuntu.com&via=matrix.org), the order when you add those handlers matters and COS Provider should in theory be added as a last handler.

To do that, we would need to move the `COSAgentProvider` to the last child class of the inheritance and I though that it would be simpler just add back the config-changed

phvalguima: Shouldn't we refresh on COS broken event as well? It is a slightly off topic question for this PR but how can we assure the dashboard is gone if we remove the COS relation from opensearch?

gabrielcocenza: I've tested and as expected (at least for me), the dashboards are removed if the relation with grafana-agent is removed

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -228,3 +231,44 @@ def mask_sensitive_information(cmd: str) -> str:
     pattern = re.compile(r"(-tspass\s+|-kspass\s+|-storepass\s+|-new\s+|pass:)(\S+)")
 
     return re.sub(pattern, r"\1" + "xxx", cmd)
+
+
+def update_grafana_dashboards_titles(charm: "OpenSearchBaseCharm") -> None:
```
Mehdi-Bendriss: Please move this to a separate helper `helper_cos`

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -228,3 +231,44 @@ def mask_sensitive_information(cmd: str) -> str:
...
+    """Update the titles in the specified directory to include the charm revision."""
+    revision = get_charm_revision(charm.model.unit)
+    path = charm.charm_dir / "src/grafana_dashboards"
+
+    for dashboard_path in path.iterdir():
+        if dashboard_path.is_file() and dashboard_path.suffix == ".json":
```
Mehdi-Bendriss: why not directly target `opensearch.json`?

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -228,3 +231,44 @@ def mask_sensitive_information(cmd: str) -> str:
...
+            try:
+                _update_dashboard_title(revision, dashboard_path)
+            except (json.JSONDecodeError, IOError) as e:
+                logger.error("Failed to process %s: %s", dashboard_path.name, str(e))
+        else:
+            logger.warning("%s is not a valid JSON file", dashboard_path.name)
```
Mehdi-Bendriss: I don't think we need to handle these

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -228,3 +231,44 @@ def mask_sensitive_information(cmd: str) -> str:
...
+
+
+def _update_dashboard_title(revision: str, dashboard_path: Path) -> None:
+    """Update the title of a Grafana dashboard file to include the charm revision."""
+    with open(dashboard_path, "r") as file:
+        dashboard = json.load(file)
+
+    old_title = dashboard.get("title")
+    if old_title:
```
Mehdi-Bendriss: nit:
```suggestion
    if old_title := dashboard.get("title"):
```

I believe we can always assume dashboards have a title and no need to test on it, otherwise it's a bug

Mehdi-Bendriss (commented):
Thanks Gabriel! Looks good, I left a couple of questions and comments.

lib/charms/opensearch/v0/helper_charm.py
gabrielcocenza: I guess it will never happen, but I was thinking that maybe the file can change or maybe in the future we can have more than one file. In such scenario the code would still work. Maybe is not necessary. What do you think? Should I make it simple and target `opensearch.json` directly?

Mehdi-Bendriss: Yeah I would just keep it simple and simply go with `opensearch.json` in this case

lib/charms/opensearch/v0/helper_charm.py
Mehdi-Bendriss: comment still relevant - the exception handling is not needed if we target `opensearch.json` - as a matter of fact we should let it crash instead of masking the issue

we can safely call `_update_dashboard_title(revision, dashboard_path)` directly

Mehdi-Bendriss (approved):

PR #523, opened 2024-12-16
Add critical alert if cluster is throttling for two hours

gabrielcocenza:
After showing the current alert rules for Managed Solutions, they asked to have a critical alert if the cluster is throttling for two hours.

### DIFF
--- a/src/alert_rules/prometheus/prometheus_alerts.yaml
+++ b/src/alert_rules/prometheus/prometheus_alerts.yaml
@@ -131,7 +131,7 @@
 
   - "alert": "OpenSearchThrottling"
     "annotations":
-      "message": "Cluster {{ $labels.cluster }} is throttling. Please optimize queries and indexing patterns or consider scale the application."
+      "message": "Cluster {{ $labels.cluster }} is throttling. Please review your indexing request rate, index lifecycle or consider scale the application."
       "summary": "OpenSearch Indexing Throttle"
     "expr": |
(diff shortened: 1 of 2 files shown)


src/alert_rules/prometheus/prometheus_alerts.yaml
```diff
@@ -138,3 +138,13 @@
     "for": "5m"
     "labels":
       "severity": "warning"
+
+  - "alert": "OpenSearchThrottlingTooLong"
+    "annotations":
+      "message": "Cluster {{ $labels.cluster }} is throttling for two hours. Please optimize queries and indexing patterns or consider scale the application."
```
phvalguima: I would recommend a slightly different text. Generally, when we get into this situation, it means the indexing request rate is higher than we anticipated OR we did not *planned* this index very well. For the later, something like lifecycle management and breaking the index into different bits (e.g. a time based index ingesting logs can be broken into a per-day index) would be more appropriate.

```suggestion
      "message": "Cluster {{ $labels.cluster }} is throttling for at least two hours. Please review your indexing request rate, index lifecycle or consider scale the application."
```

phvalguima (dismissed):
Some slight rewording on suggestions how to address throttling. Other than that it is a +1 for me

zmraul (dismissed):

phvalguima (dismissed):

zmraul (approved):

Mehdi-Bendriss (approved):

PR #531, opened 2025-01-14
[DPE-5982][DPE-6088] Azure storage

zmraul:
Implements Azure storage integration.

Changes:
- All general objects (`OpenSearchBackups<nameOfClass>`) that worked under S3 integration have been renamed to be specific about S3.
- `OpenSearchBackupBase` is now generic for both Azure and S3 integrators.
- On `OpenSearchBaseCharm`, `self.backup()` is reworked to take into account both relations existing, or not. It will then generate the correct sub class for the appropriate relation.
- Constants and error messages are a bit more generic to allow use on both relation error scenarios
- New model `AzureRelData` for the integrator databag.

### DIFF
--- a/lib/charms/opensearch/v0/models.py
+++ b/lib/charms/opensearch/v0/models.py
@@ -393,7 +393,7 @@ class AzureRelData(Model):
     storage_account: str = Field(alias="storage-account", default="")
     container: str = Field(default="")
     endpoint: Optional[str] = Field(default="")
-    path: Optional[str] = Field(default=AZURE_REPO_BASE_PATH)
+    base_path: Optional[str] = Field(alias="path", default=AZURE_REPO_BASE_PATH)
     connection_protocol: Optional[str] = Field(alias="connection-protocol", default=None)
     credentials: AzureRelDataCredentials = Field(
         alias=AZURE_CREDENTIALS, default=AzureRelDataCredentials()
--- a/lib/charms/opensearch/v0/models.py
+++ b/lib/charms/opensearch/v0/models.py
@@ -3,13 +3,14 @@
 
 """Cluster-related data structures / model classes."""
 import json
+import logging
 from abc import ABC
 from datetime import datetime
 from hashlib import md5
 from typing import Any, Dict, List, Literal, Optional
 
-from charms.opensearch.v0.constants_charm import S3_REPO_BASE_PATH
-from charms.opensearch.v0.constants_secrets import S3_CREDENTIALS
+from charms.opensearch.v0.constants_charm import AZURE_REPO_BASE_PATH, S3_REPO_BASE_PATH
+from charms.opensearch.v0.constants_secrets import AZURE_CREDENTIALS, S3_CREDENTIALS
 from charms.opensearch.v0.helper_enums import BaseStrEnum
 from pydantic import BaseModel, Field, root_validator, validator
 from pydantic.utils import ROOT_KEY
@@ -27,6 +28,9 @@ LIBPATCH = 1
 
 MIN_HEAP_SIZE = 1024 * 1024  # 1GB in KB
 MAX_HEAP_SIZE = 32 * MIN_HEAP_SIZE  # 32GB in KB
+
+
+logger = logging.getLogger(__name__)
 
 
 class Model(ABC, BaseModel):
@@ -316,7 +320,9 @@ class S3RelData(Model):
         ):
             raise ValueError("Missing fields: access_key, secret_key")
 
-        # Both bucket and endpoint must be set, or not.
+        # NOTE: Both bucket and endpoint must be set. If none of them are set,
+        # but credentials were found, this likely means that we are validating for a
+        # non cluster_manager application, which only needs credentials.
         if values.get("bucket") and not values.get("endpoint"):
             raise ValueError("Missing field: endpoint")
         if values.get("endpoint") and not values.get("bucket"):
@@ -366,6 +372,87 @@ class S3RelData(Model):
         )
 
 
+class AzureRelDataCredentials(Model):
+    """Model class for credentials passed on the Azure relation."""
+
+    storage_account: str = Field(alias="storage-account", default=None)
+    secret_key: str = Field(alias="secret-key", default=None)
+
+    class Config:
+        """Model config of this pydantic model."""
+
+        allow_population_by_field_name = True
+
+
+class AzureRelData(Model):
+    """Model class for the Azure relation data.
+
+    This model should receive the data directly from the relation and map it to a model.
+    """
+
+    storage_account: str = Field(alias="storage-account", default="")
+    container: str = Field(default="")
+    endpoint: Optional[str] = Field(default="")
+    path: Optional[str] = Field(default=AZURE_REPO_BASE_PATH)
+    connection_protocol: Optional[str] = Field(alias="connection-protocol", default=None)
+    credentials: AzureRelDataCredentials = Field(
+        alias=AZURE_CREDENTIALS, default=AzureRelDataCredentials()
+    )
+
+    class Config:
+        """Model config of this pydantic model."""
+
+        allow_population_by_field_name = True
+
+    @root_validator
+    def validate_core_fields(cls, values):  # noqa: N805
+        """Validate the core fields of the azure relation data."""
+        if (
+            not (creds := values.get("credentials"))
+            and not creds.storage_account
+            and not creds.secret_key
+        ):
+            raise ValueError("Missing fields: storage_account, secret_key")
+
+        # NOTE: Both storage_account and container must be set. If none of them are set,
+        # but credentials were found, this likely means that we are validating for a
+        # non cluster_manager application, which only needs credentials.
+        core_values = ["container"]
+        found = [item for item in core_values if values.get(item)]
+
+        if len(found) != len(core_values) and len(found) > 0:
+            raise ValueError(f"Missing fields: {set(core_values) - set(found)}")
+
+        return values
+
+    @validator(AZURE_CREDENTIALS, check_fields=False)
+    def ensure_secret_content(cls, conf: Dict[str, str] | AzureRelDataCredentials):  # noqa: N805
+        """Ensure the secret content is set."""
+        if not conf:
+            return None
+
+        data = conf
+        if isinstance(conf, dict):
+            data = AzureRelDataCredentials.from_dict(conf)
+
+        for value in data.dict().values():
+            if value.startswith("secret://"):
+                raise ValueError(f"The secret content must be passed, received {value} instead")
+        return data
+
+    @classmethod
+    def from_relation(cls, input_dict: Optional[Dict[str, Any]]):
+        """Create a new instance of this class from a json/dict repr.
+
+        This method creates a nested AzureRelDataCredentials object from the input dict.
+        """
+        if not input_dict:
+            return cls()
+
+        creds = AzureRelDataCredentials(**input_dict)
+        return cls.from_dict(dict(input_dict) | {AZURE_CREDENTIALS: creds.dict()})
+
+
 class PeerClusterRelDataCredentials(Model):
     """Model class for credentials passed on the PCluster relation."""
 
@@ -377,6 +464,7 @@ class PeerClusterRelDataCredentials(Model):
     monitor_password: Optional[str]
     admin_tls: Optional[Dict[str, Optional[str]]]
     s3: Optional[S3RelDataCredentials]
+    azure: Optional[AzureRelDataCredentials]
 
 
 class PeerClusterApp(Model):
--- a/lib/charms/opensearch/v0/models.py
+++ b/lib/charms/opensearch/v0/models.py
@@ -414,17 +414,6 @@ class AzureRelData(Model):
         ):
             raise ValueError("Missing fields: storage_account, secret_key")
 
-        # NOTE: Both storage_account and container must be set. If none of them are set,
-        # but credentials were found, this likely means that we are validating for a
-        # non cluster_manager application, which only needs credentials.
-        core_values = ["container"]
-        found = [item for item in core_values if values.get(item)]
-
-        if len(found) != len(core_values) and len(found) > 0:
-            raise ValueError(f"Missing fields: {set(core_values) - set(found)}")
-
-        return values
-
     @validator(AZURE_CREDENTIALS, check_fields=False)
     def ensure_secret_content(cls, conf: Dict[str, str] | AzureRelDataCredentials):  # noqa: N805
         """Ensure the secret content is set."""
--- a/lib/charms/opensearch/v0/models.py
+++ b/lib/charms/opensearch/v0/models.py
@@ -414,6 +414,8 @@ class AzureRelData(Model):
         ):
             raise ValueError("Missing fields: storage_account, secret_key")
 
+        return values
+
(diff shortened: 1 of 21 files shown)


lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -211,39 +213,62 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+                self.charm.on[relation].relation_created, self._on_backup_relation_created
+            )
+
         for event in [
             self.charm.on.create_backup_action,
             self.charm.on.list_backups_action,
             self.charm.on.restore_action,
         ]:
-            self.framework.observe(event, self._on_s3_relation_action)
+            self.framework.observe(event, self._on_backup_relation_action)
```
phvalguima: I think the name `_on_backup_action` makes more sense here, or simply `_on_action`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -211,39 +213,62 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+            logger.warning(
+                "Modifying relations during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
+            )
+
+    def _on_backup_relation_broken(self, event: EventBase) -> None:
+        """Defers the backup relation broken events."""
         raise NotImplementedError
 
-    def _on_s3_relation_action(self, event: EventBase) -> None:
+    def _on_backup_relation_action(self, event: EventBase) -> None:
```
phvalguima: Same comment

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -390,30 +419,67 @@ def is_idle_or_not_set(self) -> bool:
             self.is_set() or self.is_backup_in_progress() or self._is_restore_in_progress()
         )
 
+    def _list_backups(self, repository: str) -> Dict[int, str]:
```
phvalguima: So, for me to understand, we moved the  list backups to the base class. That means effectively any unit, including non-cluster-manager units, can run that right? If so, I liked the idea.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -508,42 +575,19 @@ def _on_s3_relation_event(self, event: EventBase) -> None:
             self.charm.peer_cluster_provider.refresh_relation_data(event)
 
     @override
-    def _on_s3_relation_action(self, event: EventBase) -> None:
+    def _on_backup_relation_action(self, event: EventBase) -> None:
```
phvalguima: Same comment as before.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1076,25 +1093,590 @@ def get_service_status(  # noqa: C901
...
+    def _on_list_backups_action(self, event: ActionEvent) -> None:
+        """Returns the list of available backups to the user."""
+        backups = {}
+        try:
+            backups = self._list_backups(AZURE_REPOSITORY)
+        except OpenSearchError as e:
+            event.fail(
+                f"List backups action failed - {str(e)} - check the application logs for the full stack trace."
+            )
+        if event.params.get("output").lower() == "json":
```
phvalguima, line 1188: I think we should move this logic to the main list method in the base class instead. It makes sense to have it for both types of backup.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1076,25 +1093,590 @@ def get_service_status(  # noqa: C901
...
+            event.set_results({"backups": json.dumps(backups)})
+        elif event.params.get("output").lower() == "table":
+            event.set_results({"backups": self._generate_backup_list_output(backups)})
+        else:
+            event.fail("Failed: invalid output format, must be either json or table")
+
+    def _close_indices(self, indices: Set[str]) -> bool:
```
phvalguima, line 1195: As you are also using this method, we can do one of the two:
1) We move this and `_close_indices_if_needed` to the base class, although I avoided that movement initially because only the main orchestrator backup uses it; now it makes sense to have it together
2) We create a base class for the main orchestrator where we have these common methods all in one place, then S3 and Azure inherit from this base class

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1076,25 +1093,590 @@ def get_service_status(  # noqa: C901
...
+        all_closed = all(
+            [state and state.get("closed") for state in resp.get("indices", {}).values()]
+        )
+        if not all_closed:
+            return False
+
+        # Finally, we can state it is all good
+        return True
+
+    def _close_indices_if_needed(self, backup_id: int) -> Set[str]:
```
phvalguima: Same thing, let's avoid repeated code.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1076,25 +1093,590 @@ def get_service_status(  # noqa: C901
         return status
 
 
+class OpenSearchAzureBackup(OpenSearchBackupBase):
```
phvalguima, line 1135: I am seeing a lot of repeated code when compared to the OpenSearchS3Backup. I think we could cut a lot of repetition if we put this all together and give a way to just switch the repository type.

Maybe, we could create a "manager" class that interacts with the APIs directly. In this case, it receives the type of repository as an argument at creation.

```
class OpenSearchMainBackupManager:

    def __init__(self, repository_type: s3|azure):
         ...
```

Then that class could contain methods such as "_restore" or "_backup".

The OpenSearchAzureBackup and S3Backup then focus on:
* events coming from the operator, as they may be different
* any particular case of one given backup

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1076,25 +1093,590 @@ def get_service_status(  # noqa: C901
...
+                indices_to_close.add(index)
+
+        try:
+            if not self._close_indices(indices_to_close):
+                raise OpenSearchRestoreIndexClosingError()
+        except OpenSearchError as e:
+            raise OpenSearchRestoreIndexClosingError(e)
+        return indices_to_close
+
+    def _restore(self, backup_id: int) -> Dict[str, Any]:
```
phvalguima: Same comment about the code. BTW there is a mistake I've made in the type here that got propagated: backup_id is a string.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1076,25 +1093,590 @@ def get_service_status(  # noqa: C901
...
+        try:
+            return (
+                backup_id in backups.keys()
+                and self.get_snapshot_status(backups[backup_id]["state"])
+                == BackupServiceState.SUCCESS
+            )
+        except OpenSearchListBackupError:
+            return False
+
+    def _on_restore_backup_action(self, event: ActionEvent) -> None:  # noqa #C901
```
phvalguima, line 1302: Do you think these action methods are repeated when compared to S3?  Maybe we can have a common parent for both main orchestrator classes.

phvalguima (changes requested):
Hi @zmraul, thanks for giving the first trial testing on the plugin manager and the backup component. Nice changes adding the Azure integration here.

I have some suggestions which are more targeting to reduce repeated code. I think we can just move things to be more concise so we avoid repetition. I don't think we should be changing but rather moving code around only!

I recommend we create two new classes:
1) Manager class: this class with deal with all-things API (we can discuss for the main orchestrator only
2) Base class for the main orchestrators

So, the class that is going to be the manager (let's call it `BackupManager`) receives the type of repository as an argument and save it. This class is responsible to do calls such as API calls for backup / restore. For now, we can focus on offloading the main orchestrator.

Then, we have a 2nd class which could be a common base for the main orchestrators, if we need it. This class would essentially manage common juju events such as backup and restore.

Each S3 or Azure classes would then be dedicated to the Juju events that are particular to each integrator

lib/charms/opensearch/v0/models.py
```diff
@@ -366,6 +372,87 @@ def from_relation(cls, input_dict: Optional[Dict[str, Any]]):
...
+            raise ValueError("Missing fields: storage_account, secret_key")
+
+        # NOTE: Both storage_account and container must be set. If none of them are set,
+        # but credentials were found, this likely means that we are validating for a
+        # non cluster_manager application, which only needs credentials.
+        core_values = ["container"]
+        found = [item for item in core_values if values.get(item)]
+
+        if len(found) != len(core_values) and len(found) > 0:
+            raise ValueError(f"Missing fields: {set(core_values) - set(found)}")
```
Mehdi-Bendriss: can you explain this? this would effectively be checking on the `not values.get("container")`, right?

lib/charms/opensearch/v0/models.py
```diff
@@ -366,6 +372,87 @@ def from_relation(cls, input_dict: Optional[Dict[str, Any]]):
...
+        """Ensure the secret content is set."""
+        if not conf:
+            return None
+
+        data = conf
+        if isinstance(conf, dict):
+            data = AzureRelDataCredentials.from_dict(conf)
+
+        for value in data.dict().values():
+            if value.startswith("secret://"):
```
Mehdi-Bendriss, line 430: Do we need to make this check? or can we make an assumption here?

lib/charms/opensearch/v0/models.py
```diff
@@ -366,6 +372,87 @@ def from_relation(cls, input_dict: Optional[Dict[str, Any]]):
...
+                raise ValueError(f"The secret content must be passed, received {value} instead")
+        return data
+
+    @classmethod
+    def from_relation(cls, input_dict: Optional[Dict[str, Any]]):
```
Mehdi-Bendriss, line 435: I believe you can directly use `AzureRelData.from_dict(input_dict)`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -214,38 +241,62 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
+        """Defers the backup relation broken events."""
+        raise NotImplementedError
+
+    def _on_backup_action(self, event: ActionEvent) -> None:
         """No deployment description yet, fail any actions."""
         logger.info("Deployment description not yet available, failing actions.")
         event.fail("Failed: deployment description not yet available")
 
+    @property
+    def active_relation(self) -> str | None:
```
Mehdi-Bendriss: nice!

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -285,9 +336,31 @@ def _query_restore_status(self) -> BackupServiceState:
...
+        """
+        try:
+            indices_status = self.charm.opensearch.request(
+                "GET",
+                "/_recovery?human",
+                retries=6,
+                timeout=10,
+            )
+        except OpenSearchHttpError:
+            raise OpenSearchRestoreCheckError("_is_restore_complete: failed to get indices status")
```
Mehdi-Bendriss: I believe we should not raise given the method returns a boolean. 
It seems to have been the case on the previous implementation, we will address this in the future.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -453,73 +567,79 @@ def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = PeerCluste
...
         """Processes the peer-cluster relation events."""
-        try:
-            self.charm.plugin_manager.apply_config(
-                OpenSearchBackupPlugin(
-                    charm=self.charm,
-                ).config(),
-            )
-        except (OpenSearchKeystoreNotReadyError, OpenSearchPluginMissingConfigError):
-            logger.info("Keystore not ready yet, we wait for another peer cluster.")
+        plugins = [OpenSearchS3Plugin(charm=self.charm), OpenSearchAzurePlugin(charm=self.charm)]
```
Mehdi-Bendriss, line 601: I like this!

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1077,25 +1128,551 @@ def get_service_status(  # noqa: C901
...
+
+        We run the peer cluster orchestrator's refresh on every new azure information.
+        """
+        if self.charm.opensearch_peer_cm.is_provider(typ="main"):
+            self.charm.peer_cluster_provider.refresh_relation_data(event)
+
+    @override
+    def _on_backup_action(self, event: ActionEvent) -> None:
+        """Just overloads the base method, as we process each action in this class."""
+        pass
```
Mehdi-Bendriss, line 1173: I believe we should drop this method given we overload it with nothing, so the base logic will not be invoked

lib/charms/opensearch/v0/opensearch_backups.py
Mehdi-Bendriss: I think that's sensible, the only variable here is the `AZURE_REPOSITORY`

tests/integration/ha/helpers.py
```diff
@@ -508,15 +507,15 @@ async def wait_for_backup_system_to_settle(
...
-async def delete_backup(ops_test: OpsTest, backup_id: int) -> None:
-    """Deletes a backup."""
-    # Now, check if we have finished the restore
-    unit_ip = await get_leader_unit_ip(ops_test)
-    await http_request(
-        ops_test,
-        "DELETE",
-        f"https://{unit_ip}:9200/_snapshot/{S3_REPOSITORY}/{backup_id}",
-    )
+# async def delete_backup(ops_test: OpsTest, backup_id: int, repository: str) -> None:
```
Mehdi-Bendriss, line 510: do we no longer need this?

Mehdi-Bendriss (commented):
Thanks Raul! No major concerns for me, left some comments, and I do agree that a bit of code unification for parts with no divergence between S3 and Azure would be good.

tests/integration/ha/helpers.py
zmraul: It is not used anymore. The backups are deleted directly on the storage backend at the end of the tests

lib/charms/opensearch/v0/opensearch_backups.py
zmraul: It is needed, otherwise the `_on_backup_action` in the Base class is also called in addition to the inherited ones. It leads to wrong result messages on the action

lib/charms/opensearch/v0/models.py
zmraul: It's a safeguard for the credentials, but it's probably safe to drop.

lib/charms/opensearch/v0/models.py
zmraul: Yeah, in this case is completely overkill because there is only one key which is mandatory. I think I had more keys in the past and I kept it like this when those keys were dropped from the mandatory set

Mehdi-Bendriss: Can you simplify this logic?

zmraul: This whole block can be dropped in reality, since we don't know if the model is on a `peer_cluster` relation or getting data from the azure-integrator relation

Mehdi-Bendriss (approved):

phvalguima (approved):

PR #545, opened 2025-02-08
[DPE-6534] Add TF charm and product modules for simple & large deployments

Mehdi-Bendriss:
## Issue
This PR addresses: [DPE-6534](https://warthogs.atlassian.net/browse/DPE-6534), [DPE-6535](https://warthogs.atlassian.net/browse/DPE-6535), [DPE-6536](https://warthogs.atlassian.net/browse/DPE-6536), [DPE-6537](https://warthogs.atlassian.net/browse/DPE-6537). 
Namely, this PR:
- implements the **charm** modules for according to the [TF Spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?tab=t.0): 
  - simple deployments
  - large deployments 
- implements the **product** modules for:
  - simple deployments
  - large deployments 

[DPE-6534]: https://warthogs.atlassian.net/browse/DPE-6534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-6535]: https://warthogs.atlassian.net/browse/DPE-6535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-6536]: https://warthogs.atlassian.net/browse/DPE-6536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-6537]: https://warthogs.atlassian.net/browse/DPE-6537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/ci.yaml
```diff
@@ -105,12 +97,17 @@ jobs:
           juju model-defaults logging-config='<root>=INFO; unit=DEBUG'
           juju add-model test
 
-      - name: Terraform deploy
+      - name: Terraform deploy - simple deployments charm module
```
MiaAltieri: I think its a good idea to also test the product modules. I could be wrong, but when you test the product modules you are implicitly testing the charm modules

[See how we do it in Mongo](https://github.com/canonical/mongodb-operator/blob/1c7a3f4acd7cd2f5c776d10ad63be6cf3c242fcf/.github/workflows/ci.yaml#L64-L111)

terraform/charm/large_deployment/main.tf
```diff
@@ -0,0 +1,185 @@
...
+module "opensearch_main" {
+  source = "../simple_deployment"
+
+  channel  = var.main.channel
+  revision = var.main.revision
+  base     = var.main.base
+
+  app_name          = var.main.app_name
+  units             = var.main.units
+  config            = merge(var.main.config, { "cluster_name" : var.cluster_name, "init_hold" : "false" })
```
MiaAltieri, line 39: Should we let users configure init-hold?

terraform/charm/large_deployment/variables.tf
```diff
@@ -0,0 +1,61 @@
...
+    units             = optional(number, 3)
+    constraints       = optional(string, "arch=amd64")
+    machines          = optional(list(string), [])
+    storage           = optional(map(string), {})
+    endpoint_bindings = optional(map(string), {})
+  })
+  default = null
+}
+
+variable "apps" {
```
MiaAltieri, line 45: nit: this variable would be more clear as "non-orchestors"

terraform/charm/large_deployment/variables.tf
```diff
@@ -0,0 +1,61 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+variable "cluster_name" {
+  type        = string
+  description = "The cluster name of the fleet."
+  default     = "opensearch"
+}
+
+variable "main" {
```
MiaAltieri, line 10: nit: this variable would be more clear as "main-ochestrator"

terraform/product/large_deployment/README.md
```diff
@@ -0,0 +1,47 @@
...
+| `provides` | Map of `provides` endpoints |
+| `requires` | Map of `requires` endpoints |
+
+## Usage
+
+This module is intended to be a product module, deploying all components for a proper and large opensearch deployment.
+
+It may be used as-is and directly as follows:
+```
+tf plan \
```
MiaAltieri, line 42:
```suggestion
tf init

tf plan \
```

terraform/product/large_deployment/main.tf
```diff
@@ -0,0 +1,166 @@
...
+  source       = "../../charm/large_deployment"
+  cluster_name = var.cluster_name
+  main         = var.main
+  failover     = var.failover
+  apps         = var.apps
+}
+
+# opensearch-dashboards in the main model
+module "opensearch-dashboards" {
+  source = "git::https://github.com/canonical/opensearch-dashboards-operator//terraform?ref=tf"
```
MiaAltieri: neat!

terraform/product/large_deployment/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+    channel           = optional(string, "2/stable")
+    base              = optional(string, "ubuntu@22.04")
+    revision          = optional(string, null)
+    units             = optional(number, 3)
+    constraints       = optional(string, "arch=amd64")
+    machines          = optional(list(string), [])
+    storage           = optional(map(string), {})
+    endpoint_bindings = optional(map(string), {})
+  })
+  default = null
```
MiaAltieri, line 42: defaults in the variables you define in your product modules.

So that way someone can "just go" without defining them themselves. It also leads to a friendly UX because they know what it should look like when they generate them themselves

terraform/product/large_deployment/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+  validation {
+    condition     = contains(["s3", "azure"], var.backups-integrator.storage_type)
+    error_message = "backup-integrator allows one of the values: 's3', 'azure' for storage_type."
+  }
+}
+
+variable "data-integrator" {
+  description = "Configuration for the data-integrator"
+  type        = map(string)
+  default     = { "index-name" : "test", "extra-user-roles" : "admin" }
```
MiaAltieri, line 80: oh nice, I like this

terraform/charm/large_deployment/main.tf
Mehdi-Bendriss: No, this is a config option that if not used correctly, will prevent the creation of the cluster - so we manage it ourselves

terraform/product/large_deployment/variables.tf
Mehdi-Bendriss: By defaulting to null, it also means that this variable is optional, and the user doesn't need to explicitly pass it.
Should they however decide to pass it, only 2 fields are mandatory "app_name / model"

terraform/charm/large_deployment/main.tf
```diff
@@ -0,0 +1,185 @@
...
+  apps_in_failover_model = [
+    for app in local.apps :
+    app if app.model == var.failover.model
+  ]
+  apps_not_in_failover_model = [
+    for app in local.apps :
+    app if app.model != var.failover.model
+  ]
+
+  all_models = distinct(concat(
```
phvalguima, line 18: I don't see this as a blocker but we can offer relations and consume on the same model, e.g.:
```
juju add-model test
juju deploy kafka --channel=3/edge --config profile=testing
juju deploy zookeeper --channel=3/edge
```

Then we can offer the ZK relation and consume it locally:
```
juju offer zookeeper:zookeeper zookeeper-client
juju relate kafka admin/test.zookeeper-client
```

Which results in: https://pastebin.ubuntu.com/p/XF62dDq5VW/

This way, all key relations get offered and we consume them on the models we need.

terraform/charm/large_deployment/main.tf
```diff
@@ -0,0 +1,185 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+locals {
+  apps = [
+    for app in concat(var.apps != null ? var.apps : []) : app if app != null
+  ]
+
+  apps_in_main_model = [
```
phvalguima: I guess this is not really used? Can we remove it?

terraform/charm/large_deployment/main.tf
```diff
@@ -0,0 +1,185 @@
...
+    for app in concat([var.failover], local.apps) :
+    app if app != null && app.model == var.main.model
+  ]
+  apps_not_in_main_model = [
```
phvalguima, line 9: Same

terraform/charm/large_deployment/main.tf
```diff
@@ -0,0 +1,185 @@
...
+    for app in concat([var.failover], local.apps) :
+    app if app != null && app.model != var.main.model
+  ]
+  apps_in_failover_model = [
```
phvalguima: Same

terraform/charm/large_deployment/main.tf
```diff
@@ -0,0 +1,185 @@
...
+    for app in local.apps :
+    app if app.model == var.failover.model
+  ]
+  apps_not_in_failover_model = [
```
phvalguima, line 13: Same

terraform/charm/large_deployment/main.tf
phvalguima: Re-reading the code, I believe that is what you've done with `all_models > 1`?

terraform/charm/large_deployment/variables.tf
```diff
@@ -0,0 +1,61 @@
...
+    endpoint_bindings = optional(map(string), {})
+  })
+}
+
+variable "failover" {
+  description = "Failover orchestrator app definition"
+  type = object({
+    app_name          = string
+    model             = string
+    config            = optional(map(string), { "init_hold" : "true" })
```
phvalguima, line 32: Why is `init_hold` set as default here? Shouldn't that managed by the module internally and never exposed? Why not setting a `cluster_name` default as we need it for the large deployments as well?

terraform/charm/simple_deployment/main.tf
```diff
@@ -0,0 +1,89 @@
...
+#--------------------------------------------------------
+# 1. DEPLOYMENTS
+#--------------------------------------------------------
+
+# Deploy required applications
+resource "juju_application" "opensearch" {
+  charm {
+    name     = "opensearch"
+    channel  = var.channel
+    revision = var.revision
```
phvalguima, line 18: `revision == null` works fine?

terraform/product/simple_deployment/main.tf
```diff
@@ -0,0 +1,147 @@
...
+  config            = merge(var.opensearch.config, { "init_hold" : "false" })
+  model             = var.opensearch.model
+  constraints       = var.opensearch.constraints
+  storage           = var.opensearch.storage
+  endpoint_bindings = var.opensearch.endpoint_bindings
+}
+
+# OpenSearch dashboards
+module "opensearch-dashboards" {
+  source = "git::https://github.com/canonical/opensearch-dashboards-operator//terraform?ref=tf"
```
phvalguima: I cannot find the `ref=tf` in this repo. I believe that will then default to the main branch's TF module (`2/edge`), right? In this case, we are leaving this ref as an optional?

terraform/charm/large_deployment/variables.tf
```diff
@@ -0,0 +1,61 @@
...
+  })
+  default = null
+}
+
+variable "apps" {
+  description = "Non orchestrator apps (e.g: ml, data.hot etc.)"
+  type = list(object({
+    app_name          = string
+    model             = string
+    config            = optional(map(string), { "init_hold" : "true" })
```
phvalguima, line 50: Same as for  `failover`

phvalguima (changes requested):

terraform/charm/large_deployment/main.tf
Mehdi-Bendriss: Good catch - removed

terraform/charm/large_deployment/main.tf
Mehdi-Bendriss: Yeah, my reasoning is that I would like to avoid offering opensearch or self-signed-certificates unless explicitly deployed in cross models from the get go. If not, up to the user to manually do it in the future should they want to offer those

terraform/charm/large_deployment/main.tf
Mehdi-Bendriss: same response, see above

terraform/charm/large_deployment/main.tf
Mehdi-Bendriss: same response, see above

terraform/charm/large_deployment/main.tf
Mehdi-Bendriss: same response, see above

terraform/charm/large_deployment/variables.tf
Mehdi-Bendriss: `init_hold` can never be **not** exposed, as it is a config option, the user can just fill all the config options they wish. 
This ensures that we always override the user input configs for crucial bits, such as `init_hold`. As a second safeguard, we do the same in `main.tf`.

For `cluster_name`, it is already set at `opensearch` (top of the variables.tf file). But it remains optional, as a cluster_name is auto-generated by the main orchestrator, and inherited by non-main orchestrators in large deployments. 
Similarly, if set on the config options, it will be overriden by the content of the `cluster_name` top level variable.

terraform/charm/simple_deployment/main.tf
Mehdi-Bendriss: Yes. When specifying the `channel`, the revision is optional.

terraform/product/simple_deployment/main.tf
Mehdi-Bendriss: The `ref=tf` here is a reference to the remote git branch of the dashboards TF PR I made a few days back. This now should be changed to `ref=2/edge`.

terraform/charm/large_deployment/variables.tf
Mehdi-Bendriss: I was struggling with this one, and I agree it's not ideal. The problem with adding "orchestrator" for the user is that they wouldn't know what it means.. And I don't have a better name atm tbh

terraform/product/large_deployment/variables.tf
MiaAltieri: Ah TIL, so I assume it isn't possible to specify that the variable is optional AND provide an example?

terraform/charm/large_deployment/variables.tf
MiaAltieri: Fair, @phvalguima any idea for a name here?

MiaAltieri: I could be wrong but couldn't we call it `main cluster manager`?

Mehdi-Bendriss: `cluster_manager` has a meaning in opensearch, it is a role, but it commonly refers to the "elected primary"

terraform/product/large_deployment/variables.tf
Mehdi-Bendriss: Not as far as I know, the way I know of is through "description" which is already done in the `charm/simple_deployment/` or for the non-opensearch related variables

terraform/charm/large_deployment/variables.tf
phvalguima: Ah, I see the `merge` there. Okay.

terraform/charm/large_deployment/variables.tf
phvalguima: @MiaAltieri don't look at me, I am terrible choosing names, lol.

Maybe calling it "main_orchestrator", which is a good ref. to our own docs.

terraform/product/large_deployment/variables.tf
MiaAltieri: Thanks for the context!

phvalguima (dismissed):

MiaAltieri (dismissed):
Maybe in the future if another naming scheme comes to mind it can be applied! Great work!

MiaAltieri (approved):

phvalguima (approved):

PR #546, opened 2025-02-11
update tls certificates lib to version 3 patch level 24

reneradoi:
## Issue
When a TLS related secret expires, the TLS lib might fail when trying to access this secret, as it can no longer be found.

## Solution
In version 3.24, error handling for this (and other cases) has been added.


Mehdi-Bendriss (approved):
Thanks René!

Issue #547, opened 2025-02-11
Opensearch units are deploying very slowly, takes more than 30 minutes for a 3 units cluster


dparv:
When deploying (on Azure) a 3 units opensearch cluster (+ 3 units dashboards), the opensearch units take a very long time to go to active/idle, usually 30-40 minutes, with being stuck mostly on `Waiting for TLS to be fully configured...`, waiting to get lock or restarting services, which causes the opensearch-dashboards to be waiting for a long time, then they take additional time to go to active/idle, the whole deployment takes around 60 minutes (including the Azure resources, juju bootstrap, and additional tasks like certificate generation, tf pipeline, that combined take about 22 min), which for just 2 applications is a very long time.

I see a lot of bouncing in the logs of starting, stopping and restarting the opensearch services - maybe this workflow can be optimized?

Deployment logs: https://pastebin.canonical.com/p/ykDm6Hj8ch/

Starting `unit-opensearch-1: 11:49:49 INFO juju Starting unit workers for "opensearch/1"`. First cluster green is seen at `unit-opensearch-1: 12:12:28 INFO unit.opensearch/1.juju-log Current health of cluster: green` which is already 22 minutes in, then some other restarts take place until it settles around 12:23, 33 minutes after the deployment started. Then a couple of more minutes for opensearch-dashboards to settle.

Environment:
```
opensearch  azure-westeurope  azure/westeurope  3.6.2    unsupported  13:16:44Z

App                       Version  Status  Scale  Charm                     Channel        Rev  Exposed  Message
opensearch                         active      3  opensearch                2/edge         195  no       
opensearch-dashboards              active      3  opensearch-dashboards     2/edge          26  no       
self-signed-certificates           active      1  self-signed-certificates  latest/stable  155  no       
sysconfig                          active      3  sysconfig                 latest/stable   33  no       ready
```

phvalguima:
I can see the following exception showing up twice early on your logs and then it seems it auto resolves:

```
unit-opensearch-1: 11:50:50 ERROR unit.opensearch/1.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-opensearch-1/charm/./src/charm.py", line 221, in <module>
    main(OpenSearchOperatorCharm)
  File "/var/lib/juju/agents/unit-opensearch-1/charm/venv/ops/main.py", line 553, in main
    manager.run()
[...]

PR #548, opened 2025-02-11
[DPE-6603] Add extra checks for `DeploymentDescription` existence

phvalguima:
Currently, if the cluster is not yet ready and the description not yet available, set password action throws an exception. This PR relates to: #547. 

It adds checks for the deployment description on actions and some key-points in the base charm. This PR reviews the base charm calls to the `.deployment_desc()` function and ensures we have appropriate checks each time.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_backups.py
+++ b/lib/charms/opensearch/v0/opensearch_backups.py
@@ -536,7 +536,7 @@ class OpenSearchBackup(OpenSearchBackupBase):
 
     def _on_list_backups_action(self, event: ActionEvent) -> None:
         """Returns the list of available backups to the user."""
-        if not (deployment_desc := self.charm.opensearch_peer_cm.deployment_desc()):
+        if not self.charm.opensearch_peer_cm.deployment_desc():
             event.fail("The action can be run only after the deployment is finished.")
             return
         backups = {}
@@ -684,7 +684,7 @@ class OpenSearchBackup(OpenSearchBackupBase):
 
     def _on_restore_backup_action(self, event: ActionEvent) -> None:  # noqa #C901
         """Restores a backup to the current cluster."""
-        if not (deployment_desc := self.charm.opensearch_peer_cm.deployment_desc()):
+        if not self.charm.opensearch_peer_cm.deployment_desc():
             event.fail("The action can be run only after the deployment is finished.")
             return
         if self.charm.upgrade_in_progress:
@@ -754,7 +754,7 @@ class OpenSearchBackup(OpenSearchBackupBase):
 
     def _on_create_backup_action(self, event: ActionEvent) -> None:  # noqa: C901
         """Creates a backup from the current cluster."""
-        if not (deployment_desc := self.charm.opensearch_peer_cm.deployment_desc()):
+        if not self.charm.opensearch_peer_cm.deployment_desc():
             event.fail("The action can be run only after the deployment is finished.")
             return
         if self.charm.upgrade_in_progress:
--- a/lib/charms/opensearch/v0/opensearch_backups.py
+++ b/lib/charms/opensearch/v0/opensearch_backups.py
@@ -685,7 +685,7 @@ class OpenSearchBackup(OpenSearchBackupBase):
     def _on_restore_backup_action(self, event: ActionEvent) -> None:  # noqa #C901
         """Restores a backup to the current cluster."""
         if not self.charm.opensearch_peer_cm.deployment_desc():
-            event.fail("The action can be run only after the deployment is finished.")
+            event.fail("The action can only be run once the deployment is complete.")
             return
         if self.charm.upgrade_in_progress:
             event.fail("Restore not supported while upgrade in-progress")
@@ -755,7 +755,7 @@ class OpenSearchBackup(OpenSearchBackupBase):
     def _on_create_backup_action(self, event: ActionEvent) -> None:  # noqa: C901
         """Creates a backup from the current cluster."""
         if not self.charm.opensearch_peer_cm.deployment_desc():
-            event.fail("The action can be run only after the deployment is finished.")
+            event.fail("The action can only be run once the deployment is complete.")
             return
         if self.charm.upgrade_in_progress:
             event.fail("Backup not supported while upgrade in-progress")
--- a/lib/charms/opensearch/v0/opensearch_backups.py
+++ b/lib/charms/opensearch/v0/opensearch_backups.py
@@ -537,7 +537,7 @@ class OpenSearchBackup(OpenSearchBackupBase):
     def _on_list_backups_action(self, event: ActionEvent) -> None:
         """Returns the list of available backups to the user."""
         if not self.charm.opensearch_peer_cm.deployment_desc():
-            event.fail("The action can be run only after the deployment is finished.")
+            event.fail("The action can only be run once the deployment is complete.")
             return
         backups = {}
         try:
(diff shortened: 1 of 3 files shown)


lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -536,6 +536,9 @@ def _generate_backup_list_output(self, backups: Dict[str, Any]) -> str:
 
     def _on_list_backups_action(self, event: ActionEvent) -> None:
         """Returns the list of available backups to the user."""
+        if not self.charm.opensearch_peer_cm.deployment_desc():
+            event.fail("The action can be run only after the deployment is finished.")
```
Mehdi-Bendriss: nit: phrasing on all occurrences
```suggestion
            event.fail("The action can only be run once the deployment is complete.")
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -396,7 +396,9 @@ def cleanup():
         # configure clients auth
         self.opensearch_config.set_client_auth()
 
-        deployment_desc = self.opensearch_peer_cm.deployment_desc()
+        if not (deployment_desc := self.opensearch_peer_cm.deployment_desc()):
```
Mehdi-Bendriss: i think this one's not necessary, as it is already checked as part of `_apply_peer_cm_directives_and_check_if_can_start`

Mehdi-Bendriss (commented):
Thanks Pedro, looks good. Just minor comments on the wording

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Indeed. I removed this check

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -760,7 +760,10 @@ def _on_set_password_action(self, event: ActionEvent):
         if self.upgrade_in_progress:
             event.fail("Setting password not supported while upgrade in-progress")
             return
-        if self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR:
+        if (
+            not self.opensearch_peer_cm.deployment_desc()
+            or self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR
+        ):
             event.fail("The action can be run only on the leader unit of the main cluster.")
```
smethnani: nit: make phrasing consistent with the other messages
```suggestion
            event.fail("The action can only be run on the leader unit of the main cluster.")
```

smethnani (commented):
Just added a minor follow up to Mehdi's suggestion for one more phrase

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Ah! Good catch! I copied the same message

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -760,8 +760,11 @@ def _on_set_password_action(self, event: ActionEvent):
         if self.upgrade_in_progress:
             event.fail("Setting password not supported while upgrade in-progress")
             return
-        if self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR:
-            event.fail("The action can be run only on the leader unit of the main cluster.")
+        if (
+            not self.opensearch_peer_cm.deployment_desc()
+            or self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR
+        ):
+            event.fail("The action can only be run once the deployment is complete.")
```
smethnani, line 761: Shouldn't the message be about the leader unit? `The action can only be run on the leader unit of the main cluster.`? 😅

smethnani (commented):
I think your original message was appropriate! Was just a comment on the ordering of the words to match the others. Looks good otherwise!

lib/charms/opensearch/v0/opensearch_base_charm.py
smethnani: Or do I misunderstand and the action doesn't need to be on the main cluster leader unit?

phvalguima: soo, we are not checking the leader in that `if`... If we check it, then yes we should have. I will add an extra check

phvalguima: Now yes, we are giving meaningful messages on each step of the way.

smethnani: Ah yea I see, thanks!

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -760,10 +760,12 @@ def _on_set_password_action(self, event: ActionEvent):
...
             event.fail("Setting password not supported while upgrade in-progress")
             return
-        if self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR:
-            event.fail("The action can be run only on the leader unit of the main cluster.")
+        if not self.opensearch_peer_cm.deployment_desc():
+            event.fail("The action can only be run once the deployment is complete.")
             return
-
+        if self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR:
+            event.fail("The action can be run only on the main cluster.")
```
smethnani: Sorry I missed these before!
```suggestion
            event.fail("The action can only be run on the main cluster.")
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -760,10 +760,12 @@ def _on_set_password_action(self, event: ActionEvent):
...
+            return        
         if not self.unit.is_leader():
             event.fail("The action can be run only on leader unit.")
```
smethnani: Same here 
```suggestion
            event.fail("The action can only be run on leader unit.")
```

smethnani (dismissed):
Sorry I missed a few comments with the same word ordering nitpick. Not a PR blocking issue imo so approved. 👍

smethnani (dismissed):
Thanks Pedro!

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -760,12 +760,14 @@ def _on_set_password_action(self, event: ActionEvent):
         if self.upgrade_in_progress:
             event.fail("Setting password not supported while upgrade in-progress")
             return
```
Mehdi-Bendriss: Could you move this after the deployment_desc check?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -760,12 +760,14 @@ def _on_set_password_action(self, event: ActionEvent):
...
-        if self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR:
-            event.fail("The action can be run only on the leader unit of the main cluster.")
+        if not self.opensearch_peer_cm.deployment_desc():
+            event.fail("The action can only be run once the deployment is complete.")
             return
-
+        if self.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR:
+            event.fail("The action can only be run on the main cluster.")
```
```suggestion
            event.fail("The action can only be run on the main orchestrator cluster.")
```

Mehdi-Bendriss (commented):
Thanks Pedro! just a comment on the ordering of the check compared to upgrades

Mehdi-Bendriss (approved):

smethnani (approved):

PR #551, opened 2025-02-18
Replace TLS lib V3 with TLS lib V4

saltiyazan:
This PR replaces TLS lib V3 with V4.

Main changes:
- Three cert types will have separate instances of the interface.
- The certificate signing requests and private keys are managed by the library.
- The set-private-key action is replaced with a regenrate-private-key action which will allow the library to regenerate the private key. Setting the private key should be done using juju config if needed, for simplicity I kept out of the scope of this PR but I'm happy to add it if needed.
- the `CertificateExpiringEvent` and the `CertificateInvalidatedEvent` no longer exist in the library, therefore their handlers are removed, changes around certificates for example a cert being changed because of expiry or revocation will be handled as a cert renewal in the context of the `CertificateAvailableEvent` event.


saltiyazan:
Here are a few notes on why this PR was abandoned:

- Breaking Change
This PR introduces a breaking change by moving from a single interface to three separate ones for managing certificates in the charm, one each for the API, internal communication, and admin certificates.

- Private Key Handling Incompatibility
The current charm allows setting a private key manually. In TLS library  V4, however, each unit and instance generates and uses its own private key (so 3 units would have 9 private keys), making the current action obsolete. This fundamental change in how private keys are handled requires a new design and should be led by the maintainers of the charm. Moreover, sensitive data such as private keys should be stored using Juju user secrets, which the current implementation does not support.

- State Modeling Limitations
The current model for tracking charm state is already complex yet insufficient for handling the added complexity of multiple TLS interfaces and multiple units. If each interface is integrated with a different provider, there could be three separate Certificate Authorities (CAs) in use. This makes the current flags—ca_renewing, ca_renewed, and ca_configured—inadequate. We'd need to maintain these flags separately for each certificate type. Additionally, managing multiple CA rotations across multiple units introduces high synchronization complexity, especially when coordinating restarts. A full assessment of the current state management and its limitations in OpenSearch is necessary before transitioning from TLS v3 to v4.

Issue #549, opened 2025-02-13
App not inheriting cluster name in large deployments when `--config cluster_name` not specified


smethnani:
## Steps to reproduce

1. Deploy main application
 ```
juju deploy opensearch main \
    --channel 2/edge \
    --config profile="testing"
```

2. Deploy second application
```
juju deploy opensearch failover \
    --config init_hold="true" \
    --config roles="cluster_manager"  \
    --channel 2/edge \
    --config profile="testing"
```
3. Deploy TLS and integrate
```
juju deploy self-signed-certificates --config ca-common-name="Test CA"
juju integrate self-signed-certificates main
juju integrate self-signed-certificates failover
juju integrate main:peer-cluster-orchestrator failover:peer-cluster
```
## Expected behavior
The failover app should inherit the cluster name from the main app

## Actual behavior

Failover app doesn't inherit cluster name and can't start
```
App                       Version  Status   Scale  Charm                     Channel        Rev  Exposed  Message
failover                           blocked      1  opensearch                2/edge         204  no       Cannot relate 2 clusters with different 'cluster_name' values.
```

## Versions

Operating system: `Ubuntu 24.04`

Juju CLI:  `3.5.6-genericlinux-amd64`

Juju agent: `3.5.5`

Charm revision: `204`

LXD: `5.21.2 LTS`

smethnani:
Closed by #552

PR #552, opened 2025-02-19
[DPE-6582]  App not inheriting cluster name in large deployments

smethnani:
## Issue #549

When relating two apps in a large deployment, there is a check which blocks the relation if the cluster names don't match despite inherit-cluster-name being in the pending directives.

## Solution
Add a check for inherit-cluster-name in the pending directives before blocking the relation. Additionally, store whether or not the cluster name was auto generated and only inherit the cluster name if it was auto generated.


lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -134,6 +134,8 @@ def run_with_relation_data(self, data: PeerClusterRelData) -> None:  # noqa: C90
         elif Directive.INHERIT_CLUSTER_NAME in pending_directives:
             config.cluster_name = data.cluster_name
             pending_directives.remove(Directive.INHERIT_CLUSTER_NAME)
+            if deployment_state.value == State.BLOCKED_WAITING_FOR_RELATION:
```
smethnani, line 137: Would appreciate a double check on setting the state to active here. I copied the pattern in the other "if" condition above this line but these conditions are probably why the state isn't being reset to active in #557

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -134,6 +134,8 @@ def run_with_relation_data(self, data: PeerClusterRelData) -> None:  # noqa: C90
...
+                deployment_state = DeploymentState(value=State.ACTIVE)
```
phvalguima, line 138: As I understood, that is just cleaning up a "left-over" of relation was missing, as we are effectively running within a relation hook, right?

tests/integration/ha/test_large_deployments_inherit_cluster_name.py
```diff
@@ -0,0 +1,267 @@
...
+            INVALID_FAILOVER_APP: {"units": {"active": []}},
+        },
+        wait_for_exact_units={app: units for app, units in NON_AUTOGEN_APP_UNITS.items()},
+        idle_period=IDLE_PERIOD,
+        timeout=1800,
+    )
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "xlarge"])
+@pytest.mark.group(2)
```
phvalguima: **nit** I would recommend a more meaningful name if you have two runners for the same test. It can be a string. This name shows up in the CI status, in GH as the test name.

tests/integration/ha/test_large_deployments_inherit_cluster_name.py
```diff
@@ -0,0 +1,267 @@
...
+
+AUTOGEN_APP_UNITS = {
+    MAIN_APP_AUTOGEN: 1,
+    FAILOVER_APP_AUTOGEN: 1,
+    DATA_APP_AUTOGEN: 2,
+}
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "xlarge"])
+@pytest.mark.group(1)
```
phvalguima: **nit** I would recommend a more meaningful name if you have two runners for the same test. It can be a string. This name shows up in the CI status, in GH as the test name.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1072,6 +1083,7 @@ def _error_set_from_requirer(
                 should_wait=False,
                 blocked_message=blocked_msg,
                 deployment_desc=deployment_desc,
+                cluster_name_auto_generated=peer_cluster_rel_data.cluster_name_auto_generated,
```
phvalguima: You've added that field to `PeerClusterRelData`, not `PeerClusterRelErrorData`

tests/integration/ha/test_large_deployments_inherit_cluster_name.py
```diff
@@ -0,0 +1,267 @@
...
+            my_charm,
+            application_name=INVALID_FAILOVER_APP,
+            num_units=1,
+            series=SERIES,
+            config={"init_hold": True, "roles": "cluster_manager"} | CONFIG_OPTS,
+        ),
+    )
+
+    # wait until the TLS operator is ready
+    await wait_until(
```
phvalguima, line 78: I loved the precise settings here! Congrats

lib/charms/opensearch/v0/models.py
```diff
@@ -327,6 +327,7 @@ class PeerClusterRelData(Model):
     credentials: PeerClusterRelDataCredentials
     deployment_desc: Optional[DeploymentDescription]
     security_index_initialised: bool = False
+    cluster_name_auto_generated: bool = False
```
phvalguima: It is not very clear for me when is this field set on the relation databag

phvalguima (changes requested):
Nice work! I really liked the tests and the precision with `wait_until`

I think we need to be more clear when `cluster_name_auto_generated` is set on the relation databag.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: Ouf yes of course, thank you! Removed 👍

tests/integration/ha/test_large_deployments_inherit_cluster_name.py
smethnani: Didn't realize you could use strings, thanks!

lib/charms/opensearch/v0/opensearch_peer_clusters.py
smethnani: Yes, though I'm not sure if I should be checking for other statuses as we do just above [here](https://github.com/canonical/opensearch-operator/blob/12b29e634d7204fe7412de842c043094681c03bf/lib/charms/opensearch/v0/opensearch_peer_clusters.py#L127)

lib/charms/opensearch/v0/models.py
smethnani: It's set to true in `_new_cluster_setup` if the cluster name is autogenerated and not read from the config ([here](https://github.com/canonical/opensearch-operator/blob/12b29e634d7204fe7412de842c043094681c03bf/lib/charms/opensearch/v0/opensearch_peer_clusters.py#L210))

phvalguima: Ah true! Thanks, did not see that specific line

phvalguima (dismissed):
LGTM

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -203,10 +205,9 @@ def _new_cluster_setup(self, config: PeerClusterConfig) -> DeploymentDescription
                 state=deployment_state,
             )
 
-        cluster_name = (
-            config.cluster_name.strip()
-            or f"{self._charm.app.name}-{ShortUUID().random(length=4)}".lower()
-        )
+        if not (cluster_name := config.cluster_name.strip()):
+            cluster_name = f"{self._charm.app.name}-{ShortUUID().random(length=4)}".lower()
+            self._charm.peers_data.put(Scope.APP, "cluster_name_auto_generated", True)
```
Mehdi-Bendriss: Can this be put in the `deployment_description` instead?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1058,7 +1062,14 @@ def _error_set_from_requirer(
...
             )
         elif peer_cluster_rel_data.cluster_name != deployment_desc.config.cluster_name:
-            blocked_msg = "Cannot relate 2 clusters with different 'cluster_name' values."
+            contains_inherit_directive = (
+                Directive.INHERIT_CLUSTER_NAME in deployment_desc.pending_directives
+            )
+            if not contains_inherit_directive or (
+                contains_inherit_directive
+                and not peer_cluster_rel_data.cluster_name_auto_generated
+            ):
```
Mehdi-Bendriss: very nice

Mehdi-Bendriss (commented):
Thank you Shams! great work! I left 1 comment and we should be good to go!

Mehdi-Bendriss (approved):
Thank you Shams!

phvalguima (approved):

PR #556, opened 2025-02-24
[DPE-6601] Move CI to juju 3.6.3 and fix errors

phvalguima:
There were several recurrent errors happening in our CI:
* Upgrade tests failing
* test_storage timing out
* CA rotation failing on large deployments
* HA network failing due to lxc command
* intermittent issue in Juju with `ERROR Permission Denied`

The tests have been fixed as follows:
* Upgrade tests: we are now using snap rev65, which contains wrapper scripts that bridge between old and new snap formats for 24.04 - rollback testing is fixed
* test_storage: it was missing a `profile=testing` at redeployment
* CA rotation: explained in more details below
* HA network failures: work-around in place: use the `jammy` default runner, which comes with a more recent kernel and enables LXC throttling to work. The reason why `jammy` large does not have this kernel is still under investigation by our teams
* Moved to Juju 3.6.3+, which has the latest fixes

Besides, this PR does two more changes:
* Adds CLA checker to our CI, at pull requests
* Starts to move away from `pytest-operator-cache` as it has been deprecated: now integration tests will expect the charm to be built and left on the root of the charm folder - [more details, check this other PR](https://github.com/canonical/mysql-router-k8s-operator/blob/fff0127ab49cbc061a404e99a9923f3d71bab7bb/tests/integration/conftest.py#L16)

------------------------------------------------------------------------------------------------

## CA Rotation Problem

The main issue with CA rotation test (and could be also verified locally) is the race condition between a deferred `CertificateAvailableEvent` and `SecretChangedEvent` happening within the `secret-changed` hook.

If the cert event comes first, then we will reconfigure the entire cluster with the old data in the secret (as the new data, that actually triggered the `secret-changed` in the first place will only be available after `secret.get_content(refresh=True)`).

Now, if the `SecretChangedEvent` happens before, then the refresh is correctly called and the following cert available event runs correctly.

To resolve this race condition, in the `opensearch_tls.py` module, we will only recover secrets using `peek=True`: that always gives the latest secret data, even within the `secret-changed` hook.


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -872,7 +872,17 @@ def on_tls_conf_set(
 
         # In case of renewal of the unit transport layer cert - restart opensearch
         if renewal and self.is_admin_user_configured():
+            if self.peers_data.get(Scope.UNIT, "tls_ca_renewed", False):
```
phvalguima, line 875: Only way I could find to pass this [line](https://github.com/canonical/opensearch-operator/blob/1de95171ff6688d3a290176f33d84d20ab6cdda3/lib/charms/opensearch/v0/opensearch_tls.py#L656) check

.github/workflows/ci.yaml
```diff
@@ -175,11 +175,11 @@ jobs:
     name: Integration test charm
     needs:
       - lint
-      - unit-test
+#      - unit-test
```
Mehdi-Bendriss: This should be reverted

.github/workflows/cla_checker.yaml
```diff
@@ -0,0 +1,13 @@
+
```
Mehdi-Bendriss, line 1: I would rather this to be added to the DPW instead

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -823,7 +823,7 @@ def _on_get_password_action(self, event: ActionEvent):
 
         password = self.secrets.get(Scope.APP, self.secrets.password_key(user_name))
         cert = self.secrets.get_object(
-            Scope.APP, CertType.APP_ADMIN.val
+            Scope.APP, CertType.APP_ADMIN.val, peek=True
```
Mehdi-Bendriss, line 826: I understand why we're doing this - but I'm a bit worried about the full synchronisation issues this may entail cross units.
What broader impacts are we expecting to have here?

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -64,6 +64,11 @@
 # to 0 if you are raising the major API version
 LIBPATCH = 1
 
+
+CA_ALIAS = "ca"
+OLD_ALIAS = f"old-{CA_ALIAS}"
```
Mehdi-Bendriss: For consistency:
```suggestion
OLD_CA_ALIAS = f"old-{CA_ALIAS}"
```

tests/integration/ha/conftest.py
```diff
@@ -14,6 +14,14 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture
```
Mehdi-Bendriss, line 17: This does not need to be a fixture, it will limit us in the usage when we support dual base

Mehdi-Bendriss (commented):
Thanks Pedro! I mainly have a comment on the `peek` part and the potential effects this will have on the entire workflow of secrets synchronisation

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: @Mehdi-Bendriss `peek=True` only impacts if you are running within the `secret-changed` hook

.github/workflows/cla_checker.yaml
phvalguima: I am discussing that with @carlcsaposs-canonical

lib/charms/opensearch/v0/opensearch_base_charm.py
phvalguima: Also, on our secret changed routine, [we immediately run the `get_content(refresh=True)`](https://github.com/canonical/opensearch-operator/blob/1de95171ff6688d3a290176f33d84d20ab6cdda3/lib/charms/opensearch/v0/opensearch_secrets.py#L71). That means: only a very brief set of events ends up running `get_content`, within `secret-changed` and getting the old value: the events that ran before `SecretChangedEvent` took place. We do not design these events to be dependent of that narrow window...

In summary, if we break sth, it was likely a race condition anyways and we should address that

.github/workflows/cla_checker.yaml
carlcsaposs-canonical: given that it's already in a reusable component, I think it'd be simpler to use as-is

wrapping an already reusable component in a workflow comes with a number issues. for example: https://github.com/canonical/data-platform-workflows/pull/280#pullrequestreview-2617602292

also long term, it sounds like this will be managed centrally for canonical

> On GitHub the CI action could be centrally enabled for all the repositories via [organization-wide required workflows](https://github.blog/changelog/2023-01-10-github-actions-support-for-organization-wide-required-workflows-public-beta/). This should be performed by the organisation administrator

https://docs.google.com/document/d/10jnAq1LODM141eV6s4n0UP6XYhDM4PCZF_wKX2hys-A/edit?tab=t.0

tests/integration/ha/conftest.py
phvalguima: So, I added as a fixture exactly thinking of dual base: a single function that will decide prior to any `build_and_deploy...`  calls, which charm will we use.

.github/workflows/cla_checker.yaml
Mehdi-Bendriss: I would argue - that for all DP repos to have the same set of checks that this should be in DPW regardless. Similar to the `lint` step

phvalguima: @Mehdi-Bendriss the thing is: lint, unit, integration, etc; they are all executed both during PR and nightly runs. It makes sense to bundle all them together. CLA check is executed only on pull requests. Even if we put it on DPW, we would need a different hook, hence a different call to DPW, which would be a call to CLA.

carlcsaposs-canonical: > I would argue - that for all DP repos to have the same set of checks that this should be in DPW regardless. Similar to the `lint` step

I think it's a good idea to standardize the workflow across all data repos

However, I disagree that the way to standardize the workflow should be putting it in data-platform-workflows, for the reasons I linked to around issues with wrapping an already reusable component

lib/charms/opensearch/v0/opensearch_secrets.py
```diff
@@ -210,7 +212,10 @@ def _get_juju_secret_content(self, scope: Scope, key: str) -> Optional[Dict[str,
         if not secret:
             return None
 
-        content = secret.get_content()
+        if peek:
+            content = secret.peek_content()
+        else:
+            content = secret.get_content()
         self.cached_secrets.put_content(scope, self.label(scope, key), content=content)
```
reneradoi, line 219: Please excuse my question, as I don't have all the history on the topic. But why not always calling `secret.get_content(refresh=True)` in the `_get_juju_secret_content()` method?

This would be simpler and more deterministic, from my point of view. The name of the method also suggests that the latest content of a secret is looked up. My expectation as developer would be to get the up-to-date data from a secret when I call `get_secret()` or `secret.get()`.

Is there a reason why this cannot be done in Opensearch?

reneradoi (commented):
Hey @phvalguima, I was asked to review from my context of implementing the TLS cert/CA rotation.

Regarding the workflow: It was then implemented under the impression that calls to `get_secret` would always return the latest data from a Juju secret. So I don't think there is a negative impact on the workflow itself. Rather the opposite: This would explain some issues we had over time, that I could never really understand/reproduce. So thank you for bringing this to attention!

What I would suggest though is to avoid adding a parameter `peek` but instead always get the latest content from a secret. This would be cleaner and more predictable, for my understanding.

lib/charms/opensearch/v0/opensearch_secrets.py
phvalguima: So, `peek` gets the latest value at that single call, `refresh` updates it for good. There is no real difference, in practice, for what we have. There are reasons for this breakdown but they do not really matter for OpenSearch as it is right now.

In my view, we can do either. I chose the peek just because it is less invasive, i.e. someone in the future *may* choose how and when to do refresh at their own will.

### When get/peek/refresh matter

The moment these commands should be used carefully is at, e.g. root password change. Imagine you have the following:
* Hook 1: your charm decides to change the password of your database: it changes the database and saves it to a secret
* Hook 1 fails (hence any changes to databag or relation were abandoned)
* Hook 2: your charm reruns the same event as the previous hook - it tries to open the database, but is only aware of the *old password* as the *new value* has been flushed out by the error above

You bricked your access now.

#### How to do it correctly

This is a 2-phased commit. We must first save the password to the secret, then apply:

* Hook 1: generate the new password ~~, emit a custom event~~. That event defers until the hook ID changes
* Hook 1 fails - that means we just abandon a new password that was not used for anything
* Hook 1 reruns: successfully generated another password
* Hook 2: now you have the ~~event~~ *secret-changed* above picks up again: we can see the old and new passwords with `get/peek/refresh`. We change the password of the actual database
* Hook 2: fails - we will rerun it and we still have the old/new values available

#### Conclusions

In theory, you do not even need the custom event - you can use the `secret-changed`. That problem is not limited to secrets, to be honest. But it is limited only for passwords - where you may need to fallback to the old value and you cannot generate that the same again.

phvalguima: I think, then, we should just switch the `peek` to default to `True` instead of False. It will give what you are expecting @reneradoi

reneradoi (commented):
Hey @phvalguima ,

thank you for your response! I agree with your conclusion to switch the default value of `peek` to `True`. If it wasn't for the specifics of the Opensearch charm, I'd say it would be best to get rid of it completely and always refresh the content, but I understand that it might be required esp. for password updates. 

One remark here: when updating from `get/set-password` to user secrets for passwords, it would be a good time to reconsider secret handling and avoid the caching layer, design-wise. This should have advantages regarding:
- more robustness
- less complexity
- clear responsibility (ops also has secret caching)

As for the CA rotation: I don't see a problem with this change. The workflow was always thought to be relying on up-to-date secrets anyway. I went through the implementation again, looked at the integration tests and ran it manually (with small deplyoment cluster), checking the cert files etc.

reneradoi (dismissed):

reneradoi (dismissed):

skourta (approved):
Thank you, @phvalguima. LGTM.

Mehdi-Bendriss (approved):
Thanks Pedro, and good catch! 
Hopefully the implementation for TLS v4 should reduce a bit the need for complex CA / Certs wrangling.
The rest for `peek=True` we should investigate on a secret by secret basis

PR #563, opened 2025-02-25
[DPE-6598] Move to v2.18

phvalguima:
Bump the charm to use opensearch 2.18.

Adds the new `keypassword` field for keystore passwords, [following upstream discussion and change from 2.17.x to 2.18](https://forum.opensearch.org/t/badpaddingexception-with-self-singed-certificates/22711). As can be seen [in our original CI run, all tests fail before that option was added](https://github.com/canonical/opensearch-operator/actions/runs/13523644463/job/37789565066?pr=563). Running on local test, we can see it fails with: https://pastebin.ubuntu.com/p/gQrzXXCqqF/

Test [on provider relation](https://github.com/canonical/opensearch-operator/actions/runs/13724052627/job/38468085969?pr=563) is expected to fail as we do not have an opensearch-dashboards for v2.18 yet.


lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -125,6 +125,12 @@ def set_node_tls_conf(self, cert_type: CertType, truststore_pwd: str, keystore_p
                 f"plugins.security.ssl.{target_conf_layer}.{store_type}_password",
                 pwd,
             )
+            if store_type == "keystore":
```
Mehdi-Bendriss: better move this one out of the `for` loop and directly target the corresponding key

Mehdi-Bendriss (commented):
Thanks Pedro, I have a comment

smethnani (approved):

smethnani (commented):
Thanks Pedro! Do we know the reason for the `Incompatible Opensearch and Dashboards versions` error in the `test_opensearch_provider` check?

Mehdi-Bendriss (approved):

PR #580, opened 2025-03-06
[DPE-6772] Add http.publish_host

### LINKED ISSUE #579: `opensearchpy` receives internal IPs after sniffing
When connecting from a remote endpoint to an azure deployment, I can see my client gets the internal IPs instead of public ones. Adding a debugger on the `opensearchpy`, I can see:
```
408                 will be extracted from `params`                                                                                                      
409             """                                                                                                                                      
410  ->         method, params, body, ignore, timeout = self._resolve_request_args(                                                                      
411                 method, params, body, ignore, timeout                                                                                                
412             )                                                                                                                                        
413                                                                                                                                                      
414             for attempt in range(self.max_retries + 1):                                                                                              
415                 connection = self.get_connection()                                                                                                   
(Pdb) l                                                                                                                                                  
416                                                                                                                                                      
417                 try:                                                                                                                                 
418                     status, headers_response, data = connection.perform_request(                                                                     
419                         method,                                                                                                                      
420                         url,                                                                                                                         
421                         params,                                                                                                                      
422                         body,                                                                                                                        
423                         headers=headers,                                                                                                             
424                         ignore=ignore,                                                                                                               
425                         timeout=timeout,                                                                                                             
426                     )                                                                                                                                
(Pdb) n                                                                                                                                                  
> /home/pguimaraes/Documents/Canonical/Engineering/DPE-6765-restart-node-not-up/opensearch-operator/.tox/integration/lib/python3.12/site-packages/opensea
rchpy/transport.py(414)perform_request()                                                                                                                 
-> for attempt in range(self.max_retries + 1):                                                                                                           
(Pdb)                                                                                                                                                    
> /home/pguimaraes/Documents/Canonical/Engineering/DPE-6765-restart-node-not-up/opensearch-operator/.tox/integration/lib/python3.12/site-packages/opensea
rchpy/transport.py(415)perform_request()                                                                                                                 
-> connection = self.get_connection()                                                                                                                    
(Pdb)                                                                                                                                                    
> /home/pguimaraes/Documents/Canonical/Engineering/DPE-6765-restart-node-not-up/opensearch-operator/.tox/integration/lib/python3.12/site-packages/opensea
rchpy/transport.py(417)perform_request()                                                                                                                 
-> try:                                                                                                                                                  
(Pdb) p connection                                                                                                                                       
<Urllib3HttpConnection: https://192.168.0.7:9200>
```

We are missing the `http.publish_host`, [as described here](https://www.elastic.co/blog/elasticsearch-sniffing-best-practices-what-when-why-how).

phvalguima:
Adds the http.publish_host so we can separate the IPs that are shared when sniffed.

Closes #579

### DIFF
--- a/lib/charms/opensearch/v0/helper_networking.py
+++ b/lib/charms/opensearch/v0/helper_networking.py
@@ -31,12 +31,6 @@ def get_host_ip(charm: CharmBase, peer_relation_name: str) -> str:
     return str(address)
 
 
-def get_host_public_ip(charm: CharmBase, peer_relation_name: str) -> str:
-    """Fetches the IP address of the current unit."""
-    address = charm.model.get_binding(peer_relation_name).network.ingress_address
-    return str(address)
-
-
 def get_host_public_ip() -> Optional[str]:
     """Fetches the Public IP address of the current unit."""
     cmd = "dig +short myip.opendns.com @resolver1.opendns.com"
--- a/lib/charms/opensearch/v0/helper_networking.py
+++ b/lib/charms/opensearch/v0/helper_networking.py
@@ -6,7 +6,6 @@ import logging
 import os
 import socket
 import subprocess
-from enum import Enum
 from typing import Dict, List, Optional
 
 from ops.charm import CharmBase
--- a/lib/charms/opensearch/v0/helper_networking.py
+++ b/lib/charms/opensearch/v0/helper_networking.py
@@ -6,6 +6,7 @@ import logging
 import os
 import socket
 import subprocess
+from enum import Enum
 from typing import Dict, List, Optional
 
 from ops.charm import CharmBase
@@ -25,9 +26,22 @@ LIBPATCH = 1
 logger = logging.getLogger(__name__)
 
 
-def get_host_ip(charm: CharmBase, peer_relation_name: str) -> str:
+class NetworkType(str, Enum):
+    """NetworkType Enum."""
+
+    PUBLIC = "public"
+    PRIVATE = "private"
+    BIND = "bind"
+
+
+def get_host_ip(charm: CharmBase, relation_name: str, type: NetworkType = NetworkType.BIND) -> str:
     """Fetches the IP address of the current unit."""
-    address = charm.model.get_binding(peer_relation_name).network.bind_address
+    if type == NetworkType.PRIVATE:
+        address = charm.model.get_binding(relation_name).network.private_address
+    elif type == NetworkType.PUBLIC:
+        address = charm.model.get_binding(relation_name).network.ingress_address
+    else:
+        address = charm.model.get_binding(relation_name).network.bind_address
     return str(address)
 
 
--- a/lib/charms/opensearch/v0/helper_networking.py
+++ b/lib/charms/opensearch/v0/helper_networking.py
@@ -30,14 +30,11 @@ class NetworkType(str, Enum):
     """NetworkType Enum."""
 
     PUBLIC = "public"
-    PRIVATE = "private"
     BIND = "bind"
 
 
 def get_host_ip(charm: CharmBase, relation_name: str, type: NetworkType = NetworkType.BIND) -> str:
     """Fetches the IP address of the current unit."""
-    if type == NetworkType.PRIVATE:
-        address = charm.model.get_binding(relation_name).network.private_address
     elif type == NetworkType.PUBLIC:
         address = charm.model.get_binding(relation_name).network.ingress_address
     else:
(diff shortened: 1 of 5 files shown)


smethnani (dismissed):

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -435,6 +439,11 @@ def network_hosts(self) -> List[str]:
         """All HTTP/Transport hosts for the current node."""
         return [socket.getfqdn(), self.host]
 
+    @property
+    def public_address(self) -> str:
```
Mehdi-Bendriss, line 443: this is simply acting as a proxy method, it is safe to delete

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -286,6 +291,11 @@ def update_host_if_needed(self) -> bool:
                 node.get("network.publish_host"),
                 self._opensearch.host,
             ),
+            NetworkHost(
+                "http.publish_host",
+                node.get("http.publish_host"),
+                self._opensearch.public_address,
```
Mehdi-Bendriss, line 295: This can be `None` I guess, in which case we probably want to skip it.

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -171,6 +171,11 @@ def set_node(
                 self.CONFIG_YML, "network.publish_host", self._opensearch.host
             )
 
+        if self._opensearch.public_address:
+            self._opensearch.config.put(
+                self.CONFIG_YML, "http.publish_host", self._opensearch.public_address
+            )
```
Mehdi-Bendriss: to avoid a double DNS querying, perhaps:
 ```suggestion
        if public_address := self._opensearch.public_address:
            self._opensearch.config.put(
                self.CONFIG_YML, "http.publish_host", public_address
            )
```

File: lib/charms/opensearch/v0/helper_networking.py [hunk 4]
```diff
@@ -25,9 +26,22 @@
...
+    PUBLIC = "public"
+    PRIVATE = "private"
+    BIND = "bind"
+
+
+def get_host_ip(charm: CharmBase, relation_name: str, type: NetworkType = NetworkType.BIND) -> str:
     """Fetches the IP address of the current unit."""
-    address = charm.model.get_binding(peer_relation_name).network.bind_address
+    if type == NetworkType.PRIVATE:
+        address = charm.model.get_binding(relation_name).network.private_address
```
Mehdi-Bendriss: I do not find any mention of `private_address` in the `Network` class. Can you clarify where it's coming from?

lib/charms/opensearch/v0/helper_networking.py
```diff
@@ -25,9 +26,22 @@
...
+    elif type == NetworkType.PUBLIC:
+        address = charm.model.get_binding(relation_name).network.ingress_address
```
Mehdi-Bendriss: This is not quite accurate - from the docs: 
> Due to things like public/private addresses, NAT and tunneling, the address the charm will bind() to is not always the address other people can use to connect() to the charm. This is just the first address from :attr:`.ingress_addresses`

In any way - we should avoid mixing concerns here. We have 2 separate helpers:
- `get_host_ip`
- `get_host_public_ip` 

Since it's not used elsewhere - We should probably take this out of the current method and keep it simple.

Mehdi-Bendriss (changes requested):
Thanks Pedro - I have some comments.

lib/charms/opensearch/v0/opensearch_config.py
phvalguima: That should never be none, if we do not get the DNS reply, then we should use the `ingress_address` as a fall back. At worse, we will get the same IP as the others, which is fine.

lib/charms/opensearch/v0/opensearch_distro.py
phvalguima: Added an option to fallback to ingress address if none was found with `get_host_public_ip`

lib/charms/opensearch/v0/opensearch_config.py
phvalguima: That if was removed as the idea for the `public_address` to always return an address.

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -435,6 +439,13 @@ def network_hosts(self) -> List[str]:
...
+        """Get the client bind address."""
```
Mehdi-Bendriss:
```suggestion
        """Get the public bind address of this unit."""
```

Mehdi-Bendriss (commented):
Thanks Pedro - looks good. I have 1 minor comment

Mehdi-Bendriss (approved):

smethnani (approved):

PR #581, opened 2025-03-06
[DPE-6771] Add sniff_timeout

### LINKED ISSUE #577: [CI] `continuous_write` does not set a friendly `sniff_timeout` option
When using the integration tests against OpenSearch remotely, I've noticed that `c_writes`-based test will fail and the `series_index` does not even gets created.

Looking at it closer, the test fails because the opensearch client was not able to sniff hosts.

The default sniff timeout is set at [client creation, as one of `kwargs`](https://github.com/opensearch-project/opensearch-py/blob/7336f51aa02562e19a320b1da4938abd79235a40/opensearchpy/client/__init__.py#L203). The docstring does [indicate we should use `sniffer_timeout`](https://github.com/opensearch-project/opensearch-py/blob/7336f51aa02562e19a320b1da4938abd79235a40/opensearchpy/client/__init__.py#L101), however the transport class uses [sniff_timeout](https://github.com/opensearch-project/opensearch-py/blob/7336f51aa02562e19a320b1da4938abd79235a40/opensearchpy/transport.py#L108). Fixing that name on `opensearch_client` method in our helpers does call the transport class with the right 60s value.

phvalguima:
Currently, we are using default sniff_timeout, which is 100ms.

That is not enough for the opensearchpy to a remote cluster.

This PR adds sniff_timeout and extends it to 60s.

Closes #577

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -114,7 +114,7 @@ async def run_action(
                 continue
 
             ping = subprocess.call(
-                f"nc -zv {unit.ip} 9200".split(),
+                f"nc -zv {unit.ip} 22".split(),
                 stdout=subprocess.DEVNULL,
                 stderr=subprocess.DEVNULL,
             )
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -437,7 +437,7 @@ def opensearch_client(
         sniff_on_start=True,  # sniff before doing anything
         sniff_on_connection_fail=True,  # refresh nodes after a node fails to respond
         sniffer_timeout=60.0,  # and also every 60 seconds
-        sniff_timeout=60.0,  # and also every 60 seconds
+        sniff_timeout=5.0,  # and also every 60 seconds
         use_ssl=True,  # turn on ssl
         verify_certs=True,  # make sure we verify SSL certificates
         ssl_assert_hostname=False,
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -114,7 +114,7 @@ async def run_action(
                 continue
 
             ping = subprocess.call(
-                f"ping -c 1 {unit.ip}".split(),
+                f"nc -zv {unit.ip} 9200".split(),
                 stdout=subprocess.DEVNULL,
                 stderr=subprocess.DEVNULL,
             )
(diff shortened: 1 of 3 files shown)


smethnani (dismissed):

tests/integration/helpers.py
```diff
@@ -436,7 +436,8 @@ def opensearch_client(
         http_compress=True,
         sniff_on_start=True,  # sniff before doing anything
         sniff_on_connection_fail=True,  # refresh nodes after a node fails to respond
-        sniffer_timeout=60,  # and also every 60 seconds
+        sniffer_timeout=60.0,  # and also every 60 seconds
+        sniff_timeout=60.0,  # and also every 60 seconds
```
smethnani: My only concern is that according to the comments [here](https://opensearch-project.github.io/opensearch-py/api-ref/transport.html#opensearchpy.Transport), the `sniff_timeout` should fail quickly. Going from 100ms to 60s seems like a big jump, do we need it to be the full minute?

smethnani (commented):
Thanks Pedro! Just left a question about the 60s choice

tests/integration/helpers.py
phvalguima: @smethnani fair point, I did not tune it in any way, just put the same value of 60s to move the tests forward. I will rerun it with 5s instead

phvalguima: FYI, 100ms is not enough as I have 200ms RTT with my target public cloud, which is in US... I think even 1s may be a stretch

Mehdi-Bendriss: Yeah I agree with @smethnani - 60 seconds sounds exagerated. I think it should at most be 5sec given we're running opensearch in the public cloud without immense expected latency

tests/integration/helpers.py
```diff
@@ -114,7 +114,7 @@ async def run_action(
                 continue
 
             ping = subprocess.call(
-                f"ping -c 1 {unit.ip}".split(),
+                f"nc -zv {unit.ip} 9200".split(),
```
Mehdi-Bendriss: This is not the goal of this ping. This will bias integration tests and set them in an ideal scenario. 
We are not attempting to prevent actions from being run until after the opensearch instance is up, what we want instead is to run actions after the **juju unit** is up.

Mehdi-Bendriss (changes requested):
Thanks Pedro. I have a comment.

tests/integration/helpers.py
phvalguima: @Mehdi-Bendriss why then juju reporting the unit as "running" is not enough?

phvalguima: Actually, we can target port 22 or 17070, they must be open since the start

tests/integration/helpers.py
phvalguima: Tested successfully yesterday and decreased the value to `5.0`.

smethnani (approved):

Mehdi-Bendriss (approved):

PR #585, opened 2025-03-10
[DPE-6597][DPE-6779] Remove peer-cluster observer and Unify S3/Azure integration

### LINKED ISSUE #606: [CI] Backup test occasionally fails with mismatching timestamps
It seems we are using a too precise metric to compare the upstream snapshot taken vs. the current time.

For example, this PR, we have a [run that failed](https://github.com/canonical/opensearch-operator/actions/runs/14228223476/job/39873469521?pr=585).

Looking at the error, I can see:
```
2025-04-02T20:18:09.5215254Z AssertionError: assert datetime.datetime(2025, 4, 2, 20, 17, 39) > datetime.datetime(2025, 4, 2, 20, 17, 39, 104047)
```

Which means we took the snapshot too early and the difference falls in the microseconds.

Full traceback:
```
2025-04-02T20:18:09.5130255Z =================================== FAILURES ===================================
2025-04-02T20:18:09.5131471Z [31m[1m_______________ test_create_backup_and_restore[microceph-large] ________________[0m
2025-04-02T20:18:09.5132353Z Traceback (most recent call last):
2025-04-02T20:18:09.5133758Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/runner.py", line 341, in from_call
2025-04-02T20:18:09.5135308Z     result: TResult | None = func()
2025-04-02T20:18:09.5137292Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/runner.py", line 242, in <lambda>
2025-04-02T20:18:09.5139575Z     lambda: runtest_hook(item=item, **kwds), when=when, reraise=reraise
2025-04-02T20:18:09.5142104Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_hooks.py", line 513, in __call__
2025-04-02T20:18:09.5143960Z     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
2025-04-02T20:18:09.5145972Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_manager.py", line 120, in _hookexec
2025-04-02T20:18:09.5147902Z     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
2025-04-02T20:18:09.5149607Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 182, in _multicall
2025-04-02T20:18:09.5151793Z     return outcome.get_result()
2025-04-02T20:18:09.5153115Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_result.py", line 100, in get_result
2025-04-02T20:18:09.5154533Z     raise exc.with_traceback(exc.__traceback__)
2025-04-02T20:18:09.5155990Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 167, in _multicall
2025-04-02T20:18:09.5157375Z     teardown.throw(outcome._exception)
2025-04-02T20:18:09.5158913Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/threadexception.py", line 92, in pytest_runtest_call
2025-04-02T20:18:09.5160686Z     yield from thread_exception_runtest_hook()
2025-04-02T20:18:09.5162119Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/threadexception.py", line 68, in thread_exception_runtest_hook
2025-04-02T20:18:09.5163090Z     yield
2025-04-02T20:18:09.5163924Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 167, in _multicall
2025-04-02T20:18:09.5165223Z     teardown.throw(outcome._exception)
2025-04-02T20:18:09.5166344Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/unraisableexception.py", line 95, in pytest_runtest_call
2025-04-02T20:18:09.5167453Z     yield from unraisable_exception_runtest_hook()
2025-04-02T20:18:09.5168565Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/unraisableexception.py", line 70, in unraisable_exception_runtest_hook
2025-04-02T20:18:09.5170099Z     yield
2025-04-02T20:18:09.5170975Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 167, in _multicall
2025-04-02T20:18:09.5172198Z     teardown.throw(outcome._exception)
2025-04-02T20:18:09.5173196Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/logging.py", line 846, in pytest_runtest_call
2025-04-02T20:18:09.5174145Z     yield from self._runtest_for(item, "call")
2025-04-02T20:18:09.5175057Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/logging.py", line 829, in _runtest_for
2025-04-02T20:18:09.5175920Z     yield
2025-04-02T20:18:09.5176717Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 167, in _multicall
2025-04-02T20:18:09.5177605Z     teardown.throw(outcome._exception)
2025-04-02T20:18:09.5178525Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/capture.py", line 898, in pytest_runtest_call
2025-04-02T20:18:09.5179433Z     return (yield)
2025-04-02T20:18:09.5180406Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 167, in _multicall
2025-04-02T20:18:09.5181393Z     teardown.throw(outcome._exception)
2025-04-02T20:18:09.5182328Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/skipping.py", line 257, in pytest_runtest_call
2025-04-02T20:18:09.5183227Z     return (yield)
2025-04-02T20:18:09.5184160Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 103, in _multicall
2025-04-02T20:18:09.5185247Z     res = hook_impl.function(*args)
2025-04-02T20:18:09.5186584Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/runner.py", line 174, in pytest_runtest_call
2025-04-02T20:18:09.5188003Z     item.runtest()
2025-04-02T20:18:09.5189233Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/python.py", line 1627, in runtest
2025-04-02T20:18:09.5191062Z     self.ihook.pytest_pyfunc_call(pyfuncitem=self)
2025-04-02T20:18:09.5192507Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_hooks.py", line 513, in __call__
2025-04-02T20:18:09.5194026Z     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
2025-04-02T20:18:09.5195636Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_manager.py", line 120, in _hookexec
2025-04-02T20:18:09.5197181Z     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
2025-04-02T20:18:09.5198727Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 182, in _multicall
2025-04-02T20:18:09.5200494Z     return outcome.get_result()
2025-04-02T20:18:09.5201765Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_result.py", line 100, in get_result
2025-04-02T20:18:09.5203123Z     raise exc.with_traceback(exc.__traceback__)
2025-04-02T20:18:09.5204482Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pluggy/_callers.py", line 103, in _multicall
2025-04-02T20:18:09.5206129Z     res = hook_impl.function(*args)
2025-04-02T20:18:09.5207538Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/_pytest/python.py", line 159, in pytest_pyfunc_call
2025-04-02T20:18:09.5208950Z     result = testfunction(**testargs)
2025-04-02T20:18:09.5210503Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/.tox/integration/lib/python3.10/site-packages/pytest_asyncio/plugin.py", line 529, in inner
2025-04-02T20:18:09.5211871Z     _loop.run_until_complete(task)
2025-04-02T20:18:09.5212529Z   File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
2025-04-02T20:18:09.5213060Z     return future.result()
2025-04-02T20:18:09.5213880Z   File "/home/ubuntu/actions-runner/_work/opensearch-operator/opensearch-operator/tests/integration/ha/test_backups.py", line 557, in test_create_backup_and_restore
2025-04-02T20:18:09.5214705Z     assert (
2025-04-02T20:18:09.5215254Z AssertionError: assert datetime.datetime(2025, 4, 2, 20, 17, 39) > datetime.datetime(2025, 4, 2, 20, 17, 39, 104047)
```

phvalguima:
This PR converges the API-related code in the backup components to a single class, named "BackupManager". This class will handle the primitives (backup, restore, list, clean, etc) that result in API calls to the backup system in OpenSearch.

The OpenSearchBackup class itself is then broken into three:
* OpenSearchMainBackup: for the MAIN cluster - handle all common code between S3 and azure event handling
* OpenSearch{S3,Azure}Backup: for the MAIN cluster - handle all the S3 / azure event specifics and inherit from OpenSearchMainBackup

The goal is to unify the Azure and S3 backup components as much as possible.

This PR also removes unneeded code:
* Peer cluster listener: we do not need anymore and we stick exclusively with secret changed for the opensearch non-main apps
* `OpenSearchPluginDataProvider`: any more complex data management must be done by upper classes instead of the plugin, in the case of backup, the `OpenSearchMainBackup` and its children will be the ones handling this extra data needed
* The switch code within `OpenSearch{S3,Azure}Plugin` that switches between main and non-main orchestrator: this is also handled within the `OpenSearchMainBackup` and its children

Closes #606, #623


lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -212,6 +214,107 @@ def __init__(self, charm: "OpenSearchBaseCharm", repository: str | None = None):
...
+          - OpenSearchHttpError
+        """
+        if not plugin.data:
+            return BackupServiceState.REPO_ERR_UNKNOWN
+        response = self.charm.opensearch.request(
+            "PUT",
+            f"_snapshot/{self.repository}",
+            payload={
+                "type": "s3" if self.repository == S3_REPOSITORY else "azure",
+                "settings": plugin.data.dict(exclude={"tls_ca_chain", "credentials"}),
```
zmraul: **todo**: Setting these values for Azure means that we send all of this model in dict form:
```python
storage_account: str = Field(alias="storage-account", default="")
container: str = Field(default="")
endpoint: Optional[str] = Field(default="")
base_path: Optional[str] = Field(alias="path", default=AZURE_REPO_BASE_PATH)
connection_protocol: Optional[str] = Field(alias="connection-protocol", default=None)
```
I believe all of them are fine, but `storage_account`. That one would need to be added to the `exclude` field. I _think_ `base_path` is passed with that name and not `path`. The correct setting for the azure config is `base_path`

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -406,20 +581,7 @@ def is_set(self) -> bool:
...
-        # cannot get the snapshot list.
-        response = self.charm.opensearch.request("GET", f"_snapshot/{self.repository}/_all")
-        return {
-            snapshot["snapshot"].upper(): {
-                "state": snapshot["state"],
-                "indices": snapshot.get("indices", []),
-            }
-            for snapshot in response.get("snapshots", [])
-        }
+        return not self.is_backup_in_progress() and not self.is_restore_in_progress()
```
zmraul: Very good catch!

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -874,29 +939,13 @@ def _on_create_backup_action(self, event: ActionEvent) -> None:  # noqa: C901
         if self.charm.upgrade_in_progress:
             event.fail("Backup not supported while upgrade in-progress")
             return
-        if not self._can_unit_perform_backup(event):
+        if not self.backup_manager.is_set() or self.backup_manager.is_backup_in_progress():
```
zmraul: **question**: Would `is_idle` be more appropriate here?
```suggestion
        if not self.backup_manager.is_set() or not self.backup_manager.is_idle():
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1087,564 +1120,76 @@ def _on_backup_relation_broken(self, event: RelationEvent) -> None:  # noqa: C90
...
-            response = self.charm.opensearch.request(
-                "GET",
-                f"_snapshot/{S3_REPOSITORY}",
-                retries=6,
-                timeout=10,
-            )
-            return self.get_service_status(response)
-        except OpenSearchHttpError:
-            return BackupServiceState.RESPONSE_FAILED_NETWORK
+class OpenSearchS3Backup(OpenSearchMainBackup):
```
zmraul, line 1184: Amazing! I agree this is the very minimum they differ on, just the client lib and the plugin initialization. Very neat :)

zmraul (commented):
Nice! This simplifies a lot the class and the actual split between S3 and azure is minimal. 

Left a couple of points. There is also a lot of leftover comments from the refactor on the `OpenSearchMainBackup` class that are still referencing s3, on both docstrings and logs. I would just search for `s3` keyword on the class and switch them by `self.repository` on logs and something generic for docstring.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Yes, that is indeed safer as we also avoid the case of backup run during restore!

phvalguima: I will actually break this bit into two different errors, so we can orient the user more carefully.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: @zmraul fair point, I am handling this now in a separated method. Now, each `OpenSearchMainBackup` child has a `data` method where the dictionary with the correct data is returned.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -58,14 +56,6 @@
         "class": OpenSearchKnn,
         "config": "plugin_opensearch_knn",
     },
-    "repository-s3": {
```
phvalguima, line 61: from now on, `OpenSearchBackup` classes will handle these plugins, as the place we read the config options in a given relation databag changes according to the app being main (hence reading from s3 relation) or non-main (hence, reading from secrets in the peer-cluster relation).

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1087,564 +1127,118 @@ def _on_backup_relation_broken(self, event: RelationEvent) -> None:  # noqa: C90
...
+        if endpoint_info := RequirerData(
+            model=self.charm.model,
+            relation_name=AZURE_RELATION,
+            additional_secret_fields=["secret-key"],
         ):
-            to_close = output["error"]["reason"].split("[")[2].split("]")[0]
-            raise OpenSearchRestoreIndexClosingError(f"_restore: fails to close {to_close}")
+            if (data := endpoint_info.fetch_relation_data()) and self._relation.id in data:
+                return AzureRelData.from_relation(data[self._relation.id])
+        return None
```
zmraul: Nice! I think this change is good, keeps them close together since it's not used anywhere else and only relevant within the backup class

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -630,26 +789,44 @@ def _on_backup_relation_broken(self, event: RelationEvent) -> None:
         logger.info("Non-orchestrator cluster, abandon relation event")
 
 
-class OpenSearchS3Backup(OpenSearchBackupBase):
-    """Implements backup relation and API management."""
+class OpenSearchMainBackup(OpenSearchBackupBase):
```
zmraul, line 926: I've just noticed that `on_relation_broken` should probably be moved to the `BackupBase` class. Then remove both of them whether they exist or not?. The logic on the `backup` method at the very bottom of the module would not initialize any of the subclasses for Azure/S3. Because on `relation_broken`, `backup_base.active_relation` will be False.
```python
backup_base = OpenSearchBackupBase(charm)
    if not charm.opensearch_peer_cm.deployment_desc():
        return backup_base
    elif charm.opensearch_peer_cm.deployment_desc().typ == DeploymentType.MAIN_ORCHESTRATOR:
        if not backup_base.active_relation:
            return backup_base
        if backup_base.active_relation == S3_RELATION:
            return OpenSearchS3Backup(charm)
        if backup_base.active_relation == AZURE_RELATION:
            return OpenSearchAzureBackup(charm)
return OpenSearchNonOrchestratorClusterBackup(charm)
```

zmraul (commented):
Seems good! I noticed an issue with `relation_broken`, otherwise seems good.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Nice catch! Fixed by moving the `relation_broken` all the way to the base class. It demanded some adjustments to fit both cases thou.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -651,9 +645,87 @@ def _on_backup_relation_created(self, _: RelationEvent) -> None:
...
+        if (
+            self.charm.opensearch_peer_cm.deployment_desc().typ == DeploymentType.MAIN_ORCHESTRATOR
+            and self.charm.model.get_relation(self.relation_name)
+            and self.charm.model.get_relation(self.relation_name).units
+        ):
+            # The main app units must defer this event
+            # as they are related directly with the backup integrator
+            # and they are still seeing data in the relation databag.
+            event.defer()
+            return
```
zmraul: **question**: I don't understand this block. Why defer under these conditions?

zmraul (dismissed):
Good changes. One comment on the switched broken relation

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Relation broken used to come with some relation databag / remote units. This is more to make sure we are running this logic when we really committed to remove the s3 relation.

zmraul: But here `self.relation_name` will be the `PeerClusterRelation`, not s3/azure

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -645,22 +688,62 @@ def _on_backup_relation_created(self, _: RelationEvent) -> None:
...
         if self.charm.upgrade_in_progress:
             logger.warning(
                 "Modifying relations during an upgrade is not supported. The charm may be in a broken, unrecoverable state"
             )
 
         if (
-            self.charm.opensearch_peer_cm.deployment_desc().typ == DeploymentType.MAIN_ORCHESTRATOR
-            and self.charm.model.get_relation(self.relation_name)
-            and self.charm.model.get_relation(self.relation_name).units
+            self.charm.model.get_relation(event.relation_name)
```
zmraul, line 727: :+1: good change

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -588,12 +589,52 @@ def get_snapshot_status(response: Dict[str, Any] | None) -> BackupServiceState:
         return BackupServiceState.SUCCESS
 
 
+class _DisableBackupRelationEvent(EventBase):
```
zmraul, line 577: Interesting. Is this to avoid defers/returns on the `relation_broken` event?

zmraul (dismissed):

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -212,6 +218,107 @@ def __init__(self, charm: "OpenSearchBaseCharm", repository: str | None = None):
...
+                "partial": False,  # It is the default value, but we want to avoid partial restores
+            },
+            retries=6,
+            timeout=10,
+        )
+        logger.debug(f"_restore: restore call returned {output}")
+        if (
+            BackupManager.get_service_status(output)
+            == BackupServiceState.SNAPSHOT_RESTORE_ERROR_INDEX_NOT_CLOSED
+        ):
```
reneradoi, line 257:
```suggestion
# expected format for `reason`:
# "[my-opensearch-repo:my-first-snapshot/dCK4Qth-TymRQ7Tu7Iga0g] cannot restore index [.opendistro-reports-definitions] because ..."
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -212,6 +218,107 @@ def __init__(self, charm: "OpenSearchBaseCharm", repository: str | None = None):
...
+        response = self.charm.opensearch.request("GET", f"_snapshot/{self.repository}/_all")
+        return {
+            snapshot["snapshot"].upper(): {
+                "state": snapshot["state"],
+                "indices": snapshot.get("indices", []),
+            }
+            for snapshot in response.get("snapshots", [])
+        }
+
+    def clean(self):
```
reneradoi, line 282: What is the intention here? Can't this just be removed?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -212,6 +218,107 @@ def __init__(self, charm: "OpenSearchBaseCharm", repository: str | None = None):
...
+    def is_backup_in_progress(self) -> bool:
+        """Returns True if backup is in progress, False otherwise.
+
+        This method assumes that a relation already exists, so a check for active_relation != None
+        should generally be done.
+
+        We filter the _query_backup_status() and seek for the following states:
+        - SNAPSHOT_IN_PROGRESS
+        """
+        if self._query_backup_status() in [
```
reneradoi: This could be simplified by querying `GET /_snapshot/_status` instead to see if any snapshot is currently being created, see https://docs.opensearch.org/docs/latest/api-reference/snapshots/get-snapshot-status/

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -489,8 +689,131 @@ def _on_backup_relation_created(self, _: RelationEvent) -> None:
...
-        """Defers the backup relation broken events."""
-        raise NotImplementedError
+        """Emits the backup disable event."""
+        if event.relation.name in [S3_RELATION, AZURE_RELATION]:
+            self._disable_backup_event.emit(
+                relation_name=(
+                    S3_RELATION if event.relation.name == S3_RELATION else AZURE_RELATION
+                ),
+            )
+            return
```
reneradoi, line 684: This method only seems to be executed for either `S3_RELATION` or `AZURE_RELATION`, so the whole code block after the return-statement can be removed.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1018,633 +1199,114 @@ def apply_api_config_if_needed(self) -> None:
...
-        ):
-            event.defer()
-            return
+    def __init__(self, charm: "OpenSearchBaseCharm", relation_name: str = S3_RELATION):
+        """Manager of OpenSearch backup relations."""
+        super().__init__(charm, relation_name)
+        self.client = S3Requirer(self.charm, S3_RELATION)
+        self.framework.observe(
+            self.client.on.credentials_changed, self._on_backup_credentials_changed
+        )
```
reneradoi, line 1193: Why are we not listening to the `credentials_gone` here?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -664,15 +984,12 @@ def _on_backup_action(self, event: ActionEvent) -> None:
         """Just overloads the base method, as we process each action in this class."""
         pass
 
-    @property
-    def _plugin_status(self):
-        return self.charm.plugin_manager.get_plugin_status(OpenSearchS3Plugin)
-
     def _on_list_backups_action(self, event: ActionEvent) -> None:
         """Returns the list of available backups to the user."""
         if not self.charm.opensearch_peer_cm.deployment_desc():
```
reneradoi: I believe this doesn't need to be checked again here as it is already covered by the same condition in line 1329:
```
if not self.charm.opensearch_peer_cm.deployment_desc():
    return backup_base
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -687,113 +1004,6 @@ def _on_list_backups_action(self, event: ActionEvent) -> None:
...
-                backup_id in backups.keys()
-                and self.backup_manager.get_snapshot_status(backups[backup_id]["state"])
-                == BackupServiceState.SUCCESS
-            )
-        except OpenSearchListBackupError:
-            return False
-
     def _on_restore_backup_action(self, event: ActionEvent) -> None:  # noqa #C901
         """Restores a backup to the current cluster."""
         if not self.charm.opensearch_peer_cm.deployment_desc():
```
reneradoi: Same comment, already covered by the same condition in line 1329.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -270,27 +456,7 @@ def is_restore_complete(self) -> bool:
             raise OpenSearchRestoreCheckError("is_restore_complete: failed to get indices status")
         return not self.is_restore_in_progress()
```
reneradoi: I believe the logic of this function `is_restore_complete()` is more or less a duplicate of `is_restore_in_progress()` and it would be easier to just use `is_restore_in_progress()` instead.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -825,8 +1035,8 @@ def _on_restore_backup_action(self, event: ActionEvent) -> None:  # noqa #C901
         # In case of failure, then restore action must return a list of closed indices
         closed_idx = set()
```
reneradoi: can be removed

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -489,8 +689,131 @@ def _on_backup_relation_created(self, _: RelationEvent) -> None:
...
+            event.defer()
+            return
+
+        # Now, as this is related strictly to the MAIN orchestrator,
+        # we must update any peer relation.
+        if self.charm.opensearch_peer_cm.is_provider(typ="main"):
+            self.charm.peer_cluster_provider.refresh_relation_data(event)
+
+        self.charm.status.clear(BackupInDisabling)
+        self.charm.status.clear(PluginConfigError)
```
reneradoi, line 798: This comment is related to the `_on_backup_action()` method some lines below: The error message is misleading. If there is no relation between Opensearch and S3/Azure integrator, the error message on running a backup action is "Deployment description not yet available", even though it's the missing relation that is causing the action to fail. This should be more distinctive, as the `OpenSearchBackupBase` class is used not only when the deployment isn't ready yet, but also if none or two object storage relations are available.

reneradoi (changes requested):
Hey Pedro, on Mehdi's request I've reviewed this quite big PR. There are a couple of things that I noticed in the code (see detailed comments), and there are a couple of errors I encountered when testing. Sorry for the extensive output here, please let me know if you need further details.

1) Setup of Backup service fails if region is not configured
```
unit-opensearch-0: 10:41:38 ERROR unit.opensearch/0.juju-log s3-credentials:5: Failed to setup backup service with exception: HTTP error self.response_code=500
self.response_body={'error': {'root_cause': [{'type': 'repository_verification_exception', 'reason': '[s3-repository] path [/opensearch-backups] is not accessible on cluster-manager node'}], 'type': 'repository_verification_exception', 'reason': '[s3-repository] path [/opensearch-backups] is not accessible on cluster-manager node', 'caused_by': {'type': 'i_o_exception', 'reason': 'Unable to upload object [/opensearch-backups/tests-5dODrtcmTymWtSlX4P6G4w/master.dat] using a single upload', 'caused_by': {'type': 'sdk_client_exception', 'reason': 'Unable to load region from any of the providers in the chain software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain@36812c18: [software.amazon.awssdk.regions.providers.SystemSettingsRegionProvider@4aeae607: Unable to load region from system settings. Region must be specified either via environment variable (AWS_REGION) or  system property (aws.region)., software.amazon.awssdk.regions.providers.AwsProfileRegionProvider@190ffd60: No region provided in profile: default, software.amazon.awssdk.regions.providers.InstanceProfileRegionProvider@314edb3b: Unable to retrieve region information from EC2 Metadata service. Please make sure the application is running on EC2.]'}}}, 'status': 500}
unit-opensearch-0: 10:41:38 ERROR unit.opensearch/0.juju-log s3-credentials:5: Failed to setup backup service with state Repository exception: unknown
```

If region is a mandatory parameter, it should be checked and stated to the user if missing

2) Setup of Backup service fails if connecting to Microceph with Rados-GW and TLS:

```
unit-opensearch-0: 10:45:58 ERROR unit.opensearch/0.juju-log s3-credentials:5: Failed to setup backup service with exception: HTTP error self.response_code=500
self.response_body={'error': {'root_cause': [{'type': 'repository_verification_exception', 'reason': '[s3-repository] path [/opensearch-backups] is not accessible on cluster-manager node'}], 'type': 'repository_verification_exception', 'reason': '[s3-repository]
 path [/opensearch-backups] is not accessible on cluster-manager node', 'caused_by': {'type': 'i_o_exception', 'reason': 'Unable to upload object [/opensearch-backups/tests-NaAa8hPDQTiUArP7aramTA/master.dat] using a single upload', 'caused_by': {'type': 'sdk_cli
ent_exception', 'reason': 'Unable to execute HTTP request: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target', 'caused_by': {'type': 's_s_l_handshake_exception', 're
ason': 'PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target', 'caused_by': {'type': 'validator_exception', 'reason': 'PKIX path building failed: sun.security.provider.
certpath.SunCertPathBuilderException: unable to find valid certification path to requested target', 'caused_by': {'type': 'sun_cert_path_builder_exception', 'reason': 'unable to find valid certification path to requested target'}}}, 'suppressed': [{'type': 'sdk_
client_exception', 'reason': 'Request attempt 1 failure: Unable to execute HTTP request: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target'}, {'type': 'sdk_client_ex
ception', 'reason': 'Request attempt 2 failure: Unable to execute HTTP request: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target'}, {'type': 'sdk_client_exception',
 'reason': 'Request attempt 3 failure: Unable to execute HTTP request: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target'}]}}}, 'status': 500}
unit-opensearch-0: 10:45:58 ERROR unit.opensearch/0.juju-log s3-credentials:5: Failed to setup backup service with state Repository exception: unknown
```

3) Setup of Backup service fails for Azure if container doesn't exist (assumption: same if s3 bucket doesn't exist):

```
unit-opensearch-0: 11:15:14 ERROR unit.opensearch/0.juju-log azure-credentials:6: Failed to setup backup service with exception: HTTP error self.response_code=500
self.response_body={'error': {'root_cause': [{'type': 'repository_verification_exception', 'reason': '[azure-repository] path  is not accessible on cluster-manager node'}], 'type': 'repository_verification_exception', 'reason': '[azure-repository] path  is not a
ccessible on cluster-manager node', 'caused_by': {'type': 'i_o_exception', 'reason': 'Can not write blob master.dat', 'caused_by': {'type': 'blob_storage_exception', 'reason': 'Status code 404, "\ufeff<?xml version="1.0" encoding="utf-8"?><Error><Code>ContainerN
otFound</Code><Message>The specified container does not exist.\nRequestId:4b3c30c0-501e-001d-390a-b5df13000000\nTime:2025-04-24T11:15:14.0221899Z</Message></Error>"', 'suppressed': [{'type': 'exception', 'reason': '#block terminated with an error'}]}}}, 'status'
: 500}
unit-opensearch-0: 11:15:14 ERROR unit.opensearch/0.juju-log azure-credentials:6: Failed to setup backup service with state Repository exception: unknown
```

4) `list-backups` shows wrong lenght of list: (includes headers in the count):

```
ubuntu@dev-vm:~$ juju run opensearch/leader list-backups
Running operation 7 with 1 task

Waiting for task 8...
backups: |2-
   backup-id           | backup-status
  ------------------------------------

ubuntu@dev-vm:~$ juju run opensearch/leader list-backups
Running operation 23 with 1 task
  - task 24 on unit-opensearch-0

Waiting for task 24...
backups: |2-
   backup-id           | backup-status
  ------------------------------------
  2025-04-24T11:21:58Z | success
  2025-04-24T11:22:56Z | success
  2025-04-24T11:27:32Z | success
```
The length of the output list should be computed after the list of backups was appended to the output.

phvalguima:
Hi @reneradoi, first thanks for the review, really appreciate taking the time and testing it although it being a big change. I have a couple of questions about your points:

> Setup of Backup service fails if region is not configured

When I tested, this is only required for RADOSGW as you are using an non-default EC2 region name. AWS's S3 is "Global", which does not demand region naming. So, I did not made it mandatory. We can change it thou, but I think that would be more suitable on s3-integrator itself.

> Setup of Backup service fails if connecting to Microceph with Rados-GW and TLS:

Nice catch! I will review the custom TLS bit from s3.

> list-backups shows wrong lenght of list: (includes headers in the count)

What do you mean by length? We are not printing it. Are you referring to calculating the number of `-`? That is the only part we are using the length. And `output[0]` is fixed size string:

```
    def _generate_backup_list_output(self, backups: Dict[str, Any]) -> str:
        """Generates a list of backups in a formatted table.

        List contains successful and failed backups in order of ascending time.

        Raises:
            OpenSearchError: if the list of backups errors
        """
        backup_list = []
        for id, backup in backups.items():
            state = BackupManager.get_snapshot_status(backup["state"])
            backup_list.append((id, state.value))

        output = ["{:<20s} | {:s}".format(" backup-id", "backup-status")]
        output.append("-" * len(output[0]))  ####### <<<<<<<<<<<<<<<<<< meaning this length?

        for backup_id, backup_status in backup_list:
            output.append("{:<20s} | {:s}".format(backup_id, backup_status))
        return "\n".join(output)
```

reneradoi:
> What do you mean by length? We are not printing it. Are you referring to calculating the number of `-`? That is the only part we are using the length. And `output[0]` is fixed size string:

This is the output of the action:
```
backups: |2-
   backup-id           | backup-status
  ------------------------------------
  2025-04-24T11:21:58Z | success
  2025-04-24T11:22:56Z | success
  2025-04-24T11:27:32Z | success
```
Not sure what the `2` is for, I considered it to be the amount of backups available.

reneradoi:
> > Setup of Backup service fails if region is not configured
> 
> When I tested, this is only required for RADOSGW as you are using an non-default EC2 region name. AWS's S3 is "Global", which does not demand region naming. So, I did not made it mandatory. We can change it thou, but I think that would be more suitable on s3-integrator itself.

On the s3-integrator itself it's probably not well suited, because there are implementations of the s3-interface that don't require the region to be set (for example when using boto3). But it is required by opensearch obviously if not a default AWS region, so I would recommend to check it on `s3_credentials_changed`.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: If I do so, then user-defined backups will also start blocking the charm itself. Users can still configure their own backup backend using API calls.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Yes, to defer the relation_broken safely.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: @reneradoi not really: if we have large deployments, we will receive the data from `peer-cluster` relation, which this logic after that `if` is looking for

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: We can, but essentially on both [s3](https://github.com/canonical/opensearch-operator/blob/2/edge/lib/charms/data_platform_libs/v0/s3.py#L786) and [azure](https://github.com/canonical/object-storage-integrators/blob/f0a94cb9055d7a71293a52398bd7f74d9dfabe99/azure_storage/lib/charms/data_platform_libs/v0/azure_storage.py#L219): `credentials_gone` is essentially a renaming of `-broken` relation and I am listening to -broken.

phvalguima:
@reneradoi a correction: indeed region is needed even for AWS - [tested in this dummy PR here](https://github.com/canonical/opensearch-operator/pull/621). 

I am adding an extra check in the `S3RelData` model to throw a `ValueError` in case it is not found.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Done, removed this references

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Not sure I understand your point.

So, the first thing about `_on_backup_disable` method to keep in mind: it is executed in the base class instead of the "children" classes. That is because we may not be able to define which types of classes should we create as the s3/azure/peer-cluster relation is gone now.

Therefore, we have to consider every case in this specific method. Hence, we have a break at the very early steps to return if we are not dealing with a MAIN app.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Is it just adding this comment? You are removing the parenthesis there as well

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: I prefer to keep it but leave empty. If we need to add any logic here later, then the function is already there. For now we just need the `disable()` of the plugin logic, as it removes the keys from the keystore.

lib/charms/opensearch/v0/opensearch_backups.py
reneradoi: Ups sorry, no, just add a comment.

lib/charms/opensearch/v0/opensearch_backups.py
reneradoi: Ok

lib/charms/opensearch/v0/opensearch_backups.py
reneradoi: Could a user-defined backup potentially interfere with the one we run? If so, maybe it would actually be good to block if there is any other backup running, no?

lib/charms/opensearch/v0/opensearch_backups.py
reneradoi: Thanks, that makes sense.

lib/charms/opensearch/v0/opensearch_backups.py
reneradoi: I was just wondering why you listen to the `storage_connection_info_gone` ([here](https://github.com/canonical/opensearch-operator/blob/backup-merge-main-classes/lib/charms/opensearch/v0/opensearch_backups.py#L1239)) of the azure storage lib but not on the s3 event - any particular reason?

lib/charms/opensearch/v0/opensearch_backups.py
reneradoi: Maybe we misunderstood each other: If you relate Opensearch to none of S3 and Azure storage integrator and run any action, the action will fail with `Deployment description not yet available`. This is misleading, because the deployment description is available, it's the relation that is missing. Same behaviour if Opensearch is related to both storage integrators.

In these cases, the UX needs improvement for users to find out what's actually wrong. So: correct error message and blocked status would be good.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: A backup, I doubt, but a restore can.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Not really, just we wrote the code in separate moments. I will roll back the storage-connection to use the relation-broken instead, as this is the event that I am really interested on.

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: I see, let me fix that! Thanks

lib/charms/opensearch/v0/opensearch_backups.py
phvalguima: Looking at the code, we are querying the backup of our specific repository, but we query overall recovery status for restore. I'd be inclined to keep it this way and then we can disconnect from any other backup repositories the user configured.

reneradoi (approved):
Thank you for making the adjustments @phvalguima !

For reference, the following open topics have been moved to issues and will be addressed separately:
- [Setup of Backup service fails if Azure-container or S3-bucket are not available yet](https://github.com/canonical/opensearch-operator/issues/627)
- [S3 backup/restore fails with Microceph/Rados-GW and TLS](https://github.com/canonical/opensearch-operator/issues/622)

Mehdi-Bendriss (approved):

PR #591, opened 2025-03-13
Decrease low-level `_run_cmd` logging to "DEBUG"

phvalguima:
Currently, if we have a command that fails in `_run_cmd`, we will: (1) get a ERROR level entry; and (2) raise an OpenSearchCmdError exception.

That can be confusing for an user that is following the deployment on juju debug-log, as the ERROR will point to a potential problem, but the upper classes may be handling the exception and taking the appropriate decisions afterwards.

Therefore, logging beyond "DEBUG" level should be part of the upper class source code, not the `_run_cmd` itself.


smethnani (approved):
Thanks, Pedro!

Mehdi-Bendriss (approved):

Issue #569, opened 2025-03-03
Large deployments - peer cluster relation hooks fail when removing failover app


smethnani:
## Steps to reproduce

1. Deploy apps and integrate
```
  juju deploy -n 3 \
      opensearch main \
      --config cluster_name="app" \
      --channel 2/edge \
      --config profile="testing"

  juju deploy -n 1 \
      opensearch failover \
      --config init_hold="true" \
      --config cluster_name="app" \
      --config roles="cluster_manager" \
      --channel 2/edge \
      --config profile="testing"

  juju deploy -n 1 \
      opensearch data-hot \
      --config init_hold="true" \
      --config cluster_name="app" \
      --config roles="data.hot" \
      --channel 2/edge \
      --config profile="testing"

  juju deploy self-signed-certificates --config ca-common-name="Test CA"
  juju integrate self-signed-certificates main
  juju integrate self-signed-certificates failover
  juju integrate self-signed-certificates data-hot
 
  juju integrate main:peer-cluster-orchestrator failover:peer-cluster
  juju integrate main:peer-cluster-orchestrator data-hot:peer-cluster
  juju integrate failover:peer-cluster-orchestrator data-hot:peer-cluster
```
2. Remove failover application `juju remove-application failover`
## Expected behavior
The failover app is removed

## Actual behavior
Data-hot moves to an error state with message `hook failed: "peer-cluster-relation-changed"`
Failover moves to an error state with message `hook failed: "peer-cluster-relation-departed"`

## Log output
Juju debug log: 
```
unit-data-hot-0: 07:30:41 ERROR unit.data-hot/0.juju-log peer-cluster:13: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-data-hot-0/charm/src/charm.py", line 221, in <module>
    main(OpenSearchOperatorCharm)
  File "/var/lib/juju/agents/unit-data-hot-0/charm/venv/lib/python3.10/site-packages/ops/main.py", line 553, in main
    manager.run()
  File "/var/lib/juju/agents/unit-data-hot-0/charm/venv/lib/python3.10/site-packages/ops/main.py", line 529, in run
    self._emit()
  File "/var/lib/juju/agents/unit-data-hot-0/charm/venv/lib/python3.10/site-packages/ops/main.py", line 518, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name, self._juju_context)
[...]

Issue #568, opened 2025-03-03
Large deployments - update-status hook fails after removing main app


smethnani:
## Steps to reproduce
1. Deploy apps and relate them together
```
  juju deploy -n 3 \
      opensearch main \
      --config cluster_name="app" \
      --channel 2/edge \
      --config profile="testing"

  juju deploy -n 1 \
      opensearch failover \
      --config init_hold="true" \
      --config cluster_name="app" \
      --config roles="cluster_manager" \
      --channel 2/edge \
      --config profile="testing"

  juju deploy -n 1 \
      opensearch data-hot \
      --config init_hold="true" \
      --config cluster_name="app" \
      --config roles="data.hot" \
      --channel 2/edge \
      --config profile="testing"

  juju deploy self-signed-certificates --config ca-common-name="Test CA"
  juju integrate self-signed-certificates main
  juju integrate self-signed-certificates failover
  juju integrate self-signed-certificates data-hot
 
  juju integrate main:peer-cluster-orchestrator failover:peer-cluster
  juju integrate main:peer-cluster-orchestrator data-hot:peer-cluster
  juju integrate failover:peer-cluster-orchestrator data-hot:peer-cluster
```
2. Remove main application: `juju remove-application main`

## Expected behavior
The main application is removed and failover is promoted

## Actual behavior
The main application is removed and data-hot goes into an error state once the `update-status` event is fired.

## Log output
Juju debug log: 

```
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-data-hot-0/charm/src/charm.py", line 221, in <module>
    main(OpenSearchOperatorCharm)
  File "/var/lib/juju/agents/unit-data-hot-0/charm/venv/lib/python3.10/site-packages/ops/main.py", line 553, in main
    manager.run()
  File "/var/lib/juju/agents/unit-data-hot-0/charm/venv/lib/python3.10/site-packages/ops/main.py", line 529, in run
    self._emit()
  File "/var/lib/juju/agents/unit-data-hot-0/charm/venv/lib/python3.10/site-packages/ops/main.py", line 518, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name, self._juju_context)
  File "/var/lib/juju/agents/unit-data-hot-0/charm/venv/lib/python3.10/site-packages/ops/main.py", line 139, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
[...]

Issue #564, opened 2025-02-25
Large deployments - Data node Active status without cluster manger


smethnani:
## Steps to reproduce
1. Deploy and relate applications:
```
juju deploy -n 3 opensearch main \
 --config cluster_name="app" \
 --channel 2/edge \
 --config profile="testing"

juju deploy -n 2 opensearch failover \
--config cluster_name="app" \
--config init_hold="true" \
--config roles='cluster_manager' \
--channel 2/edge \
--config profile="testing"

juju deploy -n 1 opensearch data-hot \
--config cluster_name="app" \
--config init_hold="true" \
--channel 2/edge \
--config roles='data.hot' \
--config profile="testing"

juju deploy self-signed-certificates --config ca-common-name="Test CA"
juju integrate self-signed-certificates main
juju integrate self-signed-certificates failover
juju integrate self-signed-certificates data-hot
juju integrate main:peer-cluster-orchestrator failover:peer-cluster
juju integrate main:peer-cluster-orchestrator data-hot:peer-cluster
juju integrate failover:peer-cluster-orchestrator data-hot:peer-cluster
```
2. Remove the main-data-hot relation *before* removing failover-data-hot relation:
```
juju remove-relation main data-hot
juju remove-relation failover data-hot
```

## Expected behavior
The data node should be in blocked state

## Actual behavior
The data node stays in active state

## Addional Context
This only happens if `juju remove-relation failover data-hot` is called directly after calling `juju remove-relation main data-hot`. If you wait until the units have settled to idle state before removing the failover relation, the data unit moves to an error state with `hook-failed: "opensearch-peers-relation-changed`. The debug-log for that case is below:

```
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-data-hot-0/charm/venv/lib/python3.10/site-packages/ops/model.py", line 3213, in _run
    result = subprocess.run(args, **kwargs)  # type: ignore
  File "/usr/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '('/var/lib/juju/tools/unit-data-hot-0/relation-set', '-r', '16', '--app', '--file', '-')' returned non-zero exit status 2.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-data-hot-0/charm/venv/lib/python3.10/site-packages/ops/model.py", line 3316, in relation_set
[...]

PR #593, opened 2025-03-14
[DPE-5490] [DPE-6680] [DPE-6694] [DPE-6763] [DPE-6764] Fix edge case failover app removal

### LINKED ISSUE #555: Large deployments - Peer-cluster-relation-departed hook fails when removing relation
## Steps to reproduce

1.  Deploy main/failover apps and integrate them
```
juju deploy -n 3 opensearch main \
 --config cluster_name="app" \
 --channel 2/edge \
 --config profile="testing"

juju deploy -n 3 opensearch failover \
--config cluster_name="app" \
--config init_hold="true" \
--channel 2/edge \
--config profile="testing"

juju deploy self-signed-certificates --config ca-common-name="Test CA"
juju integrate self-signed-certificates failover
juju integrate main:peer-cluster-orchestrator failover:peer-cluster
```
2. Remove the relation `juju remove-relation main failover`

## Expected behavior
The relation should be removed

## Actual behavior

The failover workload moves to an error state with message: `hook failed: "peer-cluster-relation-departed"`

## Log output

Juju debug log: 
```
unit-failover-1: 09:53:49 ERROR unit.failover/1.juju-log peer-cluster:8: Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-failover-1/charm/src/charm.py", line 221, in <module>
    main(OpenSearchOperatorCharm)
  File "/var/lib/juju/agents/unit-failover-1/charm/venv/lib/python3.10/site-packages/ops/main.py", line 553, in main
    manager.run()
  File "/var/lib/juju/agents/unit-failover-1/charm/venv/lib/python3.10/site-packages/ops/main.py", line 529, in run
    self._emit()
  File "/var/lib/juju/agents/unit-failover-1/charm/venv/lib/python3.10/site-packages/ops/main.py", line 518, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name, self._juju_context)
  File "/var/lib/juju/agents/unit-failover-1/charm/venv/lib/python3.10/site-packages/ops/main.py", line 139, in _emit_charm_event
  File "/var/lib/juju/agents/unit-failover-1/charm/venv/lib/python3.10/site-packages/ops/framework.py", line 347, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-failover-1/charm/venv/lib/python3.10/site-packages/ops/framework.py", line 853, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-failover-1/charm/venv/lib/python3.10/site-packages/ops/framework.py", line 943, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-failover-1/charm/lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py", line 763, in _on_peer_cluster_relation_departed
    not in [orchestrators.main_app.id, orchestrators.failover_app.id]
AttributeError: 'NoneType' object has no attribute 'id'
```

## Additional context

Both `main_app` and `failover_app` are None in the `orchestrators` dictionary

smethnani:
## Issue

When removing or scaling down an orchestrator in a large deployment or when removing the relations with an orchestrator, several parts of the code access properties of NoneType objects. Additionally, the failover promotes itself in cases where it shouldn't.

## Solution
- guard against accessing properties of NoneTypes
- stop the failover from promoting itself when it shouldn't
- treat scaling an orchestrator down to 0 units as deleting the orchestrator
- put peer-cluster requirer in a `Waiting` state if the main orchestrator is removed and a failover is related
- put peer-cluster requirer in a `Blocked` state if both the main and failover orchestrators are removed

This PR fixes #555 #564 #568 #569

Supersedes #440

### DIFF
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -64,7 +64,7 @@ PClusterWrongRelation = "Cluster name don't match with related cluster. Remove r
 PClusterWrongRolesProvided = "Cannot start cluster with current set of roles."
 PClusterNoDataNode = "Cannot run cluster with current roles. Waiting for data node..."
 PClusterWrongNodesCountForQuorum = (
-    "Less than 3 cluster-manager-eligible units in this cluster. Add more units."
+    "Even number of members in quorum if current unit started. Add or remove 1 unit."
 )
 PluginConfigError = "Unexpected error during plugin configuration, check the logs"
 BackupSetupFailed = "Backup setup failed, check logs for details"
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -64,7 +64,7 @@ PClusterWrongRelation = "Cluster name don't match with related cluster. Remove r
 PClusterWrongRolesProvided = "Cannot start cluster with current set of roles."
 PClusterNoDataNode = "Cannot run cluster with current roles. Waiting for data node..."
 PClusterWrongNodesCountForQuorum = (
-    "Even number of members in quorum if current unit started. Add or remove 1 unit."
+    "Less than 3 cluster-manager-eligible units in this cluster. Add more units."
 )
 PluginConfigError = "Unexpected error during plugin configuration, check the logs"
 BackupSetupFailed = "Backup setup failed, check logs for details"
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -57,11 +57,14 @@ DataRoleRemovalForbidden = (
     "Removal of data role from current deployment not allowed - the data cannot be reallocated."
 )
 PClusterNoRelation = "Cannot start. Waiting for peer cluster relation..."
+PClusterOrchestratorsRemoved = (
+    "Main-cluster-orchestrator removed, and no failover cluster related."
+)
 PClusterWrongRelation = "Cluster name don't match with related cluster. Remove relation."
 PClusterWrongRolesProvided = "Cannot start cluster with current set of roles."
 PClusterNoDataNode = "Cannot run cluster with current roles. Waiting for data node..."
 PClusterWrongNodesCountForQuorum = (
-    "Even number of members in quorum if current unit started. Add or remove 1 unit."
+    "Less than 3 cluster-manager-eligible units in this cluster. Add more units."
 )
 PluginConfigError = "Unexpected error during plugin configuration, check the logs"
 BackupSetupFailed = "Backup setup failed, check logs for details"
@@ -73,6 +76,9 @@ BackupRelUneligible = "Only orchestrator clusters should relate to backup relati
 # Wait status
 RequestUnitServiceOps = "Requesting lock on operation: {}"
 BackupDeferRelBrokenAsInProgress = "Backup service cannot be stopped: backup in progress."
+PClusterWaitingForFailoverPromotion = (
+    "Main-cluster-orchestrator removed, waiting for failover promotion."
+)
 
 
 # Maintenance statuses
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -106,7 +106,9 @@ class RelDepartureReason(BaseStrEnum):
     REL_BROKEN = "rel-broken"
 
 
-def relation_departure_reason(charm: CharmBase, relation_name: str) -> RelDepartureReason:
+def relation_departure_reason(
+    charm: CharmBase, relation_name: str, departing_app: str
+) -> RelDepartureReason:
     """Compute the reason behind a relation departed event."""
     # fetch relation info
     goal_state = charm.model._backend._run("goal-state", return_output=True, use_json=True)
@@ -116,7 +118,7 @@ def relation_departure_reason(charm: CharmBase, relation_name: str) -> RelDepart
     dying_units = [
         unit_data["status"] == "dying"
         for unit, unit_data in rel_info.items()
-        if unit != relation_name
+        if unit != departing_app and unit.split("/")[0] == departing_app
     ]
 
     # check if app removal
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -165,6 +165,10 @@ def all_units_names(charm: "OpenSearchBaseCharm") -> List[Unit]:
 
 def all_units(charm: "OpenSearchBaseCharm") -> List[Unit]:
     """Fetch the list of units for the current app."""
+    # if 0 planned units, return empty list
+    if charm.app.planned_units() == 0:
+        return []
+
     return list(charm.model.get_relation(PeerRelationName).units.union({charm.unit}))
 
 
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -165,7 +165,6 @@ def all_units_names(charm: "OpenSearchBaseCharm") -> List[Unit]:
 
 def all_units(charm: "OpenSearchBaseCharm") -> List[Unit]:
     """Fetch the list of units for the current app."""
-    # if 0 planned units, return empty list
     return list(charm.model.get_relation(PeerRelationName).units.union({charm.unit}))
 
 
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -166,9 +166,6 @@ def all_units_names(charm: "OpenSearchBaseCharm") -> List[Unit]:
 def all_units(charm: "OpenSearchBaseCharm") -> List[Unit]:
     """Fetch the list of units for the current app."""
     # if 0 planned units, return empty list
-    if charm.app.planned_units() == 0:
-        return []
-
     return list(charm.model.get_relation(PeerRelationName).units.union({charm.unit}))
 
 
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -494,6 +494,11 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
         if self.opensearch_peer_cm.is_provider():
             self.peer_cluster_provider.refresh_relation_data(event, can_defer=False)
 
+        # update any orchestrators about planned units
+        if self.opensearch_peer_cm.is_consumer():
+            self.peer_cluster_requirer.refresh_requirer_relation_data()
+            self.peer_cluster_requirer.apply_orchestrator_status()
+
         for relation in self.model.relations.get(ClientRelationName, []):
             self.opensearch_provider.update_endpoints(relation)
 
@@ -506,11 +511,6 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
         elif event.relation.data.get(event.app):
             # if app_data + app_data["nodes_config"]: Reconfigure + restart node on the unit
             self._reconfigure_and_restart_unit_if_needed()
-
-        # update any orchestrators about planned units
-        if self.opensearch_peer_cm.is_consumer():
-            self.peer_cluster_requirer.refresh_requirer_relation_data()
-            self.peer_cluster_requirer.apply_orchestrator_status()
 
         if not (unit_data := event.relation.data.get(event.unit)):
             return
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -555,7 +555,7 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             unit_name=format_unit_name(event.departing_unit.name, deployment_desc.app)
         )
 
-    def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # noqa: C901
+    def _on_opensearch_data_storage_detaching(self, event: StorageDetachingEvent):  # noqa: C901
         """Triggered when removing unit, Prior to the storage being detached."""
         if self.upgrade_in_progress:
             logger.warning(
@@ -577,9 +577,14 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
                     if node.name != self.unit_name
                 ]
                 self._compute_and_broadcast_updated_topology(remaining_nodes)
-            elif self.app.planned_units() == 0 and self.model.get_relation(PeerRelationName):
-                self.peers_data.delete(Scope.APP, "bootstrap_contributors_count")
-                self.peers_data.delete(Scope.APP, "nodes_config")
+            elif self.app.planned_units() == 0:
+                if self.model.get_relation(PeerRelationName):
+                    self.peers_data.delete(Scope.APP, "bootstrap_contributors_count")
+                    self.peers_data.delete(Scope.APP, "nodes_config")
+                if self.opensearch_peer_cm.is_provider():
+                    self.peer_cluster_provider.refresh_relation_data(event, can_defer=False)
+                if self.model.relations.get(PeerClusterRelationName):
+                    self.peer_cluster_requirer.refresh_relation_data()
 
         # we attempt to flush the translog to disk
         if self.opensearch.is_node_up():
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -1495,7 +1495,15 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
         ):
             return []
 
-        return ClusterTopology.nodes(self.opensearch, use_localhost, self.alt_hosts)
+        nodes = ClusterTopology.nodes(self.opensearch, use_localhost, self.alt_hosts)
+        if cluster_fleet_apps := self.peers_data.get_object(Scope.APP, "cluster_fleet_apps"):
+            has_planned_units = (
+                lambda node: node.app.id in cluster_fleet_apps
+                and cluster_fleet_apps[node.app.id]["planned_units"] != 0
+            )
+            nodes = [node for node in nodes if has_planned_units(node)]
+
+        return nodes
 
     def _set_node_conf(self, nodes: List[Node]) -> None:
         """Set the configuration of the current node / unit."""
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -77,7 +77,6 @@ from charms.opensearch.v0.opensearch_locking import OpenSearchNodeLock
 from charms.opensearch.v0.opensearch_nodes_exclusions import OpenSearchExclusions
 from charms.opensearch.v0.opensearch_peer_clusters import (
     OpenSearchPeerClustersManager,
-    OpenSearchProvidedRolesException,
     StartMode,
 )
 from charms.opensearch.v0.opensearch_performance_profile import OpenSearchPerformance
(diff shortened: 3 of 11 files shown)


lib/charms/opensearch/v0/constants_charm.py
```diff
@@ -57,11 +57,14 @@
...
 PClusterNoRelation = "Cannot start. Waiting for peer cluster relation..."
+PClusterOrchestratorsRemoved = (
+    "Main-cluster-orchestrator removed, and no failover cluster related."
+)
 PClusterWrongRelation = "Cluster name don't match with related cluster. Remove relation."
 PClusterWrongRolesProvided = "Cannot start cluster with current set of roles."
 PClusterNoDataNode = "Cannot run cluster with current roles. Waiting for data node..."
 PClusterWrongNodesCountForQuorum = (
-    "Even number of members in quorum if current unit started. Add or remove 1 unit."
+    "Less than 3 cluster-manager-eligible units in this cluster. Add more units."
```
phvalguima, line 67: AFAIR we support 1x unit deployments. I see this entry is only used in large deployments but I'd ask @Mehdi-Bendriss if we should support 1-manager large deployments as well in this case.

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -163,6 +165,10 @@ def all_units_names(charm: "OpenSearchBaseCharm") -> List[Unit]:
 
 def all_units(charm: "OpenSearchBaseCharm") -> List[Unit]:
     """Fetch the list of units for the current app."""
+    # if 0 planned units, return empty list
+    if charm.app.planned_units() == 0:
```
phvalguima: I'd recommend not changing this logic. We use the `all_units` in several key parts of the charm: locking mechanism, vote exclusions, etc. If we get into a situation where we have, say, 3x units of an app that is going away, they will not clean the voting exclusions correctly as they will receive `[]` [here](https://github.com/canonical/opensearch-operator/blob/79f53b2008c8601e29ca20812980c906149cfdd4/lib/charms/opensearch/v0/opensearch_nodes_exclusions.py#L147). And that behavior will only change the small deployments logic.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1019,9 +1058,6 @@ def _on_peer_cluster_relation_departed(self, event: RelationDepartedEvent):
             )
 
             rel_orchestrators.delete(event_src_cluster_type)
-            if failover_promoted:
```
phvalguima, line 1022: Why are we removing this? Where is the promotion happening now?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -584,7 +603,7 @@ def _fetch_local_cm_nodes(self, deployment_desc: DeploymentDescription) -> List[
             hosts=self.charm.alt_hosts,
         )
 
-        if not nodes:
+        if not nodes and self.charm.app.planned_units() != 0:
```
phvalguima, line 628: on `juju remove-application` I think this logic will now evaluate to `False` with this extra bit, even if we have X units around still being removed.

phvalguima (commented):
Thanks @smethnani, left some comments about your changes.

lib/charms/opensearch/v0/helper_charm.py
smethnani: Thanks, removed!

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: @phvalguima it's happening in a separate PR

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: @phvalguima can you point me to where I can read about this? Do you mean that on `juju remove-application` `planned_units()` can return something other than 0?

I don't think that will be an issue since the relation will be removed entirely once the application is removed, this is guarging against adding the unit in the relation data when scaling down to 0, so the unit is added before the unit is removed and that relation data remains since the application remains. Could `planned_units()` return something other than 0 if the application is scaled down to 0?

lib/charms/opensearch/v0/constants_charm.py
Mehdi-Bendriss: You're right, but I believe this flow is only applied post-deployment, which should work fine with 1 unit. I think this only applies on relation-departed of the main orchestrator when the overall fleet of the cluster has less than 3 CM nodes, given this is for large deployments we encourage users to scale up to not bring the whole fleet down

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -577,9 +577,14 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
...
-                self.peers_data.delete(Scope.APP, "bootstrap_contributors_count")
-                self.peers_data.delete(Scope.APP, "nodes_config")
+            elif self.app.planned_units() == 0:
+                if self.model.get_relation(PeerRelationName):
+                    self.peers_data.delete(Scope.APP, "bootstrap_contributors_count")
+                    self.peers_data.delete(Scope.APP, "nodes_config")
+                if self.opensearch_peer_cm.is_provider():
+                    self.peer_cluster_provider.refresh_relation_data(event, can_defer=False)
+                if self.model.relations.get(PeerClusterRelationName):
+                    self.peer_cluster_requirer.refresh_relation_data()
```
Mehdi-Bendriss: Given that `refresh_relation_data` is only useful for the providers, is this not already taken into account by the previous condition?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1490,7 +1495,15 @@ def _get_nodes(self, use_localhost: bool) -> List[Node]:
         ):
             return []
 
-        return ClusterTopology.nodes(self.opensearch, use_localhost, self.alt_hosts)
+        nodes = ClusterTopology.nodes(self.opensearch, use_localhost, self.alt_hosts)
+        if cluster_fleet_apps := self.peers_data.get_object(Scope.APP, "cluster_fleet_apps"):
```
Mehdi-Bendriss: Could you add a comment on why we're excluding departing units?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -336,10 +344,17 @@ def refresh_relation_data(
                 )
             else:
                 self.put_in_rel(data={"error_data": rel_data.to_str()}, rel_id=rel_id)
+                self._delete_rel_data_on_error(rel_id)
```
Mehdi-Bendriss: Why is this needed? isn't the `cluster_fleet_apps` always needed regardless of the status of the orchestrator?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -923,9 +951,13 @@ def set_security_index_initialised(self) -> None:
             rel_id=orchestrators.main_rel_id,
         )
 
-    def _put_current_app(
-        self, event: RelationEvent, deployment_desc: DeploymentDescription
-    ) -> None:
+    def refresh_relation_data(self) -> None:
```
Mehdi-Bendriss: We could perhaps change the name of this method and the one of the provider side to differentiate between the 2

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -967,36 +999,44 @@ def _on_peer_cluster_relation_departed(self, event: RelationDepartedEvent):
 
         # handle scale-down at the charm level storage detaching, or??
         if (
-            relation_departure_reason(self.charm, self.relation_name)
+            relation_departure_reason(
+                self.charm, self.relation_name, event.app.name if event.app else ""
```
Mehdi-Bendriss: when would `event.app` be None?

Mehdi-Bendriss (commented):
Thank you Shams! Looks good, I only have a few questions

lib/charms/opensearch/v0/opensearch_base_charm.py
smethnani: It's a new method added on the requirer side to update the `app` info in the relation data but perhaps I should rename it

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: I'm not really sure but the [docs](https://ops.readthedocs.io/en/latest/reference/ops.html#ops.RelationDepartedEvent) show that it can be `None`, [source](https://ops.readthedocs.io/en/latest/_modules/ops/charm.html#RelationEvent) suggests it's expected but could be the case it isn't provided

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: Ah I think this is missing a check for 0 planned units. It's for the case when relation data can be stale. Eg the app is scaled down to 0 units, the current state of cluster_fleet_apps is recorded, but later the cluster_fleet_apps is changed but not reflected in this relation data because the units are gone.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -529,7 +529,8 @@ def _on_peer_relation_departed(self, event: RelationDepartedEvent):
         remaining_nodes = [
             node
             for node in self._get_nodes(True)
-            if node.name != format_unit_name(event.departing_unit, app=current_app)
+            if node.app.id == current_app.id
+            and node.name != format_unit_name(event.departing_unit, app=current_app)
```
smethnani, line 533: This change addresses #613, so that it only considers remaining nodes from the current app

lib/charms/opensearch/v0/opensearch_base_charm.py
smethnani: This change has been removed altogether

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: This should now only delete the relation data when planned units are 0

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -241,18 +243,41 @@ def _on_peer_cluster_relation_departed(self, event: RelationDepartedEvent) -> No
...
         )
 
-    def refresh_relation_data(
+        # if the trigger app is the failover orchestrator and there are no planned units, delete it
+        if (
+            relation_departure_reason(
+                self.charm, self.relation_name, event.app.name if event.app else ""
+            )
+            == RelDepartureReason.SCALE_DOWN
+        ):
```
skourta, line 258: Usually the app should be set. If this is for typing errors you should raise instead of passing an empty string. By passing an empty string we are effectively setting up the function for undefined behaviour.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -241,18 +243,41 @@ def _on_peer_cluster_relation_departed(self, event: RelationDepartedEvent) -> No
...
+            for rel_id in all_relation_ids:
+                orchestrators = PeerClusterOrchestrators.from_dict(
+                    self.get_obj_from_rel("orchestrators", rel_id=rel_id)
+                )
+                orchestrators.delete("failover")
+                self.put_in_rel(data={"orchestrators": orchestrators.to_str()}, rel_id=rel_id)
+
+    def refresh_relation_data(  # noqa: C901
         self, event: EventBase, event_rel_id: int | None = None, can_defer: bool = True
     ) -> None:
```
skourta, line 274: Isn't this effectively fetching the **remote** app data then updating the **local** one? If so are the two synchronized somehow?

tests/integration/ha/test_large_deployments_remove_orchestrators.py
```diff
@@ -0,0 +1,172 @@
...
+        wait_for_exact_units={app: units for app, units in APP_UNITS.items()},
+        idle_period=IDLE_PERIOD,
+        timeout=1800,
+    )
+
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "xlarge"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_large_deployment_remove_app(ops_test: OpsTest) -> None:
```
skourta: I would rename the test to remove_orchestrators

skourta (commented):
Thank you @smethnani.
I left some minor comments. 
My main concern is having the failover promotion logic entirely removed. If it is going to be implemented in a future PR then perhaps the two PRs should be merged together?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: You're right and this whole block isn't needed since only the failover would have this entry in the relation data and there is no point in deleting and publishing it as the requirer will already delete it in its departed hook

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: You're right, I suppose it's better to allow it to crash if its empty for whatever reason

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: This has been removed now

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -577,9 +586,14 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
                     if node.name != self.unit_name
                 ]
                 self._compute_and_broadcast_updated_topology(remaining_nodes)
-            elif self.app.planned_units() == 0 and self.model.get_relation(PeerRelationName):
-                self.peers_data.delete(Scope.APP, "bootstrap_contributors_count")
-                self.peers_data.delete(Scope.APP, "nodes_config")
+            elif self.app.planned_units() == 0:
```
phvalguima, line 588: @smethnani I am guessing this is just a "guard rail"? B/c storage-detach runs as the very last event in the target units (the ones going away) and there is a very high chance the remote units already got the -broken/-departed events by now.

I do think we should keep this logic, but I think only on really busy clusters, where a bunch of events are running at once and remote units may not have received the -departed yet would be really make use of this logic.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
phvalguima: @smethnani the best place to get that type of confirmations is asking Juju team directly. I went over [their docs on `goal-state`](https://documentation.ubuntu.com/juju/3.6/reference/hook-command/list-of-hook-commands/goal-state/), which is the [actual command called at `planned_units`](https://github.com/canonical/operator/blob/23f482a02f821d2a357ce43a56ba5e80c8d19f2b/ops/model.py#L3783)

It is still vague for me, tbh what would happen.

phvalguima (commented):
Hi @smethnani, the PR is looking good and I liked the places you've added extra checks. I would recommend confirming ` goal-state` in Juju really returns what you expect it does before we can merge it.

smethnani:
> If it is going to be implemented in a future PR then perhaps the two PRs should be merged together?

The base is now a feature-branch so everything will be merged there first 👍

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -507,6 +507,11 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):
             # if app_data + app_data["nodes_config"]: Reconfigure + restart node on the unit
             self._reconfigure_and_restart_unit_if_needed()
 
+        # update any orchestrators about planned units
+        if self.opensearch_peer_cm.is_consumer():
```
Mehdi-Bendriss: could we move this a few lines above  below the `if self.opensearch_peer_cm.is_provider():` block with a comment, i think it makes more sense from a context standpoint

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -534,28 +549,22 @@ def _on_peer_relation_departed(self, event: RelationDepartedEvent):
 
         self.health.apply(wait_for_green_first=True)
 
-        if len(remaining_nodes) == self.app.planned_units():
+        if (
+            len([node for node in remaining_nodes if node.app.id == current_app.id])
```
Mehdi-Bendriss: nice!

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -357,6 +379,25 @@ def _notify_if_wrong_integration(
 
         return True
 
+    def _delete_rel_data(self, rel_id: int) -> None:
+        """Deletes relation data"""
+        self.delete_from_rel("cluster_fleet_apps", rel_id=rel_id)
+        self.delete_from_rel("data", rel_id=rel_id)
```
Mehdi-Bendriss, line 385: should we also delete `error_data` ?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -357,6 +379,25 @@ def _notify_if_wrong_integration(
...
+        self.delete_from_rel("rel_data_hash", rel_id=rel_id)
+
+    def _update_or_pop_from_fleet_dict(
+        self, fleet_dict: dict, app: PeerClusterApp, key: Optional[str] = None
+    ) -> None:
```
```suggestion
    def _update_fleet(
        self, fleet_dict: dict[str, dict[str, Any]], app: PeerClusterApp, key: Optional[str] = None
    ) -> None:
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
Mehdi-Bendriss: `planned_units` should return `0` on application removal - ref [here](https://github.com/canonical/operator/pull/936) :)

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -891,13 +968,13 @@ def _orchestrators(
         cm_relations = [
             rel.id
             for rel in self.model.relations[self.relation_name]
-            if rel.id != event.relation.id
+            if rel.id != event.relation.id and rel.units
```
Mehdi-Bendriss: let's be more explicit here when mixing boolean and implicit size checks, perhaps:
```suggestion
            if rel.id != event.relation.id and len(rel.units) > 0
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -891,13 +968,13 @@ def _orchestrators(
...
         ]
         for rel_id in cm_relations:
             remote_orchestrators.update(self.get_obj_from_rel(key="orchestrators", rel_id=rel_id))
 
         local_orchestrators = self.charm.peers_data.get_object(Scope.APP, "orchestrators") or {}
-        if trigger in {"main", "failover"}:
+        if trigger in {"main", "failover"} and event.relation.units:
```
```suggestion
        if trigger in {"main", "failover"} and len(event.relation.units) > 0:
```

Mehdi-Bendriss (commented):
Thank you Shams! great work, looks good! only small comments

lib/charms/opensearch/v0/opensearch_base_charm.py
smethnani: This was renamed to `refresh_requirer_relation_data`

lib/charms/opensearch/v0/opensearch_base_charm.py
smethnani: Moved up to just below the `is_provider` check

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: We decided against deleting this and addressing it in a future PR if we no longer need `error_data` on scale down to 0

Mehdi-Bendriss (approved):

skourta (approved):
Finished manual tests. I created an issue for the only problem I noticed but it is independent from this PR.  
Thank you @smethnani

PR #594, opened 2025-03-20
Change self-signed-certificates channel to latest/stable in tests

smethnani:
## Issue
CI tests failing when deploying `self-signed-certificates` using `stable` as the channel

## Solution
The recommendation is to follow the `1/stable` channel. This is currently causing errors so using `latest/stable` should be the solution for now.
Terraform should continue to use `latest/stable` for as we use a 22.04 base


Mehdi-Bendriss (dismissed):
Thank you Shams!

phvalguima (approved):

Mehdi-Bendriss (approved):

PR #595, opened 2025-03-21
pop random key if exists

smethnani:
## Issue
When loading a yaml file in the conf-setter, we're adding a random entry and immediately deleting it. If this line isn't correctly parsed for whatever reason, it will cause a crash.

## Solution
Add a guard to check that the random entry exists before deleting it

### DIFF
--- a/lib/charms/opensearch/v0/helper_conf_setter.py
+++ b/lib/charms/opensearch/v0/helper_conf_setter.py
@@ -195,7 +195,8 @@ class YamlConfigSetter(ConfigSetter):
             lines.append(f"{random_id}: {random_id}")
 
             data = self.yaml.load(StringIO("\n".join(lines)))
-            del data[random_id]
+            if random_id in data:
+                del data[random_id]
 
             return data
 


lib/charms/opensearch/v0/helper_conf_setter.py
```diff
@@ -195,7 +195,8 @@ def load(self, config_file: str) -> Dict[str, any]:
             lines.append(f"{random_id}: {random_id}")
 
             data = self.yaml.load(StringIO("\n".join(lines)))
-            del data[random_id]
+            if random_id in data:
+                del data[random_id]
```
phvalguima, line 199: Why do we need a key that gets deleted right after?

phvalguima (changes requested):
So, I think the real question here is why do we need a key that we create and then delete right after, based on a random string.

Going back to this change, I can see it was added [in this commit](https://github.com/canonical/opensearch-operator/commit/369bbae7433711173ee9c4086d4fa9a679cd5d77#diff-2a43d8614b3509c6d805d6e90003214ff6dbe59d08b069e2b1e3bb8058caf63a) several years ago. The intention was to process an empty yaml OR an yaml file that only has comments.

I installed the `ruamel.yaml` dependency that we use in our charm:
```
pip install ruamel-yaml==0.18.6
```

If we manually run that logic, we can see:
```
>>> import ruamel
>>> from io import StringIO
>>> from ruamel.yaml import YAML, CommentedSeq
>>> a=["# test", "# test2"]
>>> y=YAML()
>>> data = y.load(StringIO("\n".join(a)))
>>> data
>>> 
```

So, a file with only comments renders a `None` as data response.

If we follow the logic added by the commit, we can see:
```
>>> import uuid
>>> random_id = uuid.uuid4().hex
>>> a.append(f"{random_id}: {random_id}")
>>> data = y.load(StringIO("\n".join(a)))
>>> data
{'08d1accfd54441dab028529fc2e78dd1': '08d1accfd54441dab028529fc2e78dd1'}
>>> del data[random_id]
>>> data
{}
>>> 
```

With the `random_id` variable, now we return an empty dictionary instead. So, current `ruamel.yaml` dependency **can** process an empty yaml OR an yaml file with only comments. Now, the issue is that we want to always return a dictionary.

So, the best way would be to actually remove the entire `random_id` logic and instead return as follows:
```
            data = self.yaml.load(StringIO("\n".join(lines))) or {}
            return data
```

lib/charms/opensearch/v0/helper_conf_setter.py
phvalguima: A correction, there is a reason to keep the `random_id`: if we have comment-only file and we want to keep it this way, we need at least one entry to make it a valid yaml. Otherwise, it returns `None` at `load` call:
```
>>> import ruamel
>>> from io import StringIO
>>> from ruamel.yaml import YAML, CommentedSeq
>>> a=["# test", "# test2"]
>>> y=YAML()
>>> data = y.load(StringIO("\n".join(a)))
>>> data.ca
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'ca'
```

PR #601, opened 2025-03-27
[DPE-6858] Add expose support

phvalguima:
This PR adds expose as a variable and subsequent dynamic block. If the expose is set to true, then the dynamic block renders expose {}, which effectively exposes the application to any external access.

### DIFF
--- a/terraform/charm/simple_deployment/variables.tf
+++ b/terraform/charm/simple_deployment/variables.tf
@@ -79,7 +79,7 @@ variable "endpoint_bindings" {
 }
 
 variable "expose" {
-  description = "Expose the application"
+  description = "Expose the application for external access."
   type        = bool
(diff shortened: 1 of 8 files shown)


terraform/charm/simple_deployment/variables.tf
```diff
@@ -76,6 +78,12 @@ variable "endpoint_bindings" {
   default     = {}
 }
 
+variable "expose" {
+  description = "Expose the application"
```
Mehdi-Bendriss:
```suggestion
  description = "Expose the application for external access."
```

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

shayancanonical (approved):

PR #612, opened 2025-04-09
[DPE-6091] Add oauth integration

Zvirovyi:
# Overview

This PR adds OAuth integration to allow authorization in the OpenSearch using the clients from the [Canonical Identity Platform](https://charmhub.io/identity-platform).
By default, such clients aren't assigned to any permissions. In order to configure that, new `roles_mapping` config parameter was introduced. It allows to map IDP (Identity Provider) users to the users, that have been provided through `opensearch-client` (with inheriting their roles & permissions). Those mappings should be specified in the JSON dictionary format where keys are IDP clients ids and values are the corresponding `opensearch-client` users. If the specific mapping is wrong (mapped to the non-existent user), it will be ignored.

# Example of use

1. Deploy `opensearch` and `data-integrator` charms in the LXD model:
  ```bash
  juju deploy ./opensearch_ubuntu@24.04-amd64.charm --config profile=testing -n2
  juju deploy data-integrator --config index-name=admin-index --config extra-user-roles="admin"
  ```
2. Deploy Identity Platform in the K8s model:
  ```bash
  juju deploy identity-platform --trust --channel edge
  ```
3. Setup all the relations between `opensearch`, `data-integrator` and `identity-platform`:
  ```bash
  # K8s model
  juju offer self-signed-certificates:certificates certificates
  juju offer hydra:oauth oauth

  # LXD model
  juju consume admin/iam.certificates
  juju consume admin/iam.oauth
  juju integrate opensearch certificates
  juju integrate opensearch oauth
  juju integrate opensearch data-integrator
  ```
4. Create IDP user and acquire access token:
  ```bash
  juju run hydra/0 create-oauth-client --quiet scope="[openid,profile,email,phone,offline]" grant-types="[client_credentials]" audience="[opensearch]"
  # client-id: 960c7728-fff9-4500-8470-9978d48e3a98
  # client-secret: -.qC9yOCmiEt~Ct8kCI3C-nNFc
  
  # traefik-public address
  curl https://10.1.142.144/dev-uk8s-hydra/oauth2/token -k -d "scope=openid" -d "grant_type=client_credentials" -d "audience=opensearch" -u 960c7728-fff9-4500-8470-9978d48e3a98:-.qC9yOCmiEt~Ct8kCI3C-nNFc
  # "access_token":"ey...sc"
  ```
5. Configure role mapping:
  ```bash
  juju run data-integrator/leader get-credentials
  # username: opensearch-client_6

  juju config opensearch roles_mapping='{"960c7728-fff9-4500-8470-9978d48e3a98": "opensearch-client_6"}'
  ```
6. Test OAuth user with the access token acquired previously:
  ```bash
  # opensearch address
  curl -k -X GET https://192.168.2.163:9200/_cat/indices -H "Authorization: Bearer ey...sc"
  ```

# Key notes

* Both `opensearch` and `hydra` (`identity-platform`) should use the same certificates (correspondingly from the single `self-signed-certificates`) otherwise an error will occur on every request from the OAuth user, that will lead to unauthorized responses;
* there is the [same PR for the opensearch-dashboards charm](https://github.com/canonical/opensearch-dashboards-operator/pull/180).


lib/charms/opensearch/v0/opensearch_oauth.py
```diff
@@ -0,0 +1,118 @@
...
+                admin_secrets, self.charm.opensearch_config.SECURITY_CONFIG_YML
+            )
+        except OpenSearchCmdError as e:
+            logger.debug(f"Error when updating the security index: {e.out}")
+            event.defer()
+            return
+
+    def _on_oauth_relation_departed(self, event: RelationDepartedEvent) -> None:
+        if event.departing_unit == self.charm.unit and self.charm.peers_data is not None:
+            self.charm.peers_data.put(Scope.UNIT, "departing", True)
```
Zvirovyi: Will be there any sense to differentiate departing status by relation such as by `_depart_flag` and `_unit_departing` in `opensearch_relation_provider.py`?
@zmraul

zmraul: In the case of Hydra, is a single unit deployment afaik. So I don't think you need to differentiate between `relation_departed` and `relation_broken`. I would say you can remove the `departing` key and just run the `relation_broken` normally. In this case we don't need anything related to the oauth databag, so it should be fine.

Zvirovyi: But this check also filters out removal of the opensearch leader unit (because there will be `broken` event even if relation remains afterwards)

zmraul: You are right sorry. To your question, I don't think there is much benefit to split them, but I would still add some identifier for the module that added the flag (in this case oauth). It can let us debug issues with the charm, because we can specifically track which module failed to handle relation_departed based on what remains on the peer databag

lib/charms/opensearch/v0/opensearch_oauth.py
```diff
@@ -0,0 +1,119 @@
...
+            logger.debug("Oauth relation not yet set up")
+            return
+
+        if not self.charm.peers_data.get(Scope.APP, "security_index_initialised") or (
+            "data" not in self.charm.opensearch_peer_cm.deployment_desc().config.roles
+            and self.charm.opensearch_peer_cm.deployment_desc().start
+            != StartMode.WITH_GENERATED_ROLES
+        ):
+            logger.debug("Deferring oauth relation changed event as cluster is not ready yet")
+            event.defer()
```
Mehdi-Bendriss, line 63: Missing return I think

lib/charms/opensearch/v0/opensearch_oauth.py
```diff
@@ -0,0 +1,119 @@
...
+
+        relation = self.model.get_relation(OAUTH_RELATION)
+        if not relation.data[relation.app]:
+            logger.debug("Oauth relation not yet set up")
+            return
+
+        if not self.charm.peers_data.get(Scope.APP, "security_index_initialised") or (
+            "data" not in self.charm.opensearch_peer_cm.deployment_desc().config.roles
+            and self.charm.opensearch_peer_cm.deployment_desc().start
+            != StartMode.WITH_GENERATED_ROLES
```
Mehdi-Bendriss: I believe this is covering well the simple deployment part, however given we also support a large deployment where multiple opensearch apps are related to each others we need to ensure this logic only happens once:

(feel free to simplify the logic)
```
deployment_desc = self.charm.opensearch_peer_cm.deployment_desc()
is_small_deployment_ready = (
    deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR
    and (
        deployment_desc.start == StartMode.WITH_GENERATED_ROLES
        or "data" not in deployment_desc.config.roles
    )
)
is_main_orchestrator = self.charm.opensearch_peer_cm.is_provider(typ="main")

if (
    not deployment_desc
    or not self.charm.peers_data.get(Scope.APP, "security_index_initialised") 
    or not is_small_deployment_ready 
    or not is_main_orchestrator
):
```

lib/charms/opensearch/v0/opensearch_oauth.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+        except OpenSearchCmdError as e:
+            logger.debug(f"Error when updating the security index: {e.out}")
+            event.defer()
+            return
+
+    def _on_oauth_relation_departed(self, event: RelationDepartedEvent) -> None:
+        if event.departing_unit == self.charm.unit and self.charm.peers_data is not None:
+            self.charm.peers_data.put(Scope.UNIT, "departing_oauth", True)
+        pass
```
Mehdi-Bendriss:
```suggestion
```

lib/charms/opensearch/v0/opensearch_oauth.py
```diff
@@ -0,0 +1,119 @@
...
+        if (
+            not self.charm.unit.is_leader()
+            or self.charm.peers_data is None
+            or self.charm.peers_data.get(Scope.UNIT, "departing_oauth")
+            or not self.charm.peers_data.get(Scope.APP, "security_index_initialised")
+            or (
+                "data" not in self.charm.opensearch_peer_cm.deployment_desc().config.roles
+                and self.charm.opensearch_peer_cm.deployment_desc().start
+                != StartMode.WITH_GENERATED_ROLES
+            )
```
Mehdi-Bendriss: Similar comment as above

lib/charms/opensearch/v0/opensearch_oauth.py
```diff
@@ -0,0 +1,119 @@
...
+            "data" not in self.charm.opensearch_peer_cm.deployment_desc().config.roles
+            and self.charm.opensearch_peer_cm.deployment_desc().start
+            != StartMode.WITH_GENERATED_ROLES
+        ):
+            logger.debug("Deferring oauth relation changed event as cluster is not ready yet")
+            event.defer()
+
+        self.charm.opensearch_config.add_oidc_auth(
+            openid_connect_url=f"{relation.data[relation.app].get('issuer_url')}/.well-known/openid-configuration"
+        )
```
Mehdi-Bendriss, line 68: I think this should be added to all units in a cluster - not just the leader, in case say the leader unit gets destroyed later on etc.., right?

lib/charms/opensearch/v0/opensearch_oauth.py
```diff
@@ -0,0 +1,119 @@
...
+            or not self.charm.peers_data.get(Scope.APP, "security_index_initialised")
+            or (
+                "data" not in self.charm.opensearch_peer_cm.deployment_desc().config.roles
+                and self.charm.opensearch_peer_cm.deployment_desc().start
+                != StartMode.WITH_GENERATED_ROLES
+            )
+        ):
+            return
+
+        self.charm.opensearch_config.remove_oidc_auth()
```
Mehdi-Bendriss, line 97: Similar comment as above, i think this should run in all units

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -520,6 +523,43 @@ def remove_lingering_relation_users_and_roles(  # noqa: C901
...
                 del relation_users[rel_id]
 
         self.charm.peers_data.put_object(Scope.APP, ClientUsersDict, relation_users)
+
+    def update_relations_roles_mapping(self) -> None:
+        """Updates all the relations roles mapping due to config change."""
+        if (
+            not self.unit.is_leader()
+            or not self.opensearch.is_node_up()
+            or not self.charm.peers_data.get(Scope.APP, "security_index_initialised", False)
```
Mehdi-Bendriss: this should be implied by the previous condition `is_node_up`

lib/charms/opensearch/v0/opensearch_relation_provider.py
```diff
@@ -520,6 +523,43 @@ def remove_lingering_relation_users_and_roles(  # noqa: C901
...
+        ):
+            return
```
Mehdi-Bendriss: I think we should not fail silently here, we need to retry the operation, by raising an exception and catching it on_config changed and deferring in that case

tests/integration/relations/test_oauth.py
Mehdi-Bendriss, line 1: Could you add Docstrings in all test functions as well as inline comments to describe the flow?

tests/integration/relations/test_oauth.py
```diff
@@ -0,0 +1,253 @@
...
+    config_with_roles = config_without_roles.copy()
+    config_with_roles["roles_mapping"] = json.dumps({oauth_client_id: data_integrator_user})
+    await ops_test.model.applications["opensearch"].set_config(config_with_roles)
+    await ops_test.model.wait_for_idle(status="active")
+
+    result = requests.get(
+        opensearch_url, headers={"Authorization": f"Bearer {oauth_access_token}"}, verify=False
+    )
+    assert result.status_code == 200, "request expected to succeed with roles mapping"
+
```
Mehdi-Bendriss, line 159: Could we add the lines below this one in a new test, which should come last. And add a test where we deploy a new data-integrator instance, ensuring that the new user can also authenticate to opensearch and then deleting the first user, ensuring it got revoked while the second user is still active
Then at last, clearing the full config as you did below

Mehdi-Bendriss (changes requested):
Thank you Vladyslav, great work! I left some comments

lib/charms/opensearch/v0/opensearch_oauth.py
Zvirovyi: Done

lib/charms/opensearch/v0/opensearch_oauth.py
Zvirovyi: Done. Now config updates on all units, but applies only on the leader

lib/charms/opensearch/v0/opensearch_oauth.py
Zvirovyi: Done

lib/charms/opensearch/v0/opensearch_oauth.py
Zvirovyi: Done

lib/charms/opensearch/v0/opensearch_oauth.py
Zvirovyi: Done. Now config updates on all units, but applies only on the leader

lib/charms/opensearch/v0/opensearch_relation_provider.py
Zvirovyi: Done

lib/charms/opensearch/v0/opensearch_relation_provider.py
Zvirovyi: Done, although I did that with bool indicator

tests/integration/relations/test_oauth.py
Zvirovyi: Done. I did not put any inline comments as doc strings happened to describe all nicely

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -761,6 +763,9 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
             self.config.get(PERFORMANCE_PROFILE)
         )
 
+        if not self.opensearch_provider.update_relations_roles_mapping():
+            event.defer()
```
zmraul, line 767: Should this return here?

lib/charms/opensearch/v0/opensearch_users.py
```diff
@@ -244,6 +245,59 @@ def patch_user(self, user_name: str, patches: List[Dict[str, any]]) -> Dict[str,
...
+            mapped_users: all the users, that should be mapped to the specified role.
+
+        Raises:
+            OpenSearchUserMgmtError: If the request fails.
+        """
+        try:
+            resp = self.opensearch.request(
+                "PUT",
+                f"{ROLESMAPPING_ENDPOINT}/{role}",
+                payload={"users": mapped_users, "backend_roles": [role]},
```
zmraul, line 262: Is passing an empty list here and a backend role valid format? I noticed the output from `_get_relation_mapped_users` can be `[]`

tests/integration/relations/test_oauth.py
```diff
@@ -0,0 +1,279 @@
...
+DATA_INTEGRATOR_CONFIG = {
+    "index-name": "admin-index",
+    "extra-user-roles": "admin",
+}
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="module")
+async def microk8s_cloud(ops_test: OpsTest) -> AsyncGenerator[None, Any]:
```
zmraul: Why not have these fixtures on `conftest.py`

tests/unit/lib/test_opensearch_relation_provider.py
```diff
@@ -473,6 +499,7 @@ def test_avoid_removing_non_charmed_users_and_roles(
         mock_response_root(self.charm.unit_name, self.charm.opensearch.host)
         mock_response_nodes(self.charm.unit_name, self.charm.opensearch.host)
 
+        mock_get_relation_mapped_users.return_value = None
```
zmraul: Return value should be `[]` to match the mocked method no?

zmraul (changes requested):
Awesome thanks! Left a couple of comments

edit: aren't the oauth integration tests missing on the github CI?

lib/charms/opensearch/v0/opensearch_users.py
Zvirovyi: Yes - in such cases it will clear mappings from the role.

lib/charms/opensearch/v0/opensearch_base_charm.py
Zvirovyi: Other check right above also defers event without returning. I thought, that this approach was intentional in this part.

lib/charms/opensearch/v0/opensearch_oauth.py
Zvirovyi: Done

tests/integration/relations/test_oauth.py
Zvirovyi: Done

tests/integration/relations/test_oauth.py
Zvirovyi: Done

tests/unit/lib/test_opensearch_relation_provider.py
Zvirovyi: Done

lib/charms/opensearch/v0/opensearch_base_charm.py
zmraul: I think the idea here is to defer because the restart conditions are not yet met. I'm not sure why the previous part is not returned, at least in the `else` condition where the statuses are cleared.

@Mehdi-Bendriss in the block above, if the keystore is not ready for example, what's the point of letting the rest of the method run? I would expect that nothing happens, no restart. So might as well return early no?

lib/charms/opensearch/v0/opensearch_users.py
zmraul: Thanks

Mehdi-Bendriss (approved):

lib/charms/opensearch/v0/opensearch_base_charm.py
Mehdi-Bendriss: Hmm, I agree with the concern, but the user management doesn't warrant a restart, so it can be independent from the hooks "restart-centered" logic, I think it doesn't hurt to not have a return here

zmraul (approved):

PR #615, opened 2025-04-14
[DPE-7017] Rework validate_roles logic

smethnani:
## Issue
Validating roles previously required an odd number of cm-eligible nodes. Opensearch can intelligently handle maintaining quorum as long as there are at least 3 cm-eligible nodes in the cluster.

We would like to allow any configuration at deployment time and warn the user when removing cm-elibile units if a large deployment has less than the 3 recommended nodes.

## Solution
Rework validation logic to count the number of cm-eligible nodes on relation-departure and set a blocked status if there are less than 3.

## Checklist


lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -394,46 +393,69 @@ def demote_to_failover_orchestrator(self) -> None:
...
 
-    def validate_roles(self, nodes: List[Node], on_new_unit: bool = False) -> None:
-        """Validate full-cluster wide the quorum for CM/voting_only nodes on services start."""
+    def validate_roles(self, nodes: List[Node]) -> None:
+        """Validate cluster-wide count for CM-eligible nodes is at least 3"""
         deployment_desc = self.deployment_desc()
-        if not set(deployment_desc.config.roles) & {"cluster_manager", "voting_only"}:
+        if (
+            not set(deployment_desc.config.roles) & {"cluster_manager", "voting_only"}
+            and not deployment_desc.start == StartMode.WITH_GENERATED_ROLES
```
skourta: I belive `!=` is a lot easier to read and understand.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -394,46 +393,69 @@ def demote_to_failover_orchestrator(self) -> None:
...
             return
 
         raise OpenSearchProvidedRolesException(PClusterWrongNodesCountForQuorum)
 
+    def validate_recommended_cm_unit_count(
+        self, nodes: Optional[List[Node]] = None, only_validate_if_blocked: bool = False
+    ) -> None:
+        """Sets blocked status if validate_roles fails"""
+        if (
+            only_validate_if_blocked
```
skourta: The `only_validate_if_blocked` argument is never used.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -394,46 +393,69 @@ def demote_to_failover_orchestrator(self) -> None:
             Scope.APP, "deployment-description", deployment_desc.to_dict()
         )
 
-    def validate_roles(self, nodes: List[Node], on_new_unit: bool = False) -> None:
-        """Validate full-cluster wide the quorum for CM/voting_only nodes on services start."""
+    def validate_roles(self, nodes: List[Node]) -> None:
```
skourta: I think it is better to have the funciton return a boolean rather than `None` or `raise`. It would also simplify its call.

tests/integration/ha/test_roles_managements.py
```diff
@@ -113,30 +111,6 @@ async def test_set_roles_manually(
         assert sorted(node.roles) == ["cluster_manager", "data"], "roles unchanged"
         assert node.temperature == "cold", "Temperature unchanged."
 
-    # scale up cluster by 1 unit, this should give the new node the same roles
```
skourta, line 116: Maybe we should add a test case for the new workflow?

skourta (commented):
Thank you @smethnani . I left some comments.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
smethnani: It's used [here](https://github.com/canonical/opensearch-operator/blob/242a7ed91404dbef1d7c7b5540042e9d92f24a84/lib/charms/opensearch/v0/opensearch_base_charm.py#L502)

lib/charms/opensearch/v0/opensearch_peer_clusters.py
smethnani: Good call, thanks!

tests/integration/ha/test_roles_managements.py
smethnani: I believe the tests here are checking actual roles management as opposed to validation. There is a unit test [here](https://github.com/canonical/opensearch-operator/blob/242a7ed91404dbef1d7c7b5540042e9d92f24a84/tests/unit/lib/test_opensearch_peer_clusters.py#L121) to check the new validate roles logic

lib/charms/opensearch/v0/opensearch_peer_clusters.py
smethnani: If no objections @Mehdi-Bendriss then happy to change it

smethnani: This now returns a boolean, I changed the name to be more descriptive

tests/integration/ha/test_roles_managements.py
smethnani: I added an integration test [here](https://github.com/canonical/opensearch-operator/pull/615/files#diff-89f336c69a1a84d13f81a27201834f76b532e82cf91c998e85c91459c68023ed)

tests/integration/ha/test_large_deployments_validate_cm_count.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.mark.runner(["self-hosted", "linux", "X64", "jammy", "xlarge"])
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+@pytest.mark.skip_if_deployed
+async def test_build_and_deploy(ops_test: OpsTest) -> None:
+    """Build and deploy one unit of OpenSearch."""
+    # it is possible for users to provide their own cluster for HA testing.
+    # Hence, check if there is a pre-existing cluster.
+    my_charm = await ops_test.build_charm(".")
```
skourta, line 38: We are moving away from `build_charm` depending on when the dual base support will be merged (Should be today or early next week) you will need to rebase and refactor this.

skourta (approved):
Thank you @smethnani .
Great work. 
I ran some manual tests:
- Start with 3 nodes of a main cm app, a data app with 1 unit and then scale down to 2 units of main -> warning appears.
- Add a unit to main -> warning disappears
- Remove two units of main -> warning appears.
- Add a failover app with cm role with 2 units -> warning disappears.

Everything looks in order.

I have a non-blocking inquiry: If the cluster starts with only 2 initial CMs we currently do not show the warning (I believe this is by design). My opinion is to show it on the initial deployment as well.

tests/integration/ha/test_large_deployments_validate_cm_count.py
Mehdi-Bendriss: Let's do it when merging the feature branch to `2/edge`

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -394,45 +393,54 @@ def demote_to_failover_orchestrator(self) -> None:
...
+            # validation is only needed in large deployments
+            # we can start with any number of cms but set a blocked status
+            # if CMS < 3 when a CM-eligible application is scaled down or removed
+            return True
+
+        logger.info("Less than 3 CM-eligible units in this cluster")
+        return False
+
+    def validate_recommended_cm_unit_count(
+        self, nodes: Optional[List[Node]] = None, only_validate_if_blocked: bool = False
```
Mehdi-Bendriss: It is not very clear what is meant by `only_validate_if_blocked`, could you please rename the argument?

Mehdi-Bendriss (commented):
Thanks Shams! 1 naming related comment, but looks good

Mehdi-Bendriss (approved):

skourta (approved):

PR #619, opened 2025-04-22
[DPE-6876] - Dual base support and migration to spread

skourta:
This pull request runs integration tests using spread.

### CI/CD Workflow Updates:
* Updated `.github/workflows/ci.yaml` to use newer versions of the `lint.yaml` and `build_charm.yaml` workflows (`v31.0.1`) and switched runners from `ubuntu-22.04` to `ubuntu-latest`. [[1]](diffhunk://#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddL23-R27) [[2]](diffhunk://#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddL143-R143) [[3]](diffhunk://#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddL167-L169)
* Replaced the `integration_test_charm.yaml` workflow with a custom `.github/workflows/integration_test.yaml` workflow, inheriting secrets and simplifying permissions.
* Added a new `integration_test.yaml` workflow for integration testing, which dynamically collects and runs spread jobs, generates Allure reports, and uploads test results.

### Dependency Management:
* Removed several custom plugins (`pytest-github-secrets`, `pytest-operator-cache`, `pytest-operator-groups`, `allure-pytest-collection-report`) and replaced them with standardized or updated dependencies (e.g., `allure-pytest-default-results`). [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L73-L77) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L87-R84)

### Documentation Improvements:
* Simplified `CONTRIBUTING.md` by updating instructions for installing dependencies and running tests. Replaced `tox` commands with `charmcraft` for integration testing and removed outdated sections about Juju versions and backup testing. [[1]](diffhunk://#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R24-R43) [[2]](diffhunk://#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055L68-L120)

### Functional Changes:
* Updated `get_host_public_ip` in `helper_networking.py` to use `unit-get public-address` instead of an external DNS query for fetching the public IP.

### Miscellaneous:
* Added a new `concierge.yaml` configuration file for Juju model defaults and provider settings.


.github/workflows/integration_test.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
               _, runner, task, variant = job.split(":")
               # Example: "test_charm.py"
               task = task.removeprefix("tests/spread/")
+              task = task.replace("/", "_")
```
carlcsaposs-canonical: https://github.com/canonical/mongo-single-kernel-library/pull/24#discussion_r2028320984

.github/workflows/integration_test.yaml
```diff
@@ -84,17 +85,17 @@ jobs:
...
-          df --human-readable
+      # - name: Free up disk space
+      #   timeout-minutes: 1
+      #   run: |
+      #     printf '\nDisk usage before cleanup\n'
+      #     df --human-readable
+      #     # Based on https://github.com/actions/runner-images/issues/2840#issuecomment-790492173
+      #     rm -r /opt/hostedtoolcache/
+      #     printf '\nDisk usage after cleanup\n'
+      #     df --human-readable
```
carlcsaposs-canonical:
```suggestion
      - name: Disk usage
        timeout-minutes: 1
        run: df --human-readable
```
from https://github.com/canonical/data-platform-workflows/blob/8d0d111687006fc6653b5b6548e9563bb217872d/.github/workflows/integration_test_charm.yaml#L223

poetry.lock
carlcsaposs-canonical, line 1: large diff—is this expected? suggest considering updating unrelated deps in separate pr to minimize changes

pyproject.toml
```diff
@@ -63,22 +63,16 @@ ops = { version = ">=2.19.0", extras = ["testing"] }
...
-# TODO remove
-pytest-github-secrets = {git = "https://github.com/canonical/data-platform-workflows", tag = "v29.0.0", subdirectory = "python/pytest_plugins/github_secrets"}
 pytest-asyncio = "^0.21.2"
 pytest-operator = "^0.35.0"
-# TODO remove
-pytest-operator-cache = {git = "https://github.com/canonical/data-platform-workflows", tag = "v29.0.0", subdirectory = "python/pytest_plugins/pytest_operator_cache"}
-# TODO remove
-pytest-operator-groups = {git = "https://github.com/canonical/data-platform-workflows", tag = "v29.0.0", subdirectory = "python/pytest_plugins/pytest_operator_groups"}
-pytest-microceph = {git = "https://github.com/canonical/data-platform-workflows", tag = "v29.0.0", subdirectory = "python/pytest_plugins/microceph"}
+pytest-microceph = { git = "https://github.com/canonical/data-platform-workflows", tag = "v29.0.0", subdirectory = "python/pytest_plugins/microceph" }
```
carlcsaposs-canonical, line 75: can the microceph plugin be removed?
see
https://github.com/canonical/data-platform-workflows/blob/main/.github/workflows/integration_test_charm_deprecation_notice.md
https://github.com/canonical/mysql-operator/pull/609

spread.yaml
```diff
@@ -74,31 +74,50 @@ backends:
...
+      sudo rm -f /etc/ssh/sshd_config.d/60-cloudimg-settings.conf
+      sudo systemctl daemon-reload
+      sudo systemctl restart ssh
 
-      ADDRESS localhost
+      sudo useradd spread -s /bin/bash -m || true
+      echo "spread:$SPREAD_PASSWORD" | sudo chpasswd || true
+      echo 'spread ALL=(ALL) NOPASSWD:ALL ' | sudo tee /etc/sudoers.d/99-spread-user || true
+
+      ADDRESS "127.0.0.1"
```
carlcsaposs-canonical: why was this changed from initial commits?

spread.yaml
```diff
@@ -74,31 +74,50 @@ backends:
...
+      CHARM_UBUNTU_BASE: "$(HOST: echo $CHARM_UBUNTU_BASE)"
+      AWS_ACCESS_KEY: "$(HOST: echo $AWS_ACCESS_KEY)"
+      AWS_SECRET_KEY: "$(HOST: echo $AWS_SECRET_KEY)"
+      AZURE_STORAGE_ACCOUNT: "$(HOST: echo $AZURE_STORAGE_ACCOUNT)"
+      AZURE_SECRET_KEY: "$(HOST: echo $AZURE_SECRET_KEY)"
+      GCP_ACCESS_KEY: "$(HOST: echo $GCP_ACCESS_KEY)"
+      GCP_SECRET_KEY: "$(HOST: echo $GCP_SECRET_KEY)"
+      GCP_SERVICE_ACCOUNT: "$(HOST: echo $GCP_SERVICE_ACCOUNT)"
+      CONCIERGE_EXTRA_SNAPS: charmcraft
+      CONCIERGE_EXTRA_DEBS: pipx
```
carlcsaposs-canonical:
```suggestion
```
appears to be mistakenly duplicated

spread.yaml
```diff
@@ -74,31 +74,50 @@ backends:
...
-      CI: '$(HOST: echo $CI)'
-      CHARM_UBUNTU_BASE: '$(HOST: echo $CHARM_UBUNTU_BASE)'
-      AWS_ACCESS_KEY: '$(HOST: echo $AWS_ACCESS_KEY)'
-      AWS_SECRET_KEY: '$(HOST: echo $AWS_SECRET_KEY)'
-      AZURE_STORAGE_ACCOUNT: '$(HOST: echo $AZURE_STORAGE_ACCOUNT)'
-      AZURE_SECRET_KEY: '$(HOST: echo $AZURE_SECRET_KEY)'
-      GCP_ACCESS_KEY: '$(HOST: echo $GCP_ACCESS_KEY)'
-      GCP_SECRET_KEY: '$(HOST: echo $GCP_SECRET_KEY)'
-      GCP_SERVICE_ACCOUNT: '$(HOST: echo $GCP_SERVICE_ACCOUNT)'
+      CI: "$(HOST: echo $CI)"
```
carlcsaposs-canonical:
```suggestion
      CI: '$(HOST: echo $CI)'
```
for consistency/to minimize diff with sql repos (please apply to all env variables)

spread.yaml
```diff
@@ -74,31 +74,50 @@ backends:
...
+      GCP_SERVICE_ACCOUNT: "$(HOST: echo $GCP_SERVICE_ACCOUNT)"
+      CONCIERGE_EXTRA_SNAPS: charmcraft
+      CONCIERGE_EXTRA_DEBS: pipx
     systems:
-      - ubuntu-24.04:
-          username: runner
+      - self-hosted-linux-amd64-noble-large:
+          username: spread
+      - self-hosted-linux-amd64-noble-xlarge:
+          username: spread
```
carlcsaposs-canonical: why was username changed from sql repos?

spread.yaml
```diff
@@ -74,31 +74,50 @@ backends:
...
+  tests/spread/test_backups.py/:
+    summary: Backup test suite
+  tests/spread/test_ca_rotation.py/:
+    summary: CA Rotation test suite
+  tests/spread/test_large_deployments_inherit_cluster_name.py/:
+    summary: Large deployments inherit cluster name test suite
+  tests/spread/test_plugins.py/:
+    summary: Plugins test suite
+  tests/spread/test_small_deployment_upgrades.py/:
+    summary: Small deployment upgrades test suite
```
carlcsaposs-canonical: question: why separate suites instead of a single suite?

spread.yaml
```diff
@@ -115,9 +134,10 @@ prepare: |
   snap install --classic concierge
 
   # Install charmcraft & pipx (on lxd-vm backend)
-  concierge prepare --trace
+  concierge prepare --trace --extra-debs pipx
```
carlcsaposs-canonical:
```suggestion
  concierge prepare --trace
```

spread.yaml
```diff
@@ -115,9 +134,10 @@ prepare: |
...
 
-  pipx install tox poetry
+  pipx install tox 
+  pipx install poetry
```
carlcsaposs-canonical: why was this changed?

spread.yaml
```diff
@@ -127,9 +147,10 @@ prepare-each: |
     poetry add --lock --group integration juju@^2
   fi
   # `concierge prepare` needs to be run for each spread job in case Juju version changed
-  concierge prepare --trace
+  concierge prepare --trace --extra-debs pipx
```
carlcsaposs-canonical:
```suggestion
  concierge prepare --trace
```

spread.yaml
```diff
@@ -127,9 +147,10 @@ prepare-each: |
...
 
   # Unable to set constraint on all models because of Juju bug:
   # https://bugs.launchpad.net/juju/+bug/2065050
   juju set-model-constraints arch="$(dpkg --print-architecture)"
+
```
carlcsaposs-canonical, line 145:
```suggestion
```

tests/integration/conftest.py
```diff
@@ -1,12 +1,27 @@
+#!/usr/bin/env python3
```
carlcsaposs-canonical:
```suggestion
```
this file isn't executed directly; no need for shebang

tests/integration/conftest.py
```diff
@@ -1,12 +1,27 @@
...
+        return "jammy"
+    elif ubuntu_base == "24.04":
+        return "noble"
+    else:
+        raise NotImplementedError
+
+
 @pytest.fixture
 def charm(ubuntu_base):
     # TODO: use instead of ops_test.build_charm
```
carlcsaposs-canonical: is this still TODO?

tests/integration/ha/test_backups.py
```diff
@@ -72,16 +72,7 @@
...
-            pytest.mark.runner(
-                [
-                    "self-hosted",
-                    "linux",
-                    "X64",
-                    "jammy",
-                    "xlarge" if deploy_type == "large" else "large",
-                ]
-            ),
+            pytest.mark.group(id=f"{cloud_name}-{deploy_type}"),
```
carlcsaposs-canonical, line 75: question: why is group marker still used? given that pytest-operator-groups is deprecated & charmcraft test is the replacement

tests/integration/plugins/test_plugins.py
```diff
@@ -55,16 +54,7 @@
...
-                    "linux",
-                    "X64",
-                    "jammy",
-                    "xlarge" if deploy_type == "large" else "large",
-                ]
-            ),
+            pytest.mark.group(id=deploy_type),
         ],
     )
     for deploy_type in ["large_deployment", "small_deployment"]
```
carlcsaposs-canonical, line 60: is all this complexity still needed with the removal of the group marker? (can a simple fixture be used instead?)

tests/integration/relations/conftest.py
```diff
@@ -2,24 +2,11 @@
...
-import shutil
 
 import pytest
-from pytest_operator.plugin import OpsTest
 
 
 @pytest.fixture(scope="module")
-async def application_charm(ops_test: OpsTest):
+async def application_charm() -> str:
     """Build the application charm."""
```
carlcsaposs-canonical, line 11: nit: docstring incorrect

tests/spread/test_backups.py/all/task.yaml
```diff
@@ -0,0 +1,9 @@
+summary: test_backups.py
+environment:
+  TEST_MODULE: ha/test_backups.py
+systems:
+  - self-hosted-linux-amd64-noble-large
+execute: |
+  tox run -e integration -- "tests/integration/$TEST_MODULE" --model testing --alluredir="$SPREAD_TASK/allure-results" -m 'group(id="all")'
```
carlcsaposs-canonical: suggest using pytest fixture with pytest command line option or enviornment variable instead of using the old `pytest.mark.group` construct

think that'd be simpler

carlcsaposs-canonical: also think that might be needed for allure to work correctly (parameterizing based on a fixture will ensure different test case ids, which means the different tests won't override each other in the allure report)

carlcsaposs-canonical (commented):
initial pass based on diff from first 3 commits

didn't (and not planning to) review test changes/opensearch specific stuff closely

carlcsaposs-canonical (commented):
has allure report been tested?

poetry.lock
skourta: Yes, I removed the deprecated pytest plugins we were no longer using.

pyproject.toml
skourta: I decided against doing it in this PR. It would require additional refactoring for the microceph tests. 
@Mehdi-Bendriss what do you think?

spread.yaml
skourta: I faced connection errors when spread tries to connect to the self hosted machine. I referred to the [zinc repo](https://github.com/jnsgruk/zinc-k8s-operator/blob/a21eae8399eb3b9df4ddb934b837af25ef831976/spread.yaml#L51-L61) for this config which worked.

spread.yaml
skourta: I changed these when I had errors connecting to the self-hosted machine same as the above comment. I followed zinc for the config.

spread.yaml
skourta: Spread would not pick up on the multiple task files nested at n+1 so I added the suites to pick up the tests I separated.

spread.yaml
skourta: Missed this while cleaning changes I made while debugging. One of the errors was on this command so I separated them to see which dependency was causing the issue. I will fix it.

tests/integration/conftest.py
skourta: I believe not. I will remove it

tests/integration/ha/test_backups.py
skourta: I'm using the groups to filter the tests to run using `-m` flag of pytest. I thought the `group` nomenclature was good so I left it as the marker's name.

tests/integration/plugins/test_plugins.py
skourta: Same as last comment. If we add both backends and use a fixture to skip we would have a lot of jobs with skipped tests. Please correct me if I am wrong and let me know what you think.

tests/spread/test_backups.py/all/task.yaml
skourta: If we use a fixture to detect what type of test to run based on the runner type, we could only decide on what tests to run, right? This would not change the test case names.
I believe the test case names will be different with this setup, as we have been using it before this migration. Could you perhaps share an example of using the fixtures? I feel like I did not fully understand what you meant.

.github/workflows/integration_test.yaml
skourta: Shoud we leave the variant in the name?

spread.yaml
skourta: This is actually needed to install them on self-hosted runners

.github/workflows/integration_test.yaml
skourta: We do indeed need to add the variant otherwise, the logs of the two variants will have the same name, and the upload of the last test will fail.

poetry.lock
carlcsaposs-canonical: it seems like there are other dep changes in this pr besides that

spread.yaml
carlcsaposs-canonical: what errors? on what environment?

this is the config the other data repos use, so would like to fix it if there are issues

spread.yaml
carlcsaposs-canonical: what were the errors?

we're using the runner username to minimize changes between tests that run on github runners directly vs github runners via spread to avoid having to troubleshoot permission/group errors related to creating a new user

spread.yaml
skourta: Spread would not be able to connect to the runner through SSH. The issue only appeared when I made the switch to self-hosted runners.

spread.yaml
carlcsaposs-canonical: maybe add the combinations without nesting

e.g.
tests/spread/test_backups.py-all/task.yaml

so that a single suite can continue to be used

spread.yaml
carlcsaposs-canonical: why is charmcraft running on ci? the build should be in a separate step with build_charm.yaml

.github/workflows/integration_test.yaml
carlcsaposs-canonical: yes, the variant removal was only for mongo since they weren't using variants

spread.yaml
carlcsaposs-canonical: can you check the contents of /etc/ssh/sshd_config.d/ and /etc/ssh/sshd_config on the self-hosted runners?

something might be overriding the config

if not, you might need to add `sudo systemctl reload sshd`

tests/spread/test_backups.py/all/task.yaml
carlcsaposs-canonical: if the fixture is parameterized, the test case names will be different

tests/integration/ha/test_backups.py
carlcsaposs-canonical: when we switched to charmcraft test on the sql repos, the reasons we removed the group marker were
- it duplicates functionality now provided by spread
- it adds a significant layer of complexity/abstraction that we thought was confusing (given that charmcraft test is adding a large amount of abstraction/complexity, so we wanted to reduce complexity elsewhere) and that the abstraction the group markers added would cause more harm than benefits

I think stuff like https://github.com/canonical/opensearch-operator/blob/30b505e97b3d2149173eefc7e3df6a4ef439c6ca/tests/spread/test_large_deployments_inherit_cluster_name.py/not-autogenerated/task.yaml#L7 will be particularly painful to maintain & is more complex than what's necessary

but if that's how y'all want to write & parameterize your tests, not trying to stop you

spread.yaml
skourta: From an organisational point of view, I prefer to have them grouped. But that's a soft opinion.

spread.yaml
skourta: I removed it. I believe we only needed pipx

carlcsaposs-canonical: can you add a comment explaining it's needed for the self-hosted runners? (so that it's easy to tell what was changed in spread.yaml for the self-hosted runners)

spread.yaml
skourta: Discussed and updated to the `ubuntu` user.

pyproject.toml
skourta: Confirmed with @Mehdi-Bendriss that this is out of this PR's scope.

spread.yaml
skourta: Discussed internally and updated to `ubuntu`

tests/integration/ha/test_backups.py
skourta: I talked to @Mehdi-Bendriss and we agreed to keep the groups for now.

spread.yaml
carlcsaposs-canonical: except for using the ubuntu user, can the changes to the ssh config from the initial commits be reverted?

skourta: I did change them back. I only left what is needed. There's a config in the runner that contains conflicting params and that need to be deleted and the password for the `ubuntu` user needs to be updated to the one generated by spread.

spread.yaml
carlcsaposs-canonical: I think the suites are designed for sharing settings/separating test categories (e.g. unit & integration). For tests that are usually run at the same time or share an allure report, I'd expect them to be in the same suite

Also unless there's a good reason I'd advise not deviating from the rest of the data repos—this may lead to further divergence in the future and cause issues if we create any shared tooling in the future

spread.yaml
carlcsaposs-canonical: why does the ubuntu password need to be changed? we're already logged in as the ubuntu user

also do we need to remove the existing config in the runner if our config already has priority?

carlcsaposs-canonical: is an existing password set on the ubuntu user?

skourta: Spread will try to SSH into the machine, so you need the password anyway for that to work.
For the config, I think it is cleaner to go Jon's way and delete unwanted configs. Why have conflicting configs in the first place?

carlcsaposs-canonical: > delete unwanted configs

what happens if the config shipped on the runners changes? then this will break

I think it's better to have the minimum changes to get it to work & for the config to be portable across github-hosted, is-hosted, and other environments

carlcsaposs-canonical: > so you need the password anyway for that to work

on github runners, we don't change the password and we allow empty passwords

do the is-hosted runners set a password on the ubuntu user? if not, I think we should use the same approach as we use on the github-hosted runners

spread is not the only thing that runs on the github actions run—other things that run on github actions might not expect us to change the environment like this

skourta: Let me try without those two commands and get back to you.

skourta: With the exact same config as mysql but ubuntu as user
https://github.com/canonical/opensearch-operator/actions/runs/14633182616/job/41059393432?pr=619#step:8:1156

With the exact same config + systemctl daemon-reload
https://github.com/canonical/opensearch-operator/actions/runs/14633318767/job/41059768211?pr=619#step:8:1172

With the exact same config + systemctl daemon-reload + chpasswd
https://github.com/canonical/opensearch-operator/actions/runs/14633452112/job/41060125768?pr=619#step:8:1171

spread.yaml
skourta: I spread out the task files in `tests/spread`

spread.yaml
carlcsaposs-canonical: Looking at the contents of `/etc/shadow` on the self-hosted runner, the `ubuntu` password is empty but locked
The password can be unlocked (without changing it, in case anything else relies on the empty password) with `sudo passwd --delete ubuntu`

skourta: Done

carlcsaposs-canonical: is the daemon-reload and restart still needed? if it is needed, would recommend reload instead of restart on sshd

skourta: I tried to reload sshd but it seems on the last versions of ubuntu you need to restart ssh instead. When I restart ssh I get a warning to execute a daemon-reload.

carlcsaposs-canonical: > last versions of ubuntu you need to restart ssh instead

where do you see that?

restarting closes active connections—a full restart shouldn't be necessary to apply a config change
are you reloading `sshd` (not `ssh`)?

also, does it work if you remove the daemon-reload & restart entirely?

skourta: I see that when I execute the command. It errors out with `Failed to restart sshd.service: Unit sshd.service not found`. [Additional ref](https://serverfault.com/questions/1147379/unit-sshd-service-not-found).

For daemon-reload again once you restart the service it will issue the warning. See [here](https://github.com/canonical/opensearch-operator/actions/runs/14633182616/job/41059393432?pr=619#step:8:1156).

carlcsaposs-canonical: oh okay it looks like `sshd` was renamed to `ssh`
I think reload should still work though

skourta: Switched restart to reload. The [warning](https://github.com/canonical/opensearch-operator/actions/runs/14637395590/job/41072165063?pr=619#step:8:1022) is still present. What do you think about adding back the `daemon-reload`?

carlcsaposs-canonical: yeah think it makes sense to include

skourta: Done

carlcsaposs-canonical (dismissed):
looks good once allure test reverted

didn't review test changes closely; focused on workflow & spread config

in the future, think it might be worth revisiting https://github.com/canonical/opensearch-operator/pull/619#discussion_r2054296080 —I think the group marker will cause more problems than it solves & it could be simplified

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,310 @@
...
+          jobs = []
+          for job in spread_jobs:
+              # Example `job`: "github-ci:ubuntu-24.04:tests/spread/test_charm.py:juju36"
+              _, runner, task, variant = job.split(":")
+              # Example: "test_charm.py"
+              task = task.removeprefix("tests/spread/")
+              if runner.endswith("-arm"):
+                  architecture = "arm64"
+              else:
+                  architecture = "amd64"
```
carlcsaposs-canonical: I think this needs to be modified for the is-hosted runner names

Mehdi-Bendriss (approved):
Thank you Smail! Great work!!

carlcsaposs-canonical (approved):

PR #628, opened 2025-04-29
[DPE-7018] stabilize CA rotation in large deployments

reneradoi:
## Issue
When executing a CA rotation in a large deployments setting, the non-Main-Orchestrator units update their local CA bundle (used for API-requests to Opensearch) too early in the process. 

This results in API-errors like `Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError)` when performing requests against the unit's API. Depending on the specific API-requests executed, this can lead to failures on `is_node_up()` or `node_lock.acquire()`, causing the CA rotation workflow to fail.

## Solution
Make sure the ca bundle for requests is only updated in large deployment non-main-orchestrators if there is no CA-rotation going on.


Mehdi-Bendriss (approved):
Thank you René!

skourta (approved):

smethnani (approved):

PR #629, opened 2025-04-29
[DPE-6877] - Fix failover bootstrapping cluster

### LINKED ISSUE #518: Large Deployments - Failover fails to initialize the security index
## Steps to reproduce

1. Deploy a large deployment:
	- Main App: `cluster_manager` role only
	- Failover: `cluster_manager` and `data`

## Expected behavior
The large deployment boots up correctly.

## Actual behavior

Failover is stuck at initializing the security index.
image (2)

## Versions

Operating system: Ubuntu 24.04.1 LTS

Juju CLI: 3.6.0-genericlinux-amd64

Juju agent: 3.5.3

Charm revision: `2/edge` branch

LXD: 5.21.2 LTS

skourta:
Fixes #518 
This pull request introduces changes to improve the handling of failover orchestrators in OpenSearch deployments, specifically addressing scenarios where a failover orchestrator is the only data node in the cluster. The updates ensure proper role management to prevent unintended cluster bootstrapping and enhance the robustness of failover scenarios.

### Failover Orchestrator Role Management:
* Updated the condition in the `cleanup` method to exclude `MAIN_ORCHESTRATOR` deployment types from ignoring locks, ensuring that failover orchestrators are handled correctly.
* Added logic in `_set_node_conf` to remove the `cluster_manager` role from failover orchestrators that are the only data nodes in the cluster, preventing them from bootstrapping the cluster. This change also stores a flag indicating the role removal.
* Introduced a new helper method `_is_failover_only_data_node` to determine if the current node is a failover orchestrator and the only data node in the cluster, improving code clarity and reusability.

### Failover Recovery Enhancements:
* Added logic in `_post_start_init` to restore the `cluster_manager` role and restart the service if a failover-only data node had its role removed during initialization. This ensures proper recovery in failover scenarios.

### Documentation Update:
* Updated a comment in `_start_opensearch` to clarify that the `ignore_lock` flag is now used for both force upgrades and starting data nodes for cluster-manager-only nodes.


reneradoi (dismissed):
Great change for this very specific scenario, thank you Smail!

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1757,6 +1772,22 @@ def put_security_index_initialized(self, event: EventBase) -> None:
...
+        if (
+            deployment_desc.typ == DeploymentType.FAILOVER_ORCHESTRATOR
+            and "data" in deployment_desc.config.roles
+        ):
+            # check if it's the only data node in the cluster
+            cluster_fleet_apps = self.peers_data.get_object(Scope.APP, "cluster_fleet_apps")
+            for app in cluster_fleet_apps:
+                if self.app.name != cluster_fleet_apps[app].get("app", {}).get(
+                    "name"
+                ) and "data" in cluster_fleet_apps[app].get("roles"):
```
smethnani: Do we need to catch the case where one of the apps in the cluster has generated roles? The reported list in roles will be empty in that case. We could modify the reported roles [here](https://github.com/canonical/opensearch-operator/blob/439106663ef1c7cd142223d8a80ec7256cde57d9/lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py#L376) and [here](https://github.com/canonical/opensearch-operator/blob/439106663ef1c7cd142223d8a80ec7256cde57d9/lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py#L934) to be something like:

```
roles=deployment_desc.config.roles if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES else ClusterTopology.generated_roles()
```

smethnani (commented):
Thanks Smail! Looks good, I just left one comment about capturing the roles from `cluster_fleet_apps`

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Nice catch. I believe it should always be initialised in that fashion.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -424,7 +424,7 @@ def cleanup():
         ignore_lock = (
             "data" in deployment_desc.config.roles
```
Mehdi-Bendriss: 1. should this not also check if the roles are autogenerated? given that has an implicit `data` role.
2. How do we guarantee that this only runs for the `first data node joining`?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1789,6 +1804,22 @@ def put_security_index_initialized(self, event: EventBase) -> None:
             # notify the main orchestrator that the security index is initialized
             self.peer_cluster_requirer.set_security_index_initialised()
 
+    def _is_failover_only_data_node(self) -> bool:
```
Mehdi-Bendriss: perhaps:
```suggestion
    def _is_failover_and_only_data_node(self) -> bool:
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1535,6 +1542,14 @@ def _set_node_conf(self, nodes: List[Node]) -> None:
         else:
             computed_roles = ClusterTopology.generated_roles()
 
+        # If the failover orchestrator is the only data node in the cluster, remove the
+        # cluster-manager role from it to avoid it bootstrapping the cluster
```
Mehdi-Bendriss, line 1642: Could you add a comment about why we're doing it, something along the lines of: 
"as it is the duty of the main orchestrator which should then broadcast the security_index_initialized" flag in the peer cluster orchestrator relation"

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1789,6 +1804,22 @@ def put_security_index_initialized(self, event: EventBase) -> None:
             # notify the main orchestrator that the security index is initialized
             self.peer_cluster_requirer.set_security_index_initialised()
 
+    def _is_failover_only_data_node(self) -> bool:
+        """Check if the current node is a failover and the only data node in the cluster."""
+        deployment_desc = self.opensearch_peer_cm.deployment_desc()
+        if (
+            deployment_desc.typ == DeploymentType.FAILOVER_ORCHESTRATOR
+            and "data" in deployment_desc.config.roles
```
Mehdi-Bendriss: We should also check on the autogenerated roles

Mehdi-Bendriss (commented):
Thanks Smail, great work! I left a few comments

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: 1. ~The autogenerated roles include the data roles. So this check includes that.~ I added the check fior autogenerated roles.
2. I will run some manual tests. I only changed the condition to include failovers but I believe this should unblock all **leader** data nodes to start. Would that be an issue?

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: One of @smethnani 's comments treated this. we make sure the roles attribute is filled with all generated roles (including `data`) if the start mode is `WITH_PROVIDED_ROLES`.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: I ran several tests and can confirm all data nodes are unblocked not only one. It does not however constitute an issue, and the cluster comes up as expected.

lib/charms/opensearch/v0/opensearch_base_charm.py
smethnani: Sorry I missed the references to `deployment_desc.config.roles` in my review @skourta ! The comment was about the roles in `PeerClusterApp`, which is populated from the `deployment_desc.config.roles` (and is empty in the case of autogenerated roles).

skourta: I added the checks in [2cd024e](https://github.com/canonical/opensearch-operator/pull/629/commits/2cd024e3c21cb67fa0ebf26f268415a85ef253c1)

smethnani (commented):
Thank you @skourta! Great work! In all cases I tested, all the units eventually start and all have the correct roles. 

I think there may be some issues in the case of more than 1 data node and cluster manager in the deployment. I tested the following cases for large deployments:

A main orchestrator with `cluster_manager' role and:
- 1 failover with autogenerated roles
   - cluster manager role removed, sec-index created, failover restarts
- 1 failover with provided `cluster_manager,data` roles 
   - cluster manager role removed, sec-index created, failover restarts
- 1 failover with only cluster_manager role, 1 failover with autogenerated roles
   - Same as above, but 'Waiting for every unit in main to start` is never cleared. I don't think this is due to a change in your PR so will track it in another issue, but just wanted to make a note here

The ones with some issues:
- **2 failovers with autogenerated roles**
   - In this case, on some occasions both failovers restart, on some only 1 restarts. But in all cases the `cluster_manager_removed` flag is never cleared after all the units have restarted
- **1 failover with autogenerated roles, 1 data node**
   - in this case the failover is still restarting and is the one initializing the security index when I tested it. The `cluster_manager_removed` flag is again not cleared and remains after the units have restarted
  
For the case of 1 failover with autogenerated roles, and 1 data node. I ran accross a secret-changed hook-failure in the data node, though I'm not sure if that is due to any changes here. I've added logs below for reference and we can create a separate issue if its unrelated to this PR:
[main.log](https://pastebin.canonical.com/p/6VGX5zfqhK/) (cluster_manager only)
[failover1.log](https://pastebin.canonical.com/p/Kf2PyRbCtN/) (autogenerated roles)
[failover2.log](https://pastebin.canonical.com/p/qH3DWbrdfP/) <- this is the one that showed a hook-failure, the app is called failover2 but it is the data node and only has a data role

skourta:
> Thank you @skourta! Great work! In all cases I tested, all the units eventually start and all have the correct roles.
> 
> I think there may be some issues in the case of more than 1 data node and cluster manager in the deployment. I tested the following cases for large deployments:
> 
> A main orchestrator with `cluster_manager' role and:
> 
> * 1 failover with autogenerated roles
>   
>   * cluster manager role removed, sec-index created, failover restarts
> * 1 failover with provided `cluster_manager,data` roles
>   
>   * cluster manager role removed, sec-index created, failover restarts
> * 1 failover with only cluster_manager role, 1 failover with autogenerated roles
>   
>   * Same as above, but 'Waiting for every unit in main to start` is never cleared. I don't think this is due to a change in your PR so will track it in another issue, but just wanted to make a note here
> 
> The ones with some issues:
> 
> * **2 failovers with autogenerated roles**
>   
>   * In this case, on some occasions both failovers restart, on some only 1 restarts. But in all cases the `cluster_manager_removed` flag is never cleared after all the units have restarted
> * **1 failover with autogenerated roles, 1 data node**
>   
>   * in this case the failover is still restarting and is the one initializing the security index when I tested it. The `cluster_manager_removed` flag is again not cleared and remains after the units have restarted
> 
> For the case of 1 failover with autogenerated roles, and 1 data node. I ran accross a secret-changed hook-failure in the data node, though I'm not sure if that is due to any changes here. I've added logs below for reference and we can create a separate issue if its unrelated to this PR: [main.log](https://pastebin.canonical.com/p/6VGX5zfqhK/) (cluster_manager only) [failover1.log](https://pastebin.canonical.com/p/Kf2PyRbCtN/) (autogenerated roles) [failover2.log](https://pastebin.canonical.com/p/qH3DWbrdfP/) <- this is the one that showed a hook-failure, the app is called failover2 but it is the data node and only has a data role

Test all these scenarios and a couple more.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: I have added synchronisation between data apps when trying to start as the first data node.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -428,22 +434,60 @@ def cleanup():
...
-                not self.peers_data.get(Scope.APP, "security_index_initialised", False)
-                or (
-                    # in case all data-nodes are powered down after being previously started
-                    # ignore the lock to get a data-node started, as it holds security index
-                    self.peers_data.get(Scope.UNIT, "started")
-                    and not self.opensearch.is_service_started()
-                )
+        # if this is the a data node to join, start without getting the lock
+        # if there are multiple data apps in the cluster
+        # we synchronize the start of the first data node through peer cluster relationtoxc
```
```suggestion
        # we synchronize the start of the first data node through the peer cluster relation
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -428,22 +434,60 @@ def cleanup():
...
+        # all data nodes request to start as first data node
+        #   ->(app databag key: first_data_node on data app)
+        # main orchestrator will choose which node to start first
+        #   ->(app databag key: first_data_node on main orchestrator app)
+
+        if self._should_ignore_lock(deployment_desc):
+            # Actual start will be done in `_on_peer_cluster_relation_changed`
```
Mehdi-Bendriss: What do you mean by this?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -428,22 +434,60 @@ def cleanup():
...
             )
-        )
-        self._start_opensearch_event.emit(ignore_lock=ignore_lock)
+            self.peer_cluster_requirer.set_first_data_node(deployment_desc.app.id)
+            return
+        self._start_opensearch_event.emit()
+
+    def _should_ignore_lock(self, deployment_desc: DeploymentDescription) -> bool:
+        """Check if we should ignore the lock when starting OpenSearch."""
+        # if the security index is not initialized, a node has already started
```
Mehdi-Bendriss: "is initialized" or "is not initialized"?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -428,22 +434,60 @@ def cleanup():
...
+            logger.debug(
+                f"Requesting start as first data node without lock: {deployment_desc.app.id}"
             )
-        )
-        self._start_opensearch_event.emit(ignore_lock=ignore_lock)
+            self.peer_cluster_requirer.set_first_data_node(deployment_desc.app.id)
+            return
+        self._start_opensearch_event.emit()
+
+    def _should_ignore_lock(self, deployment_desc: DeploymentDescription) -> bool:
```
Mehdi-Bendriss: I think the logic of this method should rather be changed towards the opposite thinking for more clarity in the code, by "Only implementing the logic of what would make a unit ignore the lock."

Otherwise the branching is hard to reason with and understand what exactly constitutes an exception to the default of "always acquiring the lock"

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -428,22 +434,60 @@ def cleanup():
...
+        #   ->(app databag key: first_data_node on main orchestrator app)
+
+        if self._should_ignore_lock(deployment_desc):
+            # Actual start will be done in `_on_peer_cluster_relation_changed`
+            logger.debug(
+                f"Requesting start as first data node without lock: {deployment_desc.app.id}"
             )
-        )
-        self._start_opensearch_event.emit(ignore_lock=ignore_lock)
+            self.peer_cluster_requirer.set_first_data_node(deployment_desc.app.id)
```
Mehdi-Bendriss: I am not a fan of setting the application id as a "node" - why not use the `unit_name` here?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -809,6 +849,9 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
         if self.opensearch.is_service_started() and (
             plugin_needs_restart or perf_profile_needs_restart
         ):
+            logger.debug(
+                f"Restarting opensearch due to confuig change: plugin_needs_restart={plugin_needs_restart}, perf_profile_needs_restart={perf_profile_needs_restart}"
```
```suggestion
                f"Restarting opensearch due to config change: plugin_needs_restart={plugin_needs_restart}, perf_profile_needs_restart={perf_profile_needs_restart}"
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -141,18 +141,24 @@ class _StartOpenSearch(EventBase):
     This event will be deferred until OpenSearch starts.
     """
 
-    def __init__(self, handle, *, ignore_lock=False, after_upgrade=False):
+    def __init__(self, handle, *, ignore_lock=False, after_upgrade=False, first_data_node=False):
```
Mehdi-Bendriss: to differentiate between the `first_data_node` field
```suggestion
    def __init__(self, handle, *, ignore_lock=False, after_upgrade=False, is_first_data_node=False):
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -428,22 +434,60 @@ def cleanup():
...
-            and deployment_desc.typ == DeploymentType.OTHER
-            and (
-                not self.peers_data.get(Scope.APP, "security_index_initialised", False)
-                or (
-                    # in case all data-nodes are powered down after being previously started
-                    # ignore the lock to get a data-node started, as it holds security index
-                    self.peers_data.get(Scope.UNIT, "started")
-                    and not self.opensearch.is_service_started()
-                )
+        # if this is the a data node to join, start without getting the lock
```
Mehdi-Bendriss: We should have a bit more clarity here, and specify it's about the "first“ data node.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1056,16 +1100,14 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
         self.peers_data.delete(Scope.UNIT, "started")
 
         if event.ignore_lock:
-            # Only used for force upgrades
+            # Only used for force upgrades and starting data nodes for cluster-manager-only nodes
```
```suggestion
            # Only used for force upgrades and starting 1 data node on a large deployment where the main orchestrator has cluster-manager only nodes
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1398,11 +1446,17 @@ def _can_service_start(self) -> bool:
...
-            return self.unit.is_leader() and (
-                deployment_desc.start == StartMode.WITH_GENERATED_ROLES
-                or deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
-                or "data" in deployment_desc.config.roles
-            )
+            if self.unit.is_leader():
+                # If main orchestrator start
+                if deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR:
+                    return True
+                # If data node only start if it's the first one
```
Mehdi-Bendriss: please rephrase

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1398,11 +1446,17 @@ def _can_service_start(self) -> bool:
...
+                if (
+                    deployment_desc.start == StartMode.WITH_GENERATED_ROLES
+                    or "data" in deployment_desc.config.roles
+                ) and first_data_node:
+                    return True
+            return False
```
Mehdi-Bendriss: better group this in a single return, with comments in front of the conditions for clarity

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1818,6 +1883,25 @@ def put_security_index_initialized(self, event: EventBase) -> None:
             # notify the main orchestrator that the security index is initialized
             self.peer_cluster_requirer.set_security_index_initialised()
 
+    def _is_failover_and_no_pure_data_node(self) -> bool:
```
Mehdi-Bendriss: This runs for all failover units, is this intended?

Mehdi-Bendriss: perhaps?
```suggestion
    def _is_failover_and_sole_data_app(self) -> bool:
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1818,6 +1883,25 @@ def put_security_index_initialized(self, event: EventBase) -> None:
...
+            "data" in deployment_desc.config.roles
+            or deployment_desc.start == StartMode.WITH_GENERATED_ROLES
+        ):
+            # check if it's the only data node in the cluster
+            cluster_fleet_apps = self.peers_data.get_object(Scope.APP, "cluster_fleet_apps") or {}
+            for app in cluster_fleet_apps:
+                if (
+                    self.app.name != cluster_fleet_apps[app].get("app", {}).get("name")
+                    and "data" in cluster_fleet_apps[app].get("roles")
+                    and "cluster_manager" not in cluster_fleet_apps[app].get("roles")
```
Mehdi-Bendriss: Can you explain why we're checking on the `cluster_manager` role?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1818,6 +1883,25 @@ def put_security_index_initialized(self, event: EventBase) -> None:
...
+                ):
+                    return False
+            return True
+        return False
```
Mehdi-Bendriss: Better invert the condition to have less nestedness

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1818,6 +1883,25 @@ def put_security_index_initialized(self, event: EventBase) -> None:
             # notify the main orchestrator that the security index is initialized
             self.peer_cluster_requirer.set_security_index_initialised()
 
+    def _is_failover_and_no_pure_data_node(self) -> bool:
+        """Check if the current node is a failover and the only data node in the cluster."""
```
Mehdi-Bendriss, line 1902: it cannot be the only data node if it is part of a failover app that has the "data" role and multiple nodes

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -182,6 +182,11 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):
 
         if self._get_security_index_initialised():
             self.charm.peers_data.put(Scope.APP, "security_index_initialised", True)
+            # clean up the first data node when security index is initialised
+            self.charm.peers_data.put(Scope.APP, "first_data_node", "")
```
Mehdi-Bendriss: you can use `.delete`

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -861,6 +867,24 @@ def _get_security_index_initialised(self) -> bool:
 
         return False
 
+    def _get_first_data_node(self) -> str | None:
+        """Get the first data node from the relation data."""
+        if first_data_node := self.charm.peers_data.get(Scope.APP, "first_data_node", None):
+            return first_data_node
+
+        # check all other clusters if they have initialised the security index
```
Mehdi-Bendriss, line 876: How do we check that, given the security index flag doesn't get broadcasted from the requirer apps?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1142,7 +1181,7 @@ def set_security_index_initialised(self) -> None:
             self.charm.peers_data.get_object(Scope.APP, "orchestrators") or {}
         )
 
-        if not orchestrators:
+        if orchestrators.main_app is None:
```
Mehdi-Bendriss, line 1170: I believe we should check on both, no orchestrators and main app of orchestrators None. Any reason to omit the first check?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -930,6 +954,21 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
+            if (
+                local_first_data_node := self._get_local_first_data_node()
+            ) is not None and peer_cluster_rel_data.first_data_node is not None:
+                logger.debug(
+                    f"Local first data node: {local_first_data_node} - cluster first data node: {peer_cluster_rel_data.first_data_node}"
+                )
+                if peer_cluster_rel_data.first_data_node == local_first_data_node:
+                    self.charm._start_opensearch_event.emit(ignore_lock=True, first_data_node=True)
+                else:
+                    self.charm._start_opensearch_event.emit()
```
Mehdi-Bendriss: why is this needed?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -930,6 +954,21 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
+                self.set_first_data_node("")
```
Mehdi-Bendriss: I think we are missing a `return` here, and i'm not a fan of initiating a start from here, when the `_on_start` hook is guaranteed to be deferred until fulfilling its requirements. 
why not drop these and  let the flow continue to `refresh_relation_data` and let the charm take over?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1445,3 +1484,56 @@ def _clear_errors(self, *error_labels: str):
             error = self.charm.peers_data.get(Scope.APP, error_label, "")
             self.charm.status.clear(error, app=True)
             self.charm.peers_data.delete(Scope.APP, error_label)
+
+    def set_first_data_node(self, first_data_node: str) -> None:
```
Mehdi-Bendriss: maybe better
```suggestion
    def set_first_data_node(self, first_data_node: str | None) -> None:
```

and if None, use `delete_from_rel` instead of `put_in_rel`

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -930,6 +954,21 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
 
             self._put_main_orchestrator_registered(orchestrators.failover_rel_id, True)
 
+        if data.get("data"):
+            peer_cluster_rel_data = self.peer_cm.rel_data_from_str(data["data"])
+            local_first_data_node = self._get_local_first_data_node()
+            if (
+                local_first_data_node := self._get_local_first_data_node()
+            ) is not None and peer_cluster_rel_data.first_data_node is not None:
```
Mehdi-Bendriss: Can you explain the difference between the local and the cluster first data node?

Mehdi-Bendriss (changes requested):
Thanks Smail, I have a few comments and questions. I would also request that we review the comments and explanations left in the code in this PR.
Ideally we also have a richer unit test for the start_opensearch method.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: I updated the description

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: The requirer app sets it in the peer cluster relation on its unit data then the main orchestrator picks it up and broadcasts it.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: Local one is the name of the unit (previously app id). If set it means this unit requested to start as a first data node. 
THe cluster first data node is the name of the unit that the main orchestrator chose to start as first. This is done so only 1 data nodestarts)

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: Discussed in the call

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: Refactored

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: I updated its use now it should only run on the leader of the failover apps.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: We only consider leader units when we choose a unit to start as the first data node

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: We want to exclude cluster managers with data nodes because we want to check if there is a pure data app then we let it start instead.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Completely refactored the function

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: `from_dict` returns an uninitialized `PeerClusterOrchestrators` object if an empty dict is passed.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: updated

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -428,11 +435,59 @@ def cleanup():
         # request the start of OpenSearch
         self.status.set(WaitingStatus(RequestUnitServiceOps.format("start")))
 
-        # if this is the first data node to join, start without getting the lock
-        ignore_lock = (
-            "data" in deployment_desc.config.roles
-            and self.unit.is_leader()
-            and deployment_desc.typ == DeploymentType.OTHER
+        # In large deployments one data node needs to start to initialzie the security index
```
smethnani:
```suggestion
        # In large deployments one data node needs to start to initialize the security index
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -428,11 +435,59 @@ def cleanup():
...
+                # main orchestrator has chosen the first data node
+                if peer_cluster_rel_data.first_data_node == local_first_data_node:
+                    logger.debug(
+                        f"Local first data node: {local_first_data_node} - cluster first data node: {peer_cluster_rel_data.first_data_node}"
+                    )
+                    # this unit is the first data node chosen by the main orchestrator
+                    self._start_opensearch_event.emit(ignore_lock=True, is_first_data_node=True)
+                    self.peer_cluster_requirer.set_first_data_node(None)
+            else:
+                # main orchestratior has not chosen the first data node yet
```
smethnani:
```suggestion
                # main orchestrator has not chosen the first data node yet
```

tests/integration/ha/test_large_deployments_validate_cm_count.py
```diff
@@ -31,11 +31,10 @@
 @pytest.mark.group(1)
 @pytest.mark.abort_on_fail
 @pytest.mark.skip_if_deployed
-async def test_build_and_deploy(ops_test: OpsTest) -> None:
+async def test_build_and_deploy(charm: str, ops_test: OpsTest) -> None:
```
smethnani, line 34: 👍 thank you

lib/charms/opensearch/v0/opensearch_base_charm.py
smethnani: ~~Perhaps `_is_failover_and_first_data_app` is more descriptive since this logic can run before a join event from another data app is processed.~~

My bad, I was looking at the old name

smethnani (dismissed):
Thank you Smail! All the `cluster_manager_removed` flags were cleaned up in the tests I ran.

Just left some comments about typos

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -434,60 +435,74 @@ def cleanup():
...
-            cluster_first_data_node := self.peer_cluster_requirer.get_cluster_first_data_node()
-        ) is not None and cluster_first_data_node != deployment_desc.app.id:
-            logger.debug(
-                f"Cluster first data node is {cluster_first_data_node}, {deployment_desc.app.id} should not ignore lock."
+            self.unit.is_leader()
+            and self.opensearch_peer_cm.is_consumer()
+            and (
+                (local_first_data_node := self.peer_cluster_requirer.get_local_first_data_node())
+                is not None
             )
```
Mehdi-Bendriss: nit:
```suggestion
            and local_first_data_node := self.peer_cluster_requirer.get_local_first_data_node()
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -434,60 +435,74 @@ def cleanup():
...
+            and (
+                not self.peers_data.get(Scope.APP, "security_index_initialised", False)
+                or (
+                    # in case all data-nodes are powered down after being previously started
+                    # ignore the lock to get a data-node started, as it holds security index
+                    self.peers_data.get(Scope.UNIT, "started")
+                    and not self.opensearch.is_service_started()
+                )
+            )
+            and self.peer_cluster_requirer.get_cluster_first_data_node() is None
```
Mehdi-Bendriss, line 497: Could you explain this condition? if not cluster wide reported data node, this node should start? Shouldn't have this node reported itself already as candidate first data node? 
Also, should we not test on whether this app is a `consumer` in this case? to avoid starting it if it's not yet related to to main orchestrator?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1446,17 +1462,17 @@ def _can_service_start(self, first_data_node: bool = False) -> bool:  # noqa: C9
...
-            return False
+            return self.unit.is_leader() and (
+                deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+                # first data node in a cluster-manager-only deployment
+                or (
+                    (
+                        deployment_desc.start == StartMode.WITH_GENERATED_ROLES
+                        or "data" in deployment_desc.config.roles
+                    )
+                    and first_data_node
```
Mehdi-Bendriss: could you `s/first_data_node: bool/is_first_data_node: bool` ?

Mehdi-Bendriss (commented):
Thank you Smail! Looks good! 
I only have a couple of small questions and we should be good to go

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: 1. 1st question, we only request to start as the first data node if no other node has already been designated. If the `_should_ignore_lock` is true, the node will request to start as a first data node.
2. No need to test for the consumer explicitly. This is a data node that is **NOT** the main orchestrator. It has to be a consumer.

smethnani (approved):

Mehdi-Bendriss (approved):

PR #633, opened 2025-04-30
avoid-plugin-check-if-not-ready

reneradoi:
## Issue
When deploying Opensearch, the operator runs a plugin check and tries to enable configured plugins on the initial `config_changed` event. This is unnecessary and fails with `OpenSearchKeystoreNotReadyError` because Opensearch was not started and its API's are not available yet. 

## Solution
Wait with the plugin check/enabling until Opensearch is started.


Mehdi-Bendriss (approved):

skourta (approved):
Thank you @reneradoi

PR #635, opened 2025-05-05
fix(DPE-6869): Fix starting up after scale down to 0

### LINKED ISSUE #599: OpenSearch not starting after scaling app down to 0 units and scaling back up
## Steps to reproduce

1. Deploy applicaton:
```
juju deploy self-signed-certificates --config ca-common-name="Test CA"
juju deploy -n 3 opensearch main --channel="2/edge" --config profile="testing"
```
2. Remove all units
```
juju remove-unit main/0 main/1 main/2
``` 
3. Add units
```
juju add-unit main -n 3
```
## Expected behavior
OpenSearch should start. App units should eventually reach an active/idle state

## Actual behavior

Units get stuck `Waiting for OpenSearch to start...`
Logs show repeated `node-lock-fallback:1: Could not start opensearch service. Will retry next event.` messages

## Versions

Operating system: Ubuntu 24.04.2 LTS

Juju CLI: 3.6.4-genericlinux-amd64

Juju agent: 3.6.4

Charm revision: 213

LXD: 5.21.3 LTS

## Log output

Juju debug log: https://pastebin.canonical.com/p/M5796bG8gR/

skourta:
### Data cleanup during storage detachment:
* Added deletion of the `security_index_initialised` and `bootstrapped` keys from `peers_data` when all planned units are removed and a peer relation exists. This ensures proper cleanup of application-level data.

### Conditional logic for user credential updates:
* Modified the condition for updating the security index to check if the node is up before proceeding with credential changes. This prevents unnecessary operations when the node is not operational.


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -582,6 +582,8 @@ def _on_opensearch_data_storage_detaching(self, _: StorageDetachingEvent):  # no
             elif self.app.planned_units() == 0 and self.model.get_relation(PeerRelationName):
                 self.peers_data.delete(Scope.APP, "bootstrap_contributors_count")
                 self.peers_data.delete(Scope.APP, "nodes_config")
+                self.peers_data.delete(Scope.APP, "security_index_initialised")
+                self.peers_data.delete(Scope.APP, "bootstrapped")
```
Mehdi-Bendriss: what of the case of large deployments where there are still applications in the fleet? How would that react? Could you check?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1421,8 +1423,8 @@ def _put_or_update_internal_user_leader(
         hashed_pwd, pwd = generate_hashed_password(pwd)
 
         # Updating security index
-        # We need to do this for all credential changes
-        if secret:
+        # We need to do this for all credential changes if the node is up
+        if secret and self.opensearch.is_node_up():
             self.user_manager.update_user_password(user, hashed_pwd)
```
Mehdi-Bendriss: I think this contradicts the flow of the `set-password` action where we rely on the raised `OpenSearchError` in [`update_user_password`](https://github.com/canonical/opensearch-operator/blob/2/edge/lib/charms/opensearch/v0/opensearch_users.py#L301) to fail the action. [here ](https://github.com/canonical/opensearch-operator/blob/2/edge/lib/charms/opensearch/v0/opensearch_base_charm.py#L804-L805)

What issue are you trying to fix here?

Mehdi-Bendriss (commented):
Thanks Smail! I left a couple of comments we need to clarify

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Scaling up from 0 means that the secret exists, but OS has not been started yet. In other words, the secret exists independently of the OS service being up or not.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: It is my understanding that large deployments will have different cases to handle. We only clean the `peer` (only the application being scaled to 0) databag. The other apps will need to react based on the deployment type and the situation (failover promotion, blocking because there are no cluster managers, etc).
I believe the tests should be done after the failover rework is merged. WDYT @Mehdi-Bendriss ?

smethnani (commented):
Thanks Smail! I've tested the case of a single app and was able to scale down to 0 and back up with no issues.

In the case of large deployments, I've noticed an issue with other cluster managers not reporting themselves in `self._fetch_local_cm_nodes`. This then sets [this error](https://github.com/canonical/opensearch-operator/blob/f6fa53ee56d634d4140e8f90e3a4f28efa0dece1/lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py#L563) in the data bag.

I believe the issue is that the admin password is incorrect and `self._opensearch.is_node_up()` failing because it tries to communicate with itself but gets a 401 reponse code and message "Unauthorized".

You've mentioned in a separate comment that the large deployments case should be handled differently. I agree we should discuss the topic a bit further!

smethnani (dismissed):
Thanks @skourta! I tested a case of a large deployment with a main orchestrator and data node. The main orchestrator can be scaled to 0 and starts back up when re-using the same storage.

For the case of large deployments with a failover that promotes itself, the main orchestrator does not start on scale up, which is currerntly expected behaviour as the main orchestrator doesn't demote itself on scale to 0. That'll be addressed in a separate PR and tracked [here](https://warthogs.atlassian.net/browse/DPE-7690)

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1426,7 +1438,11 @@ def _purge_users(self):
                 self.opensearch.config.delete("opensearch-security/internal_users.yml", user)
 
     def _put_or_update_internal_user_leader(
-        self, user: str, pwd: Optional[str] = None, update: bool = True
+        self,
+        user: str,
+        pwd: Optional[str] = None,
+        update: bool = True,
+        update_password_in_opensearch: bool = True,
```
Mehdi-Bendriss: Looking at the logic, why not simply leverage the previous argument `update`?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -498,7 +498,11 @@ def _put_fleet_apps(
             app=deployment_desc.app,
             planned_units=self.charm.app.planned_units(),
             units=[format_unit_name(u, app=deployment_desc.app) for u in all_units(self.charm)],
-            roles=deployment_desc.config.roles,
+            roles=(
+                deployment_desc.config.roles
+                if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES
+                else ClusterTopology.generated_roles()
```
Mehdi-Bendriss, line 504: nice

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -926,6 +930,16 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
+            # check if valid data is present if so update the seed hosts
+            if data.get("data"):
+                # In case the main orchestrator was scaled down to 0 and back
+                # we need to update the seed hosts with the data from the relation
+                # to pick up the new IPs and enable the data node see it
+                logger.debug(
+                    "Error from provider but valid data found in relation data, updating seed hosts."
+                )
+                data = self.peer_cm.rel_data_from_str(data["data"])
+                self.charm.opensearch_config.add_seed_hosts([node.ip for node in data.cm_nodes])
```
Mehdi-Bendriss: Why not leveraging `self.charm.opensearch_peer_cm.run_with_relation_data(data)` to be broader than just targeting the seed hosts?

Mehdi-Bendriss (commented):
Thanks Smail! I have a few questions / comments

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: Nothing more than to limit the scope of changes.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Looking closely at the code, I think you are right, we can use it. I will run tests to confirm. Initially, I wanted the node to set the state of `admin_user_initialized` later on, but I ended up not clearing that state from the databag so it should be there from the previous unit.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
Mehdi-Bendriss: But that's exactly the point, now that we're broadcasting success data even on error scenarios, do we want to limit the scope of changes?

skourta: Solid point. I updated the code.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: The tests work. I updated the code.

Mehdi-Bendriss (approved):
Thanks Smail!

smethnani (approved):

PR #637, opened 2025-05-21
fix: CI - start ssh service instead of reload

skourta:
This PR updates the SSH service management command from `reload` to `start` in the spread file.


Mehdi-Bendriss (approved):
Thank you !

smethnani (approved):
Thank you Smail! The failing tests look like env issues, minus the backups test which should have an additional guard for when the deployment description isn't ready

PR #643, opened 2025-06-02
fix: Prepare the charm to handle chain order

### LINKED ISSUE #641: Wrong certificate chain order in Opensearch
The charm expects the certificate to arrive in a reversed order and then handles that [here](https://github.com/canonical/opensearch-operator/blob/2/edge/lib/charms/opensearch/v0/opensearch_tls.py#L228).

This will break if it ever gets the chain in the correct order.

I assume this came from the fact that `self-sgined-certificates` was passing the chain in a reversed order.
This has been fixed in `self-sgined-certificates`, the chain will be in the correct order from the revision `310 (1/edge)` and beyond, which will break OpenSearch.

saltiyazan:
Fixes #641 
## Issue
The charm currently depend on the certificate chain being provided in a reverse order, this is a bug, the charm will fail to set up TLS if integrated with a provider that provides the chain in the correct order.

## Solution
A new patch of the TLS library provides a public method to check the order of the chain, that function is leveraged here to reverse the chain order only if the order is wrong. (This doesn't take into account a completely random order of the chain however)

## Checklist

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_tls.py
+++ b/lib/charms/opensearch/v0/opensearch_tls.py
@@ -45,6 +45,7 @@ from charms.tls_certificates_interface.v3.tls_certificates import (
     CertificateExpiringEvent,
     CertificateInvalidatedEvent,
     TLSCertificatesRequiresV3,
+    chain_has_valid_order,
     generate_csr,
     generate_private_key,
 )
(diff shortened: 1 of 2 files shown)


saltiyazan:
Not sure what the second point in the checklist means. "I have cleaned any remaining cloud resources from my accounts."

Issue #631, opened 2025-04-30
Data node fails to start because it never receives the truststore password from main orchestrator


skourta:
The data node fails to start because it never receives the truststore password from main orchestrator in a cluster manager only setting of a large deployment.

## Steps to reproduce

1. juju deploy self-signed-certificates --channel=latest/stable
2. juju deploy ./opensearch_ubuntu@22.04-amd64.charm main --config profile=testing --config cluster_name="test" --config roles="cluster_manager"
3. juju deploy ./opensearch_ubuntu@22.04-amd64.charm failover --config profile=testing --config cluster_name="test" -n 1 --config roles="cluster_manager,data" --config init_hold=True
4. juju deploy ./opensearch_ubuntu@22.04-amd64.charm data --config profile=testing --config cluster_name="test" -n 1 --config roles="data" --config init_hold=True
5. juju integrate self-signed-certificates main
6. juju integrate self-signed-certificates failover
7. juju integrate self-signed-certificates data
8. juju integrate failover:peer-cluster main
9. juju integrate data:peer-cluster main
10. juju integrate data:peer-cluster failover

## Expected behavior
The data node should start and initialize the security index to enable the main orchestrator to start.

## Actual behavior

<img width="1721" alt="Image" src="https://github.com/user-attachments/assets/175736db-335c-43ba-acce-c6f9c89631d6" />

Data node is blocked waiting for the truststore password

## Versions

Operating system: `Ubuntu 24.04.2 LTS`

Juju CLI: `3.6.5-genericlinux-amd64`

Juju agent: `3.6.4`

Charm revision: `2/edge` branch

LXD: `5.21.3 LTS`

## Log output

Juju debug log: 

[main.log](https://github.com/user-attachments/files/19974690/main.log)
[failover.log](https://github.com/user-attachments/files/19974689/failover.log)
[data.log](https://github.com/user-attachments/files/19974688/data.log)

reneradoi:
may be related: during the initial implementation of `data-only` nodes, there was a proposal to "broadly ignore" errors on the main orchestrator side to increase robustness in such kind of race conditions, see [here](https://github.com/canonical/opensearch-operator/pull/424#discussion_r1746061957) and [here](https://github.com/canonical/opensearch-operator/blob/0ee4749c347cfbebe32044b9099882951ba3d424/lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py#L446). Could be discussed again how to handle these cases.

PR #647, opened 2025-06-10
Disable remove-orchestrator tests

smethnani:
This PR disables the `large_deployments_remove_orchestrators` tests while the #631 await a fix

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -42,6 +42,8 @@ jobs:
               _, runner, task, variant = job.split(":")
               # Example: "test_charm.py"
               task = task.removeprefix("tests/spread/")
+              if "remove_orchestrators" in task:
+                continue
               # For IS-hosted runners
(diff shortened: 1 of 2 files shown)


Mehdi-Bendriss (approved):
Thank you Shams!

.github/workflows/integration_test.yaml
```diff
@@ -42,6 +42,8 @@ jobs:
               _, runner, task, variant = job.split(":")
               # Example: "test_charm.py"
               task = task.removeprefix("tests/spread/")
+              if "remove_orchestrators" in task:
```
skourta, line 45: Can you please add a TODO comment as a reminder?

skourta (approved):
Thank you @smethnani

.github/workflows/integration_test.yaml
smethnani: It'll be tracked [here](https://warthogs.atlassian.net/browse/DPE-7527)

PR #650, opened 2025-06-11
[DPE-7487] WaitingForSpecificBusyShards status clearing

### LINKED ISSUE #645: Unit health status not updated after WaitingForSpecificBusyShards is set and resolved
On high latency environments, the `WaitingForSpecificBusyShards` waiting status may be displayed at the unit level. 

This status doesn't resolve itself even when shards have completely started.

To fix this issue, the status should not be exclsuively set by the leader unit on_peer_relation_changed and on update status.

smethnani:
## Issue

When the WaitingForSpecificBusyShards waiting status is displayed at the unit level, it sometimes doesn't get cleared.

## Solution
 The status should not be exclsuively set by the leader unit on_peer_relation_changed and on update status

closes #645/[DPE-7487]. 

[DPE-7487]: https://warthogs.atlassian.net/browse/DPE-7487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reneradoi (commented):
May be a stupid question or because I'm out of the loop a little in Opensearch, but why not using `health.apply(app=False)` instead of adding `unit` as an option?

smethnani:
@reneradoi we'd like the leader to be able to apply the status at the unit level and at the app level, not exclusively between them

reneradoi (approved):
Thank you Shams!

Mehdi-Bendriss (approved):
Thanks Shams!

PR #653, opened 2025-06-13
[DPE-7566] - Separate data and error fields in peer-cluster relation data

smethnani:
## Issue
The peer-cluster-orchestrator relation data currently sets either an error or data. This causes some issues when applications need to access information in the data field but the orchestrator has set the error field.

## Solution
Separate the error and data creation and send both, unless the relation should be severed, in which case it should only send the error data.

Addresses [DPE-7566]

[DPE-7566]: https://warthogs.atlassian.net/browse/DPE-7566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -453,18 +462,18 @@ def refresh_relation_data(  # noqa: C901
...
+        rel_err_data: Optional[PeerClusterRelErrorData],
         target_relation_ids: List[int],
     ) -> bool:
         """Check if relation is invalid and notify related sub-clusters."""
-        if not isinstance(rel_data, PeerClusterRelErrorData):
+        if rel_err_data is None:
             return False
 
-        if not rel_data.should_sever_relation:
+        if not rel_err_data.should_sever_relation:
```
skourta: nit: 
You can group the conditions to `if rel_err_data is None or not rel_err_data.should_sever_relation:`

skourta (dismissed):
Looks good to me.
I tested one of the fixes I am currently working on.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -379,18 +379,28 @@ def refresh_relation_data(  # noqa: C901
...
+            rel_data = self._rel_data(deployment_desc)
+            # replace the plaintext credentials in
+            # rel_data with their corresponding secret IDs
             rel_data_redacted_dict = self._rel_data_redacted_dict(rel_data)
 
             # grant the secrets inside the rel_data to all the related clusters
             self._grant_rel_data_secrets(rel_data_redacted_dict, all_relation_ids)
 
+        rel_err_data = self._rel_err_data(deployment_desc, orchestrators)
+        if rel_err_data is None and len(rel_data.cm_nodes) == 0:
```
Mehdi-Bendriss: nit:
```suggestion
        if not rel_err_data and not rel_data.cm_nodes:
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -379,18 +379,28 @@ def refresh_relation_data(  # noqa: C901
...
         # compute the data that needs to be broadcast to all related clusters (success or error)
         # if rel_data is an error, prepare to broadcast it to all related clusters
-        rel_data = self._rel_data(deployment_desc, orchestrators)
-
-        # if rel_data is NOT an error, we will replace the plaintext credentials in
-        # the object, with their corresponding secret IDs
-        if isinstance(rel_data, PeerClusterRelData):
+        rel_data_redacted_dict = None
+        if self.charm.is_admin_user_configured():
+            rel_data = self._rel_data(deployment_desc)
```
Mehdi-Bendriss: Can you explain why is the creation of `rel_data` contingent on whether `self.charm.is_admin_user_configured()`?
When this evaluates to False, the condition: `len(rel_data.cm_nodes)` will crash.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
Mehdi-Bendriss: I would rather that the construction of rel_err_data happens in a method instead of a mix between `_rel_err_data` and the current method

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -594,19 +603,20 @@ def _s3_credentials(
     def _rel_data(
         self,
         deployment_desc: DeploymentDescription,
-        orchestrators: PeerClusterOrchestrators,
     ) -> Union[PeerClusterRelData, PeerClusterRelErrorData]:
```
Mehdi-Bendriss: This could be adjusted to return a Tuple instead (related to a previous comments on multiple locations creating the err_data)

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -594,19 +603,20 @@ def _s3_credentials(
...
-
         # check that this cluster is fully ready, otherwise put "configuring" in
         # peer rel data for requirers to show a blocked status until it's fully
         # ready (will receive a subsequent
+        cm_nodes = []
         try:
+            cm_nodes = self._fetch_local_cm_nodes(deployment_desc)
+        except OpenSearchHttpError:
+            logger.warning(f"Could not fetch nodes in related {deployment_desc.typ} sub-cluster")
+        finally:
```
Mehdi-Bendriss: The `finally` is not needed. We can have a simple return outside of the try/except block

Mehdi-Bendriss (commented):
Thanks Shams, I left a few comments

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: `rel_error_data` would not be `None` if `self.charm.is_admin_user_configured()` returns false so `len(rel_data.cm_nodes)` would never be reached. The idea was we can't populate `PeerClusterRelDataCredentials` before the admin user is initialised but perhaps it'd be better to just have `rel_data.credentials` be contingent on `self.charm.is_admin_user_configured()`

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: I missed this, thank you!. It should just be returning `PeerClusterRelData`

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: Removed

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: I moved this to the `rel_err_data` creation method, the `rel_data` is also passed as an argument and the appropriate error message is filled there if cm_nodes is empty

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: I moved the `if self.charm.is_admin_user_configured()` to a new method which returns the peer rel data credentials to make this more obvious. If it can't be filled then `rel_data` will be None

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: Moved this to the `rel_err_data` creation method

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: Thanks! Done

Mehdi-Bendriss (dismissed):
This is clean, I like this change. Thank you Shams!

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -601,50 +587,61 @@ def _s3_credentials(
...
         self,
-        deployment_desc: DeploymentDescription,
-    ) -> Union[PeerClusterRelData, PeerClusterRelErrorData]:
+        deployment_desc: Optional[DeploymentDescription],
+    ) -> Optional[PeerClusterRelData]:
         """Build and return the peer cluster rel data to be shared with requirer sub-clusters."""
-        # check that this cluster is fully ready, otherwise put "configuring" in
-        # peer rel data for requirers to show a blocked status until it's fully
-        # ready (will receive a subsequent
+        # returns non if this cluster is not fully ready, or if the admin user
```
skourta: typo

skourta (dismissed):
Thank you @smethnani .
Just one small typo

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: Thank you! Fixed

Mehdi-Bendriss (approved):

skourta (approved):

PR #662, opened 2025-06-18
[DPE-7574] Fix interpolated pattern match

smethnani:
## Issue
The `charm.status.clear` regex check is matching the template message against itself rather than the status message for the case of `Status.CheckPattern.Interpolated`

## Solution
Match against `context.status.message`


Mehdi-Bendriss (approved):
Good catch! Thank you Shams!

reneradoi (approved):
Nice find, thank you Shams!

PR #665, opened 2025-06-24
Change Grafana channel to 1/stable

smethnani:
## Issue
The plugins tests are failing to deploy`grafana-agent` because the channel that we use has been removed. 

## Solution
We should move to an existing channel. This PR changes the channel to `1/stable`


Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):
Good catch, thanks Shams!

PR #683, opened 2025-08-08
[DPE-7880] Wrap credential assignments for backwards compatibility

smethnani:
The credentials in the peer cluster relation data were not passed as secrets in older revisions, breaking upgrades from those revisions in large deployments. 

This PR adds a wrapper to resolve the credentials as secrets only if they are uri's using the `secret://` scheme.

The value in `credentials["admin_tls"]` was previously a dictionary, not a string. The case is handled directly in `rel_data_from_str` instead of using the wrapper.

Addresses [DPE-7880]

[DPE-7880]: https://warthogs.atlassian.net/browse/DPE-7880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_peer_clusters.py
+++ b/lib/charms/opensearch/v0/opensearch_peer_clusters.py
@@ -680,75 +680,77 @@ class OpenSearchPeerClustersManager:
             )
         return PeerClusterRelData.from_dict(content)
 
+    def _resolve_credential(
+        self,
+        credential: str,
+        *,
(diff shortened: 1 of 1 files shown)


skourta (approved):
Thank you, @smethnani. Great find!

Mehdi-Bendriss (approved):
Thank you Shams! Great work!

PR #685, opened 2025-08-18
[DPE-7832] - Rework profiles

skourta:
This pull request refactors the way OpenSearch performance profiles are managed and applied throughout the codebase. The main changes involve removing the legacy `OpenSearchPerfProfile` model, consolidating profile management logic into a new `ProfilesManager`, and updating both configuration and documentation to reflect the removal of the "staging" profile and the new default behaviors.

**Performance profile management refactor:**

* Removed the `OpenSearchPerfProfile` class and all related logic from `lib/charms/opensearch/v0/models.py`, including constants for heap size and the "staging" profile option. The new approach uses the `ProfilesManager` for profile handling. [[1]](diffhunk://#diff-fd5b4ae87f76121ac1429c6ef3b3350ca00b3a9395cf2a26a383580a8bb2e60aL549-L605) [[2]](diffhunk://#diff-fd5b4ae87f76121ac1429c6ef3b3350ca00b3a9395cf2a26a383580a8bb2e60aL170) [[3]](diffhunk://#diff-fd5b4ae87f76121ac1429c6ef3b3350ca00b3a9395cf2a26a383580a8bb2e60aL28-L32)
* Updated `lib/charms/opensearch/v0/opensearch_base_charm.py` to use `ProfilesManager` and `OpenSearchClusterState` for profile management, replacing the previous `performance_profile` logic. Profile requirements are now checked and enforced via `ProfilesManager`, and JVM heap size is set directly via the new method. [[1]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43L83-R85) [[2]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R225-R227) [[3]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43L267) [[4]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43L708-R721) [[5]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43L848-R871) [[6]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R1155-R1160)

**Configuration and API updates:**

* Changed the default `profile` value in `config.yaml` from `"production"` to `"testing"` and removed references to the "staging" profile in both configuration and documentation.
* Updated `lib/charms/opensearch/v0/opensearch_config.py` to remove use of `OpenSearchPerfProfile` and refactored the method `set_jvm_heap_size` to use the size directly. [[1]](diffhunk://#diff-051187028ed5323ea871a59612ae7a65952495f9bf26c43df5bd680d47c8527eL12-R12) [[2]](diffhunk://#diff-051187028ed5323ea871a59612ae7a65952495f9bf26c43df5bd680d47c8527eL111-L129)

**Documentation updates:**

* Updated documentation in `docs/how-to/h-large-deployment.md` and `docs/how-to/h-minor-upgrade.md` to remove the "staging" profile, adjust RAM limits for "production" profile from 32G to 31G, and reflect the new default of "testing". [[1]](diffhunk://#diff-0819b46f37ba83761171ae2b4741b167263c025f0cf48e612a5e9dbed9807167L70-R73) [[2]](diffhunk://#diff-a1fd4f4805e6cc2970d7e39fdb96221071beadbde04afcde3a5fcdd2eab65703L125-R125)

**Error handling and startup logic:**

* Improved error handling and logging in the OpenSearch startup and upgrade routines to reflect the new profile management process and ensure requirements are checked before starting the service. [[1]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R1135-L1130) [[2]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R1188) [[3]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R1200) [[4]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43L1454-R1476)


.gitignore
```diff
@@ -63,3 +63,4 @@ override.tf.json
 # Ignore CLI configuration files
 .terraformrc
 terraform.rc
+.env
```
gatici: Need EOL here.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -589,6 +593,13 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):  # noqa C901
             contributor_count = self.peers_data.get(Scope.APP, "bootstrap_contributors_count", 0)
             self.peers_data.put(Scope.APP, "bootstrap_contributors_count", contributor_count + 1)
 
+        # check requirements
+        if self.state.unit.is_started and (
+            missing_requirements := self.profiles_manager.check_all_requirements()
+        ):
+            logger.error(f"Missing profile requirements: {missing_requirements}")
```
gatici: Using f-trings in the logging is not recommended.  Even the log level setting is different, f-string will be formatted which means the line runs every time. However, parameterized logging line only runs for the appropriate log levels.  This is also included in the charm style guide: https://docs.google.com/document/d/1H5W2oi7cuTaz86taTQiITv49jpJNKzzDZ2G25IQpE4w/edit?tab=t.0

This command applies to several logging lines.

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,286 @@
+# Copyright 2024 Canonical Ltd.
```
gatici:
```suggestion
# Copyright 2025 Canonical Ltd.
```

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,286 @@
...
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 2
+
+
+logger = logging.getLogger(__name__)
+
+
+_1GB_IN_KB = 1024 * 1024  # 1GB in KB
+MAX_HEAP_SIZE = 31 * _1GB_IN_KB  # 32GB in KB
```
gatici:
```suggestion
MAX_HEAP_SIZE = 31 * _1GB_IN_KB  # 31GB in KB
```

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,286 @@
...
+        """Apply a system requirement."""
+        try:
+            self.workload._run_cmd(f"sysctl -w {system_requirement}={value}")
+            return int(self.workload._run_cmd(f"sysctl -n {system_requirement}")) == value
+        except OpenSearchCmdError:
+            return False
+
+    def _get_kernel_property_value(self, prop: str) -> int:
+        """Get the value of a kernel parameter."""
+        return int(self.workload._run_cmd(f"sysctl -n {prop}"))
```
gatici: We are not handling errors here, I will suggest to use a try except block not to put charm in unexpected error state.

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,286 @@
...
+        if self._get_kernel_property_value(prop) < val and not self._apply_system_requirement(
+            prop, val
+        ):
+            missing_requirements.append(f"{prop} should be at least {val}")
+
+        prop, val = "vm.swappiness", 1
+        if self._get_kernel_property_value(prop) > val and not self._apply_system_requirement(
+            prop, 0
+        ):
+            missing_requirements.append(f"{prop} should be at most 1")
```
gatici: I see an inconsistency between what we suggest and do here.  We say prop should be at most 1 and  set it to 0.

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,286 @@
...
+    def check_memory_requirements(self, profile: OpenSearchProfile) -> List[str]:
+        """Checks memory requirements for the unit."""
+        memory_size = self.workload.meminfo()["MemTotal"]
+
+        if (
+            profile.memory_requirements.memory_size
+            and memory_size < profile.memory_requirements.memory_size
+        ):
+            logger.error(
+                f"Insufficient memory: {memory_size} < {profile.memory_requirements.memory_size}"
```
gatici: We should not use f-strings while logging.

lib/charms/opensearch/v0/state.py
```diff
@@ -0,0 +1,109 @@
...
+    def deployment_description(self) -> DeploymentDescription | None:
+        """Return DeploymentDescription from peer relation"""
+        deployment_desc_dict = self.relation_data.get_object(self.scope, "deployment-description")
+        if not deployment_desc_dict:
+            return None
+        return DeploymentDescription.from_dict(deployment_desc_dict)
+
+
+class OpenSearchUnit:
+    """State/Relation data collection for an opensearch node (juju uni)."""
```
gatici:
```suggestion
    """State/Relation data collection for an opensearch node (juju unit)."""
```

lib/charms/opensearch/v0/state.py
```diff
@@ -0,0 +1,109 @@
...
+
+class OpenSearchClusterState(Object):
+    """Global state object for an opensearch cluster"""
+
+    def __init__(self, charm: "OpenSearchBaseCharm"):
+        super().__init__(parent=charm, key="charm_state")
+        self.charm = charm
+        self.config = charm.config
+
+    @property
```
gatici:
```suggestion
   @cached_property 
```

We can use  @cached_property from  functools  to avoid creating fresh wrappers in each access.

lib/charms/opensearch/v0/state.py
```diff
@@ -0,0 +1,109 @@
...
+    def app(self) -> OpenSearchApp:
+        """Get state of the local opensearch app."""
+        return OpenSearchApp(
+            charm=self.charm,
+        )
+
+    @property
```
gatici: Same here.

tests/integration/ha/test_large_deployments_remove_orchestrators.py
```diff
@@ -189,7 +189,7 @@ async def test_large_deployment_remove_orchestrators(ops_test: OpsTest) -> None:
         apps_full_statuses={
             DATA_APP: {"blocked": [PClusterOrchestratorsRemoved]},
         },
-        units_statuses=["active"],
+        units_statuses=["blocked"],
```
gatici: I did not get why this needs to be changed as blocked.

gatici (commented):
Generally looks good. I commented for some minor issues.
Review is not finished yet, checking tests.

gatici:
In test_opensearch_peer_clusters.py we have a test as below which sets profile attribute. However I do not see this attribute in the class definition: 

 PeerClusterConfig does not have profile parameter.
```
class PeerClusterConfig(Model):
    """Model class for the multi-clusters related config set by the user."""

    cluster_name: str
    init_hold: bool
    roles: List[str]
    # We have a breaking change in the model
    # For older charms, this field will not exist and they will be set in the
    # profile called "testing".
    data_temperature: Optional[str] = None
```

Test class:
```
class TestOpenSearchPeerClustersManager(unittest.TestCase):
    BASE_LIB_PATH = "charms.opensearch.v0"
    BASE_CHARM_CLASS = f"{BASE_LIB_PATH}.opensearch_base_charm.OpenSearchBaseCharm"
    PEER_CLUSTERS_MANAGER = (
        f"{BASE_LIB_PATH}.opensearch_peer_clusters.OpenSearchPeerClustersManager"
    )

    user_configs = {
        "default": PeerClusterConfig(
            cluster_name="", init_hold=False, roles=[], profile="production"
        ),
        "name": PeerClusterConfig(
            cluster_name="logs", init_hold=False, roles=[], profile="production"
        ),
        "init_hold": PeerClusterConfig(
            cluster_name="", init_hold=True, roles=[], profile="production"
        ),
        "roles_ok": PeerClusterConfig(
            cluster_name="",
            init_hold=False,
            roles=["cluster_manager", "data"],
            profile="production",
        ),
        "roles_ko": PeerClusterConfig(
            cluster_name="", init_hold=False, roles=["data"], profile="production"
        ),
        "roles_temp": PeerClusterConfig(
            cluster_name="", init_hold=True, roles=["data.hot"], profile="production"
        ),
    }

    p_units = [
        PatchedUnit(name="opensearch/0"),
        PatchedUnit(name="opensearch/1"),
        PatchedUnit(name="opensearch/2"),
        PatchedUnit(name="opensearch/3"),
        PatchedUnit(name="opensearch/4"),
    ]
 ```

**Comment by gatici** [comment]
> Generally looks good. I commented for some minor issues. Review is not finished yet, checking tests.

Tests are fine. No more comments for tests.

File: lib/charms/opensearch/v0/opensearch_profile.py (hunk 5 above)
Comment (skourta) [reply lib/charms/opensearch/v0/opensearch_profile.py]: I think it is ok to raise in this case. If we cannot execute a sysctl command on the unit then we should raise an error to try again.

File: lib/charms/opensearch/v0/state.py (hunk 9 above)
Comment (skourta) [reply lib/charms/opensearch/v0/state.py]: Thank you I didn't know these existed.

File: tests/integration/ha/test_large_deployments_remove_orchestrators.py (hunk 11 above)
Comment (skourta) [reply tests/integration/ha/test_large_deployments_remove_orchestrators.py]: This is because of the new testing profile will block the unit if it detects that there are no cluster manager nodes. In testing the minimal number of CM nodes is set to 1.

Comment (gatici) [reply tests/integration/ha/test_large_deployments_remove_orchestrators.py]: OK

**Review by gatici — DISMISSED** [review]
LGTM

File: lib/charms/opensearch/v0/opensearch_profile.py [hunk 12]
```diff
@@ -0,0 +1,289 @@
...
+        return ProfileMemoryRequirements(
+            memory_size=None,
+            jvm_heap_percentage=None,
+        )
+
+    @property
+    def cluster_topology_requirements(self) -> ClusterTopologyRequirements:
+        """Get the cluster topology requirements for this profile."""
+        return ClusterTopologyRequirements(
+            cluster_managers=1,
```
smethnani, line 145: This currently conflicts with the `opensearch_peer_clusters::validate_recommended_cm_count` which sets a blocked status if the number of CM units is less than 3 on scale down. We should remove that now that we have this imo (in a future PR).

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,289 @@
...
+                % (memory_size, profile.memory_requirements.memory_size)
+            ]
+
+        return []
+
+    def check_cluster_topology(self, profile: OpenSearchProfile) -> List[str]:
+        """Check the cluster topology requirements."""
+        cluster_fleet_apps = self.state.app.cluster_fleet_apps
+        current_app = self._current_peer_cluster_app()
+        # backwards compatibility for revisions that do not set generated roles
```
smethnani, line 194: 👍

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,289 @@
...
+    Ensures cluster meets production minimal requirements
+    """
+
+    type = PerformanceType.PRODUCTION
+
+    @property
+    def memory_requirements(self) -> ProfileMemoryRequirements:
+        """Get the memory requirements for this profile."""
+        return ProfileMemoryRequirements(
+            memory_size=8 * _1GB_IN_KB,  # 4GB in KB
```
smethnani:
```suggestion
            memory_size=8 * _1GB_IN_KB,  # 8GB in KB
```

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,289 @@
...
+        missing_requirements.extend(self.check_memory_requirements(profile))
+        missing_requirements.extend(self.check_cluster_topology(profile))
+
+        return missing_requirements
+
+    def get_config_profile(self) -> OpenSearchProfile:
+        """Get the current config profile."""
+        return (
+            ProductionProfile()
+            if PerformanceType(self.state.config.get("profile")) == PerformanceType.PRODUCTION
```
smethnani: I think it would be good to validate the input here. A typo from the user would cause a crash.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1139,6 +1165,12 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
             event.defer()
             return
 
+        profile = self.profiles_manager.profile
+        self.opensearch_config.set_jvm_heap_size(
+            profile.get_jvm_heap_size(self.opensearch.meminfo()["MemTotal"])
+        )
+        self.state.unit.profile = profile
```
smethnani: I think this may be overwriting the unit profile with stale values set during init in the manager (I'm not sure dispatch is called for custom events but I would check if this what's causing the config inconsistency I'm seeing in tests).

smethnani (commented):
Thank you Smail! Great work!

There seems to be a bug in changing and applying profiles. I lowered the min memory to 2 and jvm_heap_percentage to 0.1 for the production profile and ran a simple deployment with 3 units. The `-Xms/x` for all nodes from /_nodes/jvm was 10% of what meminfo showed in the juju units.

I then changed the profile to testing. In the peers data, 2 units had profile=testing and 1 still had profile=production. Only 1 node showed Xms and Xmx change to 1g.

Before changing config:
```
# peers data
{
  "main/0": {
    "local-unit": {
      "data": {
        "profile": "production"
      }
    },
    "related-units": {
      "main/1": {
        "data": {
          "profile": "production"
        }
      },
      "main/2": {
        "data": {
          "profile": "production"
        }
      }
    }
  }
}

# GET /_nodes/jvm
{
  "name": "main-0.cbf",
  "-Xms": "-Xms3208746k",
  "-Xmx": "-Xmx3208746k"
}
{
  "name": "main-2.cbf",
  "-Xms": "-Xms3208746k",
  "-Xmx": "-Xmx3208746k"
}
{
  "name": "main-1.cbf",
  "-Xms": "-Xms3208746k",
  "-Xmx": "-Xmx3208746k"
}
```
After config=testing
```
# peers data
{
  "main/0": {
    "local-unit": {
      "data": {
        "profile": "production"
      }
    },
    "related-units": {
      "main/1": {
        "data": {
          "profile": "testing"
        }
      },
      "main/2": {
        "data": {
          "profile": "testing"
        }
      }
    }
  }
}


# GET /_nodes/jvm
{
  "name": "main-0.cbf",
  "-Xms": "-Xms3208746k",
  "-Xmx": "-Xmx3208746k"
}
{
  "name": "main-2.cbf",
  "-Xms": "-Xms3208746k",
  "-Xmx": "-Xmx3208746k"
}
{
  "name": "main-1.cbf",
  "-Xms": "-Xms1048576k",
  "-Xmx": "-Xmx1048576k"
}
```

I then changed config back to production:
```
# peers data
{
  "main/0": {
    "local-unit": {
      "data": {
        "profile": "production"
      }
    },
    "related-units": {
      "main/1": {
        "data": {
          "profile": "production"
        }
      },
      "main/2": {
        "data": {
          "profile": "testing"
        }
      }
    }
  }
}

# /_nodes/jvm
{
  "name": "main-0.cbf",
  "-Xms": "-Xms3208746k",
  "-Xmx": "-Xmx3208746k"
}
{
  "name": "main-2.cbf",
  "-Xms": "-Xms1048576k",
  "-Xmx": "-Xmx1048576k"
}
{
  "name": "main-1.cbf",
  "-Xms": "-Xms3208746k",
  "-Xmx": "-Xmx3208746k"
}
```
Logs: https://pastebin.canonical.com/p/TvndpwkKM4/

lib/charms/opensearch/v0/opensearch_profile.py
skourta: If there's any typo we would default to testing as it is the safest profile. WDYT @Mehdi-Bendriss should we add validation, block and report?

Mehdi-Bendriss: For invalid config options we should rather put the charm in a Blocked state to get the user to set a correct value. 
In the meantime, we should not change the previous (correct) value set in the charm

skourta: I just finished a call with @gatici about validating config option values. I think she has a very elegant proposition that would fit perfectly the DP charm pattern.

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -589,6 +593,13 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):  # noqa C901
             contributor_count = self.peers_data.get(Scope.APP, "bootstrap_contributors_count", 0)
             self.peers_data.put(Scope.APP, "bootstrap_contributors_count", contributor_count + 1)
 
+        # check requirements
+        if self.state.unit.is_started and (
```
Mehdi-Bendriss: I have 3 questions here:
1. Why check on whether the unit is started? 
2. Why not check it's started directly from the service? `self.opensearch.is_started` ? since this is the most accurate source of truth on the current state of the service 
3. Many of missing requirements could prevent the unit from starting, so this will never run

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -705,16 +716,18 @@ def _on_update_status(self, event: UpdateStatusEvent):  # noqa: C901
...
+        2- Do we have users that need to be deleted, and if so we need to delete them.
         3- every 6 hours check if certs are expiring soon (in 7 days),
             as a safeguard in case relation broken. As there will be data loss
             without the user noticing in case the cert of the unit transport layer expires.
             So we want to stop opensearch in that case, since it cannot be recovered from.
         """
-        # if there are missing system requirements defer
-        if len(missing_sys_reqs := self.opensearch.missing_sys_requirements()) > 0:
-            self.status.set(BlockedStatus(" - ".join(missing_sys_reqs)))
+        if self.state.unit.is_started and (
```
Mehdi-Bendriss: I'm not sure I understand the the need for the `is_started` check here since many missing requirements would prevent the unit from starting at all - therefore ever preventing the Blocked status to be set

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -845,18 +850,38 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
             if original_status:
                 self.status.set(original_status)
 
-        perf_profile_needs_restart = self.performance_profile.apply(
-            self.config.get(PERFORMANCE_PROFILE)
-        )
+        profile_restart_needed = False
+        if self.state.unit.is_started:
```
Mehdi-Bendriss: why is this check needed? This is not taking into account deploy-time passed config options (`juju deploy ... --config profile=...`)
Why not restrict this check (on the daemon as mentioned in a previous comment, not the relation data) to the restart part?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -845,18 +850,38 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
...
+            if missing_requirements:
+                logger.error(f"Missing profile requirements: {missing_requirements}")
+                self.status.set(BlockedStatus(" - ".join(missing_requirements)))
+                return
+
+            # if the profile hasn't been applied before
+            logger.debug(
+                "current profile: %s, config profile: %s", current_profile, config_profile
+            )
+            if current_profile is None or current_profile != config_profile:
```
Mehdi-Bendriss: 1. we need some validation in here
2. shouldn't the `set_jvm..` be better set in the profile manager?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1139,6 +1165,12 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
             event.defer()
             return
 
+        profile = self.profiles_manager.profile
+        self.opensearch_config.set_jvm_heap_size(
+            profile.get_jvm_heap_size(self.opensearch.meminfo()["MemTotal"])
+        )
```
Mehdi-Bendriss: similarly, this seems too "specific" to be set from here

lib/charms/opensearch/v0/opensearch_distro.py
```diff
@@ -569,3 +542,16 @@ def version(self) -> str:
         output = self.run_bin("opensearch-bin", "--version 2>/dev/null")
         logger.debug(f"version call output: {output}")
         return output.split(", ")[0].split(": ")[1]
+
+    def meminfo(self) -> dict[str, float]:
```
Mehdi-Bendriss, line 546: any argument against having this near the `check_missing_requirements`? in the same sense we check for swappiness and other kernel parameters

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,293 @@
...
+            return min(int(self.memory_requirements.jvm_heap_percentage * mem_size), MAX_HEAP_SIZE)
+        return _1GB_IN_KB
+
+    def __hash__(self):
+        """Get the hash of the profile."""
+        return hash(self.type)
+
+    def __eq__(self, value: object) -> bool:
+        """Check equality with another OpenSearchProfile."""
+        return self.type == value.type if isinstance(value, OpenSearchProfile) else False
```
Mehdi-Bendriss, line 97: why not extend Model to get this for free?

lib/charms/opensearch/v0/opensearch_profile.py
Mehdi-Bendriss: Good catch! I would argue we should remove it now, as we're introducing an inconsistency in the behavior

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,293 @@
...
+            self.workload._run_cmd(f"sysctl -w {system_requirement}={value}")
+            return int(self.workload._run_cmd(f"sysctl -n {system_requirement}")) == value
+        except OpenSearchCmdError:
+            return False
+
+    def _get_kernel_property_value(self, prop: str) -> int:
+        """Get the value of a kernel parameter."""
+        return int(self.workload._run_cmd(f"sysctl -n {prop}"))
+
+    def check_missing_system_requirements(self) -> List[str]:
```
Mehdi-Bendriss, line 166: A note on these, given the length of the potential missing system requirements exceeding the juju status 120 chars limit, we need to log this as an error (in addition to the BlockedStatus)

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def get_config_profile(self) -> OpenSearchProfile:
+        """Get the current config profile."""
+        return (
+            ProductionProfile()
+            if PerformanceType(self.state.config.get("profile")) == PerformanceType.PRODUCTION
+            else TestingProfile()
+        )
+
+    def _current_peer_cluster_app(self) -> PeerClusterApp:
```
Mehdi-Bendriss, line 259: It's not clear to me why is this needed to be manually constructed?

lib/charms/opensearch/v0/state.py
```diff
@@ -0,0 +1,110 @@
...
+        return None
+
+    @profile.setter
+    def profile(self, new_profile: OpenSearchProfile):
+        """Set the performance profile for the unit."""
+        logger.debug("Setting performance profile: %s", new_profile.type.value)
+        self.relation_data.put(self.scope, PERFORMANCE_PROFILE, new_profile.type.value)
+
+    @property
+    def is_started(self) -> bool:
```
Mehdi-Bendriss: I'm not convinced by this as a state construct, when we can easily tap in the source for an accurate state

src/charm.py
```diff
@@ -139,12 +138,7 @@ def _set_upgrade_status(self):
 
     def _on_upgrade_charm(self, _):
         update_grafana_dashboards_title(self)
-        if not self.performance_profile.current:
-            # We are running (1) install or (2) an upgrade on instance that pre-dates profile
-            # First, we set this unit's effective profile -> 1G heap and no index templates.
-            # Our goal is to make sure this value exists once the refresh is finished
-            # and it represents the accurate value for this unit.
-            self.performance_profile.current = PerformanceType.TESTING
+        # TODO check backwards compatibility for profiles
```
Mehdi-Bendriss, line 141: what is the plan for this one? :)

tests/integration/ha/test_large_deployments_remove_orchestrators.py
Mehdi-Bendriss: Could you be explicit on the expected status message?

Mehdi-Bendriss (commented):
Thanks Smail! This rework was greatly needed! I left a few comments

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: The check on `is_started` is not to check on the workload it is to check if the `start_opensearch` method has been run successfully. In `start_opensearch` event we also do the same checks so I disabled the checks while we have executed that event yet. This is the same for `update_status` and `config_changed`.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: 1. I will add validation.
2. No, a manager cannot call another manager per our pattern. Managers only take state and workload as arguments. The managers are supposed to be called from the event handlers.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Same argument as before.

lib/charms/opensearch/v0/state.py
skourta: This is just a shortcut for what is already implemented. We are using the started flag in multiple operations. I consider this flag not an indicator of the workload being active but that we have ran the start event.

lib/charms/opensearch/v0/opensearch_profile.py
skourta: I wanted to add this to the state but we cannot do that because at the unit state we do not have access to the charm.

lib/charms/opensearch/v0/opensearch_distro.py
skourta: This is workload specific I believe.

src/charm.py
skourta: Remove it once we are done investigating upgrades.

lib/charms/opensearch/v0/opensearch_profile.py
skourta: Simplicity mostly.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: It is taken into account on the start opensearch event.

smethnani (commented):
Thanks Smail!

I tested again with the same set up as before. I set an invalid config option, which set the blocked status and message. However this status isn't immediately removed when the invalid config option is corrected:
```
juju config main profile="invalid"
```
All units go into blocked status with "Invalid profile configuration option"
```
juju config main profile="production"
```
Only one unit's status changes:

Screenshot 2025-09-17 at 11 07 43 AM

After the first unit has restarted, the second unit's status is cleared:

Screenshot 2025-09-17 at 11 08 08 AM

However the units are then all idle, and the final unit's status still has not been cleared, and will not be cleared until an update-status event occurs. The unit will also not set the new JVM options until the update status event occurs.

Before update-status:

Screenshot 2025-09-17 at 11 08 42 AM

```
# /_nodes/jvm
{
  "name": "main-0.9cd",
  "-Xms": "-Xms1048576k",
  "-Xmx": "-Xmx1048576k"
}
{
  "name": "main-1.9cd",
  "-Xms": "-Xms3208745k",
  "-Xmx": "-Xmx3208745k"
}
{
  "name": "main-2.9cd",
  "-Xms": "-Xms3208745k",
  "-Xmx": "-Xmx3208745k"
}
```
After update-status:
Status cleared and JVM options set correctly
```
{
  "name": "main-0.9cd",
  "-Xms": "-Xms3208745k",
  "-Xmx": "-Xmx3208745k"
}
{
  "name": "main-1.9cd",
  "-Xms": "-Xms3208745k",
  "-Xmx": "-Xmx3208745k"
}
{
  "name": "main-2.9cd",
  "-Xms": "-Xms3208745k",
  "-Xmx": "-Xmx3208745k"
}
```

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,274 @@
...
+            for app in cluster_fleet_apps.values()
+            if "cluster_manager" in app.roles
+        )
+        nbr_data_nodes = sum(
+            app.planned_units for app in cluster_fleet_apps.values() if "data" in app.roles
+        )
+
+        if nbr_cm_nodes < profile.cluster_topology_requirements.cluster_managers:
+            missing_requirements.append(
+                f"At least {profile.cluster_topology_requirements.cluster_managers} cluster manager nodes are required. Found only {nbr_cm_nodes}."
```
akram09: @Mehdi-Bendriss proposed to update this message to "At least: 3 cluster manager and 3 data nodes required." In the [How-To-Guide](https://docs.google.com/document/d/1yoGA-6_eVnR9U7QhpMQdakTNITnPj6J2W9Pbqiu40lE/edit?tab=t.0)

smethnani (dismissed):
Great work @skourta ! All manual tests I ran worked as expected. The units all restart in sequence and an invalid config message is cleared immediately when the config input is corrected.

gatici:
Hello Smail,
I was manually testing it I hit an issue as following.

Reproduce:

1. Deploy Opensearch from this branch 
2. Deploy Self-signed certificates
3. Relate opensearch to self-signed-certificates
4. Opensearch become active
5. Set config to production for Opensearch
6. Opensearch charm status set to Blocked with a message ` At least 3 cluster manager
 nodes are required` which is expected.
 7. Added 2 more units to Opensearch
 8. All the Opensearch units become active
 9. Problem: Opensearch charm status is stuck at blocked with the same message that 3  cluster manager is needed.
 Expected: Charm should continue to process by recovering from Blocked status.


image

lib/charms/opensearch/v0/opensearch_profile.py
gatici: Not only profile type but all the charm config options should be validated before we start to use them. We should not continue on processing upon a  wrong profile setting and just set to charm to Blocked.  I think we can create a separate PR to implement  validation for all configs later.

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,273 @@
...
+    def get_config_profile(self) -> OpenSearchProfile:
+        """Get the current config profile."""
+        return (
+            ProductionProfile()
+            if PerformanceType(self.state.config.get("profile")) == PerformanceType.PRODUCTION
+            else TestingProfile()
+        )
+
+    def _current_peer_cluster_app(self) -> PeerClusterApp:
+        deployment_desc = self.state.app.deployment_description
```
gatici, line 260: This can be None.

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,273 @@
...
+            planned_units=self.state.charm.app.planned_units(),
+            units=[
+                format_unit_name(u, app=deployment_desc.app) for u in all_units(self.state.charm)
+            ],
+            roles=(
+                deployment_desc.config.roles
+                if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES
+                else ClusterTopology.generated_roles()
+            ),
+        )
```
gatici, line 272:
```suggestion
        deployment_desc = self.state.app.deployment_description
        if deployment_desc is None:
            return PeerClusterApp(
                app=self.state.charm.app,
                planned_units=self.state.charm.app.planned_units(),
                units=[format_unit_name(u, app=self.state.charm.app) for u in all_units(self.state.charm)],
                roles=ClusterTopology.generated_roles(),
            )
        return PeerClusterApp(
            app=deployment_desc.app,
            planned_units=self.state.charm.app.planned_units(),
            units=[
                format_unit_name(u, app=deployment_desc.app) for u in all_units(self.state.charm)
            ],
            roles=(
                deployment_desc.config.roles
                if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES
                else ClusterTopology.generated_roles()
            ),
        )
```

lib/charms/opensearch/v0/state.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @property
+    def profile(self) -> Optional[OpenSearchProfile]:
+        """Current profile of the unit"""
+        if profile_str := self.relation_data.get(self.scope, PERFORMANCE_PROFILE, None):
+            return (
+                ProductionProfile()
+                if PerformanceType(profile_str) == PerformanceType.PRODUCTION
+                else TestingProfile()
+            )
```
gatici, line 75:
```suggestion
            if PerformanceType(profile_str) == PerformanceType.PRODUCTION:
                return ProductionProfile()
            elif PerformanceType(profile_str) == PerformanceType.TESTING:
                return TestingProfile()
```

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -0,0 +1,273 @@
...
+
+# The unique Charmhub library identifier, never change it
+LIBID = "8b7aa39016e748ea908787df1d7fb089"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 2
```
gatici, line 44: Why do we start from version 2 ?

gatici (commented):
It looks good to me. Great job @skourta. Some minor comments are added.

lib/charms/opensearch/v0/opensearch_profile.py
skourta: There was a lib before we just renamed the file and refactored it.

lib/charms/opensearch/v0/opensearch_profile.py
skourta: It should raise if it is None. In the charm code if this is executed then the `deployment_desc` should have been set already.

lib/charms/opensearch/v0/opensearch_profile.py
skourta: The `charm.app` is of a different type than that of the `PeerClusterApp` and `DeploymentDescription`.
The deployment description is set in the charm code before any code of the profiles is executed.

lib/charms/opensearch/v0/opensearch_profile.py
skourta: If the profile is invalid we never set it and we return before we apply it in the `config_changed` event.

lib/charms/opensearch/v0/state.py
skourta: We never store the invalid config option in the relation data. This is not needed.

lib/charms/opensearch/v0/opensearch_profile.py
gatici: OK

lib/charms/opensearch/v0/state.py
gatici: OK, As this is  the current state it is fine.

lib/charms/opensearch/v0/opensearch_profile.py
gatici: The proposed code may not be totally correct, just written to handle dployment_desc None situations.

lib/charms/opensearch/v0/opensearch_profile.py
gatici: If deployment_desc must not be None, why we allowed to return None in below method ? And if it should raise where it is raising ?
```
    @property
    def deployment_description(self) -> DeploymentDescription | None:
        """Return DeploymentDescription from peer relation"""
        deployment_desc_dict = self.relation_data.get_object(self.scope, "deployment-description")
        if not deployment_desc_dict:
            return None
        return DeploymentDescription.from_dict(deployment_desc_dict)
````

**Comment by smethnani** [comment]
Did some further testing with the following requirements for production profile:
```
     def memory_requirements(self) -> ProfileMemoryRequirements:
         """Get the memory requirements for this profile."""
         return ProfileMemoryRequirements(
            memory_size=2 * _1GB_IN_KB, 
            jvm_heap_percentage=0.1,
         )

     @property
     def cluster_topology_requirements(self) -> ClusterTopologyRequirements:
         """Get the cluster topology requirements for this profile."""
         return ClusterTopologyRequirements(
            cluster_managers=2,
            data=2,
         )
```

I set up a large deployment with:
```
juju deploy -n 2 ./opensearch main  --config cluster_name="app" --config profile="production" --config roles="cluster_manager" 
juju deploy -n 2 ./opensearch data  --config cluster_name="app" --config profile="production" --config roles="data" --config init_hold="true"

juju deploy self-signed-certificates --config ca-common-name="Test CA" --channel="1/stable"
juju integrate self-signed-certificates data
```

After relating main and data, there still doesn't seem to be a `profile` flag in any of the units.  The JVM options are not set as expected for a production profile:
```
{
  "name": "main-0.0eb",
  "-Xms": "-Xms1g",
  "-Xmx": "-Xmx1g"
}
{
  "name": "data-0.916",
  "-Xms": "-Xms1g",
  "-Xmx": "-Xmx1g"
}
{
  "name": "data-1.916",
  "-Xms": "-Xms1g",
  "-Xmx": "-Xmx1g"
}
{
  "name": "main-1.0eb",
  "-Xms": "-Xms1g",
  "-Xmx": "-Xmx1g"
}
```
Setting the config option for each app again to either "production" or "testing" triggers a config-changed event, sets the profile flag and restarts all the units. If the deployment was started with "testing" profile set, then the flag-setting and restart only happens if the config is set to "production" after deployment (ie no config-changed event if config is set explicitly to "testing" which I believe is what we want)

Scenario 2:
Set up a large deployment with 1 main node and 2 data nodes, testing profile. Relate app and wait for idle. Then:
- Set `juju config main profile="production"`
The units all block with the "2 cluster manager nodes required" message
- Add 1 main unit
All existing units remain blocked state with the new cluster manager node, only the new main unit has a profile flag set, the applied profile for the existing main unit (main/0) remains as "testing" (1g mem) while the new main unit (main/1) has a production profile applied (3g mem):
```
# profile flag
main/0  not set
main/1  production
data/0  not set
data/1  not set

# /_nodes/jvm
{
  "name": "main-0.726",
  "-Xms": "-Xms1g",
  "-Xmx": "-Xmx1g"
}
{
  "name": "main-1.726",
  "-Xms": "-Xms3208746k",
  "-Xmx": "-Xmx3208746k"
}
{
  "name": "data-1.a26",
  "-Xms": "-Xms1g",
  "-Xmx": "-Xmx1g"
}
{
  "name": "data-0.a26",
  "-Xms": "-Xms1g",
  "-Xmx": "-Xmx1g"
}

```

Screenshot 2025-09-29 at 11 28 00 AM

File: lib/charms/opensearch/v0/opensearch_profile.py [hunk 33]
```diff
@@ -0,0 +1,273 @@
...
+                deployment_desc.config.roles
+                if deployment_desc.start == StartMode.WITH_PROVIDED_ROLES
+                else ClusterTopology.generated_roles()
+            ),
+        )
+
+    @property
+    def profile(self) -> OpenSearchProfile:
+        """Get the current profile."""
+        return self.state.unit.profile or self.get_config_profile()
```
gatici: self.state.unit.profile is existing profile and  self.get_config_profile is new requested profile by config. (Correct me if I am wrong)
As the `or`  short-circuits, unless the unit.profile is set to non-None, it will always return unit-profile. 

Do we really want this ?

```
Unit=Production, Config=Testing -> returns Production

Unit=Testing, Config=Production -> returns Testing

Unit=None, Config=Production -> returns Production

Unit=None, Config=Testing -> returns Testing
```

skourta: Yes, that is what we want. The idea is to return whatever profile is applied if it has been applied before (thus stored in the databag) otherwise return what is configured to be applied.

gatici: Thank you!

smethnani (commented):
I observed a few issues while manually testing. One of the issues seems to be flaky and not happen every time:

```
juju deploy -n 2 ./opensearch main --config cluster_name="app" --config profile="testing" --config roles = "cluster_manager" --constraints "mem=2G"
juju deploy -n 2 ./opensearch data --config cluster_name="app" --config profile="testing" --config roles = "data" --constraints "mem=2G"
```
Set `juju config main profile="production"`, the main units will display "Insufficient Memory" message.
Add a unit with `juju add-unit main`
The config option is set to "production" and the units all have the testing profile applied. 
**Issue**: Only the new unit will have the "Insufficient Memory" status message while the other units in main will not.

One another occasion, I run the same set up:
- the second main unit did not start and does not show up when hitting the `/_cat/nodes/` endpoint. I'm not sure if this is related to this PR or not but I can see that `_can_service_start` initially failed and `start_opensearch` was deferred. 
**Issue**: Once the start opensearch event is processed, the node isn't able to join the cluster and opensearch logs on the node show "Wait for cluster to be available .."
- I then set `juju config main profile="production"` and the node was able to join the cluster and shows up in `/_cat/nodes` (with the testing profile applied)
- I then ran `juju add-unit main` and the new main unit again does not start.

Logs for this case:

- [Juju logs - before setting production profile config option](https://pastebin.canonical.com/p/bDnrjsytWJ/)
- [Juju logs - after that point](https://pastebin.canonical.com/p/mbxsMZpr7H/)
- [Opensearch logs on main/1 which could not join cluster](https://pastebin.canonical.com/p/4BW74tWYbJ/)

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -575,15 +581,16 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent):  # noqa C901
...
-                self.opensearch_peer_cm.validate_recommended_cm_unit_count()
             if self.model.relations[PeerClusterRelationName]:
                 self.peer_cluster_requirer.apply_orchestrator_status()
         elif event.relation.data.get(event.app):
             # if app_data + app_data["nodes_config"]: Reconfigure + restart node on the unit
             self._reconfigure_and_restart_unit_if_needed()
 
+        # check requirements
+        if self.state.app.deployment_description:
+            self.profiles_manager.check_all_requirements()
```
Mehdi-Bendriss: I would rename this method `check_missing_requirements`

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -804,13 +797,46 @@ def _on_config_changed(self, event: ConfigChangedEvent):  # noqa C901
...
+            self.status.set(BlockedStatus(InvalidProfileConfigOption))
+            return
+
+        if self.profiles_manager.check_all_requirements(config_profile):
+            event.defer()
+            return
+
+        # if the profile hasn't been applied before
+        logger.debug("current profile: %s, config profile: %s", current_profile, config_profile)
+        if current_profile is None or current_profile != config_profile:
```
Mehdi-Bendriss, line 842: I'm not sure I understand why we're checking if the current profile is none, doesn't this mean that it's a new deployment (where this hook will run before the _on_start)? in which case no restart needed

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1112,6 +1116,13 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
             # TODO: remove this IF condition once LP#2076599 is fixed in Juju.
             return
 
+        # update any orchestrators about planned units this is useful when a new data node joins
+        # the orchestrator needs to know the planned units to determine if it can start or not
```
Mehdi-Bendriss, line 1132: in which case, shouldn't this only happen _after_ the unit starts? meaning at the end of the start process?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1196,25 +1201,24 @@ def _start_opensearch(self, event: _StartOpenSearch) -> None:  # noqa: C901
...
+        except OpenSearchNotFullyReadyError as e:
+            self.node_lock.release()
+            logger.debug("Node started but not fully ready: %s", e)
+            event.defer()
+        finally:
             # In large deployments with cluster-manager-only-nodes, the startup might fail
             # for the cluster-manager if a joining data node did not yet initialize the
             # security index. We still want to update and broadcast the latest relation data.
             if self.opensearch_peer_cm.is_provider(typ="main"):
                 self.peer_cluster_provider.refresh_relation_data(event, can_defer=False)
```
Mehdi-Bendriss, line 1236: We should also have that as a finally for the event where the start was deferred due to a delayed start

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1280,7 +1290,11 @@ def _post_start_init(self, event: _StartOpenSearch):  # noqa: C901
                 event.defer()
                 return
 
-        self.peers_data.put(Scope.UNIT, "started", True)
+        # Add a timestamp instead of just a boolean to be able to trigger
+        # a relation changed event in case of restarts
+        # if you remove the flag and add it again in the same hook the
+        # relation changed event won't be triggered
```
Mehdi-Bendriss: I think no need to add this, as the rel data change only happens after the hook completes as per juju functioning

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -723,17 +720,13 @@ def _on_update_status(self, event: UpdateStatusEvent):  # noqa: C901
...
-            logger.error("Missing profile requirements: %s", missing_requirements)
-            self.status.set(BlockedStatus(" - ".join(missing_requirements)))
-            return
-
         # if node already shutdown - leave
         if not self.opensearch.is_node_up():
             return
 
+        if self.profiles_manager.check_all_requirements():
+            return
```
Mehdi-Bendriss: Why not set this at the beginning of the hook before any subsequent check?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1932,6 +1945,21 @@ def handle_joining_data_node(self) -> None:
         if self.peers_data.get(Scope.UNIT, "started", False):
             self.status.clear(PClusterNoDataNode)
         else:
```
```suggestion
        if self.peers_data.get(Scope.UNIT, "started", False):
            self.status.clear(PClusterNoDataNode)
            return
            
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1932,6 +1945,21 @@ def handle_joining_data_node(self) -> None:
...
+            try:
+                config_profile = self.profiles_manager.get_config_profile()
+            except ValueError:
+                return
```
Mehdi-Bendriss: Is this assuming that the status was already set as part of the `config_change` hook?

lib/charms/opensearch/v0/opensearch_profile.py
```diff
@@ -234,34 +208,55 @@ def check_cluster_topology(self, profile: OpenSearchProfile) -> List[str]:
...
+                error_message = f"At least {profile.cluster_topology_requirements.cluster_managers} cluster manager nodes are required."
+            case (False, True):
+                error_message = f"At least {profile.cluster_topology_requirements.data} data nodes are required."
+            case _:
+                return []
+
+        logger.error("Missing cluster topology requirements: %s", error_message)
+        return [error_message]
+
+    def check_all_requirements(
```
Mehdi-Bendriss: We should rename this method, as we are checking for missing requirements (as also returned)

lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -108,8 +108,88 @@ def remove_oidc_auth(self):
             self.SECURITY_CONFIG_YML, "config/dynamic/authc/openid_auth_domain"
         )
 
+    def set_jwt_auth(self, jwt_config: JWTAuthConfiguration) -> None:
```
Mehdi-Bendriss, line 111: why is this part of this PR?

Mehdi-Bendriss (commented):
Thanks Smail! I left a few questions

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Because it uses `deployment_description` and we would need to add checks inside of it for its existence. I prefered to move it down here and avoid adding the checks inside.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Yes and becuase we check that opensearch is up before we emit the restart event we won't emit a restart event in that scenario

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: This is here only because of merge with main I did not add it as part of this PR. I will create an issue and add it in a separate PR if you do not mind.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Can you please explain? I added this to fix an issue where we were removing and adding back the started flag in the same juju hook and when that happens in the end the databag did not change so there's no relation_changed event fired.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Yes, this is only to apply the JVM options because when the data node joins we emit a custom start event.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: I saw experimentally it takes a long time for the start event to fully execute and that leaves the orchestrators hanging because they wouldn't know about the existing data nodes. I believe from a charm perspective as soon as we have the units we should unblock the orchestrators. If the data nodes fail to start then they will be blocked and so will the orchestrators.

lib/charms/opensearch/v0/opensearch_config.py
skourta: I'm not sure why this is highlighted in your review. It is not in mine. This comes from the merge of the 2/edge into this branch but it shouldn't be highlighted as it wasn't modified by this PR and we are merging into 2/edge

lib/charms/opensearch/v0/opensearch_profile.py
skourta: Because the state properties are only based on the unit databag. There are usecases where the deployment description might be None for example if update_status gets triggered before config_changed.

Mehdi-Bendriss (dismissed):

smethnani (dismissed):
Thank you @skourta! The statuses are now displayed as expected

gatici (dismissed):
Thanks for great work @skourta

Mehdi-Bendriss (approved):

smethnani (approved):

PR #686, opened 2025-08-21
[DPE-7996] - Fix OpenSearchStartTimeoutError being raised all the time

skourta:
Refactor the logic for waiting for the OpenSearch service to start in `opensearch_distro.py`. Only raise a timeout error if the connection failed.

Service startup logic improvements:

* Changed the loop to use assignment expression (`connected := _is_connected()`).
* Replaced the `else` clause with an explicit `if not connected:` check after the loop, and added a debug log message when OpenSearch fails to start within the timeout period.


smethnani (approved):
Nice catch!

Mehdi-Bendriss (approved):
Thanks Smail! Nice catch! 🙈

PR #689, opened 2025-08-29
[DPE-7828]: Fix main orchestrator demotion

skourta:
This pull request introduces several enhancements and refactorings to the OpenSearch peer cluster management code, focusing on improved status handling, more robust blocking logic, and increased debugging visibility. The changes help clarify orchestrator roles, ensure correct cluster startup conditions, and provide better traceability for relation events and state transitions.

Fixes the following issues:
- Main orchestrator not blocked when it is also requiror
- Main orchestrator not demoted when scaled down [DPE-7828]
- Main takes over when scaled back up from 0
- Main cannot start when scaled back up from 0
- Deployment type wrongly initialized for existing deployment when scaling back up from 0
- Failover not reelected after first failover promotion [DPE-7827]
- Failover seeing itself as both main and failover in certain scenarios

**Status and Blocking Logic Improvements**

* Added logic to block the main orchestrator if it is incorrectly acting as a requirer, setting a `BlockedStatus` with the new message `PClusterMainIsRequirer`. The status is cleared when the condition no longer applies. (`opensearch_relation_peer_cluster.py`, `constants_charm.py`) [[1]](diffhunk://#diff-03a06629bc952081d1fd57adce9d458306154c5da201a8bca5bf83b7fd137b6dR271-R309) [[2]](diffhunk://#diff-b4a07b5e0ba3f40e62941435fbef81c9b9d541a641c9d274e8d82616750878baR81)
* Implemented `_block_if_waiting_for_peer_cluster` to block operations if the cluster is waiting for a peer, using the `PClusterNoRelation` message, and to clear the status when resolved. (`opensearch_relation_peer_cluster.py`)

**Cluster Startup and Event Handling**

* Refactored the cluster startup sequence to check service start conditions earlier and defer events if requirements are not met, improving reliability during orchestration. (`opensearch_base_charm.py`)
* Enhanced the handling of storage detachment for orchestrators, ensuring proper demotion and cleanup of relation data when the provider role changes. (`opensearch_base_charm.py`)

**Debugging and Traceability Enhancements**

* Added extensive debug logging throughout peer cluster management and relation event handlers for better traceability of orchestration actions, cluster setup, and directive evaluation. (`opensearch_peer_clusters.py`, `opensearch_relation_peer_cluster.py`, `opensearch_base_charm.py`) [[1]](diffhunk://#diff-ff2e2a1916e8d863212137b2a0fb586f2fee51453c83d7b2a82fa4e5c590bc28R79) [[2]](diffhunk://#diff-ff2e2a1916e8d863212137b2a0fb586f2fee51453c83d7b2a82fa4e5c590bc28R114) [[3]](diffhunk://#diff-03a06629bc952081d1fd57adce9d458306154c5da201a8bca5bf83b7fd137b6dR154-R183) [[4]](diffhunk://#diff-03a06629bc952081d1fd57adce9d458306154c5da201a8bca5bf83b7fd137b6dR192-R221) [[5]](diffhunk://#diff-03a06629bc952081d1fd57adce9d458306154c5da201a8bca5bf83b7fd137b6dR271-R309) [[6]](diffhunk://#diff-03a06629bc952081d1fd57adce9d458306154c5da201a8bca5bf83b7fd137b6dR382) [[7]](diffhunk://#diff-03a06629bc952081d1fd57adce9d458306154c5da201a8bca5bf83b7fd137b6dL423-R502) [[8]](diffhunk://#diff-03a06629bc952081d1fd57adce9d458306154c5da201a8bca5bf83b7fd137b6dR1045-R1058) [[9]](diffhunk://#diff-03a06629bc952081d1fd57adce9d458306154c5da201a8bca5bf83b7fd137b6dR1078) [[10]](diffhunk://#diff-03a06629bc952081d1fd57adce9d458306154c5da201a8bca5bf83b7fd137b6dR1096) [[11]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R388) [[12]](diffhunk://#diff-5c0d89f838ffd5a9ecf2288b3698320fda2d2b2b737361b6e10adb55d1929f43R513) [[13]](diffhunk://#diff-ff2e2a1916e8d863212137b2a0fb586f2fee51453c83d7b2a82fa4e5c590bc28R182) [[14]](diffhunk://#diff-ff2e2a1916e8d863212137b2a0fb586f2fee51453c83d7b2a82fa4e5c590bc28R249-R252) [[15]](diffhunk://#diff-ff2e2a1916e8d863212137b2a0fb586f2fee51453c83d7b2a82fa4e5c590bc28R324-R327)

**Relation Data Management**

* Improved relation data cleanup by ensuring that all related keys (including error and orchestrator data) are deleted when a relation is removed, preventing stale state. (`opensearch_relation_peer_cluster.py`)
* Added a `clean_all_relation_data` method to facilitate complete cleanup of relation data across all relations for a given app. (`opensearch_relation_peer_cluster.py`)

**Deployment Description Handling**

* Refined deployment description logic to better handle failover orchestrator states and pending directives, ensuring the cluster is properly blocked when waiting for peer relations. (`opensearch_peer_clusters.py`)

[DPE-7828]: https://warthogs.atlassian.net/browse/DPE-7828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-7827]: https://warthogs.atlassian.net/browse/DPE-7827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -404,6 +406,8 @@ def cleanup():
         self.status.clear(AdminUserNotConfigured)
         self.status.clear(TLSNotFullyConfigured)
         self.status.clear(TLSRelationMissing)
+        if self.unit.is_leader():
+            self.status.clear(PClusterNoRelation, app=True)
```
Mehdi-Bendriss, line 410: Are we guaranteed to have the PClusterRelation by the time the code reached here? `_apply_peer_cm_directives_and_check_if_can_start` would be the gatekeeper, correct?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -669,7 +673,11 @@ def _on_opensearch_data_storage_detaching(self, event: StorageDetachingEvent):
                         self.peers_data.delete(Scope.APP, "bootstrapped")
                 if self.opensearch_peer_cm.is_provider():
                     self.peer_cluster_provider.refresh_relation_data(event, can_defer=False)
-                if self.opensearch_peer_cm.is_consumer():
+                    logger.debug("demoting main orhcestrator")
```
```suggestion
                    logger.debug("demoting main orchestrator")
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -1271,6 +1280,8 @@ def _post_start_init(self, event: _StartOpenSearch):  # noqa: C901
         self.status.clear(WaitingToStart)
         self.status.clear(ServiceStartError)
         self.status.clear(PClusterNoDataNode)
+        if self.unit.is_leader():
+            self.status.clear(PClusterNoRelation, app=True)
```
Mehdi-Bendriss: why is this needed? given it was cleared in the `_on_start` hook?

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -91,12 +92,12 @@ def run(self) -> None:
...
-            # TODO: Should we add an entry on DeploymentDesc "errors" to reflect on status?
-            self._charm.peers_data.put_object(
-                Scope.APP, "deployment-description", deployment_desc.to_dict()
-            )
+        # if deployment_desc.state.value == State.ACTIVE:
+        # we only update the deployment desc if all is well. WHY??
+        # TODO: Should we add an entry on DeploymentDesc "errors" to reflect on status?
+        self._charm.peers_data.put_object(
+            Scope.APP, "deployment-description", deployment_desc.to_dict()
+        )
```
Mehdi-Bendriss: Could you explain this? I'm not very clear about the change

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -272,7 +289,17 @@ def _existing_cluster_setup(
             directives.append(Directive.SHOW_STATUS)
             directives.remove(Directive.WAIT_FOR_PEER_CLUSTER_RELATION)
 
-        deployment_type = self._deployment_type(config, start_mode)
+        if (
+            prev_deployment.typ == DeploymentType.FAILOVER_ORCHESTRATOR
+            and not self._charm.model.relations[PeerClusterRelationName]
+        ):
```
Mehdi-Bendriss: the deployment state should have already been set as `BLOCKED_WAITING_FOR_RELATION` no?, why is it needed to add this specific check?

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -110,14 +111,23 @@ def run(self) -> None:
...
         config = current_deployment_desc.config
         deployment_state = current_deployment_desc.state
         pending_directives = current_deployment_desc.pending_directives
 
+        logger.debug("deployment_desc; %s", current_deployment_desc)
+
         if Directive.WAIT_FOR_PEER_CLUSTER_RELATION in pending_directives:
+            logger.debug("clearing WAIT_FOR_PEER_CLUSTER_RELATION directive")
             pending_directives.remove(Directive.WAIT_FOR_PEER_CLUSTER_RELATION)
+            if deployment_state.value == State.BLOCKED_WAITING_FOR_RELATION:
```
Mehdi-Bendriss, line 127: Is this not already inferred from the `WAIT_FOR_PEER_CLUSTER_RELATION` directive?

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -110,14 +111,23 @@ def run(self) -> None:
...
+                logger.debug("clearing deployment_state")
+                deployment_state = DeploymentState(value=State.ACTIVE)
+            if self._charm.unit.is_leader():
+                self._charm.status.clear(PClusterNoRelation)
```
Mehdi-Bendriss: Is this not already handled in the `_on_start` hook?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -148,13 +151,36 @@ def __init__(self, charm: "OpenSearchBaseCharm"):
     def _on_peer_cluster_relation_joined(self, event: RelationJoinedEvent):
         """Received by all units in main/failover clusters when new sub-cluster joins the rel."""
         if not self.charm.unit.is_leader():
+            logger.debug("Node not a leader. Skipping refresh relation data")
+            return
+
+        if not self.charm.opensearch_peer_cm.is_provider():
```
Mehdi-Bendriss: isn't this check effectively preventing the broadcast of the error: `Related to a non orchestrator application`?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -163,20 +189,36 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
 
+        logger.debug("refreshing relation data")
+        self.refresh_relation_data(event, event_rel_id=event.relation.id)
+
         # only the main-orchestrator is able to designate a failover
         if deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
             return
 
+        if self._block_if_main_orchestrator_is_requirer():
+            return
```
Mehdi-Bendriss: Couldn't we handle this in the `set_error` method where we are testing against other possible failure scenarios? or better handle this at the requirer level (`error_from_requirer`)?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -226,18 +268,42 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
         orchestrators.failover_app = candidate_failover_app
         orchestrators.failover_rel_id = event.relation.id
         self.charm.peers_data.put_object(Scope.APP, "orchestrators", orchestrators.to_dict())
 
         self._broadcast_new_failover_app(peer_cluster_app, target_relation_ids)
 
+    def _block_if_main_orchestrator_is_requirer(self) -> bool:
+        """Block the charm if main orchestrator is a requirer"""
+        if (
+            not self.charm.unit.is_leader()
```
Mehdi-Bendriss: This method always runs on the leader

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -226,18 +268,42 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
+        if peer_cluster_requirer_relations:
+            self.charm.status.set(
+                BlockedStatus(PClusterMainIsRequirer),
+                app=True,
+            )
+            return True
+        # clean the status if it is set
+        else:
+            self.charm.status.clear(PClusterMainIsRequirer, app=True)
+        return False
```
```suggestion
        self.charm.status.clear(PClusterMainIsRequirer, app=True)
        return False
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -886,6 +965,36 @@ def _get_first_data_node(self) -> str | None:
...
+                app=True,
+            )
+            return True
+        else:
+            self.charm.status.clear(
+                PClusterNoRelation,
+                app=True,
+            )
+
+        return False
```
```suggestion
        self.charm.status.clear(PClusterNoRelation, app=True)
        return False
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -886,6 +965,36 @@ def _get_first_data_node(self) -> str | None:
...
 
+    def clean_all_relation_data(self) -> None:
+        """Clean up all relation data."""
+        if not self.charm.unit.is_leader():
+            return
+
+        for rel in self.charm.model.relations[self.relation_name]:
+            self._delete_rel_data(rel.id)
+
+    def _block_if_waiting_for_peer_cluster(self) -> bool:
```
Mehdi-Bendriss: Similar comment, on whether this could not have been part of the error setting on the provider side to centralize error logic in a single method

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -908,7 +1017,8 @@ def __init__(self, charm: "OpenSearchBaseCharm"):
 
     def _on_peer_cluster_relation_joined(self, event: RelationJoinedEvent):
         """Event received when a new main-failover cluster unit joins the fleet."""
-        pass
+        if self.charm.unit.is_leader():
+            self.charm.status.clear(PClusterNoRelation, app=True)
```
Mehdi-Bendriss: this is not entirely accurate, if there provider is an erroneous provider, right?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1528,3 +1657,32 @@ def get_cluster_first_data_node(self) -> str | None:
...
+        ]
+        # clean the status if it is set
+        if (
+            not peer_cluster_requirer_relations
+            and self.charm.app.status.message == PClusterMainIsRequirer
+        ):
+            self.charm.status.set(
+                ActiveStatus(),
+                app=True,
+            )
```
Mehdi-Bendriss: we shouldn't set directly an ActiveStatus, which is too far reaching, rather `status.clear`

Mehdi-Bendriss (commented):
Thanks Smail, great work! I have a few comments mainly on the scattering of the error handling logic on both the provider and requirers of the PCluster relation.
(we should perhaps reduce a bit the debug logs where possible)

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: The error in that case should be different and detected on the relation.

lib/charms/opensearch/v0/opensearch_base_charm.py
skourta: Correct. `opensearch_peer_cm.can_start` to be precise.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
skourta: So we actually save the deployment-decription resulting from `_existing_cluster_setup`. I'm not sure the reason why it was only saved on state being active.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
skourta: When scaling back up from 0 with the main orchestrator demoted to failover. We won't have the `BLOCKED_WAITING_FOR_RELATION` so we need to set it.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -163,12 +175,23 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
+        logger.debug("refreshing relation data")
+        is_failover_waiting_for_peer_relation = (
+            Directive.WAIT_FOR_PEER_CLUSTER_RELATION in deployment_desc.pending_directives
+        )
+        self.refresh_relation_data(
+            event,
+            event_rel_id=event.relation.id,
+            can_defer=not is_failover_waiting_for_peer_relation,
+        )
+        if is_failover_waiting_for_peer_relation:
```
smethnani: I think this early return before the failover does the `should_promote` check is what's preventing the demoted main orchestrator from promoting itself when it should in the test mentioned

smethnani (commented):
Great work @skourta! As discussed, I'm seeing an issue in the following manual test for reference:

I set up a large deployment with 1 main unit and 2 data units. I then scaled main down to 0 without destroying storage. Scaling back up with storage re-attached, the main orchestrator remains a failover and doesn't promote itself when it should:
Screenshot 2025-09-23 at 2 07 03 PM

Testing with a failover included works as expected. The main orchestrator demotes itself when scaled down to 0 and the failover then takes over. On scaling main back up, the error message is "Waiting for a peer-cluster-relation" which, from our discussion, I believe is expected and the main-failover relation must be severed and re-created with the demoted main as the requirer.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: It was I moved the promotion logic and had to tweak the logic of registering the main orchestrators to make sure rescaling a previously demoted main orchestrator detects that there is a new main registered and does not promote itself

smethnani:
> Great work @skourta! As discussed, I'm seeing an issue in the following manual test for reference:
> 
> I set up a large deployment with 1 main unit and 2 data units. I then scaled main down to 0 without destroying storage. Scaling back up with storage re-attached, the main orchestrator remains a failover and doesn't promote itself when it should

@skourta I retested this case with the latest changes and the main orchestrator now promotes itself on scaling back up 🎉

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -76,27 +76,29 @@ def __init__(self, charm: "OpenSearchBaseCharm"):
...
             return
 
-        if deployment_desc.state.value == State.ACTIVE:
-            # we only update the deployment desc if all is well.
-            # TODO: Should we add an entry on DeploymentDesc "errors" to reflect on status?
-            self._charm.peers_data.put_object(
-                Scope.APP, "deployment-description", deployment_desc.to_dict()
-            )
+        # TODO: Should we add an entry on DeploymentDesc "errors" to reflect on status?
+        self._charm.peers_data.put_object(
```
Mehdi-Bendriss, line 99: nice catch!

lib/charms/opensearch/v0/opensearch_peer_clusters.py
Mehdi-Bendriss: question still relevant

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -272,7 +288,19 @@ def _existing_cluster_setup(
             directives.append(Directive.SHOW_STATUS)
             directives.remove(Directive.WAIT_FOR_PEER_CLUSTER_RELATION)
 
-        deployment_type = self._deployment_type(config, start_mode)
+        # When scaling back up from 0 with the main orchestrator demoted to failover.
```
Mehdi-Bendriss: Can you rephrase this?

lib/charms/opensearch/v0/opensearch_peer_clusters.py
Mehdi-Bendriss: Not sure I understand, isn't the flow?
1. Main orchestrator (current app) demoted its deployment_type to failover when scaling down to 0
2. scaled back up again - deployment type still set to failover 
3. the deployment state "waiting for relation" should be automatically set

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -272,7 +288,19 @@ def _existing_cluster_setup(
             directives.append(Directive.SHOW_STATUS)
             directives.remove(Directive.WAIT_FOR_PEER_CLUSTER_RELATION)
 
-        deployment_type = self._deployment_type(config, start_mode)
```
Mehdi-Bendriss, line 275: I don't understand why we're removing this and basing the whole logic exclusively on the previous deployment type instead of computing a new one?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -168,15 +184,30 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
             deployment_desc.typ == DeploymentType.FAILOVER_ORCHESTRATOR
             and self.should_promote_failover_to_main()
         ):
-            logger.info("Promoting failover orchestrator to main orchestrator")
```
Mehdi-Bendriss, line 171: why remove this?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -168,15 +184,30 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
             self._promote_failover()
             self.refresh_relation_data(event)
             return
```
Mehdi-Bendriss, line 190: Please add a comment here

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -168,15 +184,30 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
+            Directive.WAIT_FOR_PEER_CLUSTER_RELATION in deployment_desc.pending_directives
+        )
+        self.refresh_relation_data(
+            event,
+            event_rel_id=event.relation.id,
+            can_defer=not is_failover_waiting_for_peer_relation,
+        )
+
+        if is_failover_waiting_for_peer_relation:
+            return
```
Mehdi-Bendriss: how do we know this is currently a failover and not an app that could have been mistakenly integrated with?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -168,15 +184,30 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
...
+
         # only the main-orchestrator is able to designate a failover
         if deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
             return
 
+        # if the main orchestrator is a requirer return
+        if self.model.relations[PeerClusterRelationName]:
```
Mehdi-Bendriss: 1. wouldn't this effectively also discard all relation changed events (no matter how accurate they are) if the current "main" has a wrong relation?
2. wouldn't this also apply to the failover?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1226,13 +1308,14 @@ def _on_peer_cluster_relation_departed(self, event: RelationDepartedEvent):
             self.charm.peers_data.get_object(Scope.APP, "orchestrators")
         )
 
-        # a cluster of type "other" is departing (wrong relation), or, the current is a main
+        # a non registered cluster is departing (wrong relation), or, the current is a main
```
```suggestion
        # a non elected orchestrator cluster is departing (wrong relation), or, the current is a main
```

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1226,13 +1308,14 @@ def _on_peer_cluster_relation_departed(self, event: RelationDepartedEvent):
...
         # orchestrator and a failover is departing, we can safely ignore.
         if event.relation.id not in [
             orchestrators.main_rel_id,
             orchestrators.failover_rel_id,
         ]:
             self._clear_errors(f"error_from_requirer-{event.relation.id}")
+            self._clear_errors(f"error_from_providers-{event.relation.id}")
```
Mehdi-Bendriss, line 1316: nice

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1528,3 +1610,26 @@ def get_cluster_first_data_node(self) -> str | None:
             return None
 
         return self.peer_cm.rel_data_from_str(peer_cluster_data).first_data_node
+
+    def clean_is_candidate_failover(self) -> None:
```
Mehdi-Bendriss: is this still used?

Mehdi-Bendriss (commented):
Thanks Smail, good work. I left some comments

lib/charms/opensearch/v0/opensearch_peer_clusters.py
skourta: Previously, BLOCKED_WAITING_FOR_RELATION was only set on `new_cluster_setup`. When scaling back we already have the deployment_desc already exists so we run the `existing_cluster_setup`. That is why we need to also add it here in the `existing_cluster_setup`.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: To clean redundant logs we have logs inside the `_promote_failover` function

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: I feel like I am missing something. If it's an app that was mistakenly integrated with it will set the errors in the refresh_relation_data and we should exit in that case as well, no? And even then, there's no use in deferring the event imo

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
skourta: 1. I'm not sure I completely understand this. If it has a wrong relation then it will set the error (main orchestrator is a requirer) on `refresh_relation_data` at this stage we know the integrations are not correctly set up so we should exit.
2. No this would not affect failovers because of the check on the previous line

lib/charms/opensearch/v0/opensearch_peer_clusters.py
skourta: This is an existing cluster and the `_deployment_type` is based on `init_hold` which should only be used at deployment. If the cluster already exists, then we can no longer base the deployment type on `init_hold` because it can be demoted or promoted during its lifecycle.

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
smethnani: @skourta I don't think we should exit if the current main has a wrong relation. Shouldn't there just be a message telling the user to sever it? Consider the following:
```
juju deploy ./opensearch main --config roles="cluster_manager" --config cluster_name="app" 
juju deploy ./opensearch failover --config roles="cluster_manager" --config init_hold="true"  --config cluster_name="app"
juju deploy ./opensearch data --config roles="data" --config init_hold="true" --config cluster_name="app" 

juju integrate failover:peer-cluster-orchestrator main
juju integrate main:peer-cluster-orchestrator data
```

In the above case, main will never run `handle_joining_data_node` and won't emit a StartOpenSearch event because it exits here and so the deployment will be stuck

skourta: I have removed this part of the code. I think I'm missing a commit on GitHub.

smethnani (dismissed):
Great work @skourta! Tested manually and all worked as expected.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -124,9 +124,7 @@ def run_with_relation_data(self, data: PeerClusterRelData) -> None:  # noqa: C90
         if Directive.WAIT_FOR_PEER_CLUSTER_RELATION in pending_directives:
             logger.debug("clearing WAIT_FOR_PEER_CLUSTER_RELATION directive")
             pending_directives.remove(Directive.WAIT_FOR_PEER_CLUSTER_RELATION)
-            if deployment_state.value == State.BLOCKED_WAITING_FOR_RELATION:
-                logger.debug("clearing deployment_state")
-                deployment_state = DeploymentState(value=State.ACTIVE)
+            deployment_state = DeploymentState(value=State.ACTIVE)
```
Mehdi-Bendriss: Is this not too aggressive? in the sense that there are other states other than `BLOCKED_WAITING_FOR_RELATION`

Mehdi-Bendriss (commented):
Thanks Smail! I have 1 question left

lib/charms/opensearch/v0/opensearch_peer_clusters.py
skourta: We do have checks for other directives under this. Other than that, I don't see a way to clear this particular state if it set and needs to be cleared.

gatici (approved):

gatici:
@skourta  Thank you!

Mehdi-Bendriss (approved):

smethnani (approved):

PR #690, opened 2025-09-02
[DPE 7933] - Plugin Manager Rework

smethnani:
This PR simplifies the plugin manager component and adds support for the notifications plugin

More specifically this PR:
- Removes logic handling the knn_plugin config option
- Simplifies the OpenSearch keystore manager by removing the base class and custom errors
- Adds a PluginEvents component which handles SMTP events and can be extended to add other plugin relation event handlers
- Adds a `plugins` field to `PeerClusterRelData` to transfer plugin secrets
- Updates the backups component to use the new plugin manager api when adding or removing keys

The logic to transfer secret key entries to units in sub-clusters differs from how the backup credentials are transferred (that method doesn't work corrrectly, see #693). The keys are transferred as follows:
- Each app leader maintains a list of plugin secrets in their app data bags
- Units in subclusters additionally store a list of secrets they are tracking in their unit data bags
- The main orchestrator leader creates a secret with the keys, grants it to subclusters, writes the secret id to the peer-cluster-relation data
- Subcluster leaders read the list transferred from main and compute a diff with their own list and add or remove from the list accordingly
- Subcluster units compute a diff between the secrets they are tracking and the secrets the app is tracking and add or remove from their list (and keystores) accordingly. 
- The sequence above handles the cases when a secret is created or removed. Changes that happen between adding or removing the secret are handled during the `secret-changed` event.
```mermaid
sequenceDiagram
    Main(leader)->>+Main(leader): create_secret(keys)
    Main(leader)->>+Main(leader): state.add_secret_id(secret_id)
    Main(leader)->>+Main(leader): refresh_rel_data
    Main(leader)-->>-peer-cluster: peer-cluster-relation-changed
    peer-cluster->>+peer-cluster: state.add_secret_id(secret_id)
    peer-cluster-->>+opensearch-peers: opensearch-peers-relation-changed
    opensearch-peers->>+opensearch-peers: get_secret(secret_id)
    opensearch-peers->>+opensearch-peers: update_keystore(keys)
```
- [ x] Manually
- [ ] Unit tests
- [ ] Integration tests


lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -231,3 +231,13 @@ def mask_sensitive_information(cmd: str) -> str:
     pattern = re.compile(r"(-tspass\s+|-kspass\s+|-storepass\s+|-new\s+|pass:)(\S+)")
 
     return re.sub(pattern, r"\1" + "xxx", cmd)
+
+
+def diff(desired: list[str], current: list[str]):
```
Mehdi-Bendriss: nit:
```suggestion
def diff(desired: list[str], current: list[str]) -> set[str]:
```

lib/charms/opensearch/v0/models.py
```diff
@@ -501,6 +501,35 @@ def __getitem__(self, item):
...
+class PluginConfigType(BaseStrEnum):
+    """Type of config (key, cert, etc)."""
+
+    KEYS = "keys"
+
+
+class PluginConfigRemovalInfo(Model):
+    """Model class for representing data needed to remove plugin configuration"""
+
+    typ: PluginConfigType
```
Mehdi-Bendriss: We could perhaps replace this enum with a string literal type directly `Literal["keys", ...]`

lib/charms/opensearch/v0/models.py
```diff
@@ -501,6 +501,35 @@ def __getitem__(self, item):
         return self.__root__[item]
 
 
+class PluginConfigType(BaseStrEnum):
+    """Type of config (key, cert, etc)."""
```
Mehdi-Bendriss: Could you explain what this data structure represent? and its possible values

lib/charms/opensearch/v0/models.py
```diff
@@ -501,6 +501,35 @@ def __getitem__(self, item):
...
+
+    KEYS = "keys"
+
+
+class PluginConfigRemovalInfo(Model):
+    """Model class for representing data needed to remove plugin configuration"""
+
+    typ: PluginConfigType
+    content: List[str] = Field(default_factory=list)
```
Mehdi-Bendriss: 1. should we make it into a `set` to get rid of `add` ?
2. what would `content` look like for non secrets based plugins?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -697,16 +692,8 @@ def _on_backup_relation_broken(self, event: RelationEvent) -> None:
...
-            try:
-                self.charm.plugin_manager.apply_config(plug.disable())
-            except OpenSearchKeystoreNotReadyError:
-                if self.charm.opensearch.is_service_started():
-                    # We have an application up and running but not responding
-                    # We cannot defer either, so we will fail and retry later
-                    raise
-                logger.debug("Backup broken relation: keystore not ready yet")
-                # No need to defer, we already cleaned the keystore.
-                return
```
Mehdi-Bendriss, line 709: I'm very pleased to see this!

lib/charms/opensearch/v0/models.py
```diff
@@ -501,6 +501,35 @@ def __getitem__(self, item):
...
+        for item in items:
+            if item not in current:
+                self.content.append(item)
+                current.add(item)
+
+
+class PluginConfigAddInfo(Model):
+    """Model class for representing data needed to configure plugins"""
+
+    relation_name: Optional[str]
```
```suggestion
    relation_name: Optional[str] = None
```

lib/charms/opensearch/v0/models.py
```diff
@@ -510,6 +539,7 @@ class PeerClusterRelData(Model):
     deployment_desc: Optional[DeploymentDescription]
     security_index_initialised: bool = False
     first_data_node: Optional[str] = None
+    plugins: Optional[Dict[str, PluginConfigAddInfo]]
```
```suggestion
    plugins: Optional[Dict[str, PluginConfigAddInfo]] = None
```

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -781,12 +768,8 @@ def _on_backup_disable(self, event: _DisableBackupRelationEvent) -> None:  # noq
...
         try:
-            self.charm.plugin_manager.apply_config(plug.disable())
-        except OpenSearchKeystoreError:
-            # The plugin is not present, we can ignore it
-            logger.info(f"Error while removing {plug.name}")
-            event.defer()
-            return
+            keys_to_remove = plug.disable().secret_entries.keys()
+            self.charm.plugin_manager.remove_from_keystore(keys_to_remove)
         except OpenSearchError as e:
```
Mehdi-Bendriss: is this still relevant?

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -29,122 +26,51 @@
...
-
-            self._clean_cache_if_needed()
+            self._opensearch.run_bin(KEYSTORE, f"remove {key}")
         except OpenSearchCmdError as e:
             if "does not exist in the keystore" in str(e):
-                logger.info(
-                    "opensearch_keystore._delete:"
-                    f" Key {key} not found in keystore, continuing..."
-                )
+                logger.info("Keystore command 'delete' failed for key: %s. Key not found.", key)
```
Mehdi-Bendriss: This is okay for now - we will however improve this class in a future PR, to reduce the possible failures and exceptions thrown when easy to avoid, like [here](https://github.com/canonical/opensearch-operator/compare/2/edge...rework-backups#diff-3a66047320fe54743e3c9ebbf37d7f43c1f00f8dc94ef62cf888373d6e4bebfaR43)

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -29,122 +26,51 @@
...
         Raises:
             OpenSearchHttpError: If the reload fails.
         """
-        response = self._opensearch.request("POST", "_nodes/reload_secure_settings")
-        logger.debug(f"_update_keystore_and_reload: response received {response}")
+        try:
+            response = self._opensearch.request("POST", "_nodes/_local/reload_secure_settings")
+        except OpenSearchHttpError:
+            logger.info("Could not request secure settings reload.")
+            return False
```
Mehdi-Bendriss:
```suggestion
        except OpenSearchHttpError as e:
            logger.error("Could not reload secure settings: %s", e)
            return False
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+if TYPE_CHECKING:
+    from charms.opensearch.v0.opensearch_base_charm import OpenSearchBaseCharm
 
-ConfigExposedPlugins = {
-    "opensearch-knn": {
-        "class": OpenSearchKnn,
-        "config": "plugin_opensearch_knn",
-    },
-}
+SMTP_SECRET_LABEL = "notifications-plugin"
```
Mehdi-Bendriss: could we invert this for better traversal:
```suggestion
SMTP_SECRET_LABEL = "plugin-notifications"
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
-                    "in both but different values": {
-                        k: v
-                        for k, v in new_conf.items()
-                        if k in current_settings.keys() and current_settings[k] != v
-                    },
-                }
-            )
-        )
-        return current_settings, new_conf
+    def _on_smtp_credentials_changed(self, event):
```
Mehdi-Bendriss: nit: type hints missing in this and subsequent methods (args, and return types)

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+        if not self.charm.opensearch.is_started():
+            # node must be reachable to reload settings after adding keys
+            event.defer()
+            return
 
-    def apply_config(self, config: OpenSearchPluginConfig) -> bool:  # noqa: C901
-        """Runs the configuration changes as passed via OpenSearchPluginConfig.
+        # return if no relation data
+        if not (parameters := self.smtp.get_relation_data()):
+            logger.debug("Missing relation %s", self.relation_name)
```
Mehdi-Bendriss: 1. we are missing the relation data, and not the relation, correct?
2. should we set the charm in a blocked state?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
 
-        For each: configuration and secret
-        1) Remove the entries to be deleted
-        2) Add entries, if available
-        Returns True if a configuration change was performed on opensearch.yml only
-        and a restart is needed.
+        # validate data
+        required_params = ["user", "password"]
+        if missing_parameters := [p for p in required_params if not getattr(parameters, p)]:
+            logger.error("Parameters missing from smtp-intgrator: %s" % missing_parameters)
```
Mehdi-Bendriss: Similar question on Blocked status, should we set it?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
-        if config.config_entries:
-            self._opensearch_config.update_plugin(config.config_entries)
+        self.charm.state.unit.put_plugin_config_removal_info(
+            self.secret_label, PluginConfigType.KEYS, list(entries.keys())
+        )
+        if not self.charm.unit.is_leader():
+            return
+
+        self.charm.secrets.put(Scope.APP, self.secret_label, json.dumps(entries))
+        if secret_id := self.charm.secrets.get_secret_id(Scope.APP, self.secret_label):
```
Mehdi-Bendriss: why do we check on its existence when we just set it?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
-                plugin.config().config_entries[k] == existing_setup[k]
-                for k in plugin.config().config_entries.keys()
-            ]
+            return
+
+        try:
+            self.charm.secrets.delete(Scope.APP, self.secret_label)
+            self.charm.state.app.remove_plugin_config_info(self.secret_label)
+
+            if self.charm.opensearch_peer_cm.is_provider(typ="main"):
```
Mehdi-Bendriss, line 150: this is redundant, as the check was performed earlier

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+                self.charm.peer_cluster_provider.refresh_relation_data(event)
+        except SecretNotFoundError:
+            logger.debug("Can't find secret %s", self.secret_label)
+
+    def _on_secret_changed(self, event):
+        """Handles secret changes"""
+        label = self.charm.secrets.label(Scope.APP, self.secret_label)
+        if event.secret.label == label:
```
Mehdi-Bendriss: better invert the condition and return

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+            secret = event.secret.get_content(refresh=True)
+            raw = secret.get(self.secret_label)
+            keys = json.loads(raw)
+
+            # update keys to remove as they may change here
```
Mehdi-Bendriss: Could you explain this comment?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+            self.charm.on[PeerRelationName].relation_changed, self._on_peer_relation_changed
         )
 
-    def _needs_upgrade(self, plugin: OpenSearchPlugin) -> bool:
-        """Returns true if plugin needs upgrade."""
-        plugin_version = plugin.version.split(".")
-        version = self._opensearch.version.split(".")
-        num_points = min(len(plugin_version), len(version))
-        return version[:num_points] != plugin_version[:num_points]
+    def _on_peer_relation_changed(self, event):
```
Mehdi-Bendriss: missing docstrings :)

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+        # if this is a subcluster, all units must add plugin keys from secrets to their keystores
+        if self.charm.opensearch_peer_cm.is_consumer(of="main"):
```
Mehdi-Bendriss: Better invert the condition and return, to reduce a level of nestedness

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+        removal_info = self.charm.state.unit.plugin_config_removal_info.get(self.secret_label)
+        keys = removal_info.content
+
+        self.charm.plugin_manager.remove_from_keystore(keys)
+        self.charm.state.unit.remove_plugin_config_removal_info(self.secret_label)
+
+        # if unit is main orchestrator leader, remove secrets transferred to subclusters
+        if not self.charm.unit.is_leader() or not self.charm.opensearch_peer_cm.is_provider(
+            typ="main"
         ):
```
Mehdi-Bendriss: just to double check, this won't check if it is a simple deployment, in which case the way to go would be to test on the Deployment Type of the deployment description

lib/charms/opensearch/v0/state.py
Mehdi-Bendriss, line 1: There is a PR by Smail for performance profiles, where he also introduced a state module, could you please validate that both your approaches are aligned?

lib/charms/opensearch/v0/state.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(
+        self,
+        charm: "OpenSearchBaseCharm",
+    ):
+        self.scope = Scope.APP
+        self.relation_data = RelationDataStore(charm, PeerRelationName)
+
+    @property
+    def plugin_config_info(self) -> Dict[str, PluginConfigAddInfo]:
+        """Returns plugin secrets this app is tracking"""
```
Mehdi-Bendriss: is it generic to all plugin info, or just secrets? if only secrets, I would probably rename these to use the `secret_labels` term
applies to subsequent methods

lib/charms/opensearch/v0/state.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __init__(
+        self,
+        charm: "OpenSearchBaseCharm",
+    ):
+        self.scope = Scope.UNIT
+        self.relation_data = RelationDataStore(charm, PeerRelationName)
+
+    @property
+    def plugin_config_removal_info(self) -> Dict[str, PluginConfigRemovalInfo]:
```
Mehdi-Bendriss: Related to the consolidation discussion we had, and the "content" part

applies to subsequent methods

lib/charms/opensearch/v0/models.py
```diff
@@ -501,6 +501,35 @@ def __getitem__(self, item):
...
+            if item not in current:
+                self.content.append(item)
+                current.add(item)
+
+
+class PluginConfigAddInfo(Model):
+    """Model class for representing data needed to configure plugins"""
+
+    relation_name: Optional[str]
+    secret_id: str
```
Mehdi-Bendriss: will all plugins require a secret_id? (non core plugins)

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+                )
+                keys_to_write.update(keys_to_add)
 
-    def _remove_if_needed(self, plugin: OpenSearchPlugin) -> bool:
-        """If disabled, removes plugin configuration or sets it to other values."""
-        return False
+            for label in removed:
+                # this unit should delete the keys it wrote as the app secret has been removed
+                removal_info = self.charm.state.unit.plugin_config_removal_info[label]
+                keys_to_write.update({k: None for k in removal_info.content})
```
Mehdi-Bendriss: Note: better direct removal than None

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+                logger.info("Could not update keystore. Deferring event.")
+                event.defer()
+                return
+
+            # remove plugin secret and label from unit if no longer tracking
+            for label in removed:
+                self.charm.state.unit.remove_plugin_config_removal_info(label)
+
+
+class OpenSearchPluginManager:
```
Mehdi-Bendriss, line 251: As discussed, let's drop

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -626,8 +628,24 @@ def _rel_data(
...
             security_index_initialised=self._get_security_index_initialised(),
             first_data_node=self._get_first_data_node(),
+            plugins=(
+                self._plugin_config_info()
+                if deployment_desc.typ == DeploymentType.MAIN_ORCHESTRATOR
+                else None
+            ),
         )
 
+    def _plugin_config_info(self):
```
Mehdi-Bendriss: nit: missing return type

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -971,6 +989,10 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
 
         # fetch the success data
         data = self.peer_cm.rel_data_from_str(data["data"])
+
+        if (plugin_configs := data.plugins) is not None:
```
Mehdi-Bendriss, line 1003: could you please put a comment on None vs empty dict?

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -1019,6 +1041,24 @@ def _on_peer_cluster_relation_changed(self, event: RelationChangedEvent):  # noq
         # recompute the deployment desc
         self.charm.opensearch_peer_cm.run_with_relation_data(data)
 
+    def _track_plugin_secrets(self, configs_from_relation):
```
Mehdi-Bendriss: missing type hints (args / return)

lib/charms/opensearch/v0/opensearch_secrets.py
```diff
@@ -374,12 +374,37 @@ def delete(self, scope: Scope, key: str) -> None:
 
         logging.debug(f"Deleted secret {scope}:{key}")
 
+    def track_secret(self, secret_id: str, scope: Scope, key: str) -> Optional[Secret]:
+        """Track a granted secret and add it to the cache"""
+        label = self.label(scope, key)
+        cached_secret_meta = self.cached_secrets.get_meta(scope, label)
+        if cached_secret_meta:
```
Mehdi-Bendriss: nit: walrus operator

tests/integration/plugins/test_plugins.py
Mehdi-Bendriss, line 1: For another PR: could you add integration tests for all plugins, reflecting the excel sheet you had created and the test flows you had already drafted

Mehdi-Bendriss (commented):
Thank you Shams! Great work! I have some comments

lib/charms/opensearch/v0/state.py
```diff
@@ -0,0 +1,117 @@
...
+    @property
+    def plugin_config_info(self) -> Dict[str, PluginConfigAddInfo]:
+        """Returns plugin secrets this app is tracking"""
+        plugin_config_info = self.relation_data.get_object(self.scope, "plugin_config_info") or {}
+        return {
+            label: PluginConfigAddInfo.from_dict(plugin)
+            for label, plugin in plugin_config_info.items()
+        }
+
+    def put_plugin_config_info(
```
reneradoi: A comment to the `State` in the charm structure pattern: It should be a representation of (peer) relation data, but I would not add methods for data modification to the state itself. Reasoning: There are so many of them in the different places of the code base, your state objects will be very much clogged with all of them. Rather leave these methods to the managers if you want to bundle them. Not all of the modifications have to be in separate functions btw, for most of them it will be sufficient to have `self.state.OpenSearchApp.update({key: value})` as in [this example](https://github.com/canonical/charmed-etcd-operator/blob/3.6/edge/src/events/etcd.py#L156).

lib/charms/opensearch/v0/state.py
```diff
@@ -0,0 +1,117 @@
...
+        self, label: str, secret_id: str, relation_name: str, typ: PluginConfigType
+    ) -> None:
+        """Adds plugin secret being tracked by app"""
+        plugin_config = self.plugin_config_info
+        plugin_config[label] = PluginConfigAddInfo(
+            relation_name=relation_name, secret_id=secret_id, typ=typ
+        )
+        self.relation_data.put_object(self.scope, "plugin_config_info", plugin_config)
+
+    def remove_plugin_config_info(self, label: str) -> None:
```
reneradoi: Same comment regarding methods to modify relation data in the state.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +39,185 @@
...
+                event.defer()
+
+
+class OpenSearchPluginEvents(Object):
+    """Events handler for OpenSearch plugin events"""
+
+    def __init__(self, charm: "OpenSearchBaseCharm"):
+        super().__init__(charm, "plugins")
+        self.charm = charm
+        self.smtp_events = SmtpEvents(self.charm)
```
reneradoi: I would recommend to not do that. It created dependencies between each other in the event handlers, and that should be avoided. Better to instantiate the `SmtpEvents` directly in the base charm (in case of OpenSearch) or the `charm.py` (if already following the structure pattern).

lib/charms/opensearch/v0/opensearch_plugin_manager.py
reneradoi: Out of interest: Why?

reneradoi (commented):
Hey Shams! This is a significant improval on the previous implementation for plugins, I really like it!

As Mehdi has already focussed a lot on the implementation details, I'd like to give some more general feedback on the structure changes (see detailed comments).

One thing I noticed when testing (but not sure if related to this PR because I don't see changes that could have impact): When a simple deployment of 3 units starts up, the first unit fails quite a few times before finally coming up. This can be observed with the error message `OpenSearch already started, but post-start init failed.` appearing in the log. On multiple tries, this happened up to ten times to me until the first node finally came up. The status of that unit will be `An error occurred during the start of the OpenSearch service.`

Did you also see this happening? I'm not sure why this is caused, at least the OpenSearch logs themselves don't give a clear indication. Could this be related to the longer list of plugins loaded at the start (after the recent changes to the snap)?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
Mehdi-Bendriss: @reneradoi because this manager only serves as a proxy to the KeystoreManager, there is no "plugin manager" specific logic per se

reneradoi:
> One thing I noticed when testing (but not sure if related to this PR because I don't see changes that could have impact): When a simple deployment of 3 units starts up, the first unit fails quite a few times before finally coming up. This can be observed with the error message `OpenSearch already started, but post-start init failed.` appearing in the log. On multiple tries, this happened up to ten times to me until the first node finally came up. The status of that unit will be `An error occurred during the start of the OpenSearch service.`
> 
> Did you also see this happening? I'm not sure why this is caused, at least the OpenSearch logs themselves don't give a clear indication. Could this be related to the longer list of plugins loaded at the start (after the recent changes to the snap)?

@smethnani 
On the startup issue: I've tested with `2/edge` and the same also happens there. I've created an issue for this (#700) so it can be troubleshooted separately. I consider it not related to the changes in your PR.

smethnani:
Thanks for checking and letting me know @reneradoi!

lib/charms/opensearch/v0/models.py
smethnani: 1. A set isn't json serializable so this is still a list
2. I've renamed "content" to "cleanup". Other than keys, this could contain the names of certificates to remove from the JVM truststore (if we add `typ="cert"`).

tests/integration/plugins/test_plugins.py
smethnani: Yes, will be tracked [here](https://warthogs.atlassian.net/browse/DPE-7937)

lib/charms/opensearch/v0/models.py
smethnani: This has been replaced with the string literal you suggested in another comment.

For reference, the idea was for it to contain the type of configuration to infer what needs to be done with the config data. Eg if PluginConfigType is "keys", then keys need to be added to the keystore. If the PluginConfigType was something else like "cert", then a certificate needs to be added to the jvm truststore.

lib/charms/opensearch/v0/state.py
smethnani: Moved state modification to the manager!

lib/charms/opensearch/v0/opensearch_plugin_manager.py
smethnani: Thanks! Moved to the base charm

lib/charms/opensearch/v0/models.py
smethnani: Changed to literal

lib/charms/opensearch/v0/helper_charm.py
smethnani: Thanks! Added return type `tuple[set[str], set[str]]`

lib/charms/opensearch/v0/opensearch_plugin_manager.py
smethnani: Charm now blocks if relation data or required parameters are missing

lib/charms/opensearch/v0/opensearch_plugin_manager.py
smethnani: Thanks! Removed earlier check

lib/charms/opensearch/v0/opensearch_plugin_manager.py
smethnani: I've kept the plugin manager to handle modifying state with plugin-related data and moved the keystore manager to the base charm

reneradoi (approved):
The changes that I wanted to see are done, thank you @smethnani. Excited to see the integration tests!

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -693,7 +690,8 @@ def _on_backup_relation_broken(self, event: RelationEvent) -> None:
             # other units and the keystore is clean.
             # Any other exception should result in actual errors.
             keys_to_remove = plug.disable().secret_entries.keys()
-            self.charm.plugin_manager.remove_from_keystore(keys_to_remove)
+            self.charm.keystore.remove_entries(keys_to_remove)
+            self.charm.keystore.reload()
```
Mehdi-Bendriss: it looks like the failed reload of the keystore is not handled (in all its references), what should be the course of action here?

Mehdi-Bendriss (commented):
Thanks Shams! I left a comment on the non-handling of failed keystore reload
(Unit tests and integration tests will follow in a subsequent PR)

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -29,122 +26,51 @@
...
-                )
-                return
-            raise OpenSearchKeystoreError(str(e))
+        self._opensearch.run_bin(KEYSTORE, "create")
 
-    def reload_keystore(self) -> None:
-        """Updates the keystore value (adding or removing) and reload.
-
-        This method targets only the local unit as alt_hosts is not set.
+    def add_entries(self, entries: dict[str, str]) -> None:
```
gatici, line 47: I am assuming that this will automatically create the keystore path.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +37,218 @@
...
+            logger.debug("Missing relation data from %s", self.relation_name)
+            self.charm.status.set(BlockedStatus(SmtpNoRelationData))
+            return
+
+        self.charm.status.clear(SmtpNoRelationData)
+
+        # validate data
+        required_params = ["user", "password"]
+        if missing_parameters := [p for p in required_params if not getattr(parameters, p)]:
+            logger.error("Parameters missing from smtp-intgrator: %s" % missing_parameters)
```
gatici: There is a typo in integrator and missing_parameters is a  list, we can first create string to make it prettier.
```suggestion
             logger.error("Parameters missing from smtp-integrator: %s", ", ".join(missing_parameters))
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +37,218 @@
...
+            self.charm.status.set(
+                BlockedStatus(SmtpMissingRequiredParameters.format(missing_parameters))
```
gatici: Missing parameters are list,  we can first create string to make it prettier.

 msg = SmtpMissingRequiredParameters.format(", ".join(missing_parameters))
 BlockedStatus(msg)

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +37,218 @@
...
+
+    def _on_secret_changed(self, event) -> None:
+        """Handles secret changes"""
+        label = self.charm.secrets.label(Scope.APP, self.secret_label)
+        if label != event.secret.label:
+            return
+
+        secret = event.secret.get_content(refresh=True)
+        raw = secret.get(self.secret_label)
+        keys = json.loads(raw)
```
gatici: We can do it  more gracefully to prevent unexpected errors.
```suggestion
        content = event.secret.get_content(refresh=True)
        raw = content.get(self.secret_label)
        if not raw:
            logger.warning("Secret %s has no %s payload", event.secret.label, self.secret_label)
            return

        try:
            keys = json.loads(raw)
        except json.JSONDecodeError:
            logger.error("Malformed JSON in secret %s", event.secret.label)
            return
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -50,399 +37,218 @@
...
+            plugin = app_plugins[label]
+            if not plugin.secret_id:
+                continue
+
+            # start locally tracking secret and write transferred keys to keystore
+            keys_to_add = json.loads(
+                self.charm.secrets.track_secret(plugin.secret_id, Scope.APP, label)
+                .get_content()
+                .get(label)
+            )
```
gatici: We can handle more gracefully here.
```suggestion
            content = self.charm.secrets.track_secret(plugin.secret_id, Scope.APP, label).get_content()
            raw = content.get(label)
            if not raw:
                continue
                
            try:
                keys_to_add = json.loads(raw)
            except json.JSONDecodeError:
                logger.error("Invalid JSON in secret for label %s", label)
                continue

```

lib/charms/opensearch/v0/opensearch_secrets.py
```diff
@@ -374,12 +376,36 @@ def delete(self, scope: Scope, key: str) -> None:
 
         logging.debug(f"Deleted secret {scope}:{key}")
 
+    def track_secret(self, secret_id: str, scope: Scope, key: str) -> Optional[Secret]:
```
gatici: As this return a  secret object, function name does not reflect the behavior. Can we rename it as `get_tracked_secret` ?

gatici (commented):
Looks good. Thank you @smethnani  I just added some minor comments. Looking forward to see the integration tests.

lib/charms/opensearch/v0/opensearch_keystore.py
smethnani: That's right, the `--force` flag creates it automatically

lib/charms/opensearch/v0/opensearch_plugin_manager.py
smethnani: Thanks! Added

lib/charms/opensearch/v0/opensearch_plugin_manager.py
smethnani: Thanks! Added

lib/charms/opensearch/v0/opensearch_backups.py
smethnani: In each case, the event is now deferred if the keystore reloading fails

gatici (approved):

reneradoi (approved):
Re-approval, changes added after last review look fine.

Mehdi-Bendriss (approved):
Thank you Shams!

PR #696, opened 2025-09-18
[DPE-8017] Add jwt authentication

reneradoi:
This PR adds JSON Web Token (JWT) authentication to Opensearch.

## Solution:
The configuration of JWT authentication is provided by the `jwt-integrator` charm via a new relation called `jwt-configuration`. The configuration parameters are shared via relation data and relation secrets.

The business logic and event handling is bundled in a new class called `JwtHandler`. On the relevant events, the logic applies the configuration to the Opensearch security configuration file and updates the security index by running the `securityadmin` script.

In large deployments, the relation must be configured to the Main-Orchestrator app. If not, the events will be ignored and a blocked status will be displayed.

- [X] Manually
- [ ] Unit tests
- [X] Integration tests

Manual tests have been executed with both small and large deployments in remote-VM environments provisioned with testflinger.

The provided integration test in `test_jwt_auth.py` covers:
- small deployment
- large deployment with data-only nodes
  - invalid relation: blocked status and no access with the configured JWT
  - valid relation: data can be accessed with the configured JWT

## Checklist
- [X] I have added or updated any relevant documentation -> will happen in a Discourse post once this PR is merged.
- [X] I have cleaned any remaining cloud resources from my accounts.


lib/charms/opensearch/v0/opensearch_config.py
```diff
@@ -108,6 +108,70 @@ def remove_oidc_auth(self):
...
+            self.SECURITY_CONFIG_YML,
+            "config/dynamic/authc/jwt_auth_domain/http_enabled",
+            False,
+        )
+
+        self._opensearch.config.put(
+            self.SECURITY_CONFIG_YML,
+            "config/dynamic/authc/jwt_auth_domain/transport_enabled",
+            False,
+        )
```
Mehdi-Bendriss, line 175: i think it makes sense to directly remove _all_ keys related to jwt. `self._opensearch.config.delete`

lib/charms/opensearch/v0/opensearch_jwt.py
```diff
@@ -0,0 +1,192 @@
...
+
+        relation_data = self.jwt_requires.fetch_relation_data([self.jwt_relation.id])
+        jwt_config = relation_data[self.jwt_relation.id]
+
+        self._validate_and_apply_jwt_auth_config(jwt_config)
+
+    def _on_jwt_relation_broken(self, event: ops.RelationBrokenEvent) -> None:
+        """Handle the removal of the relation."""
+        if self.charm.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR:
+            self.charm.status.clear(JWTRelationInvalid)
```
Mehdi-Bendriss: I was looking for this when i saw the status earlier, good job :)

tests/integration/relations/test_jwt_auth.py
```diff
@@ -0,0 +1,224 @@
...
+        ops_test,
+        apps=[APP_NAME],
+        apps_statuses=["active"],
+        units_statuses=["active"],
+        wait_for_exact_units=DEFAULT_NUM_UNITS,
+    )
+
+    # todo: replace with charm name once published
+    await ops_test.model.deploy("jwt-integrator", channel="1/edge")
+    await wait_until(ops_test, apps=[JWT_APP_NAME], apps_statuses=["blocked"])
```
Mehdi-Bendriss, line 68: Could you add the exact expected status message?

tests/integration/relations/test_jwt_auth.py
```diff
@@ -0,0 +1,224 @@
...
+
+    logger.info(f"Integrating {DATA_APP} with {JWT_APP_NAME} - this will result in blocked status")
+    await ops_test.model.integrate(
+        f"{JWT_APP_NAME}:{JWT_CONFIG_RELATION}",
+        f"{DATA_APP}:{JWT_CONFIG_RELATION}",
+    )
+    await wait_until(
+        ops_test,
+        apps=[DATA_APP],
+        units_statuses=["blocked"],
```
Mehdi-Bendriss: could you add the exact status messages?

lib/charms/opensearch/v0/opensearch_jwt.py
```diff
@@ -0,0 +1,192 @@
...
+    def jwt_relation(self) -> ops.Relation | None:
+        """Return the jwt relation if present."""
+        return self.jwt_requires.relations[0] if len(self.jwt_requires.relations) else None
+
+    def _on_jwt_relation_created(self, event: ops.RelationCreatedEvent) -> None:
+        """Handle relation creation."""
+        if self.charm.opensearch_peer_cm.deployment_desc().typ != DeploymentType.MAIN_ORCHESTRATOR:
+            # in large deployments, JWT configuration must only be handled by the main orchestrator
+            # this is a safeguard to avoid different sources for applying security configuration
+            self.charm.status.set(ops.BlockedStatus(JWTRelationInvalid))
```
Mehdi-Bendriss: shouldn't these statuses rather be application statuses?

tests/integration/relations/test_jwt_auth.py
```diff
@@ -0,0 +1,224 @@
...
+
+    logger.info("Test access to `/_cat/nodes` with JWT")
+    ip_address = await get_leader_unit_ip(ops_test, app=APP_NAME)
+    url = f"https://{ip_address}:9200/_cat/nodes"
+    result = requests.get(
+        url, headers={"Authorization": f"Bearer {generated_jwt['token']}"}, verify=False
+    )
+    assert result.status_code == 200, "Request failed"
+    logger.info("Access with JWT successful")
+
```
Mehdi-Bendriss, line 111: Could you ensure here that regular basic auth still works? to ensure that jwt didn't overrule basic auth

tests/integration/relations/test_jwt_auth.py
```diff
@@ -0,0 +1,224 @@
...
+    # remove Opensearch to allow for follow-up test
```
Mehdi-Bendriss, line 140: Could you add a relation JWT removal here, and ensure that jwt auth has been disabled? while basic auth still works

Mehdi-Bendriss (commented):
Thank you René! Great work! I have left some small comments, but the core logic is solid

lib/charms/opensearch/v0/opensearch_config.py
Mehdi-Bendriss: On second thoughts, I think it doesn't make sense to _remove_ them, but only set them back to their default values and only remove the ones that were added

reneradoi: Updated to reset all previously set config options.

tests/integration/relations/test_jwt_auth.py
reneradoi: I wouldn't do it in this specific case, as this is a status that is from JWT integrator. It is:
- not part of the Opensearch code base and not scope of testing here, we just want to wait for that situation to arrive
- in advanced-statuses format and therefore won't fit into the current `wait_until` implementation

tests/integration/relations/test_jwt_auth.py
reneradoi: updated

lib/charms/opensearch/v0/opensearch_jwt.py
reneradoi: changed all statuses to app statuses.

tests/integration/relations/test_jwt_auth.py
reneradoi: updated

tests/integration/relations/test_jwt_auth.py
reneradoi: updated

Mehdi-Bendriss (approved):
Thank you René!

PR #698, opened 2025-09-22
Group Common Code of `get_application_units` and `get_application_subordinate_units`

### LINKED ISSUE #488: Group common code in test helpers for application fetching
We need to remove duplicate code between `get_application_units` and  `get_application_subordinate_units` and extract common code.

akram09:
This PR provides a fix #488. Function `get_application_units` had some shared logic with `get_application_subordinate_units`. I have moved the shared logic to a separate function that will be called by the two functions. This provides a cleaner way of handling that logic.

- [ ] Manually
- [ ] Unit tests
- [X] Integration tests

## Checklist
- [ ] I have updated mainly integration tests helpers in `tests/integration/helpers_deployment`

### DIFF
--- a/tests/integration/helpers_deployments.py
+++ b/tests/integration/helpers_deployments.py
@@ -185,9 +185,7 @@ async def get_application_subordinate_units(
     # `get_unit_ip` should be replaced with `.public_address`
     raw_app = get_raw_application(ops_test, app)
     units = []
-    for principal_unit in get_raw_application(ops_test, principal_app)[
-        "units"
-    ].values():
+    for principal_unit in get_raw_application(ops_test, principal_app)["units"].values():
         u_name, raw_unit = None, None
         for u_name, raw_unit in principal_unit["subordinates"].items():
             if u_name.startswith(f"${app}/"):
@@ -199,9 +197,7 @@ async def get_application_subordinate_units(
             # unit not ready yet...
             continue
 
-        units.append(
-            _get_unit(ops_test, app, raw_app, u_name, raw_unit, subordinate=True)
-        )
+        units.append(_get_unit(ops_test, app, raw_app, u_name, raw_unit, subordinate=True))
     return await asyncio.gather(*units) if units else []
 
 
@@ -230,8 +226,7 @@ def _is_every_condition_on_app_met(
         any_match = False
         for status_val, messages in apps_full_statuses[app].items():
             any_match = any_match or (
-                app_status.value == status_val
-                and app_status.message in (messages or ["", None])
+                app_status.value == status_val and app_status.message in (messages or ["", None])
             )
         if not any_match:
             return False
@@ -300,9 +295,7 @@ async def _is_every_condition_met(
             units = await get_application_units(ops_test, app)
 
         if -1 < expected_units != len(units):
-            logger.info(
-                f"{app} -- expected units: {expected_units} -- current: {len(units)}"
-            )
+            logger.info(f"{app} -- expected units: {expected_units} -- current: {len(units)}")
             return False
 
         if (apps_statuses or apps_full_statuses) and not _is_every_condition_on_app_met(
@@ -368,9 +361,7 @@ async def wait_until(  # noqa: C901
     """
     if not apps:
         raise ValueError("apps must be specified.")
-    if not (
-        apps_statuses or apps_full_statuses or units_statuses or units_full_statuses
-    ):
+    if not (apps_statuses or apps_full_statuses or units_statuses or units_full_statuses):
         apps_statuses = ["active"]
         units_statuses = ["active"]
     if isinstance(wait_for_exact_units, int):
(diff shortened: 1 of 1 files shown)


tests/integration/helpers_deployments.py
```diff
@@ -119,46 +119,61 @@ async def get_unit_hostname(ops_test: OpsTest, unit_id: int, app: str) -> str:
...
 
 
+async def unit_from_raw(
+    ops_test: OpsTest,
+    unit_name: str,
+    raw_unit: Dict[str, Any],
+    app_name: str,
+    raw_app: Dict[str, Any],
+    subordinate: bool = False,
+) -> Unit:
```
Mehdi-Bendriss: I don't think this this function needs to be async.

nit: Perhaps, for consistency over the order and name of arguments (`app` is set as the first argument after ops_test in all test helpers)  
```suggestion
def _get_unit(
    ops_test: OpsTest,
    app: str,
    raw_app: dict[str, Any],
    unit_name: str,
    raw_unit: dict[str, Any],
    subordinate: bool = False,
) -> Unit:
```

tests/integration/helpers_deployments.py
```diff
@@ -119,46 +119,61 @@ async def get_unit_hostname(ops_test: OpsTest, unit_id: int, app: str) -> str:
...
+    """Create a Unit object from raw unit data."""
+    unit_id = int(unit_name.split("/")[-1])
+
+    app_id = f"{ops_test.model.uuid}/{app_name}"
+    app_short_id = md5(app_id.encode()).hexdigest()[:3]
+    if subordinate:
+        machine_id = -1
+    else:
+        machine_id = int(raw_unit["machine"])
```
Mehdi-Bendriss: nit, for conciseness:
```suggestion
    machine_id = -1 if subordinate else int(raw_unit["machine"])
```

Mehdi-Bendriss (commented):
Thanks Akram, I have 2 minor comments

tests/integration/helpers_deployments.py
akram09: Actually, I made it async since there is an `await get_unit_hostname(ops_test, unit_id, app_name)` in the function.

Mehdi-Bendriss (dismissed):

tests/integration/helpers_deployments.py
```diff
@@ -174,45 +186,17 @@ async def get_application_subordinate_units(
...
-                since=unit["juju-status"]["since"],
-            ),
-            app_status=Status(
-                value=raw_app["application-status"]["current"],
-                since=raw_app["application-status"]["since"],
-                message=raw_app["application-status"].get("message"),
-            ),
-        )
-
+        unit = await _get_unit(ops_test, app, raw_app, u_name, raw_unit, subordinate=True)
```
gatici: Here I would suggest to use asyncio.gather instead of awaiting one by one. This significantly shorten the processing time as gather run many independent I/O coroutines at the same time.

```
tasks = []
tasks.append(_get_unit(ops_test, app, raw_app, u_name, raw_unit))
return await asyncio.gather(*tasks) if tasks else []
```

tests/integration/helpers_deployments.py
```diff
@@ -174,45 +186,17 @@ async def get_application_subordinate_units(
     raw_app = get_raw_application(ops_test, app)
     units = []
     for principal_unit in get_raw_application(ops_test, principal_app)["units"].values():
-        u_name, unit = None, None
-        for u_name, unit in principal_unit["subordinates"].items():
+        u_name, raw_unit = None, None
+        for u_name, raw_unit in principal_unit["subordinates"].items():
             if app in u_name:
```
gatici: The statement `if app in u_name` may  pick wrong subs if names are overlapping.
For example:

subs = {"gcs-requirer/0": {}, "gcs/0": {}}
app = "gcs"
if we look for gcs in gcs-requirer/0, it will also match although we are looking for another.
Instead we can use `u_name.startswith(f"{app}/")`.

tests/integration/helpers_deployments.py
```diff
@@ -119,46 +119,58 @@ async def get_unit_hostname(ops_test: OpsTest, unit_id: int, app: str) -> str:
...
-                since=unit["juju-status"]["since"],
-            ),
-            app_status=Status(
-                value=raw_app["application-status"]["current"],
-                since=raw_app["application-status"]["since"],
-                message=raw_app["application-status"].get("message"),
-            ),
-        )
-
+        unit = await _get_unit(ops_test, app, raw_app, u_name, raw_unit)
```
gatici: Same comment here, we can use asyncio.gather which is more efficient.

gatici (commented):
Good job. I added some suggestions  regarding efficiency and edge cases.

gatici (dismissed):

Mehdi-Bendriss (dismissed):

gatici (approved):
LGTM

Mehdi-Bendriss (approved):

PR #705, opened 2025-10-02
Propagate backup credentials in large deployments

### LINKED ISSUE #693: Backup credentials not correctly transferred to subcluster keystores in large deployments
The backup credentials need to be added to keystores in all units of subclusters. The current implementation is only able to do so during a secret-changed event and not when the secret is initially transferred.

## Steps to reproduce
Tested with microceph and rgw enabled
1. Set up large deployment
```
juju deploy -n 1 opensearch main --channel 2/edge --config profile="testing"
juju deploy -n 1 opensearch data --channel 2/edge --config profile="testing" --config init_hold="true" --config roles="data"
juju deploy self-signed-certificates --config ca-common-name="Test CA" --channel="latest/stable"
juju integrate self-signed-certificates data
juju integrate main:peer-cluster-orchestrator data:peer-cluster
juju deploy s3-integrator
```
2. Add s3 credentials to s3-integrator
3. Relate s3-integrator with main
```
juju integrate s3-integrator main
```
## Expected behavior
Backups manager is set up correctly

## Actual behavior

Main enters blocked state with `Backup setup failed, check logs for details` message

If you test the same credentials with a simple deployment, it will work
## Versions
Charm revision: 274

smethnani:
Backup configuration is not working correctly in large deployments (see #693), resulting in a "Failed to set up backup" blocked state despite credentials being correct. 

This PR:
- Adds helper methods to store and load plugin related secrets
- Replaces code specific to S3 or Azure credentials in large deployments with methods for general plugin-related credentials
- Adds a `config-changed` event handler to `OpenSearchMainBackup` for the case when a previous revision of the charm is upgraded with a backup relation already established to create and propagate a secret with the credentials
- For the case of promoted failovers: adds a `missing_relations` flag to the peers data if credentials are found with missing relations. The flag is cleared along with the blocked status message after all missing relations are established with the new main orchestrator

Closes #693/[DPE-8339]

- [x] Manually
- [ ] Unit tests
- [x] Integration tests

[DPE-8339]: https://warthogs.atlassian.net/browse/DPE-8339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/opensearch/v0/helper_plugins.py
```diff
@@ -0,0 +1,60 @@
...
+    if not charm.unit.is_leader():
+        return
+
+    if (
+        charm.opensearch_peer_cm.deployment_desc().typ == DeploymentType.MAIN_ORCHESTRATOR
+        and charm.secrets.has(Scope.APP, label)
+        and not charm.secrets.is_secret_owner(Scope.APP, label)
+    ):
+        # current app is a promoted failover that does not own secret
+        charm.secrets.delete(Scope.APP, label)
```
reneradoi: I'm not sure if this will work - I thought only the owner of a secret can delete it?

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1129,9 +1115,18 @@ def apply_api_config_if_needed(self) -> None:
...
+        except ValueError as e:
             logger.error(f"Failed to setup backup service with exception: {e}")
             state = BackupServiceState.REPO_ERR_UNKNOWN
+        except OpenSearchHttpError as e:
+            logger.error(f"Failed to setup backup service with exception: {e.response_text}")
+            state = BackupServiceState.REPO_ERR_UNKNOWN
+            if NODES_MISSING_CREDS_ERR in e.response_text:
+                state = BackupServiceState.NODES_MISSING_CREDENTIALS
+
+        if state == BackupServiceState.NODES_MISSING_CREDENTIALS:
```
reneradoi: You could get rid of this condition (and by that, potentially the `noqa` too) and raise directly in the error handling right above.

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -66,6 +66,10 @@ def reload(self):
         Raises:
             OpenSearchHttpError: If the reload fails.
         """
+        if not self._opensearch.is_started():
+            # service not running, settings will be picked up at startup
+            return True
```
reneradoi, line 71: I was confused because the method doesn't have a return type - could you please add it, too?

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -527,7 +527,7 @@ def is_peer_cluster_orchestrator_relation_set(self) -> bool:
             is not None
         )
 
-    def rel_data(self, peek_secrets: bool = False) -> Optional[PeerClusterRelData]:
```
reneradoi, line 530: Nice that `peek_secrets` can be removed. I was not a big fan of this workaround.

reneradoi (commented):
Hi Shams! Just a high level feedback because of time constraints, hope that is ok?

lib/charms/opensearch/v0/helper_plugins.py
smethnani: 🙈 yes

smethnani: Removed + added integration test for backups after a failover promotion

lib/charms/opensearch/v0/opensearch_backups.py
smethnani: Remove `noqa`, thank you!

lib/charms/opensearch/v0/helper_plugins.py
```diff
@@ -0,0 +1,53 @@
...
+from charms.opensearch.v0.opensearch_internal_data import Scope
+from ops import ModelError, SecretNotFoundError
+
+if TYPE_CHECKING:
+    from charms.opensearch.v0.opensearch_base_charm import OpenSearchBaseCharm
+
+logger = logging.getLogger(__name__)
+
+
+def store_plugin_secret(
```
skourta, line 18: We can migrate these in the future to the plugin manager.

lib/charms/opensearch/v0/opensearch_backups.py
```diff
@@ -1129,9 +1114,15 @@ def apply_api_config_if_needed(self) -> None:
...
             snapshot_repo_resp = self.backup_manager.register(self.data)
             state = self.get_service_status(snapshot_repo_resp)
-        except (ValueError, OpenSearchHttpError) as e:
+        except ValueError as e:
             logger.error(f"Failed to setup backup service with exception: {e}")
             state = BackupServiceState.REPO_ERR_UNKNOWN
+        except OpenSearchHttpError as e:
+            logger.error(f"Failed to setup backup service with exception: {e.response_text}")
+            state = BackupServiceState.REPO_ERR_UNKNOWN
+            if NODES_MISSING_CREDS_ERR in e.response_text:
```
skourta, line 1123: Can you please explain this to me?

skourta (commented):
Thank you Shams. I left a question. I will do some manual tests by EOD

lib/charms/opensearch/v0/helper_plugins.py
smethnani: This requires OpensearchSecrets which needs a charm reference that the plugin manager doesn't have. That's why they are in helper methods. The plugin manager only has methods for state modification, not secret creation.

lib/charms/opensearch/v0/opensearch_backups.py
smethnani: This condition means the credentials are correct but that some nodes don't have them in their keystores yet. This early raise avoids the status displaying "Failed to set up backup" and then switching back to "Configuring backup..." before clearing. It will just show "Configuring backup..." and then be cleared when all nodes receive the credentials.

lib/charms/opensearch/v0/helper_plugins.py
skourta: Yes, that is why I mentioned in the future. My thinking is that we will have access to the secrets as part of the state in the future after refactoring.

lib/charms/opensearch/v0/opensearch_backups.py
skourta: And we detect this directly from OpenSearch?

smethnani: From the response text of the http call, yes

lib/charms/opensearch/v0/helper_plugins.py
smethnani: Don't secrets need to access the charm model though?

lib/charms/opensearch/v0/helper_plugins.py
```diff
@@ -0,0 +1,53 @@
...
+        logger.debug("Can't find secret '%s'", label)
+    except ModelError as e:
+        logger.debug("Cannot delete secret %s", label, e)
+    charm.plugin_manager.remove_plugin_config(Scope.APP, label)
+
+
+def decode_plugin_secret_content(content: dict, label: str) -> Optional[dict]:
+    """Decodes JSON payload from plugin secret"""
+    if not (raw := content.get(label)):
+        logger.warning("Secret %s has no %s payload", label, label)
```
gatici:
```suggestion
        logger.warning("Secret key %s not found in the content", label)
```

gatici (approved):
Good job Shams, that works as expected.
Thanks you!

PR #710, opened 2025-10-08
[DPE-8500] - V1 of the keystore rework

Mehdi-Bendriss:
The current keystore implementation needs improvement in:
- functionality (adding support for file)
- unnecessary handling of exceptions  

## Solution:
Rework the keystore implementation

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

## Checklist

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_keystore.py
+++ b/lib/charms/opensearch/v0/opensearch_keystore.py
@@ -8,9 +8,9 @@ This module manages OpenSearch keystore access and lifecycle.
 import functools
 import logging
 import os
-from abc import ABC, abstractmethod
 from typing import Any, Dict, List
 
+from charms.opensearch.v0.opensearch_distro import OpenSearchDistribution
 from charms.opensearch.v0.opensearch_exceptions import (
     OpenSearchCmdError,
     OpenSearchError,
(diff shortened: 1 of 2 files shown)


gatici (approved):
Looks good.

gatici (approved):

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -38,54 +38,62 @@ class OpenSearchKeystoreNotReadyError(OpenSearchKeystoreError):
...
-        ...
+        self._opensearch.run_bin("keystore", "create")
 
-    @functools.cached_property
-    @abstractmethod
-    def list(self) -> List[str]:
-        """Lists the keys available in opensearch's keystore."""
-        ...
+    def put_entries(self, entries: dict[str, str]) -> None:
+        """Add new key/val entries on the keystore."""
```
skourta, line 58: The remove, list, and reload methods have `_create_if_needed` but the put ones don't is this by design?

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -148,3 +160,9 @@ def reload_keystore(self) -> None:
         """
         response = self._opensearch.request("POST", "_nodes/reload_secure_settings")
         logger.debug(f"_update_keystore_and_reload: response received {response}")
+
+    # TODO delete once backups rework fully merged
+    def _clean_cache_if_needed(self):
```
skourta, line 165: Is reading the keystore keys a costly operation? I'm not sure the cache is needed in the first place.

skourta (commented):
Thank you. I left a couple of questions.

smethnani (approved):
Thanks @Mehdi-Bendriss ! Tested with a simple deployment and the s3-integrator. The current implementation of backups works as expected.

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -38,54 +38,62 @@ class OpenSearchKeystoreNotReadyError(OpenSearchKeystoreError):
...
+    def _create_if_needed(self) -> None:
+        """Creates the keystore if not already present."""
+        if os.path.exists(self._keystore_path):
+            return
 
-        Raises:
-            OpenSearchHttpError: If the reload fails.
-        """
-        ...
+        self._opensearch.run_bin("keystore", "create")
```
akram09, line 55: Is it fine to not verify again if the keystore has indeed been created ?

akram09 (commented):
Seems good to me, just one minor question

lib/charms/opensearch/v0/opensearch_keystore.py
Mehdi-Bendriss: There is no need when adding entries, as we're adding the `--force` flag which will create the keystore if it doesn't exist 
(You can see the explanation in a comment a couple of lines below)

lib/charms/opensearch/v0/opensearch_keystore.py
Mehdi-Bendriss: The cache is not needed indeed, and its removal will come in the plugins manager PR.
 
In this PR, it has been relocated from the removed abstract class to this one. This PR is the V1 of the rework, limited to the things we immediately need in the new backups rework (without breaking the legacy backup module), while Shams' feature branch gets merged and brings the latest version of the Keystore with these methods removed.

lib/charms/opensearch/v0/opensearch_keystore.py
Mehdi-Bendriss: I think the promise here is that if it is not created the command would fail, therefore propagating an exception `OpenSearchCmdError`

gatici (approved):

akram09 (approved):

PR #711, opened 2025-10-08
[DPE-8527] - extract certificates key/trust stores openssl logic into helpers

Mehdi-Bendriss:
The openssl operations on PKCS12 certificates and trust stores related operations currently in the TLS module will be reused by the backups modules. 

There was a bug in the previous TLS implementation that lost the whole cert chain when storing in the trust store with the exception of the leaf certificate.  

## Solution:
- Extract those into helpers
- Fix the bug described above by breaking down CA chains into individual entries in the trust store 

- [ ] Manually
- [x] Unit tests
- [x] Integration tests

## Checklist

### DIFF
--- a/lib/charms/opensearch/v0/helper_security.py
+++ b/lib/charms/opensearch/v0/helper_security.py
@@ -212,8 +212,8 @@ def list_cas(store_pwd: str, store_path: str) -> Optional[dict[str, str]]:
 
     certs = {}
     for cert in certificates:
-        alias = [line for line in cert if line.startswith("friendlyName:")][0]
-        alias = alias.split("friendlyName:")[-1]
+        alias = [line for line in cert.split("\n") if line.strip().startswith("friendlyName:")][0]
+        alias = alias.split("friendlyName:")[-1].strip()
 
         certs[alias] = f"{start_cert_marker}{cert.split(start_cert_marker)[1]}{end_cert_marker}"
 
--- a/lib/charms/opensearch/v0/helper_security.py
+++ b/lib/charms/opensearch/v0/helper_security.py
@@ -2,6 +2,7 @@
 # See LICENSE file for licensing details.
 
 """Helpers for security related operations, such as password generation etc."""
+import logging
 import math
 import os
 import secrets
@@ -9,9 +10,12 @@ import string
 import subprocess
 import tempfile
 from datetime import datetime
+from os.path import exists
 from typing import Optional, Tuple
 
 import bcrypt
+from charms.opensearch.v0.helper_charm import run_cmd
+from charms.opensearch.v0.opensearch_exceptions import OpenSearchCmdError
 from cryptography import x509
 
 # The unique Charmhub library identifier, never change it
@@ -23,6 +27,8 @@ LIBAPI = 0
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 LIBPATCH = 1
+
+logger = logging.getLogger(__name__)
 
 
 def hash_string(string: str) -> str:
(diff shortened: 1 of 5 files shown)


lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +123,213 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
     finally:
         os.unlink(tmp_key.name)
         os.unlink(tmp_pkcs8_key.name)
+
+
+def store_ca(
+    alias: str, store_pwd: str, store_path: str, ca: str, keep_previous: bool = True
+) -> bool:
+    """Add new CA cert to trust store."""
+    keytool = "opensearch.keytool"
```
gatici: This can be defined as a global string as it is locally defined in 3 different functions.
KEYTOOL = "opensearch.keytool"

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +123,213 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+                -keystore {store_path} \
+                -file {ca_tmp_file.name} \
+                -storetype PKCS12
+                """,
+                f"-storepass {store_pwd}",
+            )
+            run_cmd(f"sudo chmod +r {store_path}")
+            logger.info("New CA was added to truststore.")
+        except OpenSearchCmdError as e:
+            logging.error(f"Error storing the ca-cert: {e}")
```
gatici: There is an already defined logger, we should use it.
```suggestion
            logger.error(f"Error storing the ca-cert: {e}")
```

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +123,213 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+        return None
+
+    keytool = "opensearch.keytool"
+
+    # we fetch the list of stored aliases
+    cmd = f"{keytool} -list -keystore {store_path} -storetype PKCS12"
+    args = f"-storepass {store_pwd}"
+
+    try:
+        resp = run_cmd(cmd, args).out.split()
```
gatici: By default, .split() will divide using the empty spaces as a separator, then the rest of the code does not work as expected. 
We can use:
        resp = run_cmd(cmd, args).out.split("\n")  or
        resp = run_cmd(cmd, args).out.splitlines()

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +123,213 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+        return [
+            line.split("Alias name:")[-1].strip()
+            for line in resp
+            if line.startswith("Alias name:")
+        ]
+    except OpenSearchCmdError as e:
+        logging.error(f"Error reading the current truststore: {e}")
```
gatici:
```suggestion
        logger.error(f"Error reading the current truststore: {e}")
```

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +123,213 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+    """List the CAs current stored in a trust store."""
+    if not exists(store_path):
+        return None
+
+    cmd = f"openssl pkcs12 -in {store_path}"
+    args = f"-passin pass:{store_pwd}"
+    try:
+        stored_certs = run_cmd(cmd, args).out
+    except OpenSearchCmdError as e:
+        logging.error(f"Error reading the current truststore: {e}")
```
gatici:
```suggestion
        logger.error(f"Error reading the current truststore: {e}")
```

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +123,213 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+    cmd = f"openssl pkcs12 -export -in {tmp_cert.name} -inkey {tmp_key.name} -out {store_path} -name {name}"
+    args = f"-passout pass:{store_pwd}"
+    if key_pwd:
+        args = f"{args} -passin pass:{key_pwd}"
+
+    try:
+        run_cmd(cmd, args)
+        run_cmd(f"sudo chmod +r {store_path}")
+    except OpenSearchCmdError as e:
+        logging.error(f"Error storing the TLS certificates for {name}: {e}")
```
gatici:
```suggestion
        logger.error(f"Error storing the TLS certificates for {name}: {e}")
```

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +123,213 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+    tmp_ca_file.flush()
+    tmp_ca_file.seek(0)
+
+    try:
+        return run_cmd(f"openssl x509 -in {tmp_ca_file.name} -noout -issuer").out
+    except OpenSearchCmdError as e:
+        logger.error(f"Error reading the current truststore: {e}")
+        return None
+    finally:
+        tmp_ca_file.close()
```
gatici, line 328: This is also good but we can prefer to use context manager for guaranteed cleanup and file closing.
```suggestion
    with tempfile.NamedTemporaryFile(mode="w+t", dir="/tmp", delete=False) as f:
        f.write(cert)
        f.flush()
        path = f.name
    try:
        return run_cmd(f"openssl x509 -in {path} -noout -issuer").out
    except OpenSearchCmdError as e:
        logger.error(f"Error reading the current truststore: {e}")
        return None
    finally:
        os.remove(path)
```

gatici (commented):
Looks good. Some minor changes are suggested for improvement.

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,236 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
     finally:
         os.unlink(tmp_key.name)
         os.unlink(tmp_pkcs8_key.name)
+
+
+def split_ca_chain(pem_content: str) -> list[str]:
```
skourta, line 132: This can be simplified if you split by `-----END CERTIFICATE-----` and add it back.

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,236 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+
+    return certs
+
+
+def store_ca(
+    alias: str, store_pwd: str, store_path: str, ca: str, keep_previous: bool = True
+) -> bool:
+    """Add new CA cert to trust store."""
+    # This loop and split are to handle when the CA is a chain with intermediate certs
+    certs = list(reversed(split_ca_chain(ca)))
```
skourta, line 144: Should we keep reversing the order [given SSC fixed their bug](https://github.com/canonical/self-signed-certificates-operator/pull/349)?

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,236 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+    for line in pem_content.split("\n"):
+        current_cert += line + "\n"
+        if "-----END CERTIFICATE-----" in line:
+            certs.append(current_cert.strip())
+            current_cert = ""
+
+    return certs
+
+
+def store_ca(
```
skourta, line 139: Shouldn't these be keystore methods instead?

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,236 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+        if keep_previous:
+            cmd = f"{KEYTOOL} -changealias -alias {alias}-{index} -destalias old-{alias}-{index} -keystore {store_path} -storetype PKCS12"
+            args = f"-storepass {store_pwd}"
+            try:
+                run_cmd(cmd, args)
+                logger.info(f"Current CA {alias}-{index} was renamed to old-{alias}-{index}.")
+            except OpenSearchCmdError as e:
+                # This message means there was no "ca" alias or store before, if it happens ignore
+                if not (
+                    f"Alias <{alias}-{index}> does not exist" in e.out
```
skourta: This will raise if e.out is None

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,236 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+                    -keystore {store_path} \
+                    -file {ca_tmp_file.name} \
+                    -storetype PKCS12
+                    """,
+                    f"-storepass {store_pwd}",
+                )
+                run_cmd(f"sudo chmod +r {store_path}")
+                logger.info("New CA was added to truststore.")
+            except OpenSearchCmdError as e:
+                logging.error(f"Error storing the ca-cert: {e}")
```
skourta: Another logging to be changed.

skourta: Also consider [this rule](https://docs.astral.sh/ruff/rules/logging-f-string/)

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,236 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+    """Load stored CA cert."""
+    return (list_cas(store_pwd, store_path) or {}).get(alias)
+
+
+def remove_ca(alias: str, store_pwd: str, store_path: str) -> None:
+    """Remove old CA cert from trust store."""
+    if not exists(store_path):
+        return
+
+    list_cmd = f"{KEYTOOL} -list -keystore {store_path} -alias {alias} -storetype PKCS12"
```
skourta, line 260: Why list first? We can potentially delete and handle the raised exception just like we do for listing, no?

lib/charms/opensearch/v0/helper_security.py
skourta: same for `store_key_pair`

lib/charms/opensearch/v0/opensearch_tls.py
```diff
@@ -552,50 +558,15 @@ def store_new_ca(self, secrets: Dict[str, Any]) -> bool:  # noqa: C901
...
-                logging.error(f"Error storing the ca-cert: {e}")
-                return False
+        is_ca_stored = store_ca(
+            alias=CA_ALIAS,
+            store_pwd=admin_secrets.get("truststore-password"),
+            store_path=f"{self.certs_path}/{CA_ALIAS}.p12",
+            ca=secrets.get("ca-cert"),
+            keep_previous=True,
+        )
+        if not is_ca_stored:
```
skourta:
```suggestion
        if not store_ca(
            alias=CA_ALIAS,
            store_pwd=admin_secrets.get("truststore-password"),
            store_path=f"{self.certs_path}/{CA_ALIAS}.p12",
            ca=secrets.get("ca-cert"),
            keep_previous=True,
        ):
```

skourta (commented):
Thank you Mehdi. I left some small comments.

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,236 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+                    """,
+                    f"-storepass {store_pwd}",
+                )
+                run_cmd(f"sudo chmod +r {store_path}")
+                logger.info("New CA was added to truststore.")
+            except OpenSearchCmdError as e:
+                logging.error(f"Error storing the ca-cert: {e}")
+                return False
+
+    return True
```
smethnani, line 188: Can the certs list be empty and if so should we return `True`?

tests/unit/lib/test_opensearch_tls.py
```diff
@@ -742,7 +742,7 @@ def test_on_certificate_available_any_node_unit_cert_full_workflow(
     )
     @patch("charms.opensearch.v0.opensearch_tls.OpenSearchTLS._add_ca_to_request_bundle")
     @patch("charms.opensearch.v0.opensearch_tls.tempfile.NamedTemporaryFile")
```
smethnani, line 744: Some changes needed for this test to pass (and possibly similar changes to others):
```
# L790
new_ca = """-----BEGIN CERTIFICATE-----
new_ca
+-----END CERTIFICATE-----"""
```

```
# L839
-  "keytool *-changealias *-alias ca *-destalias old-ca",
+  "keytool *-changealias *-alias ca-0 *-destalias old-ca-0",
```

smethnani (commented):
Thanks @Mehdi-Bendriss ! I left a question

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,236 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+                    f"""{KEYTOOL} -importcert \
+                    -noprompt \
+                    -alias {alias}-{index} \
+                    -keystore {store_path} \
+                    -file {ca_tmp_file.name} \
+                    -storetype PKCS12
+                    """,
+                    f"-storepass {store_pwd}",
+                )
+                run_cmd(f"sudo chmod +r {store_path}")
```
gatici, line 182: We do not need to run a  command instead we can use os lib to change the access rights:
os.chmod(store_path, 0o640)

lib/charms/opensearch/v0/helper_security.py
gatici: We should avoid using f-strings during logging. f-strings will always be interpreted even if we really do not need to call them.

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,236 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+
+    # parse output to retrieve the current CA (in case there are many)
+    certificates = split_ca_chain(stored_certs)
+
+    start_cert_marker = "-----BEGIN CERTIFICATE-----"
+
+    certs = {}
+    for cert in certificates:
+        alias = [line for line in cert.split("\n") if line.strip().startswith("friendlyName:")][0]
+        alias = alias.split("friendlyName:")[-1].strip()
```
gatici, line 233: How can we be sure that [line for line in cert.split("\n") if line.strip().startswith("friendlyName:")] always exits ?
I think `friendlyName` may optionally exists in PKCS12 certs.

lib/charms/opensearch/v0/helper_security.py
gatici: The clause seems correct as it will ignore not found cases.

tests/unit/lib/test_opensearch_tls.py
gatici: Absolutely agree.
L839-843:
```
        assert re.search(
            "keytool *-changealias *-alias ca-0 *-destalias old-ca-0",
            run_cmd.call_args_list[0].args[0],
        )
        assert re.search("keytool *-importcert.* *-alias ca-0", run_cmd.call_args_list[1].args[0])
```

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: This is accounting for that fix. This is essentially to give the index 0 to the root CA (which comes last in the CA chain that's correctly ordered) in the trust store (since we iterate over `range(len)`)

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: No, the keystore is used to store credentials, based on `bin/opensearch-keystore`.
This is about PKCS12 trust stores for CA certificates.

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: good catch!

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: In this particular occurrence I believe the addition of a context manager is not very helpful, as the deletion is still warranted in the `finally`.
In addition to this, in the `store_key_pair` 2 context managers are needed, therefore 2 `os.remove` calls needed in the `finally`. Which ultimately would just add 1 level of indentation to the code

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: Smail is right for the case where the `run_cmd` raises an exception due to one `TimeoutError, subprocess.TimeoutExpired` (though rare in these local operations), in which case, we do not build the exception with `out=`, as we don't have the output of the `subprocess.run`.

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: Yes! I'm aware of this topic, however those were moved over from the TLS module 😬

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: Though I agree with your reasoning looking at the code, this was copied over from the TLS module, and I only reworked what I specifically needed to expose for both TLS **and** Backups. 
As well as the fix on +1-cert ca-chains only the leaf was stored in the trust store. The rest is simply relocated and I'd rather keep it as-is for the time being, given it was tested as such.

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: We are the ones storing entries in the trust store. And we always set an alias for inserted certificates. So it is guaranteed to always have values set in the friendlyName.

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: i think it is the same, run_cmd is a wrapper over `subprocess.run`

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: from the code, the validation always happens upfront when testing on the existence of the `ca-cert` secret, so this should not occur

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,232 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+def store_ca(
+    alias: str, store_pwd: str, store_path: str, ca: str, keep_previous: bool = True
+) -> bool:
+    """Add new CA cert to trust store."""
+    # This loop and split are to handle when the CA is a chain with intermediate certs
+    certs = list(reversed(split_ca_chain(ca)))
+
+    for index in range(len(certs)):
+        if keep_previous:
+            cmd = f"{KEYTOOL} -changealias -alias {alias}-{index} -destalias old-{alias}-{index} -keystore {store_path} -storetype PKCS12"
```
akram09: Question: What will happen to the stored old ca ? Are we just going to store them in case of need or there is a different process that will leverage them ? If that is the case, I think the name of the alias should be normalized in a separate constant instead of specifying "old-{alias}-{index}"

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -117,3 +126,232 @@ def to_pkcs8(private_key: str, password: Optional[str] = None) -> str:
...
+            for line in resp
+            if line.startswith("Alias name:")
+        ]
+    except OpenSearchCmdError as e:
+        logger.error("Error reading the current truststore: %s", e)
+        return None
+
+
+def list_cas(store_pwd: str, store_path: str) -> Optional[dict[str, str]]:
+    """List the CAs current stored in a trust store."""
```
akram09: nit: "List the CAs currently stored in a trust store"

gatici (dismissed):
LGTM

lib/charms/opensearch/v0/helper_security.py
akram09: I got my answer when reading the tls module, but still think that if we are using the function from another module we would still need to be aware of this. 
I would personally have a constant OLD_CA_PREFIX="old" . But this is very minor

akram09 (dismissed):
Seems good for me

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: good idea, done!

gatici (dismissed):

akram09 (dismissed):

akram09 (approved):

gatici (approved):

PR #724, opened 2025-10-20
Pin lxc image for jammy

smethnani:
There is an issue with the jammy image used by lxd which is causing charms running on 22.04 (eg self-signed-certificates) to be stuck in "waiting for machine" state and breaking most tests.

## Solution:
Pin the image to a working one while the issue is being fixed

- [ ] Manually
- [ ] Unit tests
- [x] Integration tests

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -140,12 +140,6 @@ jobs:
       - timeout-minutes: 1
         if: ${{ success() || (failure() && steps.spread.outcome == 'failure') }}
         run: snap list
-      - name: lxc image list
-        timeout-minutes: 1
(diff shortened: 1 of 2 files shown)


skourta:
Why not just deploy SSC with 24 base?

smethnani:
> Why not just deploy SSC with 24 base?

@skourta `latest/stable` just runs on 22.04. Same goes for the s3-integrator `1/stable` that we use in the tests. ~~Preferrable to change the channels?~~ Azure-storage-integrator only runs on 22.04

Mehdi-Bendriss (dismissed):

Mehdi-Bendriss (approved):

PR #730, opened 2025-10-22
[DPE-8405] Restrict relation with OAuth provider to main-orchestrator

### LINKED ISSUE #699: Restrict OAuth relation to main-orchestrator in large deployments
If OAuth is enabled in a large deployment of Opensearch, inconsistencies could occur if users move the relation between the Opensearch applications. This should be restricted to only integrate the main-orchestrator with the OAuth provider.

In order to enable OAuth, the `securityadmin` script is run on the leader unit of the application that has the `oauth` relation. If an `oauth`-relation is created on multiple applications, this could lead to issues on updating the security index.

Currently the configuration for OAuth is not handled on the `peer-cluster` relation and not really supported yet for large deployments. By restricting the `oauth`-relation to the main-orchestrator, this would allow for support of OIDC in large deployments.

The same restriction has been implemented for JWT auth, see [here](https://github.com/canonical/opensearch-operator/blob/2/edge/lib/charms/opensearch/v0/opensearch_jwt.py#L78-L80).

reneradoi:
Fixes https://github.com/canonical/opensearch-operator/issues/699.

## Solution:
This PR restricts the OAuth relation in large deployments of Opensearch to only be created with the main-orchestrator application of the cluster. Otherwise the OAuth configuration will not be applied to Opensearch and a `blocked` status will be displayed.

The solution is in line with the one for JWT authentication (implementation and UX are the same).

- [X] Manually
- [ ] Unit tests
- [X] Integration tests

The PR also adds integration test coverage for this use case:
- creating a large deployment consisting of cluster-manager-only and data-only nodes
- verifying `blocked` status and OAuth access not possible if relation is created with `Data` application
- when the relation is established with the `Main` application, OAuth access is possible on the `Data` nodes.

Furthermore, manual tests for both small and large deployments have been conducted, ensuring that OAuth and JWT authentication can be used in parallel and both work.

### DIFF
--- a/lib/charms/opensearch/v0/constants_charm.py
+++ b/lib/charms/opensearch/v0/constants_charm.py
@@ -76,6 +76,7 @@ JWTAuthConfigInvalid = (
     "Configuration for JWT authentication is invalid. Check and correct parameters."
 )
 JWTRelationInvalid = "JWT relation must be created with Main-cluster-orchestrator"
+OAuthRelationInvalid = "OAuth relation must be created with Main-cluster-orchestrator"
 SecurityIndexUpdateError = "Failed to update security configuration, check logs for details."
 
 # Wait status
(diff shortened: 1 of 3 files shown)


Mehdi-Bendriss (approved):
Great change! thank you René!

akram09 (approved):
Nice work @reneradoi !

PR #733, opened 2025-10-28
[DPE-7059] Raise on the last cluster manager unit

skourta:

This pull request introduces a new error handling mechanism for the scenario where no cluster manager units are available in an OpenSearch cluster. The changes ensure that the system will raise a specific exception and log an error if all cluster managers are lost, helping to prevent cluster state loss and improve operational robustness.

## Solution:

**Error handling improvements:**

* Added a new exception class `OpenSearchNoClusterManagersError` to `opensearch_exceptions.py` for cases when there are no cluster managers in the cluster.
* Updated the imports in `opensearch_base_charm.py` to include `OpenSearchNoClusterManagersError`.
* In the `_on_opensearch_data_storage_detaching` method, added logic to check for the presence of cluster managers and raise `OpenSearchNoClusterManagersError` if none are up, with an appropriate error log message.

**Cluster manager status checks:**

* Added the method `does_cluster_have_cm_up` to `opensearch_peer_clusters.py` to determine if at least one cluster manager unit is up in the fleet.

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

## Checklist

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -690,13 +690,17 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
                     self.peer_cluster_provider.clean_all_relation_data()
                 elif self.opensearch_peer_cm.is_consumer():
                     self.peer_cluster_requirer.refresh_requirer_relation_data()
-                
+
             # No cluster managers left in the cluster fleet
             # raise so we do not lose the cluster state
-            if len(self.opensearch_peer_cm.apps_in_fleet()) > 0 and not self.opensearch_peer_cm.does_cluster_have_cm_up():
-                logger.error("No cluster managers left in the cluster fleet. Please scale up your cluster manager units.")
+            if (
+                len(self.opensearch_peer_cm.apps_in_fleet()) > 0
+                and not self.opensearch_peer_cm.does_cluster_have_cm_up()
+            ):
+                logger.error(
+                    "No cluster managers left in the cluster fleet. Please scale up your cluster manager units."
+                )
                 raise OpenSearchNoClusterManagersError()
-                
 
         # we attempt to flush the translog to disk
         if self.opensearch.is_node_up():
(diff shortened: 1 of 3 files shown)


lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -690,6 +691,24 @@ def _on_opensearch_data_storage_detaching(self, event: StorageDetachingEvent):
...
+                    ]
+                )
+                > 0
+                and not self.opensearch_peer_cm.does_cluster_have_cm_up()
+            ):
+                logger.error(
+                    "No cluster managers left in the cluster fleet. Please scale up your cluster manager units."
+                )
+                raise OpenSearchNoClusterManagersError()
+
```
gatici, line 711: This only raises if there are other Apps in the fleet but no CM is up. If there are no other apps and CM is down, it will not raise. But this situation may still  happen. 
What do you think about changing the condition like if cluster is initialized but there is no CM, raise ?
```
if cluster_initialized and not cm_up:
       raise OpenSearchNoClusterManagersError()
```

skourta: We do not want to block a small deployment app from being removed. If there are no other apps then this is a small deployment. That was my line of thinking.

gatici: OK, that's reasonable.

gatici (dismissed):

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -768,3 +768,14 @@ def rel_data_from_str(self, redacted_dict_str: str) -> PeerClusterRelData:
             )
 
         return PeerClusterRelData.from_dict(content)
+
+    def does_cluster_have_cm_up(self) -> bool:
```
Mehdi-Bendriss: nit, better naming:
```suggestion
    def is_any_cm_node_up_in_cluster(self) -> bool:
```

Mehdi-Bendriss (commented):
Thanks Smail! Looks good. Just a nit on naming

Mehdi-Bendriss (approved):

gatici (approved):
LGTM

PR #718, opened 2025-10-15
[DPE-7823] Snapshots rework

gatici:
## Description of issue or feature:

The existing backup flow is very difficult to digest and to welcome new changes. That's why a new backup/restore flow is implemented to increase the flexibility and simplify the operations.

## Solution:

**Design**

A single events manager (OpenSearchSnapshotsEvents) owns all snapshot actions and storage relations.

A OpenSearchSnapshotsManager is the interface which talks to OpenSearch for all necessary operations(create repo/snapshot/restore, list, status, etc).The OpenSearchSnapshotsManager supports the existing charm actions which are create-backup, list-backups and restore.

Only one active object-storage backend is acceptable otherwise conflict occurs and it will be blocked untill the conflict is resolved by user.

S3 Rados GW is supported and S3 custom CA  operations are handled (store/remove CA in cacerts.p12). Opensearch restart necessity is detected and performed.

All actions run after pre-requisites are meet and pre-requisities are properly defined and actionable errors are thrown.

Main orchestrator publishes storage “triggers” + secret URIs and non-main orchestrators apply locally (keystore, optional CA, repo ensure).

**Handling Events**

Single cluster and main orchestrator acts on _on_s3_credentials_changed and _on_azure_credentials_changed events.
- publishes effective storage creds to peer-clusters
- persist snapshot-object-storage-type in peers data
- write credentials to keystore and reload
- create/ensure repository existence
- handle S3 CA optionally and trigger a restart if needed
- publishes secret URIs (not raw creds) and grants access, sub-clusters resolve via model.get_secret.

Single cluster and main orchestrator acts on _on_s3_credentials_gone/departed and _on_azure_credentials_departed events.
- remove repo (best-effort), delete keystore entries, reload, clear peers flag.
- remove S3 CA optionally (and request restart if needed).

Large/non-main Orchestrators acts on peer_orchestrator_relation_changed/gone events. events.
- get the credentials from peer cluster orchestrator relation databag
- store S3 or Azure credentials in the keystore
- store S3 CA and restart Opensearch if needed
- ensure repo existence

Large/non-main Orchestrators acts on custom object_storage_changed events.
- remove repo + keystore entries (and S3 CA if present).
- optionally request restart to clean up CA usage.

## How was this change tested?
- [x] Manually
- [x] Unit tests
- [x] Integration tests

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

PR #738, opened 2025-11-18
Update list_cas and remove_ca to account for CA index

smethnani:
`list_cas` expects stored CA aliases to be suffixed with an index. Older revisions of the charm do not use this format for CA aliases and so this causes a ValueError to be raised after upgrading.

Additionally, the logic to remove CAs after a CA rotation does not account for the index format and only looks to remove a CA with alias `old-ca`, so the old CA/chain isn't actually removed.

## Solution
`list_cas`: Check first if the ca alias ends with a digit before attempting to cast it to an int.
`remove_ca`: Match CAs which start with the alias to remove all `<alias>-i`

### DIFF
--- a/lib/charms/opensearch/v0/helper_security.py
+++ b/lib/charms/opensearch/v0/helper_security.py
@@ -259,7 +259,7 @@ def remove_ca(alias: str, store_pwd: str, store_path: str) -> None:
         return
 
     aliases = list_aliases(store_pwd, store_path) or []
-    to_remove = [a for a in aliases if a == alias or a.startswith(f"{alias}-")]
+    to_remove = [a for a in aliases if a.startswith(f"{alias}-")]
 
     for a in to_remove:
         del_cmd = f"{KEYTOOL} -delete -keystore {store_path} -alias {a} -storetype PKCS12"
--- a/lib/charms/opensearch/v0/helper_security.py
+++ b/lib/charms/opensearch/v0/helper_security.py
@@ -258,19 +258,14 @@ def remove_ca(alias: str, store_pwd: str, store_path: str) -> None:
     if not exists(store_path):
         return
 
-    list_cmd = f"{KEYTOOL} -list -keystore {store_path} -alias {alias} -storetype PKCS12"
-    list_args = f"-storepass {store_pwd}"
-    try:
-        run_cmd(list_cmd, list_args)
-    except OpenSearchCmdError as e:
-        # This message means there was no "ca" alias or store before, if it happens ignore
-        if e.out and f"Alias <{alias}> does not exist" in e.out:
-            return
+    aliases = list_aliases(store_pwd, store_path) or []
+    to_remove = [a for a in aliases if a == alias or a.startswith(f"{alias}-")]
 
-    del_cmd = f"{KEYTOOL} -delete -keystore {store_path} -alias {alias} -storetype PKCS12"
-    del_args = f"-storepass {store_pwd}"
-    run_cmd(del_cmd, del_args)
-    logger.info("Removed %s from truststore.", alias)
+    for a in to_remove:
+        del_cmd = f"{KEYTOOL} -delete -keystore {store_path} -alias {a} -storetype PKCS12"
+        del_args = f"-storepass {store_pwd}"
+        run_cmd(del_cmd, del_args)
+        logger.info("Removed %s from truststore.", a)
 
 
 def store_key_pair(
--- a/lib/charms/opensearch/v0/helper_security.py
+++ b/lib/charms/opensearch/v0/helper_security.py
@@ -259,7 +259,7 @@ def remove_ca(alias: str, store_pwd: str, store_path: str) -> None:
         return
 
     aliases = list_aliases(store_pwd, store_path) or []
-    to_remove = [a for a in aliases if a.startswith(f"{alias}")]
+    to_remove = [a for a in aliases if a.startswith(alias)]
 
     for a in to_remove:
         del_cmd = f"{KEYTOOL} -delete -keystore {store_path} -alias {a} -storetype PKCS12"
--- a/lib/charms/opensearch/v0/helper_security.py
+++ b/lib/charms/opensearch/v0/helper_security.py
@@ -233,8 +233,9 @@ def list_cas(store_pwd: str, store_path: str) -> Optional[dict[str, str]]:
         alias = alias.split("friendlyName:")[-1].strip()
 
         alias_split = alias.split("-")  # support for CA chains with multiple intermediate CAs
-        ca_index = int(alias_split[-1])
-        ca_alias = "-".join(alias_split[:-1])
+        last = alias_split[-1]
+        ca_index = int(last) if last.isdigit() else 0
+        ca_alias = "-".join(alias_split[:-1]) if last.isdigit() else alias
         certs.setdefault(ca_alias, []).insert(
             ca_index, f"{start_cert_marker}{cert.split(start_cert_marker)[1]}"
         )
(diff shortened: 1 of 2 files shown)


lib/charms/opensearch/v0/helper_security.py
```diff
@@ -257,19 +258,14 @@ def remove_ca(alias: str, store_pwd: str, store_path: str) -> None:
...
-        # This message means there was no "ca" alias or store before, if it happens ignore
-        if e.out and f"Alias <{alias}> does not exist" in e.out:
-            return
-
-    del_cmd = f"{KEYTOOL} -delete -keystore {store_path} -alias {alias} -storetype PKCS12"
-    del_args = f"-storepass {store_pwd}"
-    run_cmd(del_cmd, del_args)
-    logger.info("Removed %s from truststore.", alias)
+    aliases = list_aliases(store_pwd, store_path) or []
+    to_remove = [a for a in aliases if a == alias or a.startswith(f"{alias}-")]
```
skourta: You can simplify the condition to `a.startswith(alias)` no?

skourta (commented):
LGTM. I left a small comment. We should check with @gatici because I believe she fixed the same issue in her branch as well

lib/charms/opensearch/v0/helper_security.py
smethnani: True, thank you!

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -257,19 +258,14 @@ def remove_ca(alias: str, store_pwd: str, store_path: str) -> None:
...
-        # This message means there was no "ca" alias or store before, if it happens ignore
-        if e.out and f"Alias <{alias}> does not exist" in e.out:
-            return
-
-    del_cmd = f"{KEYTOOL} -delete -keystore {store_path} -alias {alias} -storetype PKCS12"
-    del_args = f"-storepass {store_pwd}"
-    run_cmd(del_cmd, del_args)
-    logger.info("Removed %s from truststore.", alias)
+    aliases = list_aliases(store_pwd, store_path) or []
+    to_remove = [a for a in aliases if a.startswith(alias)]
```
gatici, line 262: The prefix matching may be to broad. If there are some similar CA names, this may remove all of them. For example if we would like to remove `myalias` but in the truststore there is also myalias-xxx, both of them will be deleted.  We also need to consider the old CAs which has  a suffix OLD_CA_PREFIX. def store_ca allows to keep previous versions of CAs using this prefix.

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -257,19 +258,14 @@ def remove_ca(alias: str, store_pwd: str, store_path: str) -> None:
...
+
+    for a in to_remove:
+        del_cmd = f"{KEYTOOL} -delete -keystore {store_path} -alias {a} -storetype PKCS12"
+        del_args = f"-storepass {store_pwd}"
+        run_cmd(del_cmd, del_args)
```
gatici, line 267: We need to handle OpenSearchCmdError for the cases that CA's in the `to_remove` is removed before deleting them.  If the alias disappear before we delete it, it should not cause a failure.

lib/charms/opensearch/v0/helper_security.py
gatici, line 262: We may add some logging with debug level such as if to_remove is empty, there is nothing to remove.

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -233,8 +233,9 @@ def list_cas(store_pwd: str, store_path: str) -> Optional[dict[str, str]]:
         alias = alias.split("friendlyName:")[-1].strip()
 
         alias_split = alias.split("-")  # support for CA chains with multiple intermediate CAs
-        ca_index = int(alias_split[-1])
-        ca_alias = "-".join(alias_split[:-1])
+        last = alias_split[-1]
+        ca_index = int(last) if last.isdigit() else 0
```
gatici, line 237: We are assuming that indices always start with 0  and continuous. In case indices are missing or out of order, certs may include empty slots or unexpected order. 
In the loop, instead of insert, we can use append and then sort the items before joining them such as:
```
chains.setdefault(ca_alias, []).append(
    (ca_index, f"{start_cert_marker}{cert.split(start_cert_marker, 1)[1]}".strip())
)
later:
for alias, items in chains.items():
    items.sort(key=lambda t: t[0])
    cas[alias] = "\n".join(pem for _, pem in items if pem)
    ```

gatici (changes requested):
Thank you Shams! I commented regarding handling some possible edge cases.

lib/charms/opensearch/v0/helper_security.py
smethnani: Shouldn't all stored ca's have an index now? The only case they will not is if the charm was upgraded from a revision where we did not use indices, in which case isn't it okay to just set the index to 0 as there is no chain? The rest leaves the logic unchanged.

lib/charms/opensearch/v0/helper_security.py
smethnani: If the prefix is `OLD_CA_PREFIX`, then shouldn't that be passed as an argument to`remove_ca`? 

We currently only call this with `remove_ca(OLD_CA_ALIAS)` after a ca rotation. It accounts for `OLD_CA_PREFIX` but does not account for the format with the indices. This just ensures that they are actually removed, the current implementation does not remove the old ca (or chain) since the alias no longer matches

lib/charms/opensearch/v0/helper_security.py
smethnani: `to_remove` is created from `list_aliases` which returns aliases that currently exist, isn't that enough to be sure they are there? In which case would they disappear?

lib/charms/opensearch/v0/helper_security.py
gatici: OK, thank you for clarification. My comment on too broad prefix matching is still valid. According to line 148 old CA will be in a format of {OLD_CA_PREFIX}{alias}-{index}.  For example if remove_ca("old-ca") is called,  it will remove all the patterns that starts with "old-ca" and would remove `old-ca-other-1`, `old-ca2`, `old-cafe`.  We need make sure that last part of the split is only the digits.

lib/charms/opensearch/v0/helper_security.py
gatici: Although it is rare situation, there is a small delta between listing resources and deleting the resources. 
Another hook of the charm might be removing the same alias simultaneously. Or a user might be manipulating things manually by deleting aliases.

lib/charms/opensearch/v0/helper_security.py
gatici: Normally, we are expecting that all the ca's will have index in order. But this not the reality. https://docs.openssl.org/3.5/man1/openssl-pkcs12/#pkcs12-output-export-options There is note there: ` There is no guarantee that the first certificate present is the one corresponding to the private key. Certain software which tries to get a private key and the corresponding certificate might assume that the first certificate in the file is the one corresponding to the private key, but that may not always be the case. `
The order of certificates in the output is not deterministic. From the OpenSSL documentation, the order depends on:
- internal hash table iteration order
- the order they were originally inserted
- PKCS12 file structure (which doesn't guarantee ordering)
That's why we sort them at the end. For removal, the order is not important but TLS validation, they are expected to be in order such as root->intermediate-> leaf or  in the reverse order.
Also if we consider upgrade scenario:

alias = "ca"
ca_index = 0  # defaulted here

new certs
alias = "ca", ca_index = 0
alias = "ca", ca_index = 1

legacy ca ->  defaults to index 0
new ca-0 ->  explicitly index 0

One of them will override the another one as both of them claiming the index 0.
Instead if we append them to the list, both version will be maintained.

smethnani:
Superseded by #718

PR #743, opened 2025-11-27
[DPE-9022] Add rollback docs

smethnani:
Rollbacks aren't supported in upstream Opensearch. A charm user may run `juju refresh` to an older revision mid-upgrade, which will install a lower version of OpenSearch on an already-upgraded unit which will then fail to start. 

This PR:
- adds instructions to restore the cluster to an operable state
- removes suggestions to rollback mid-upgrade
- fixes some typos and adds custom words to `docs/.custom_wordlist.txt` to satisfy the spell-checker
## Checklist


docs/how-to/upgrade.md
```diff
@@ -96,8 +92,6 @@ If the deployment is of a local charm, save a copy of the current `.charm` file.
 Optionally, it is recommended to scale the application up by one unit before upgrading.
 
 The new unit will be the first one to be updated, and it will assert that the upgrade is possible.
-In case of failure, having the extra unit will ease the rollback procedure,
```
Mehdi-Bendriss, line 108: i believe we can keep this, with the emphasize that this will be dropped, no?

docs/how-to/upgrade/recover-from-rollback.md
```diff
@@ -0,0 +1,190 @@
...
+      "node_decision": "no",
+      "store": {
+        "found": false
+      }
+    }
+  ]
+}
+```
+
+Delete the index. Replace `index1` with the name of the index identified in the previous step:
```
akram09: I think we should mention that we need to have made already a backup to make sure there is no data loss.

docs/how-to/upgrade/recover-from-rollback.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Add a New Unit
+Add a replacement unit to restore the desired scale for your application:
+
+```
+juju add-unit opensearch -n 1
+```
+
+## Remove Rolled Back Unit
+Remove the rolled back unit. Replace `opensearch/2` with the unit that was rolled back:
```
akram09: The cluster state needs to be in a green state, and the node has not started. I have followed the guide and since my cluster wasn't red at the begining I didn't have to delete any index. However when I deleted the unit, the cluster became in a red state. I think we should talk about this case in this guide since users may fall into it

akram09 (changes requested):

docs/how-to/upgrade/recover-from-rollback.md
smethnani: The rolled back node shouldn't have been able to start, so I don't see how deleting it would have led to a red cluster state (it was likely already in a red cluster state if this is what you saw). Do you happen to still have any logs from this test or info about which indices were lost? Or maybe a record of the commands you called?

docs/how-to/upgrade/recover-from-rollback.md
smethnani: I added a warning before the command to clarify that the index won't be recoverable without a snapshot

akram09 (dismissed):
Very Nice, thanks @smethnani

gatici (dismissed):
Thanks Shams!

docs/how-to/upgrade/recover-from-rollback.md
```diff
@@ -0,0 +1,224 @@
...
+During the upgrade process, the routing allocation setting may be restricted to `primaries`.
+Restore normal allocation by enabling all routing:
+
+```text
+PUT _cluster/settings
+{
+  "persistent": {
+    "cluster.routing.allocation.enable": "all"
+  }
+}
```
izmalk: Question: The JSON is part of the PUT query, right?

izmalk (dismissed):
LGTM now.

I've implemented some cosmetic changes in the how-to guide:
1. Fixed formatting issues.
2. Improved wording.

Please bear in mind that, according to our style guide, headings use sentence case, there is a blank line before and after each heading, and every code block should specify a syntax highlighting language.

docs/how-to/upgrade/recover-from-rollback.md
smethnani: Indeed it is. Thanks @izmalk !

akram09 (dismissed):

skourta:
The docs on opensearch for any version under 3 seem to return 404 for now. I'll monitor them for a while and see what we can do.

docs/how-to/upgrade.md
```diff
@@ -461,10 +461,67 @@ Machine  State    Address         Inst id        Base          AZ  Message
...
 
+#### Rollback a charm revision with a different workload version
+
+If the charm revision you are rolling back to has a different workload version, the rollback process will attempt to roll back the workload version as well. However, since OpenSearch does not support downgrading, the rollback process will rollback the charm code then attempt a best-effort rollback of the OpenSearch workload. 
+
+This is a dangerous operation that may lead to an unhealthy OpenSearch cluster. It is recommended to instead perform a backup and restore of the cluster to a new deployment with the desired OpenSearch version.
+
+We do not recommend performing a rollback to a charm revision with a different OpenSearch version. However, if you choose to proceed, you can follow the same steps as rolling back to a charm revision with the same workload version. There are two possible outcomes after the rollback:
+
+##### It is possible to perform a best-effort rollback between the two workload revisions
```
izmalk: Suggestion: Let's try to make headings smaller than that. Also, revision is for a charm. Workload has a version. Just to make it a bit less confusing.


```suggestion
##### If the rollback between the versions is possible
```

docs/how-to/upgrade.md
```diff
@@ -461,10 +461,67 @@ Machine  State    Address         Inst id        Base          AZ  Message
 
 Notice that the OpenSearch charm is now at revision **144**.
 
+#### Rollback a charm revision with a different workload version
+
+If the charm revision you are rolling back to has a different workload version, the rollback process will attempt to roll back the workload version as well. However, since OpenSearch does not support downgrading, the rollback process will rollback the charm code then attempt a best-effort rollback of the OpenSearch workload.
```
izmalk: Suggestion: Wording improvements for clarity.

```suggestion
If you roll back to a charm revision with a different workload version, the process will roll back the charm code and then make a best-effort attempt to roll back the workload, since OpenSearch does not support downgrades.
```

docs/how-to/upgrade.md
```diff
@@ -461,10 +461,67 @@ Machine  State    Address         Inst id        Base          AZ  Message
 
 Notice that the OpenSearch charm is now at revision **144**.
 
+#### Rollback a charm revision with a different workload version
+
+If the charm revision you are rolling back to has a different workload version, the rollback process will attempt to roll back the workload version as well. However, since OpenSearch does not support downgrading, the rollback process will rollback the charm code then attempt a best-effort rollback of the OpenSearch workload. 
+
+This is a dangerous operation that may lead to an unhealthy OpenSearch cluster. It is recommended to instead perform a backup and restore of the cluster to a new deployment with the desired OpenSearch version.
+
+We do not recommend performing a rollback to a charm revision with a different OpenSearch version. However, if you choose to proceed, you can follow the same steps as rolling back to a charm revision with the same workload version. There are two possible outcomes after the rollback:
```
izmalk: Suggestion: This seems to be redundant. No new information here. And we can't use an introductory sentence with a colon at the end for headings. If you want it, you can put a list after it with the headings (either links or just text).

I'd suggest just deleting this to make everything shorter.


```suggestion
```

docs/how-to/upgrade.md
```diff
@@ -461,10 +461,67 @@ Machine  State    Address         Inst id        Base          AZ  Message
...
+
+##### It is possible to perform a best-effort rollback between the two workload revisions
+
+In this case both the charm code and the workload will be rolled back to the previous version. However, since this is rollback is a dangerous operation, manual intervention is required to rollback the workload version. The charm will be blocked and a message will be displayed indicating that you need to run the `force-refresh-start` action with the `check-compatibility=false` to continue with the best-effort rollback of the workload version.
```
izmalk:
```suggestion
In this case, both the charm code and the workload will be rolled back to the previous version. However, because rollback is a risky operation, rolling back the workload requires manual intervention. The charm will enter a blocked state and display a message instructing you to run the `force-refresh-start` action with `check-compatibility=false` to continue the best-effort workload rollback.
```

docs/how-to/upgrade.md
```diff
@@ -461,10 +461,67 @@ Machine  State    Address         Inst id        Base          AZ  Message
...
+self-signed-certificates/0*  active    idle   0        10.149.40.252
+
+Machine  State    Address        Inst id        Base          AZ   Message
+0        started  10.149.40.252  juju-f44a9a-0  ubuntu@24.04  xof  Running
+1        started  10.149.40.7    juju-f44a9a-1  ubuntu@24.04  xof  Running
+2        started  10.149.40.93   juju-f44a9a-2  ubuntu@24.04  xof  Running
+3        started  10.149.40.126  juju-f44a9a-3  ubuntu@24.04  xof  Running
+```
+
+##### It is not possible to perform a rollback between the two workload revisions
```
izmalk: Suggestion: Shorter heading and an empty line after a heading.

```suggestion
##### If the rollback between the versions is not possible

```

docs/how-to/upgrade.md
```diff
@@ -461,10 +461,67 @@ Machine  State    Address         Inst id        Base          AZ  Message
...
+
+Machine  State    Address        Inst id        Base          AZ   Message
+0        started  10.149.40.252  juju-f44a9a-0  ubuntu@24.04  xof  Running
+1        started  10.149.40.7    juju-f44a9a-1  ubuntu@24.04  xof  Running
+2        started  10.149.40.93   juju-f44a9a-2  ubuntu@24.04  xof  Running
+3        started  10.149.40.126  juju-f44a9a-3  ubuntu@24.04  xof  Running
+```
+
+##### It is not possible to perform a rollback between the two workload revisions 
+In this case, the charm code will be rolled back but the OpenSearch workload will remain in the newer version. The charm will be blocked and a message will be displayed indicating that you need to refresh back to a charm revision with the same workload version or perform a backup and restore to a new deployment.
```
izmalk:
```suggestion
In this case, the charm code will be rolled back, but the OpenSearch workload will remain on the newer version. The charm will enter a blocked state and display a message instructing you to either refresh to a charm revision with the same workload version or perform a backup and restore to a new deployment.
```

docs/how-to/upgrade.md
```diff
@@ -461,10 +461,67 @@ Machine  State    Address         Inst id        Base          AZ  Message
...
+8        started  10.149.40.64   juju-0bfd52-8  ubuntu@24.04  xof  Running
+9        started  10.149.40.31   juju-0bfd52-9  ubuntu@24.04  xof  Running
+```
+
 ### Check the cluster's health
 
-Once the charm is rolled back, it is important to check the cluster's health to ensure it is healthy.
-OpenSearch's upstream documentation
+Once the charm is rolled back, it is important to check the cluster’s health to ensure it is healthy.
+OpenSearch’s upstream documentation
```
izmalk: Question: Why do we do this? The rendered version is the same.

docs/how-to/upgrade.md
```diff
@@ -494,3 +551,228 @@ The response should look similar to the following example:
...
 ```
+
+(how-to-recover-rollback)=
+## Recovering from a rollback
+
+OpenSearch does not support rolling back to a previous version.
+If a unit has already been upgraded, performing `juju refresh`
+to a previous revision might result in OpenSearch failing to start on that unit.
+In this case, a manual recovery is required.
+Follow the steps in this section to restore the cluster to a healthy, operational state.
```
izmalk: Suggestion: Wording improvements for clarity. Please double-check the meaning is preserved.

```suggestion
OpenSearch does not support downgrades.
Running `juju refresh` to a previous revision may cause OpenSearch to fail to start.
In that case, manual recovery is required.
Follow the steps in this section to restore the cluster to a healthy state.
```

docs/how-to/upgrade.md
```diff
@@ -494,3 +551,228 @@ The response should look similar to the following example:
...
+
+For more information, please refer to the upstream
+[OpenSearch documentation about rolling upgrades](https://docs.opensearch.org/latest/migrate-or-upgrade/rolling-upgrade/#preparing-to-upgrade).
+
+### Check Juju status
+
+Check Juju model status:
```
izmalk:
```suggestion
First, check Juju model status:
```

docs/how-to/upgrade.md
```diff
@@ -494,3 +551,228 @@ The response should look similar to the following example:
...
+
+Note the blocked unit; in this example, it is `opensearch/2`.
+This unit will not recover automatically, and additional steps are required to replace it.
+
+### Check cluster health
+
+Retrieve the cluster health:
+
+```text
+GET _cluster/health?pretty
```
izmalk: Question: Should we replace it with a relevant `curl` command?

docs/how-to/upgrade.md
```diff
@@ -494,3 +551,228 @@ The response should look similar to the following example:
...
+      }
+    }
+  ]
+}
+```
+
+Delete the problematic index identified in the previous step:
+
+```{warning}
+If you do not have a snapshot containing this index, it will not be recoverable.
```
izmalk: Suggestion: Wording improvement just for clarity.

```suggestion
If you do not have a snapshot containing this index, the data will be lost!
```

izmalk (changes requested):
A few suggestions for improvement. Most notably, headings can be shorter for better readability. A couple of questions too.

docs/how-to/upgrade.md
skourta: We do not need it. This is a mishap from my part when copying and pasting from Grammarly.

docs/how-to/upgrade.md
skourta: We can. Do you prefer that? This is just to make it tool agnostic. You can even execute these on OS dashboards

izmalk: That is an idea I have; the decision is up to you. I, personally, would prefer to have a copy-pasteable curl command. Curl is multi-platform and popular enough to be universal.
We can provide both, if you want. Leave the existing code block and add a collapsed block ([see syntax of the summary HTML block](https://github.com/canonical/opensearch-operator/blob/2ce3ea4186d193b7bef0ba5baa06d900168b8d67/docs/how-to/back-up-and-restore/index.md?plain=1#L57-L68)) with the text "See curl example".

docs/how-to/upgrade.md
```diff
@@ -461,6 +461,63 @@ Machine  State    Address         Inst id        Base          AZ  Message
 
 Notice that the OpenSearch charm is now at revision **144**.
 
+#### Rollback a charm revision with a different workload version
+
+If you roll back to a charm revision with a different workload version, the process will roll back the charm code and then make a best-effort attempt to roll back the workload, since OpenSearch does not support downgrades.
+
+This is a dangerous operation that may lead to an unhealthy OpenSearch cluster. It is recommended to instead perform a backup and restore of the cluster to a new deployment with the desired OpenSearch version.
```
Mehdi-Bendriss: prefix by **Note:**

docs/how-to/upgrade.md
```diff
@@ -494,3 +551,227 @@ The response should look similar to the following example:
...
+{
+  "persistent": {
+    "cluster.routing.allocation.enable": "all"
+  }
+}
+```
+
+### Add a new unit
+
+Add a replacement unit to restore the original scale of the application:
```
Mehdi-Bendriss: Which reminds me that we should ask the user to scale up by 1 first at the beginning of the upgrade, right?

Mehdi-Bendriss (commented):
Thanks Smail! Looks good, i have a couple of comments

docs/how-to/upgrade.md
izmalk: Suggestion: I'd suggest using a proper admonition syntax. And since it's about dangerous operation, I think we should use a warning or a caution.

I also added some wording improvements to make the message clearer and avoid passive voice.

````suggestion
```{caution}
This operation is risky and may cause the OpenSearch cluster to become unhealthy. We recommend performing a backup and restore to a new deployment with the required OpenSearch version instead.
```
````

izmalk (commented):
Added a suggestion for a proper note block.

docs/how-to/upgrade.md
skourta: I will switch to curl

docs/how-to/upgrade.md
skourta: We do yes. [See](https://canonical-charmed-opensearch.readthedocs-hosted.com/2/how-to/upgrade/#upgrade-the-opensearch-cluster)

docs/how-to/upgrade.md
skourta: I belive we should avoid talking about rollbacks unless necessary.

izmalk (dismissed):
Fixed a few minor issues. LGTM.

docs/how-to/upgrade.md
Mehdi-Bendriss: Good point, we can say "manual recovery" instead of rollback in the `Optionally, ...` previous section, as the user needs to understand why it's so important

docs/how-to/upgrade.md
Mehdi-Bendriss: Okay, in that case - can we write `(Optional but highly advised)`

skourta: Added

docs/how-to/upgrade.md
skourta: Added

Mehdi-Bendriss (approved):

PR #750, opened 2025-12-04
chore: use 1/stable track for self-signed-certificates

gatici:

The Self-signed-certificates charm latest/stable track is deprecated and removed soon. We should use 1/stable track in Opensearch Operator. This PR is also fixing an issue in the integration tests which started to appear after we move to 1/stable track. The real fix should be done in self-signed-certificates but we avoid to hit the edge cases in the integration tests now.

- [ ] Manually
- [ ] Unit tests
- [ ] Integration tests

## Checklist


gatici:
This is superseded by https://github.com/canonical/opensearch-operator/pull/751.

PR #751, opened 2025-12-04
SSC 1/stable track support

skourta:
This pull request introduces several updates to improve reliability and maintainability across both the OpenSearch charm logic and its integration tests. The changes focus on ensuring correct cluster state handling, refining test application removal, and updating channel references for TLS certificates.

### OpenSearch charm logic:

* Added a check to ensure `self.opensearch.is_started()` before evaluating the number of cluster managers during storage detaching, which helps prevent cluster state loss when the service is not running.

### Integration tests:

* Updated the removal of invalid applications in `test_invalid_conditions` to use a `timeout` parameter instead of `force`, `destroy_storage`, and `no_wait`, streamlining the cleanup process and improving test reliability.

### TLS certificates configuration:

* Changed the stable channel for TLS certificates from `"latest/stable"` to `"1/stable"` and added a TODO to update documentation once the new channel is released, ensuring consistency with current deployment practices.

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -700,7 +700,8 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             # No cluster managers left in the cluster fleet
             # raise so we do not lose the cluster state
             if (
-                len(
+                self.opensearch.is_started()
+                and len(
                     [
                         app
                         for app in self.opensearch_peer_cm.apps_in_fleet()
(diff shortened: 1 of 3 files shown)


gatici (approved):

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -700,7 +700,8 @@ def _on_opensearch_data_storage_detaching(self, event: StorageDetachingEvent):
             # No cluster managers left in the cluster fleet
             # raise so we do not lose the cluster state
             if (
-                len(
+                self.opensearch.is_started()
```
smethnani: Do we need to check that opensearch is running successfully here? If so should we add `and not self.opensearch.is_failed()`?

skourta: This is to handle the case where OS is off. In one of the tests we remove a unit before we even start OS. The following lines make an HTTP request to OS which fails.

smethnani (approved):
Thanks @skourta !

PR #752, opened 2025-12-05
Extract microk8s to spread

skourta:
This pull request updates the setup and testing workflow for MicroK8s integration, focusing on streamlining configuration and removing redundant code. The main changes include adding support for a Docker Hub mirror, improving MicroK8s setup logic in the CI pipeline, and simplifying integration test fixtures by removing duplicate MicroK8s installation/configuration steps.

**CI/CD Pipeline Improvements**
* Added support for configuring a Docker Hub mirror via the `DOCKERHUB_MIRROR` environment variable in the `spread.yaml` backend configuration.
* Updated the `prepare` script in `spread.yaml` to conditionally install and configure MicroK8s with Docker Hub mirror settings when running OAuth-related tasks on IS-hosted runners. This includes waiting for MicroK8s to be fully initialized and ensuring all pods are running before proceeding.

**Integration Test Simplification**
* Removed the `microk8s_cloud` fixture from `tests/integration/relations/conftest.py`, which previously handled MicroK8s installation and configuration. The logic is now handled in the CI pipeline, reducing redundancy and simplifying test setup.
* Cleaned up unused imports in `tests/integration/relations/conftest.py` that were only required for the removed fixture.

**Documentation/Comment Updates**
* Clarified comments in `spread.yaml` regarding backend restoration logic, specifying that restoration is only needed for the LXD backend and not for CI.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -98,6 +98,7 @@ backends:
       GCP_ACCESS_KEY: '$(HOST: echo $GCP_ACCESS_KEY)'
       GCP_SECRET_KEY: '$(HOST: echo $GCP_SECRET_KEY)'
       GCP_SERVICE_ACCOUNT: '$(HOST: echo $GCP_SERVICE_ACCOUNT)'
+      DOCKERHUB_MIRROR: '$(HOST: echo $DOCKERHUB_MIRROR)'
       # To install pipx on self-hosted runners
       CONCIERGE_EXTRA_DEBS: pipx
     systems:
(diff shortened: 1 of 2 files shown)


gatici (approved):

spread.yaml
```diff
@@ -125,9 +126,48 @@ prepare: |
   cd "$SPREAD_PATH"
   snap install --classic concierge
 
+
   # Install charmcraft & pipx (on lxd-vm backend)
   concierge prepare --trace
 
+  if [[ -n "$DOCKERHUB_MIRROR" ]] && ([[ $SPREAD_TASK == *"oauth"* ]])
```
akram09: I believe it is better to have a separate `if (spread_task = *"oauth"` for cases where we are not running in self hosted runners. So we check if the job is oauth, then inside the check we verify if we have dockerhub_mirror and configure it.

spread.yaml
```diff
@@ -125,9 +126,48 @@ prepare: |
...
+    microk8s config > ~/.kube/config
+
+    
+    echo "Waiting for all pods to be in 'Running' phase..."
+    until kubectl get po -A --field-selector=status.phase!=Running 2>&1 | grep -q "No resources found"; do
+      echo "Some pods are not running. Retrying in 5 seconds..."
+      sleep 5
+    done
+    echo "Success: All pods are running!"
+    /snap/bin/juju add-k8s uk8s --client --controller $(/snap/bin/juju show-controller | yq "keys | .[0]")
```
akram09, line 175: Why `/snap/bin/juju` ?

akram09 (changes requested):

spread.yaml
skourta: Because /snap/bin is not in Path

spread.yaml
```diff
@@ -125,10 +126,55 @@ prepare: |
...
+      do
+        echo "MicroK8s has not yet configured iptables."
+        sleep 10
+      done
+      tee /var/snap/microk8s/current/args/certs.d/docker.io/hosts.toml << EOF
+      server = "$DOCKERHUB_MIRROR"
+      [host."${DOCKERHUB_MIRROR#'https://'}"]
+      capabilities = ["pull", "resolve"]
+  EOF
+    fi
```
gatici, line 151:
```suggestion
     tee /var/snap/microk8s/current/args/certs.d/docker.io/hosts.toml <<EOF
server = "$DOCKERHUB_MIRROR"
[host."${DOCKERHUB_MIRROR#'https://'}"]
capabilities = ["pull", "resolve"]
EOF
    fi
```
The lines in the EOF should not have any empty space in the beginning of line.

spread.yaml
```diff
@@ -125,10 +126,55 @@ prepare: |
   cd "$SPREAD_PATH"
   snap install --classic concierge
 
+  if [[ $SPREAD_TASK == *"oauth"* ]]
+  then
+    # Running on IS-hosted runner; configure microk8s to use Docker Hub mirror
+    # Run before concierge prepare because of https://github.com/canonical/concierge/issues/75
+    sudo snap install microk8s  --classic
+    sudo snap install kubectl --classic 
+    sudo snap install yq
```
gatici: We do not use `sudo` to install concierge or change ownership but use sudo to install microk8s.  Why ?
```
  chown -R root:root "$SPREAD_PATH"
  snap install --classic concierge
```

gatici (changes requested):

spread.yaml
skourta: we run using root there's no need for sudo

spread.yaml
skourta: Fixed. Thank you.

gatici (approved):

akram09 (approved):

PR #762, opened 2025-12-08
fix: only data node ignores lock if restarting

skourta:
If there is only one data node and is restarting it will be stuck waiting for the lock because after we stop opensearch we cannot get the lock from the index.

## Solution:

This PR introduces an `ignore_lock` parameter when emitting the `_start_opensearch_event`, allowing the lock to be bypassed if the current unit is the only data app and is the leader.

OpenSearch restart logic improvements:

* Added logic to determine if the current unit is the only data node and the leader, and set an `ignore_lock` flag accordingly when restarting OpenSearch. This flag is passed to `_start_opensearch_event.emit()` to allow the restart to proceed without waiting for a lock in single-node scenarios.

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -1489,7 +1489,23 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             self.status.set(WaitingStatus(ServiceIsStopping))
             return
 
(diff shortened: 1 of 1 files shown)


smethnani (approved):
Thank you @skourta

gatici (approved):

gatici:
Many thanks @skourta !

PR #765, opened 2025-12-08
Fixing Terraform Flaky Tests due to Limited storage on github runners

akram09:
This Pull Request adds a fix for flaky Terraform tests caused by limited storage availability on GitHub runners. The solution is to mount /dev/sdb and use it as the LXD storage pool. This approach was previously used in the [data-integrator](https://github.com/canonical/data-integrator/blob/9cd98d71a90e55696201003acce0cbc7b0c09de5/.github/workflows/integration_test_charm.yaml#L215C7-L220C90) charm to resolve a similar issue.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -70,6 +70,12 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: (Github hosted) Mount /mnt to have more storage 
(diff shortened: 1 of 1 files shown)


smethnani (approved):
Thank you @akram09!

gatici (approved):

PR #766, opened 2025-12-08
feat: support GCS repository for snapshot operations

gatici:

This PR allows Opensearch to use with Google Cloud storage to store snapshots or restore them.
With this change, operators can use a gcs-integrator relation to:
- create backups (snapshots) into a GCS bucket
- list existing backups
- restore backups from GCS
in the same way they already do with S3 and Azure.

## Solution:
GCS Integration
- Introduced a gcs relation endpoint (GCS_RELATION) and integrated it into the snapshots flow alongside the existing S3 and Azure backends.
- Extended OpenSearchSnapshotsManager to support ObjectStorageType.GCS:
  - build the repository settings for GCS based on the relation data.
  - call ensure_repository + verify_repository(ObjectStorageType.GCS) to validate connectivity and permissions.
  - reused the existing pre-check logic (leader checks, health, running operations) so GCS behaves consistently with S3/Azure.

- [x] Manually
- [x] Unit tests
- [x] Integration tests

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,4 +64,4 @@ override.tf.json
 .terraformrc
 terraform.rc
 .env
-internal_users.yml
+/tests/unit/resources/config/opensearch-security/internal_users.yml
--- a/.gitignore
+++ b/.gitignore
@@ -64,4 +64,3 @@ override.tf.json
 .terraformrc
 terraform.rc
 .env
-/tests/unit/resources/config/opensearch-security/internal_users.yml
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,4 @@ override.tf.json
 .terraformrc
 terraform.rc
 .env
+internal_users.yml
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -262,7 +262,6 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
 
 
 def write_gcs_service_account_json(
-    *,
     secret_key: str,
     dst: str = "/tmp/service_account.json",
     mode: int = 0o600,
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -263,8 +263,7 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
 
 def write_gcs_service_account_json(
     secret_key: str,
-    dst: str = "/tmp/service_account.json",
-    mode: int = 0o600,
+    dst: str = "/var/snap/opensearch/common/service_account.json",
 ) -> Path:
     """Write GCS service account JSON (from relation secret_key) to a file.
 
@@ -291,12 +290,11 @@ def write_gcs_service_account_json(
 
     path = Path(dst)
     path.write_text(content, encoding="utf-8")
-    path.chmod(mode)
     return path
 
 
 def remove_gcs_service_account_json(
-    dst: str = "/tmp/service_account.json",
+    dst: str = "/var/snap/opensearch/common/service_account.json",
 ) -> None:
     """Remove the GCS service account JSON file.
 
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -2,18 +2,15 @@
 # See LICENSE file for licensing details.
 
 """Utility functions for charms related operations."""
-import json
 import logging
 import os
 import re
 import subprocess
-from pathlib import Path
 from time import time_ns
 from types import SimpleNamespace
 from typing import TYPE_CHECKING, Iterable, List, Union
 
 from charms.opensearch.v0.constants_charm import (
-    GCS_SERVICE_ACCOUNT_JSON,
     PeerRelationName,
 )
 from charms.opensearch.v0.helper_enums import BaseStrEnum
@@ -262,54 +259,3 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
     add = desired_labels - current_labels
     remove = current_labels - desired_labels
     return add, remove
-
-
-def write_gcs_service_account_json(
-    secret_key: str,
-    dst: str = GCS_SERVICE_ACCOUNT_JSON,
-) -> Path:
-    """Write GCS service account JSON (from relation secret_key) to a file.
-
-    Args:
-        secret_key: JSON string content of the service account.
-        dst: Destination path.
-
-    Returns:
-        Path to the written file.
-
-    Raises:
-        ValueError: if secret_key is empty or not valid JSON.
-        OSError: if writing fails.
-    """
-    if not secret_key:
-        raise ValueError("Missing GCS secret_key (service account JSON).")
-
-    try:
-        obj = json.loads(secret_key)
-        content = json.dumps(obj)
-    except json.JSONDecodeError as e:
-        raise ValueError("GCS secret_key is not valid JSON.") from e
-
-    path = Path(dst)
-    path.write_text(content, encoding="utf-8")
-    return path
-
-
-def remove_gcs_service_account_json(
-    dst: str = "/var/snap/opensearch/common/home/snap_daemon/service_account.json",
-) -> None:
-    """Remove the GCS service account JSON file.
-
-    Args:
-        dst: Path to the service account file.
-
-    Raises:
-        FileNotFoundError: if missing_ok is False and file doesn't exist.
-        OSError: if deletion fails for other reasons.
-    """
-    path = Path(dst)
-    try:
-        path.unlink()
-    except FileNotFoundError:
-        logger.warning("GCS service account JSON file not found, skipping removal: %s", dst)
-        return
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -263,7 +263,7 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
 
 def write_gcs_service_account_json(
     secret_key: str,
-    dst: str = "/var/snap/opensearch/common/service_account.json",
+    dst: str = "/var/snap/opensearch/common/home/snap_daemon/service_account.json",
 ) -> Path:
     """Write GCS service account JSON (from relation secret_key) to a file.
 
@@ -294,7 +294,7 @@ def write_gcs_service_account_json(
 
 
 def remove_gcs_service_account_json(
-    dst: str = "/var/snap/opensearch/common/service_account.json",
+    dst: str = "/var/snap/opensearch/common/home/snap_daemon/service_account.json",
 ) -> None:
     """Remove the GCS service account JSON file.
 
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -2,10 +2,12 @@
 # See LICENSE file for licensing details.
 
 """Utility functions for charms related operations."""
+import json
 import logging
 import os
 import re
 import subprocess
+from pathlib import Path
 from time import time_ns
 from types import SimpleNamespace
 from typing import TYPE_CHECKING, Iterable, List, Union
@@ -257,3 +259,58 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
     add = desired_labels - current_labels
     remove = current_labels - desired_labels
     return add, remove
+
+
+def write_gcs_service_account_json(
+    *,
+    secret_key: str,
+    dst: str = "/tmp/service_account.json",
+    mode: int = 0o600,
+) -> Path:
+    """Write GCS service account JSON (from relation secret_key) to a file.
+
+    Args:
+        secret_key: JSON string content of the service account.
+        dst: Destination path.
+        mode: File permissions.
+
+    Returns:
+        Path to the written file.
+
+    Raises:
+        ValueError: if secret_key is empty or not valid JSON.
+        OSError: if writing fails.
+    """
+    if not secret_key:
+        raise ValueError("Missing GCS secret_key (service account JSON).")
+
+    try:
+        obj = json.loads(secret_key)
+        content = json.dumps(obj)
+    except json.JSONDecodeError as e:
+        raise ValueError("GCS secret_key is not valid JSON.") from e
+
+    path = Path(dst)
+    path.write_text(content, encoding="utf-8")
+    path.chmod(mode)
+    return path
+
+
+def remove_gcs_service_account_json(
+    dst: str = "/tmp/service_account.json",
+) -> None:
+    """Remove the GCS service account JSON file.
+
+    Args:
+        dst: Path to the service account file.
+
+    Raises:
+        FileNotFoundError: if missing_ok is False and file doesn't exist.
+        OSError: if deletion fails for other reasons.
+    """
+    path = Path(dst)
+    try:
+        path.unlink()
+    except FileNotFoundError:
+        logger.warning("GCS service account JSON file not found, skipping removal: %s", dst)
+        return
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -2,6 +2,7 @@
 # See LICENSE file for licensing details.
 
 """Utility functions for charms related operations."""
+import base64
 import json
 import logging
 import os
@@ -261,6 +262,31 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
     return add, remove
 
 
+def b64decode_to_text(content: str) -> str:
+    """If content looks like base64-encoded text, decode it; otherwise return as-is.
+
+    Args:
+        content: The string to decode.
+
+    Returns:
+        decoded_text: The decoded string.
+
+    Only returns decoded text if decoding works *and* the result is valid UTF-8.
+    """
+    s_stripped = content.strip()
+
+    # if it already looks like JSON, don't try base64.
+    if s_stripped.startswith("{") and s_stripped.endswith("}"):
+        return content
+
+    try:
+        decoded_bytes = base64.b64decode(s_stripped, validate=True)
+        decoded_text = decoded_bytes.decode("utf-8")
+        return decoded_text
+    except (ValueError, UnicodeDecodeError):
+        return content
+
+
 def write_gcs_service_account_json(
     secret_key: str,
     dst: str = "/var/snap/opensearch/common/home/snap_daemon/service_account.json",
@@ -270,7 +296,6 @@ def write_gcs_service_account_json(
     Args:
         secret_key: JSON string content of the service account.
         dst: Destination path.
-        mode: File permissions.
 
     Returns:
         Path to the written file.
@@ -279,11 +304,13 @@ def write_gcs_service_account_json(
         ValueError: if secret_key is empty or not valid JSON.
         OSError: if writing fails.
     """
-    if not secret_key:
+    if not secret_key or not secret_key.strip():
         raise ValueError("Missing GCS secret_key (service account JSON).")
 
+    candidate = b64decode_to_text(secret_key)
+
     try:
-        obj = json.loads(secret_key)
+        obj = json.loads(candidate)
         content = json.dumps(obj)
     except json.JSONDecodeError as e:
         raise ValueError("GCS secret_key is not valid JSON.") from e
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -12,7 +12,10 @@ from time import time_ns
 from types import SimpleNamespace
 from typing import TYPE_CHECKING, Iterable, List, Union
 
-from charms.opensearch.v0.constants_charm import PeerRelationName
+from charms.opensearch.v0.constants_charm import (
+    GCS_SERVICE_ACCOUNT_JSON,
+    PeerRelationName,
+)
 from charms.opensearch.v0.helper_enums import BaseStrEnum
 from charms.opensearch.v0.models import App, PeerClusterApp
 from charms.opensearch.v0.opensearch_exceptions import OpenSearchCmdError
@@ -263,7 +266,7 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
 
 def write_gcs_service_account_json(
     secret_key: str,
-    dst: str = "/var/snap/opensearch/common/home/snap_daemon/service_account.json",
+    dst: str = GCS_SERVICE_ACCOUNT_JSON,
 ) -> Path:
     """Write GCS service account JSON (from relation secret_key) to a file.
 
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -270,7 +270,6 @@ def write_gcs_service_account_json(
     Args:
         secret_key: JSON string content of the service account.
         dst: Destination path.
-        mode: File permissions.
 
     Returns:
         Path to the written file.
--- a/lib/charms/opensearch/v0/helper_charm.py
+++ b/lib/charms/opensearch/v0/helper_charm.py
@@ -2,7 +2,6 @@
 # See LICENSE file for licensing details.
 
 """Utility functions for charms related operations."""
-import base64
 import json
 import logging
 import os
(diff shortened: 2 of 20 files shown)


tests/unit/resources/config/opensearch-security/internal_users.yml
```diff
@@ -9,16 +9,16 @@ _meta:
 
 ## Demo users
 
+kibanaserver:
```
skourta: please remove this from the PR. It will keep appearing everytime we run the tests. Should we add it to the gitignore? @Mehdi-Bendriss

tests/unit/lib/test_snapshots.py
```diff
@@ -30,14 +30,19 @@
...
 
 class TestCreateBackup(SnapshotsUnitTestFixtures):
-    @pytest.mark.parametrize("backend", ["s3", "azure"])
+    @pytest.mark.parametrize("backend", ["s3", "azure", "gcs"])
     def test_create_backup_when_manager_raises_http_error_then_action_fails(self, backend):
         if backend == "s3":
             self.use_s3()
             rels = {self.s3_relation()}
-        else:
+        elif backend == "azure":
```
skourta: Can't we extract this portion into a fixture or a function? it is being reused all over the file.

tests/spread/test_backups.py-gcs-large/task.yaml
```diff
@@ -0,0 +1,9 @@
+summary: test_backups.py
+environment:
+  TEST_MODULE: ha/test_backups.py
+systems:
+  - self-hosted-linux-amd64-noble-xlarge
```
skourta: Do we need an xlarge runner? It's best to use a standard runner to avoid long waiting queues.

tests/integration/ha/test_backups.py
```diff
@@ -306,6 +325,30 @@ async def _configure_azure(
...
 
 
+async def _configure_gcs(
+    ops_test: OpsTest,
+    config: Dict[str, str],
+    credentials: Dict[str, str],
+) -> None:
+    """Configure gcs-integrator with bucket/path and service account JSON (via Juju secret)."""
+    logger.info("Adding Juju secret for GCS service account JSON")
+    local_label = "".join(random.choice(string.ascii_letters) for _ in range(10))
```
skourta, line 335: Why is the label random?

tests/integration/ha/test_backups.py
```diff
@@ -306,6 +325,30 @@ async def _configure_azure(
...
+        {"secret-key": credentials["secret-key"]},
+    )
+    logger.info(f"Juju secret for GCS credentials added. Secret URI: {credentials_secret_uri}")
+
+    full_cfg = deepcopy(config)
+    full_cfg.update({"credentials": credentials_secret_uri})
+
+    logger.info("Setting up configuration for gcs-integrator charm...")
+    await ops_test.model.applications[GCS_INTEGRATOR].set_config(full_cfg)
+    await ops_test.model.wait_for_idle(apps=[GCS_INTEGRATOR], timeout=TIMEOUT)
```
skourta, line 349: Use `wait_until`

tests/integration/ha/test_backups.py
```diff
@@ -911,6 +1015,32 @@ async def _drop_azure_relation_if_any(ops_test: OpsTest, app: str) -> None:
...
+        timeout=TIMEOUT,
+    )
+    logger.info(
+        "Dropped GCS relation %s -> %s.",
+        f"{app}:{GCS_RELATION}",
+        f"{GCS_INTEGRATOR}:{GCS_RELATION}",
+    )
+
+
 async def _ensure_only_s3_integrator_related(
```
skourta, line 1044: we need to add gcs top dropped here and in the azure function

terraform/charm/large_deployment/main.tf
```diff
@@ -12,7 +12,7 @@ locals {
   ]
   apps_not_in_failover_model = [
     for app in local.apps :
-    app if app.model_uuid != var.failover.model_uuid
+    app if var.failover == null || app.model_uuid != var.failover.model_uuid
```
skourta, line 15: Aren't these on another PR?

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -257,3 +259,54 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
     add = desired_labels - current_labels
     remove = current_labels - desired_labels
     return add, remove
+
+
+def write_gcs_service_account_json(
+    secret_key: str,
+    dst: str = "/var/snap/opensearch/common/home/snap_daemon/service_account.json",
```
skourta: This looks more like a charm constant no?

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -779,3 +782,69 @@ def verify_azure_credentials(cfg: ObjectStorageConfig) -> bool:
...
+    Args:
+        cfg: ObjectStorageConfig object
+
+    Returns:
+        True if we can access the configured bucket, False otherwise.
+
+    cfg.gcs.credentials.secret_key to contain a service-account JSON
+    (as a string), and cfg.gcs.bucket to contain the bucket name.
+    """
+    gcs_cfg = cfg.gcs
```
skourta: I think you should pass directly the GCS config only

lib/charms/opensearch/v0/models.py
```diff
@@ -494,13 +496,39 @@ def from_relation(cls, input_dict: Optional[Dict[str, Any]]):
...
 
         allow_population_by_field_name = True
 
+    @validator("secret_key", pre=True)
+    def _normalize_secret_key(cls, v):  # noqa: N805
+        """Accept either raw JSON or base64-encoded JSON"""
+        if v is None:
+            return None
+
+        content = v.decode() if isinstance(v, (bytes, bytearray)) else str(v)
```
skourta: when can this happen?

lib/charms/opensearch/v0/models.py
```diff
@@ -511,7 +539,56 @@ class GcsRelData(Model):
     bucket: str = Field(default="")
     base_path: Optional[str] = Field(alias="path", default=None)
     storage_class: Optional[str] = Field(alias="storage-class", default=None)
-    credentials: GcsRelDataCredentials = Field(alias=GCS_CREDENTIALS, default=None)
+    credentials: GcsRelDataCredentials = Field(
+        alias=GCS_CREDENTIALS, default_factory=GcsRelDataCredentials
+    )
+
+    class Config:
```
skourta, line 545: I do not understand this class declaration

lib/charms/opensearch/v0/opensearch_peer_clusters.py
```diff
@@ -820,6 +826,11 @@ def rel_data_from_str(self, redacted_dict_str: str) -> PeerClusterRelData:
                 credentials["azure"]["secret-key"], content_key="azure-secret-key"
             )
 
+        if credentials.get("gcs") and credentials["gcs"].get("secret-key"):
```
skourta: You can do the same check as line 744 for consistency

lib/charms/opensearch/v0/opensearch_snapshots.py
```diff
@@ -259,17 +282,24 @@ def _on_backup_credentials_changed(  # noqa C901
                     "azure.client.default.key": object_storage_config.azure.credentials.secret_key,
                 }
             )
+        else:
```
skourta: Might be better to use an elif and raise on alternative

lib/charms/opensearch/v0/opensearch_snapshots.py
```diff
@@ -259,17 +282,24 @@ def _on_backup_credentials_changed(  # noqa C901
...
+            )
 
         self.charm.opensearch_keystore_events.reload_event.emit()
 
         if not self.charm.unit.is_leader():
             return
 
         try:
-            self.charm.snapshots_manager.ensure_repository(
+            if self.charm.snapshots_manager.ensure_repository(
```
skourta, line 299: Shouldn't we also error if `ensure_repository` returns false?

lib/charms/opensearch/v0/opensearch_snapshots.py
```diff
@@ -298,17 +328,18 @@ def _on_backup_credentials_changed(  # noqa C901
...
     ) -> None:
         """Handler for s3 credentials gone event."""
-        object_storage_type = (
-            ObjectStorageType.S3
-            if isinstance(event, CredentialsGoneEvent)
-            else ObjectStorageType.AZURE
-        )
-
-        if not object_storage_type:
+        if isinstance(event, CredentialsGoneEvent):
```
skourta, line 335: Can you please explain this condition?

lib/charms/opensearch/v0/opensearch_snapshots.py
```diff
@@ -365,42 +398,71 @@ def _on_peer_clusters_relation_changed_for_snapshots(self, event) -> None:  # no
...
+        if s3_info:
+            info_to_save = s3_info
+        elif azure_info:
+            info_to_save = azure_info
+        elif gcs_info:
+            info_to_save = gcs_info
+        else:
+            info_to_save = None
+
+        if s3_info:
```
skourta: You can fuse these two blocks of code together to only have the conditions once

skourta (commented):
Thank you @gatici! I left some comments

lib/charms/opensearch/v0/helper_charm.py
akram09: nit; may seem extra but would name the file gcs_service_account.json to not have future conflicts

lib/charms/opensearch/v0/helper_security.py
akram09: Agree

lib/charms/opensearch/v0/models.py
```diff
@@ -511,7 +539,56 @@ class GcsRelData(Model):
     bucket: str = Field(default="")
     base_path: Optional[str] = Field(alias="path", default=None)
     storage_class: Optional[str] = Field(alias="storage-class", default=None)
```
akram09, line 540: Question; Why do we need storage_class ?

akram09 (commented):
Thanks @gatici !

tests/unit/resources/config/opensearch-security/internal_users.yml
gatici: Done, I added the file to .gitignore.

tests/unit/lib/test_snapshots.py
gatici: Done, a fixture is used.

tests/spread/test_backups.py-gcs-large/task.yaml
gatici: I changed it to large runner, if it is fine than we can keep it.

tests/integration/ha/test_backups.py
gatici: This is the label for juju secret and it just providing us a guarantee that every run creates a fresh secret. This is not a big deal for runners but can be important for local runs.

tests/integration/ha/test_backups.py
gatici: The `wait_until` expects an app status and if we do not provide, it is counted as `active`. Here we do not want to check the app status as this is a generic helper. We only want to check if juju finishes all work and model reaches a stable state. The status checks are done in the test cases.

tests/integration/ha/test_backups.py
gatici: Good catch, done

terraform/charm/large_deployment/main.tf
gatici: When I rebase it, this part will be removed. This is added to pass the integration tests.

lib/charms/opensearch/v0/helper_charm.py
gatici: It is added as constant.

lib/charms/opensearch/v0/helper_security.py
gatici: Done.

lib/charms/opensearch/v0/models.py
gatici: This is the content that we got from juju secret and juju secret may store the secret content as str or bytes or bytearray. So, we should expect any format which can be retrived as a secret.

lib/charms/opensearch/v0/models.py
gatici: This is not a real class but a convention for Pydantic v1 to make a meta config.  It configures allow_population_by_field_name = True to outer class which is meaning that attach those configuration values to GcsRelData model.

lib/charms/opensearch/v0/opensearch_peer_clusters.py
gatici: Done

lib/charms/opensearch/v0/opensearch_snapshots.py
gatici: Done.

lib/charms/opensearch/v0/helper_charm.py
Mehdi-Bendriss: I don't think we need this and next function as a broad helpers, as we're ultimately just writing/deleting a text file (which is also not k8s compatible), feel free to use `opensearch.write_file` as those will ultimately be moved to the Workload class

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -257,3 +259,54 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
...
+        content = json.dumps(obj)
+    except json.JSONDecodeError as e:
+        raise ValueError("GCS secret_key is not valid JSON.") from e
+
+    path = Path(dst)
+    path.write_text(content, encoding="utf-8")
+    return path
+
+
+def remove_gcs_service_account_json(
```
Mehdi-Bendriss: similar comment as above

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -779,3 +782,69 @@ def verify_azure_credentials(cfg: ObjectStorageConfig) -> bool:
...
+        logger.error(
+            "GCS credential validation failed: secret_key is not valid JSON: %s",
+            e,
+        )
+        return False
+
+    project_id = sa_info.get("project_id")
+
+    client_kwargs: dict = {}
+    if project_id:
```
Mehdi-Bendriss:
```suggestion
    client_kwargs: dict = {}
    if project_id := sa_info.get("project_id"):
```

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -779,3 +782,69 @@ def verify_azure_credentials(cfg: ObjectStorageConfig) -> bool:
             exc_info=e,
         )
         return False
+
+
+def verify_gcs_credentials(cfg: ObjectStorageConfig) -> bool:  # noqa: C901
```
Mehdi-Bendriss:
```suggestion
def verify_gcs_credentials(object_storage_config: ObjectStorageConfig) -> bool:  # noqa: C901
```

Could we use plain names for variables? it's often difficult to figure out what an abbreviation means

lib/charms/opensearch/v0/helper_security.py
```diff
@@ -779,3 +782,69 @@ def verify_azure_credentials(cfg: ObjectStorageConfig) -> bool:
...
+
+    if not service_account_json:
+        logger.error("GCS credential validation failed: secret_key is empty.")
+        return False
+    if not bucket_name:
+        logger.error("GCS credential validation failed: bucket name is missing.")
+        return False
+
+    try:
+        sa_info = json.loads(service_account_json)
```
Mehdi-Bendriss:
```suggestion
        service_account = json.loads(service_account_json)
```

lib/charms/opensearch/v0/models.py
```diff
@@ -494,13 +496,39 @@ def from_relation(cls, input_dict: Optional[Dict[str, Any]]):
...
-    secret_key: str = Field(alias="secret-key", default=None)
+    secret_key: Optional[str] = Field(alias="secret-key", default=None)
 
     class Config:
         """Model config of this pydantic model."""
 
         allow_population_by_field_name = True
 
+    @validator("secret_key", pre=True)
+    def _normalize_secret_key(cls, v):  # noqa: N805
```
```suggestion
    def _normalize_secret_key(cls, values):  # noqa: N805
```

lib/charms/opensearch/v0/models.py
```diff
@@ -494,13 +496,39 @@ def from_relation(cls, input_dict: Optional[Dict[str, Any]]):
...
+        """Accept either raw JSON or base64-encoded JSON"""
+        if v is None:
+            return None
+
+        content = v.decode() if isinstance(v, (bytes, bytearray)) else str(v)
+        content = content.strip()
+        if not content:
```
```suggestion
        if not (content := content.strip()):
```

lib/charms/opensearch/v0/opensearch_snapshots.py
gatici: It already does. The `ensure_repository` method does a call using `opensearch.request` and if anything fails it raises OpensearchHttpError and it is already catched and  status is set to Blocked in the following lines.

lib/charms/opensearch/v0/opensearch_snapshots.py
gatici: On the S3 integrator emits the CredentialsGoneEvent when a backup relation is broken. So, it is checking the event using `isinstance` method to detect the storage_type.

lib/charms/opensearch/v0/opensearch_snapshots.py
gatici: Good catch. Done.

lib/charms/opensearch/v0/helper_charm.py
gatici: Done.

lib/charms/opensearch/v0/models.py
gatici: Storage classes is an optional specification  that user can provide and it helps to optimize cost and access patterns to buckets.
It has some options like hot data, warm data, cold data, very cold data.

Mehdi-Bendriss: @akram09 Some object stores like AWS S3 provide "storage classes" so users can optimize costs vs reliability+speed for the stored data (e.g frequently accessed data on S3 standard or express and archived on S3 glacier)

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -604,6 +611,40 @@ def _put_fleet_apps(
...
+            secret_key = gcs.credentials.secret_key
+
+            # set the secrets in the charm
+            self.charm.secrets.put(Scope.APP, "gcs-secret-key", secret_key)
+
+            return GcsRelDataCredentials(secret_key=secret_key)
+
+        # Non-main orchestrators: only return creds if we already have them
+        secret_key = self.charm.secrets.get(Scope.APP, "gcs-secret-key")
+        if not secret_key:
```
Mehdi-Bendriss: nit, walrus

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
```diff
@@ -970,6 +1017,13 @@ def _grant_rel_data_secrets(  # noqa: C901
                             self.secrets.grant_secret_to_relation(
                                 secret_id["secret-key"], relation
                             )
+
```
Mehdi-Bendriss, line 1019:
```suggestion
```

lib/charms/opensearch/v0/models.py
```diff
@@ -494,13 +496,39 @@ def from_relation(cls, input_dict: Optional[Dict[str, Any]]):
...
+
+        # already JSON
+        if content.startswith("{") and content.endswith("}"):
+            # validate JSON shape
+            json.loads(content)
+            return content
+
+        # base64 (urlsafe)
+        try:
+            decoded_bytes = base64.b64decode(content, altchars=b"-_", validate=True)
```
Mehdi-Bendriss, line 524: could you explain when we'd expect these 2 different cases?

lib/charms/opensearch/v0/helper_charm.py
```diff
@@ -257,3 +259,54 @@ def diff(desired: Iterable[str], current: Iterable[str]) -> tuple[set[str], set[
     add = desired_labels - current_labels
     remove = current_labels - desired_labels
     return add, remove
+
+
+def write_gcs_service_account_json(
+    secret_key: str,
+    dst: str = "/var/snap/opensearch/common/home/snap_daemon/service_account.json",
+) -> Path:
```
Mehdi-Bendriss: I don't think we need to have a helper for this one, given that this helper is VM only and only writes data to a text file. 
This will be moved to the Workload class eventually, but for now I think you can use `opensearch.write_file` instead

lib/charms/opensearch/v0/opensearch_snapshots.py
```diff
@@ -259,17 +282,26 @@ def _on_backup_credentials_changed(  # noqa C901
                     "azure.client.default.key": object_storage_config.azure.credentials.secret_key,
                 }
             )
+        elif object_storage_type == ObjectStorageType.GCS:
+            sa_path = write_gcs_service_account_json(
```
Mehdi-Bendriss:
```suggestion
            service_account_path = write_gcs_service_account_json(
```

Mehdi-Bendriss (commented):
Thanks Gulsum, great work! We're already seeing the result of the refactor :)

lib/charms/opensearch/v0/helper_security.py
Mehdi-Bendriss: + could we rename sa_info with `service_account`

lib/charms/opensearch/v0/helper_charm.py
gatici: Done

lib/charms/opensearch/v0/helper_security.py
gatici: Done

lib/charms/opensearch/v0/helper_security.py
gatici: Done

lib/charms/opensearch/v0/helper_security.py
gatici: Done

lib/charms/opensearch/v0/models.py
gatici: Done

lib/charms/opensearch/v0/models.py
gatici: Done

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
gatici: Done

lib/charms/opensearch/v0/opensearch_relation_peer_cluster.py
gatici: Done

lib/charms/opensearch/v0/helper_charm.py
gatici: Done

lib/charms/opensearch/v0/opensearch_snapshots.py
gatici: Done

lib/charms/opensearch/v0/models.py
gatici: We are using both options currently and both of them valid.
Operators may provide GCS service account as a plaintext json or it can be in a base64 encoded format. In Github CI, is appears as base64 encoded data and in our manual test we are using in the json format. Both approaches are valid and supported.

.gitignore
```diff
@@ -64,3 +64,4 @@ override.tf.json
 .terraformrc
 terraform.rc
 .env
+/tests/unit/resources/config/opensearch-security/internal_users.yml
```
Mehdi-Bendriss: why is this added to the .gitignore? don't we want to keep it?

Mehdi-Bendriss (dismissed):
Thanks Gulsum! I do have one question

skourta (dismissed):
Thank you for the hard work @gatici!

.gitignore
skourta: This keeps changing when we run the unit tests.

gatici: No, we want to keep it. I will revert this line.

skourta (approved):

.gitignore
gatici: The change that we did here does not solve the issue @skourta. That's why I reverted.

Mehdi-Bendriss (approved):

PR #772, opened 2025-12-10
fix: remove verify repository from requirer error checks

gatici:

OpensearchPeerClusterRequirer performs some checks regarding peer-cluster-relation. There was a check for repository verification but this is not useful. This was aiming not publish creds to subclusters if main orhestrator do not finish checks and preparations. However, `self.refresh_relation_data` method is called in several places and triggers peer_relation_changed event  in the subclusters. By this way, they may get the creds any time. This change does not prevent it, so removing.

## Solution:

We are removing the verify_repository check from the method `_error_set_from_requirer` which is not useful.

- [x] Manually
- [ ] Unit tests
- [x] Integration tests

## Checklist


smethnani (approved):
Thank you @gatici!

skourta (approved):
Thanks @gatici

PR #773, opened 2025-12-10
chore: remove restart after CA add/remove

gatici:

Were are restarting Opensearch service if a new CA added/changed/removed. In the upstream guides, there is no any hot-reload mechanism. But in reality, hot-reload helps the CA updates takes effect. We tested that although we remove all the service restarts after CA changes and run a  hot-reload, it all works well. 

## Solution:

This PR removes all the Opensearch service restarts and just runs `keystore_manager.reload()`.

- [x] Manually
- [ ] Unit tests
- [x] Integration tests

## Checklist

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_base_charm.py
+++ b/lib/charms/opensearch/v0/opensearch_base_charm.py
@@ -2049,20 +2049,6 @@ class OpenSearchBaseCharm(CharmBase, abc.ABC):
             )
         )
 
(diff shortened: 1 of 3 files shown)


smethnani:
I believe you still need to reload secure settings after adding the CA

gatici:
This is superseded by https://github.com/canonical/opensearch-operator/pull/774.

PR #774, opened 2025-12-10
Refactor code and verify repository after all units save creds

### LINKED ISSUE #771: Verification of repository if OS elected cluster manager does not backup credentials
Repository verification may fail despite credentials on the main orchestartor being correct. As seen [here](https://github.com/canonical/opensearch-operator/actions/runs/20072030625/job/57619932590)

The credentials must be stored on the elected cluster manager node before this call is successful. The elected cluster manager in OpenSearch may be a different node from the leader unit of the main orchestrator.
## Steps to reproduce

1. Deploy large deployments
2. trigger change of elected CM in OS to have an different elected CM than the main orchestrator

## Expected behavior
Verification of repository succeeds

## Actual behavior

Verification of respository fails

Image

skourta:

Fixes #771 
If the leader of the main orchestrator is different from the elected CM in OS the verification of the repository will fail. We need to propagate the credentials to all units of OS to make sure the elected CM has the credentials before we actually verify that the credentials are valid.
## Solution:

We propagate credentials to all units. Whenver a unit saves the credentials it addes its hash into the unit data bag. Once all units have credentials, the leader of the main orchestrator will verify the credentials.

This PR also includes refactoring code to use a single event handler for both s3/azure backends.
- [x] Manually
- [ ] Unit tests
- [x] Integration tests

### DIFF
--- a/lib/charms/opensearch/v0/opensearch_snapshots.py
+++ b/lib/charms/opensearch/v0/opensearch_snapshots.py
@@ -7,7 +7,7 @@ import json
 import logging
 from datetime import datetime
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Optional, Tuple, Union
 
 from charms.data_platform_libs.v0.azure_storage import (
     AzureStorageRequires,
@@ -124,17 +124,17 @@ class OpenSearchSnapshotEvents(Object):
         self.azure_requirer = AzureStorageRequires(charm, AZURE_RELATION)
 
         # simple deployments or main orchestrator
-        self.framework.observe(
-            self.s3_requirer.on.credentials_changed, self._on_s3_credentials_changed
-        )
-        self.framework.observe(self.s3_requirer.on.credentials_gone, self._on_s3_credentials_gone)
-        self.framework.observe(
+        for event in [
+            self.s3_requirer.on.credentials_changed,
             self.azure_requirer.on.storage_connection_info_changed,
-            self._on_azure_credentials_changed,
-        )
-        self.framework.observe(
-            self.azure_requirer.on.storage_connection_info_gone, self._on_azure_credentials_gone
-        )
+        ]:
+            self.framework.observe(event, self._on_backup_credentials_changed)
+
+        for event in [
+            self.s3_requirer.on.credentials_gone,
+            self.azure_requirer.on.storage_connection_info_gone,
+        ]:
+            self.framework.observe(event, self._on_backup_credentials_gone)
 
         # large deployments with non-main orchestrator
         self.framework.observe(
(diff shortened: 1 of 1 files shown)


lib/charms/opensearch/v0/opensearch_snapshots.py
```diff
@@ -109,9 +111,15 @@ class ObjectStorageType(str, Enum):
     CONFLICT = "conflict"
 
 
+class VerifyBackupCredentialsEvent(EventBase):
+    """Event for cleaning up old CAs."""
```
smethnani: Nit: is this doc string accurate?

lib/charms/opensearch/v0/opensearch_snapshots.py
```diff
@@ -463,6 +338,12 @@ def _on_azure_credentials_gone(self, event: StorageConnectionInfoGoneEvent) -> N
             event.defer()
             return
 
+        if (
+            object_storage_type == ObjectStorageType.S3
+            and self.charm.snapshots_manager.is_custom_s3_ca_stored()
+        ):
+            self.charm.snapshots_manager.remove_s3_ca()
```
gatici, line 382: We are missing a  keystore reload just after this.

skourta: I also found a couple of other places and added them

gatici:
Great work @skourta ! I manually tested and it worked well with the demotion promotion scenario.

lib/charms/opensearch/v0/opensearch_snapshots.py
```diff
@@ -278,183 +312,58 @@ def _on_s3_credentials_changed(self, event: CredentialsChangedEvent) -> None:  #
...
-                    app=True,
-                )
+            self.charm.status.set(
+                BlockedStatus(
+                    BackupMisconfiguration.format(
+                        object_storage_type.value, f"{object_storage_type.value} integrator"
+                    )
+                ),
+                app=True,
+            )
```
gatici, line 322: In line 326, we are clearing status only for Azure. This should be generic to work for any storage type.

```
        self.charm.status.clear(
            BackupMisconfiguration.format("azure", "azure integrator"), app=True
        )
   ```

gatici (dismissed):
1 minor comment is provided. Looks good.

smethnani (dismissed):
Thank you @skourta, great work! Tested manually for simple and large deployments (2 main units, 1 failover with `cluster_manager` role, 2 units with `data` role) and was able to create and restore a snapshot, conflicts were handled correctly and the missing relations status on failover promotion was cleared when a backup relation was established with it.

lib/charms/opensearch/v0/opensearch_snapshots.py
skourta: Fixed

smethnani (approved):

akram09 (approved):

PR #776, opened 2025-12-11
docs: add Dashboards documentation link to the Nav Menu

izmalk:

We are thinking on a way to publish the OpenSearch Dashboards documentation with the OpenSearch documentation together, on the same web portal.

## Solution:

Add a link to OpenSearch Dashboards documentation to the main Navigation Menu for discoverability.

I've tried a few options and none of them worked. I'll continue to look for a way to implement this, but meanwhile, I want to introduce this link as a temporary workaround. 

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

Local build works and looks fine.

## Checklist


docs/index.md
```diff
@@ -69,4 +69,5 @@ tutorial/index
 how-to/index
 reference/index
 explanation/index
+Dashboards charm docs<https://canonical-charmed-opensearch-dashboards.readthedocs-hosted.com/>
```
skourta, line 72: nit: We should use the full name OpenSearch Dashboards

izmalk: The full item would be:
"OpenSearch Dashboards charm documentation".
Which is a bit bigger, than fits in a single line.
Given the context of it being Charmed OpenSearch docs, I thought we can drop OpenSearch in the Nav Menu. Just like we do with most Nav Menu items (Deploy instead of Deploy OpenSearch, etc.).
Screenshot from 2026-01-06 11-25-41.
Do you think we can settle with that?

Mehdi-Bendriss (approved):

gatici (approved):

Issue #768, opened 2025-12-09
Pin terraform product of opensearch-dashboards to a commit or release


akram09:
The opensearch terraform product is[ sourcing](https://github.com/canonical/opensearch-operator/blob/2/edge/terraform/product/simple_deployment/main.tf#L31) `2/edge` branch of opensearch-dashboard which can leads to non reproducable and unreliable production. The recommendation is to use a commit tag or a release on stable channels.

PR #778, opened 2025-12-15
Pin Terraform Version 

akram09:
This Pull Request pins the version of opensearch dashboard terraform module that is used in opensearch terraform products. This is a fix to #768


skourta (approved):
Thank you @akram09

smethnani (approved):

PR #781, opened 2026-01-05
fix: handle the situation that opensearch_failover does not exist

gatici:

This PR fixes Terraform evaluation errors in the large deployment configuration when the failover orchestrator is optional.

When failover is not provided, Terraform was attempting to dereference var.failover.model_uuid and access outputs from modules created with for_each, causing failures during terraform module validation.

## Solution

- Guarded access to var.failover.model_uuid in locals [1](https://github.com/canonical/opensearch-operator/pull/781/changes#diff-a05ae41a3894f6cd0e8a3198516ce7ca6fca5616d285b6c7a8ec2b3f0fcb57efR15)
- Fixed output handling by using try(..., null) to avoid failures when failover is not deployed [2](https://github.com/canonical/opensearch-operator/pull/781/changes#diff-d9a43fd5511c20c497a0b452603dcb9adcf6c11506eb026792f9722ffea81ea3R27) [3](https://github.com/canonical/opensearch-operator/pull/781/changes#diff-d9a43fd5511c20c497a0b452603dcb9adcf6c11506eb026792f9722ffea81ea3R29)

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

## Checklist

### DIFF
--- a/terraform/charm/large_deployment/main.tf
+++ b/terraform/charm/large_deployment/main.tf
@@ -12,7 +12,7 @@ locals {
   ]
   apps_not_in_failover_model = [
     for app in local.apps :
-    app if app.model_uuid != var.failover.model_uuid
+    app if var.failover == null || app.model_uuid != var.failover.model_uuid
   ]
 
   all_models = distinct(concat(
--- a/terraform/charm/large_deployment/outputs.tf
+++ b/terraform/charm/large_deployment/outputs.tf
@@ -24,9 +24,9 @@ output "app_names" {
   description = "Output of all deployed application names."
   value = {
     opensearch_main     = module.opensearch_main.app_names["opensearch"]
-    opensearch_failover = module.opensearch_failover.app_names["opensearch"]
+    opensearch_failover = try(module.opensearch_failover["deployed"].app_names["opensearch"], null)
     opensearch_apps = [
-      for app_module in module.opensearch_non_orchestrator_apps :
+      for app_module in values(module.opensearch_non_orchestrator_apps) :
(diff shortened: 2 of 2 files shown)


akram09 (dismissed):
Thanks @gatici

akram09:
Small comment. 
I was thinking that we should fix the terraform provider version so that it don't take other versions and make the pipeline fails again

akram09 (dismissed):

terraform/charm/large_deployment/versions.tf
```diff
@@ -7,7 +7,7 @@ terraform {
   required_providers {
     juju = {
       source  = "juju/juju"
-      version = "~> 1.0"
+      version = "~= 1.1.0"
```
Mehdi-Bendriss: I think we should rather revert this change, reason being that this module will most likely be part of a bigger solution module where we don't have control over the exact version used there (assume COS uses 1.2.0, and opensearch 1.1.0 - it will fail) - we should perhaps revert the approach and assume that if something breaks because of a new juju TF provider we get notified first

(cc @akram09)

Mehdi-Bendriss (commented):
Thanks Gulsum! just a minor comment on the TF version

terraform/charm/large_deployment/versions.tf
gatici: Thank you Mehdi. The last commit is reverted.

akram09 (approved):

Mehdi-Bendriss (approved):

PR #789, opened 2026-01-21
[DPE-9144] add smtp support

### LINKED ISSUE #777: Add back SMTP support
Refer to this [commit](https://github.com/canonical/opensearch-operator/pull/775/commits/5e3c3002e9755f18a65510b1bdf6a1a5539cfcdd) to add back SMTP support

gatici:
Implements [DA231](https://docs.google.com/document/d/1LaMAF5_vuRhcmqam2LbUnOeq-Gy2Le7fKc3lXHJXGmY/edit?tab=t.0)

Fixes https://github.com/canonical/opensearch-operator/issues/777.

## Solution:
This PR adds SMTP notifications lifecycle in the OpenSearch charm.

The SMTP-related notification configs are created automatically by Opensearch when the smtp-integrator relation is established.
- SMTP sender (smtp_account)
- Email group (email_group)
- Email channel (email)

## How it works
 Integrate Opensearch with smtp-integrator
 Configure the SMTP integrator with SMTP credentials:

```shell
juju config smtp-integrator \
  host="<smtp-host>" \
  port="587" \
  user="<smtp-username>" \
  password="<smtp-password>" \
  smtp-sender="<sender-email>" \
  recipients="a@example.com,b@example.com"
```

The smtp-sender, email-group and email-channel is automatically created. The config-id is derived from relation-id.
```
 sudo curl --cacert /var/snap/opensearch/98/etc/opensearch/certificates/chain.pem   -u "admin:${ADMIN_PASSWORD}"   "https://10.127.127.111:9200/_plugins/_notifications/configs/" | jq .
{
  "start_index": 0,
  "total_hits": 3,
  "total_hit_relation": "eq",
  "config_list": [
    {
      "config_id": "smtp-85",
      "last_updated_time_ms": 1769774765184,
      "created_time_ms": 1769774765184,
      "config": {
        "name": "smtp-85",
        "description": "SMTP sender: (smtp-85)",
        "config_type": "smtp_account",
        "is_enabled": true,
        "smtp_account": {
          "host": "smtp.gmail.com",
          "port": 587,
          "method": "start_tls",
          "from_address": "h68805701@gmail.com"
        }
      }
    },
    {
      "config_id": "smtp-85_recipients",
      "last_updated_time_ms": 1769774767882,
      "created_time_ms": 1769774767882,
      "config": {
        "name": "smtp-85_recipients",
        "description": "Email group managed by (smtp-85_recipients)",
        "config_type": "email_group",
        "is_enabled": true,
        "email_group": {
          "recipient_list": [
            {
              "recipient": "gulsumatici@gmail.com"
            }
          ]
        }
      }
    },
    {
      "config_id": "smtp-85_email-channel",
      "last_updated_time_ms": 1769774768149,
      "created_time_ms": 1769774768149,
      "config": {
        "name": "smtp-85_email-channel",
        "description": "Email channel: (smtp-85_email-channel)",
        "config_type": "email",
        "is_enabled": true,
        "email": {
          "email_account_id": "smtp-85",
          "recipient_list": [],
          "email_group_id_list": [
            "smtp-85_recipients"
          ]
        }
      }
    }
  ]
}

```

image

```
ADMIN_PASSWORD="xxxxx"
OS_IP="10.127.127.128"
CFG_ID="smtp-85_email-channel"

sudo curl --cacert /var/snap/opensearch/current/etc/opensearch/certificates/chain.pem \
  -u "admin:${ADMIN_PASSWORD}" \
  -X POST "https://${OS_IP}:9200/_plugins/_notifications/feature/test/${CFG_ID}" \
  -H "Content-Type: application/json" \
  -d '{"subject":"SMTP integration test","message":"Sent via smtp-integrator + OpenSearch"}'

{"event_source":{"title":"Test Message Title-smtp-85_email-channel","reference_id":"smtp-85_email-channel","severity":"info","tags":[]},"status_list":[{"config_id":"smtp-85","config_type":"email","config_name":"smtp-85_email-channel","email_recipient_status":[{"recipient":"gulsum.atici@canonical.com","delivery_status":{"status_code":"200","status_text":"Success"}}],"delivery_status":{"status_code":"200","status_text":"Success"}}]}
```

image

- [x] Manually
- [x] Unit tests
- [x] Integration tests

## Checklist


lib/charms/opensearch/v0/constants_charm.py
```diff
@@ -77,6 +77,7 @@
 BackupMisconfiguration = "opensearch {} repository setup failed. Check the {} config."
 BackupRelUneligible = "Only orchestrator clusters should relate to backup relation."
 SecretAccessError = "Failed to access secret, please check permissions."
+SmtpRelationInvalid = "SMTP relation must be created with the main-orchestrator cluster."
```
Mehdi-Bendriss:
```suggestion
SmtpRelationInvalid = "SMTP relation must be established with the main-orchestrator cluster."
```

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -230,6 +234,9 @@ def __init__(self, *args, distro: Type[OpenSearchDistribution] = None):
 
         self.plugin_manager = OpenSearchPluginManager(self.state)
         self.plugin_events = OpenSearchPluginEvents(self)
+        self.notifications = OpenSearchNotificationsClient(self.opensearch)
```
Mehdi-Bendriss: Why not use `Manager` directly instead of Client?

lib/charms/opensearch/v0/constants_charm.py
```diff
@@ -94,7 +95,11 @@
 InvalidProfileConfigOption = (
     "Invalid profile configuration option. Only `production` and `testing` values are allowed."
 )
-
+SmtpWaitingRecipients = (
+    "SMTP sender configured; waiting for recipients to create email group/channel."
+)
+SmtpNoRelationData = "No relation data found. Please check the relation with smtp-integrator."
```
Mehdi-Bendriss: todo, review first part

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,257 @@
...
+
+        Returns:
+            True if config exists, False if 404.
+        """
+        try:
+            self.opensearch.request("GET", f"/_plugins/_notifications/configs/{config_id}")
+            return True
+        except OpenSearchHttpError as exc:
+            code = getattr(exc, "response_code", None)
+            if code in (404, 400):
```
Mehdi-Bendriss: what does 400 mean in this context?

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,257 @@
...
+
+    if val == "none":
+        return "none"
+    if val == "tls":
+        return "tls"
+
+    raise ValueError(f"Unsupported transport_security: {val}")
+
+
+class OpenSearchNotificationsClient:
```
Mehdi-Bendriss: Could you use a tenacity retry decorator on all methods performing http requests ?

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,257 @@
...
+            port: the smtp port
+            transport_security: an OpenSearchDistribution instance
+            from_address: the smtp address
+            description: the smtp description
+            is_enabled: whether the smtp sender is enabled
+
+        Returns:
+            NotificationsConfigRef: the notification config
+        """
+        method = check_transport_security(transport_security)
```
Mehdi-Bendriss: related to a previous point on the need to have it as a helper at all

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,257 @@
...
+        """
+        self.opensearch = opensearch
+
+    def apply_smtp_sender(
+        self,
+        *,
+        sender_id: str,
+        host: str,
+        port: int,
+        transport_security: Union[str, Any],
```
Mehdi-Bendriss: why not just `str`?

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,257 @@
...
+    Raises:
+        ValueError
+    """
+    val = getattr(transport_security, "value", transport_security)
+    val = str(val).strip().lower()
+
+    if val == "none":
+        return "none"
+    if val == "tls":
+        return "tls"
```
Mehdi-Bendriss: I have 2 points here:

1. Given that this helper is only used once, I think we can afford doing this directly in `apply_smtp_sender`, and simply going with:
```
(transport_security or "").lower() in {"none", "tls"}
```

2. from the docs it seems like we also have `start_tls` - why restrict to just none and tls?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+            event.defer()
+            return
+
+        if deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
+            if self.charm.unit.is_leader():
+                self.charm.status.set(BlockedStatus(SmtpRelationInvalid), app=True)
+            return
+
+        if not self.charm.opensearch.is_started():
+            # service must be reachable
```
Mehdi-Bendriss: nit, could we add a log line?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+            logger.debug("Deployment not ready. Deferring event.")
+            event.defer()
+            return
+
+        if deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
+            if self.charm.unit.is_leader():
+                self.charm.status.set(BlockedStatus(SmtpRelationInvalid), app=True)
+            return
+
+        if not self.charm.opensearch.is_started():
```
Mehdi-Bendriss: this should be `is_node_up` to ensure the http endpoint is up

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+            # service must be reachable
+            event.defer()
+            return
+
+        try:
+            parameters = self.smtp.get_relation_data_from_relation(event.relation)
+        except Exception as exc:
```
Mehdi-Bendriss: Could we have a more specific exception?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+                self.charm.status.set(BlockedStatus(msg), app=True)
+                return
+
+        if self.charm.unit.is_leader():
+            self.charm.status.clear(
+                SmtpMissingRequiredParameters, pattern=Status.CheckPattern.Interpolated, app=True
+            )
+
+        duplicated_sender_emails = self._has_duplicate_sender_emails()
+        if duplicated_sender_emails:
```
Mehdi-Bendriss: I do not understand this condition, why is it problematic to have a duplicate sender email across relations / use cases?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+                    "Duplicate SMTP sender email across smtp relations:",
+                    app=True,
+                    pattern=Status.CheckPattern.Start,
+                )
+
+        sender_email = str(parameters.smtp_sender)
+        sender_id = self._sender_id_from_email(sender_email)
+        label = self._label(event.relation.id)
+        group_id = self._recipient_group_id(sender_id)
+        channel_id = self._email_channel_id(sender_id)
```
Mehdi-Bendriss: These methods should be in the manager, and how about merging them all together in a single method and returning a dict or a namedtuple?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+            # clear any previous duplicate warning
+            if self.charm.unit.is_leader():
+                self.charm.status.clear(
+                    "Duplicate SMTP sender email across smtp relations:",
+                    app=True,
+                    pattern=Status.CheckPattern.Start,
+                )
+
+        sender_email = str(parameters.smtp_sender)
+        sender_id = self._sender_id_from_email(sender_email)
```
Mehdi-Bendriss: Why not use the relation id as sender id?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+
+        # store keystore creds on every unit
+        entries = {
+            f"opensearch.notifications.core.email.{sender_id}.username": parameters.user,
+            f"opensearch.notifications.core.email.{sender_id}.password": parameters.password,
+        }
+        self.charm.keystore_manager.put_entries(entries)
+
+        # reload secure settings
+        if not self.charm.keystore_manager.reload():
```
Mehdi-Bendriss: why do we emit a custom reload event in snapshots and call it directly here?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+
+        sender_email = str(parameters.smtp_sender)
+        sender_id = self._sender_id_from_email(sender_email)
+        label = self._label(event.relation.id)
+        group_id = self._recipient_group_id(sender_id)
+        channel_id = self._email_channel_id(sender_id)
+
+        # create/update SMTP sender config
+        if self.charm.unit.is_leader():
+            self.charm.notifications.apply_smtp_sender(
```
Mehdi-Bendriss: missing exception handling for this and other methods below that make an HTTP request

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+
+        # store cleanup info per relation
+        cleanup = {
+            "keys": list(entries.keys()),
+            "notification_config_ids": [sender_id, group_id, channel_id],
+        }
+        self.charm.plugin_manager.put_plugin_config(scope=Scope.UNIT, label=label, cleanup=cleanup)
+
+        # leader stores secret for subclusters for per relation
+        if self.charm.unit.is_leader():
```
Mehdi-Bendriss: Should we add a condition about this being a large deployment?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+            event.defer()
+            return
+
+        # store cleanup info per relation
+        cleanup = {
+            "keys": list(entries.keys()),
+            "notification_config_ids": [sender_id, group_id, channel_id],
+        }
+        self.charm.plugin_manager.put_plugin_config(scope=Scope.UNIT, label=label, cleanup=cleanup)
+
```
Mehdi-Bendriss: can we exit here if the unit is not the leader and reduce 1 level of indentation?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+                    recipients=[str(r) for r in parameters.recipients],
+                )
+                self.charm.notifications.apply_email_channel(
+                    channel_id=channel_id,
+                    sender_id=sender_id,
+                    email_group_ids=[group_id],
+                    fallback_recipients=[],
+                )
+                self.charm.status.clear(SmtpWaitingRecipients, app=True)
+            if self.charm.unit.is_leader() and not parameters.recipients:
```
Mehdi-Bendriss:
```suggestion
            else:
```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+
+        if deployment_desc.typ != DeploymentType.MAIN_ORCHESTRATOR:
+            if self.charm.unit.is_leader():
+                self.charm.status.clear(SmtpRelationInvalid, app=True)
+            return
+
+        label = self._label(event.relation.id)
+
+        plugin_config = self.charm.state.unit.plugin_config_info.get(label)
+        if not plugin_config:
```
Mehdi-Bendriss: nit: walrus

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+        self.charm.plugin_manager.remove_plugin_config(scope=Scope.UNIT, label=label)
+
+        if not self.charm.unit.is_leader():
+            return
+
+        # remove secrets and OpenSearch configs created by this relation
+        if self.charm.unit.is_leader():
+            remove_plugin_secret(self.charm, label)
+            # delete channel first, then group, then sender
+            for config_id in reversed(config_ids):
```
Mehdi-Bendriss: i'm afraid the order is not maintained when putting lists in the relation data / secrets, so this may not have the expected behavior

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+                try:
+                    self.charm.notifications.delete_config(config_id)
+                except Exception:
```
Mehdi-Bendriss: can we have a narrower exception?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+            self.charm.peer_cluster_provider.refresh_relation_data(event)
+
+    def _on_secret_changed(self, event: SecretChangedEvent) -> None:
+        """Handles secret changes (support multiple smtp relations).
+
+        Args:
+            event: SecretChangedEvent
+        """
+        secret = getattr(event, "secret", None)
+        label = getattr(secret, "label", None)
```
Mehdi-Bendriss: let's just directly access properties of the event

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+
+        if not label or SMTP_SECRET_LABEL not in label:
+            return
+
+        content = event.secret.get_content(refresh=True)
+
+        m = re.search(r"(plugin-notifications-\d+)", event.secret.label)
+        if not m:
```
Mehdi-Bendriss: nit: walrus + variable name

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+            return
+        label = m.group(1)
+
+        if not (plugin_config := decode_plugin_secret_content(content, label)):
+            return
+
+        keys = plugin_config.get("keys")
+        if not keys:
```
Mehdi-Bendriss: nit walrus

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,257 @@
...
+            "is_enabled": is_enabled,
+            "smtp_account": {
+                "host": host,
+                "port": int(port),
+                "method": method,
+                "from_address": from_address,
+            },
+        }
+        self._create_or_update_config(config_id=sender_id, name=sender_id, config=config)
+        return NotificationsConfigRef(config_id=sender_id, name=sender_id)
```
Mehdi-Bendriss: Do we need a return in all the public methods?

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def __init__(self, opensearch):
+        """The constructor for OpenSearchNotificationsClient.
+
+        Args:
+            opensearch: an OpenSearchDistribution instance
+        """
+        self.opensearch = opensearch
+
+    def apply_smtp_sender(
```
Mehdi-Bendriss: Could we replace the prefix `apply_` by `put_`? as "put" is easier understood as create/update

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,257 @@
...
+        self,
+        *,
+        sender_id: str,
+        host: str,
+        port: int,
+        transport_security: Union[str, Any],
+        from_address: str,
+        description: str = "",
+        is_enabled: bool = True,
```
Mehdi-Bendriss: should `is_enabled` ever be False? I don't think we need to ever consider keeping an SMTP related resource a disabled instead of deleting it, right?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,333 @@
...
+            event.defer()
+            return
+
+        self.charm.plugin_manager.remove_plugin_config(scope=Scope.UNIT, label=label)
+
+        if not self.charm.unit.is_leader():
+            return
+
+        # remove secrets and OpenSearch configs created by this relation
+        if self.charm.unit.is_leader():
```
Mehdi-Bendriss: condition not necessary, exitted earlier if non leader

tests/integration/plugins/test_plugins.py
```diff
@@ -135,6 +138,62 @@ async def _wait_for_units(
         )
 
 
+async def _notifications_list_configs(ops_test: OpsTest, base_url: str) -> dict[str, Any]:
```
Mehdi-Bendriss, line 152: missing docstrings in this module

tests/integration/plugins/test_plugins.py
```diff
@@ -135,6 +138,62 @@ async def _wait_for_units(
...
+    return await http_request(ops_test, "GET", f"{base_url}/_plugins/_notifications/configs")
+
+
+def _cfg_name(item: dict) -> str | None:
```
Mehdi-Bendriss, line 165: is this a notifications specific helper? if yes, we could prefix it with it or if "plugins" the same
(nit: `config` instead of _cfg)

Mehdi-Bendriss (commented):
Thank you Gulsum! This is great work! I left some comments and questions

lib/charms/opensearch/v0/constants_charm.py
gatici: Done

lib/charms/opensearch/v0/opensearch_base_charm.py
gatici: I renamed it as  OpenSearchNotificationsManager.

lib/charms/opensearch/v0/opensearch_notifications_client.py
gatici: Done

lib/charms/opensearch/v0/opensearch_notifications_client.py
gatici: 400 Bad request means that request format or payload is not correct. I only kept 404 here and if 400 happens, it will raise.

lib/charms/opensearch/v0/opensearch_notifications_client.py
gatici: check_transport_security method is removed.

lib/charms/opensearch/v0/opensearch_notifications_client.py
gatici: I defined it as Enum after Akram's suggestion.

lib/charms/opensearch/v0/opensearch_notifications_client.py
gatici: this method is removed and implementation is changed as suggested.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Done

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Done

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Used SecretError here.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: We are using a configuration object (smtp_account) identified with its config ID. The config id is derived from sender email address and every sender-email address generates the same config id. The config id must be unique as smtp_account is treated as stateful and the  credentials live in the key store also keyed with this config id. Besides, the email channel and the group is also reference this ID.  Our implementation assumes that one sender_config = one SMTP identity. That's why we do not allow duplicate sender emails.
```
{
  "config_id": "smtp-sender-h68805701-gmail-com",
  "config_type": "smtp_account",
  "smtp_account": {
    "host": "smtp.gmail.com",
    "port": 587,
    "method": "start_tls",
    "from_address": "h68805701@gmail.com"
  }
}
```

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,241 @@
...
+    """
+
+    config_id: str
+    name: str
+
+
+class OpenSearchNotificationsClient:
+    """Notifications plugin API client using OpenSearchDistribution.request()."""
+
+    def __init__(self, opensearch):
```
akram09: **nit:** Missing type definition

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,241 @@
...
+This client wraps OpenSearchDistribution.request() to manage notifications configs:
+- smtp sender (config_type: smtp_account)
+- email group (config_type: email_group)
+- email channel (config_type: email)
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Dict, Iterable, List, Optional, Union
```
akram09: **nit:** to integrate it easily in the single kernel we would need to abstain from using typing

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        self.opensearch = opensearch
+
+    def apply_smtp_sender(
+        self,
+        *,
+        sender_id: str,
+        host: str,
+        port: int,
+        transport_security: Union[str, Any],
```
akram09: **nit:** Why not enum ?

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,241 @@
...
+        description: str = "",
+        is_enabled: bool = True,
+    ) -> NotificationsConfigRef:
+        """Apply smtp sender configuration.
+
+        Args:
+            sender_id: the id of the sender
+            host: the smtp host
+            port: the smtp port
+            transport_security: an OpenSearchDistribution instance
```
akram09: I think here the description is wrong

lib/charms/opensearch/v0/opensearch_notifications_client.py
akram09: Right I think we should raise in case of 400

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,241 @@
...
+            config_id: Notification Config ID
+            name: Notification Name
+            config: Notification Config
+        """
+        if self._exists(config_id):
+            self._update_config(config_id=config_id, config=config)
+        else:
+            self._create_config(config_id=config_id, name=name, config=config)
+
+    def _exists(self, config_id: str) -> bool:
```
akram09: **nit: _config_exists**

lib/charms/opensearch/v0/opensearch_notifications_client.py
```diff
@@ -0,0 +1,241 @@
...
+            "email": {
+                "email_account_id": sender_id,
+                "recipient_list": [{"recipient": r} for r in (fallback_recipients or [])],
+                "email_group_id_list": list(email_group_ids),
+            },
+        }
+        self._create_or_update_config(config_id=channel_id, name=channel_id, config=config)
+        return NotificationsConfigRef(config_id=channel_id, name=channel_id)
+
+    def delete_config(self, config_id: str) -> None:
```
akram09: Should we check if the config exists before ?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,335 @@
...
+        self.smtp = SmtpRequires(self.charm, self.relation_name)
+
+        self.framework.observe(self.smtp.on.smtp_data_available, self._on_smtp_credentials_changed)
+        self.framework.observe(
+            self.charm.on[self.relation_name].relation_broken, self._on_smtp_credentials_gone
+        )
+        self.framework.observe(self.charm.on.secret_changed, self._on_secret_changed)
+
+    @staticmethod
+    def _label(relation_id: int) -> str:
```
akram09: I would like to have these static methods at least in the manager.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
akram09, line 1: Why the file is called `opensearch_plugin_manager` ? shouldn't we have a separate smtp_events.py file ?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,335 @@
...
+        for rel in self.charm.model.relations[self.relation_name]:
+            try:
+                data = self.smtp.get_relation_data_from_relation(rel)
+            except Exception:
+                # ignore the error, handler will take care of it
+                continue
+            if not data or not data.smtp_sender:
+                continue
+            sender = str(data.smtp_sender)
+            if sender in seen and seen[sender] != rel.id:
```
akram09: Is there a case where we have more than one duplicate ? Meaning that if we continue the loop we would find another duplicate that we are not reaching with this method

lib/charms/opensearch/v0/opensearch_plugin_manager.py
akram09: Nice catch

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,335 @@
...
+
+        if not self.charm.opensearch.is_started():
+            # service must be reachable
+            event.defer()
+            return
+
+        try:
+            parameters = self.smtp.get_relation_data_from_relation(event.relation)
+        except Exception as exc:
+            msg = f"Could not read smtp relation data: {exc}"
```
akram09: This should be in the constants

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,335 @@
...
+            logger.error(msg)
+            if self.charm.unit.is_leader():
+                self.charm.status.set(BlockedStatus(msg), app=True)
+                return
+
+            if not parameters:
```
akram09: This is always the case when we have an exception no ? Or this should not be in the try except block

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,335 @@
...
+            self.charm.status.clear(
+                SmtpMissingRequiredParameters, pattern=Status.CheckPattern.Interpolated, app=True
+            )
+
+        duplicated_sender_emails = self._has_duplicate_sender_emails()
+        if duplicated_sender_emails:
+            if self.charm.unit.is_leader():
+                self.charm.status.set(
+                    BlockedStatus(
+                        f"Duplicate SMTP sender email across smtp relations: {duplicated_sender_emails}"
```
akram09: Add this to constants_charm.py

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,335 @@
...
+        # store cleanup info per relation
+        cleanup = {
+            "keys": list(entries.keys()),
+            "notification_config_ids": [sender_id, group_id, channel_id],
+        }
+        self.charm.plugin_manager.put_plugin_config(scope=Scope.UNIT, label=label, cleanup=cleanup)
+
+        # leader stores secret for subclusters for per relation
+        if self.charm.unit.is_leader():
+            store_plugin_secret(
```
akram09: I am personally not a big fan of helpers functions that takes the charm as parameter.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
akram09: I think we are already blocking at the start of the function if we are not a main orchestrator. Wouldn't be that enough ?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Done

lib/charms/opensearch/v0/opensearch_plugin_manager.py
```diff
@@ -39,6 +53,335 @@
...
+        label = self._label(event.relation.id)
+        group_id = self._recipient_group_id(sender_id)
+        channel_id = self._email_channel_id(sender_id)
+        ts = getattr(parameters.transport_security, "value", parameters.transport_security)
+        transport_security = str(ts).strip().lower()
+
+        # create/update SMTP sender config
+        if self.charm.unit.is_leader():
+            self.charm.notifications.apply_smtp_sender(
+                sender_id=sender_id,
```
akram09: **very nit:** In upstream I saw they are calling it `smtp_account` maybe we could keep the same naming instead of `sender_id` which was a bit confusing to me at first.

akram09 (commented):
Seems good for me, thank you @gatici . I will approve once I finish my manual testing.

docs/how-to/add-smtp-credentials.md
```diff
@@ -0,0 +1,92 @@
+(how-to-guides-add-smtp-credentials)=
```
akram09, line 1: Shouldn't we request a review from Technical Authors ?

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: The relation id is not deterministic and we can not find and update using the relation_id. The user may update part of config and we need to look for existing config it it does not exist, we will create it. That's why each sender, recipient and email channel config ids and names are created using a deterministic config_id, derived from sender_email.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: The possible NotificationsClientError and OpensearchHttpError's are handled.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: We already checking the deployment type at the beginning of the method  and it never reaches here if it is not main orchestrator.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Done

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Done

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Done

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Done

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Done

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Done

docs/how-to/add-smtp-credentials.md
gatici: I requested review from @izmalk .

lib/charms/opensearch/v0/opensearch_notifications_client.py
gatici: Removed return values.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: We started to emit keystore.reload_event here.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: OK, this is fixed by getting channel_id and group_id by getting sender_id.
```
            # delete in dependency order: channel, then group, then sender
            sender_id = sender_ids[0]
            channel_id = self.charm.notifications.email_channel_id(sender_id)
            group_id = self.charm.notifications.recipient_group_id(sender_id)
            for config_id in (channel_id, group_id, sender_id):
                try:
                    self.charm.notifications.delete_config(config_id)
                except NotificationsClientError:
                    logger.exception("Failed deleting notifications config %s", config_id)
 ```

File: lib/charms/opensearch/v0/opensearch_plugin_manager.py (hunk 24 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_plugin_manager.py]: Done

File: lib/charms/opensearch/v0/opensearch_notifications_client.py (hunk 27 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_notifications_client.py]: Done

File: lib/charms/opensearch/v0/opensearch_notifications_client.py (hunk 28 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_notifications_client.py]: Removed is_enabled parameters.

File: lib/charms/opensearch/v0/opensearch_plugin_manager.py (hunk 29 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_plugin_manager.py]: Done

File: tests/integration/plugins/test_plugins.py (hunk 30 above)
Comment (gatici) [reply tests/integration/plugins/test_plugins.py:152]: Added.

File: tests/integration/plugins/test_plugins.py (hunk 31 above)
Comment (gatici) [reply tests/integration/plugins/test_plugins.py:165]: Yes, it is notifications specific helper and renamed as _notifications_config_name.

File: lib/charms/opensearch/v0/opensearch_notifications_client.py (hunk 32 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_notifications_client.py]: Added.

File: lib/charms/opensearch/v0/opensearch_notifications_client.py (hunk 33 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_notifications_client.py]: OK, avoided importing from typing.

File: lib/charms/opensearch/v0/opensearch_notifications_client.py (hunk 34 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_notifications_client.py]: transport_security is defined as Enum.

File: lib/charms/opensearch/v0/opensearch_notifications_client.py (hunk 35 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_notifications_client.py]: yes, I corrected it.

File: lib/charms/opensearch/v0/opensearch_notifications_client.py (hunk 36 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_notifications_client.py]: Renamed.

File: lib/charms/opensearch/v0/opensearch_notifications_client.py (hunk 37 above)
Comment (gatici) [reply lib/charms/opensearch/v0/opensearch_notifications_client.py]: we do not need to check the existence as we handle the exception and log in case it is not found.
```
            try:
                self.charm.notifications.delete_config(config_id)
            except OpensearchHttpError:
                logger.exception("Failed deleting notifications config %s", config_id)
  ```

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: They are moved to Manager.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Moved smtp_events to a new file.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Practically, this scenario should never happen. Each smtp relation has  a single sender and normally we block when we find a single duplicate. As we never allow having  a duplicate, finding a single duplicate is enough.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Added to constants.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: This is outside of the try/except block as it runs only when the try completed without raising.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: Added to constants.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: I added store_plugin_secret method to base_charm. By this way, we do not need to give the charm as a parameter.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: I replaced all sender_id's with smtp_account_id's.

lib/charms/opensearch/v0/opensearch_notifications_client.py
gatici: yes, we did.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
Mehdi-Bendriss: Yes, I understood that a bit later in the PR, but I don't think this should be a limitation for us if we use the relation_id as sender id https://github.com/canonical/opensearch-operator/pull/789#discussion_r2736529262

assume:
- smtp rel 1: 
  - sender: sre-alerts@example.com
  - recipients: service1@example.com
- smtp rel 2:
  - sender: sre-alerts@example.com
  - recipients: service2@example.com, security@example.com

lib/charms/opensearch/v0/opensearch_plugin_manager.py
Mehdi-Bendriss: Could you explain what you mean by "the relation id is non deterministic"? it is a unique integer per smtp-integrator relation

gatici: Yes, it is unique but non-deterministic. We can not derive or predict the relation id from the inputs like sender-email. The user may update part of the config and we need to look for existing config. If it does not exist, we create it. We need to track and check which relation has which config and duplicate smtp-accout detection is difficult.  If we use relation-id to create notification's config, implementation is simpler as not sender-change-cleanup and we are only tight to relation-ids but notification config is less readable. Besides, if we use sender_id as config_id, configs across all relations is more stable as same_email creates same config_id.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: We removed this method.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: The implementation is changed to use relation_id as notification config_id.

lib/charms/opensearch/v0/opensearch_plugin_manager.py
gatici: We started to use relation_id as sender_id.

lib/charms/opensearch/v0/constants_charm.py
Mehdi-Bendriss: @gatici should we have a message that's more actionable, for ex: Missing smtp configuration data from the smtp-integrator.

lib/charms/opensearch/v0/opensearch_keystore.py
```diff
@@ -73,7 +73,9 @@ def remove_entries(self, keys: list[str]) -> None:
             try:
                 self._opensearch.run_bin("keystore", f"remove {key}")
             except OpenSearchCmdError as e:
-                if e.err and "does not exist in the keystore" in e.err:
+                # distro passes stderr as first arg (e.cmd); some callers set e.err
+                err_text = e.err or e.cmd or ""
```
Mehdi-Bendriss: why `e.cmd`? isn't it just the command that was typed?

lib/charms/opensearch/v0/opensearch_notifications_manager.py
```diff
@@ -0,0 +1,324 @@
...
+
+class NotificationsClientError(RuntimeError):
+    """Raises when Notifications API operations fail."""
+
+
+class TransportSecurity(str, Enum):
+    """SMTP transport security protocol."""
+
+    NONE = "none"
+    STARTTLS = "starttls"
```
Mehdi-Bendriss, line 37: Any reason not to have `start_tls` directly?

lib/charms/opensearch/v0/opensearch_notifications.py
```diff
@@ -0,0 +1,324 @@
...
+    STARTTLS = "starttls"
+    TLS = "tls"
+
+    def api_method(self) -> str:
+        """Return the OpenSearch Notifications API method string."""
+        return {"none": "none", "starttls": "start_tls", "tls": "ssl"}[self.value]
+
+
+@dataclass(frozen=True)
+class NotificationsConfigRef:
```
Mehdi-Bendriss: We can delete

lib/charms/opensearch/v0/opensearch_notifications_manager.py
```diff
@@ -0,0 +1,324 @@
...
+@dataclass(frozen=True)
+class SmtpConfig:
+    """SMTP-related config derived from relation data."""
+
+    sender_email: str
+    smtp_account_id: str
+    label: str
+    group_id: str
+    channel_id: str
+    transport_security: TransportSecurity
```
Mehdi-Bendriss, line 63: Could we review optional fields if any and adjust the types and default values?

lib/charms/opensearch/v0/opensearch_notifications.py
```diff
@@ -0,0 +1,324 @@
...
+        """
+        try:
+            self.opensearch.request("DELETE", f"/_plugins/_notifications/configs/{config_id}")
+        except OpenSearchHttpError as exc:
+            code = getattr(exc, "response_code", None)
+            if code == 404:
+                return
+            raise
+
+    def _create_or_update_config(
```
Mehdi-Bendriss: could you add a tenacity retry here?

lib/charms/opensearch/v0/opensearch_base_charm.py
```diff
@@ -230,6 +236,9 @@ def __init__(self, *args, distro: Type[OpenSearchDistribution] = None):
 
         self.plugin_manager = OpenSearchPluginManager(self.state)
         self.plugin_events = OpenSearchPluginEvents(self)
+        self.notifications = OpenSearchNotificationsManager(self.opensearch)
```
Mehdi-Bendriss: nit, should we perhaps for consistency, name it "notifications_manager"?

lib/charms/opensearch/v0/opensearch_notifications_events.py
Mehdi-Bendriss, line 1: just a note on the module name, we have 2 related modules with different names, `smtp` and `notifications`

lib/charms/opensearch/v0/opensearch_notifications_events.py
```diff
@@ -0,0 +1,319 @@
...
+            self.charm.status.clear(SmtpRelationInvalid, app=True)
+            self.charm.status.clear(SMTPConfigurationError, app=True)
+            self.charm.status.clear(SmtpNoRelationData, app=True)
+            self.charm.status.clear(
+                SmtpMissingRequiredParameters, pattern=Status.CheckPattern.Interpolated, app=True
+            )
+
+        label = self.charm.notifications.label(event.relation.id)
+        plugin_config = self.charm.state.unit.plugin_config_info.get(label)
+        if not plugin_config:
```
Mehdi-Bendriss, line 245: nit, walrus

lib/charms/opensearch/v0/opensearch_smtp.py
```diff
@@ -0,0 +1,319 @@
...
+            return
+
+        self.charm.remove_plugin_secret(label)
+
+        # delete in dependency order: channel, then group, then smtp account
+        channel_id = self.charm.notifications.email_channel_id(smtp_account_id)
+        group_id = self.charm.notifications.recipient_group_id(smtp_account_id)
+        for config_id in (channel_id, group_id, smtp_account_id):
+            try:
+                self.charm.notifications.delete_config(config_id)
```
Mehdi-Bendriss: question on the order of operations, should this come after or before the keystore cleanup?

tests/integration/plugins/test_plugins.py
Mehdi-Bendriss: i think this was missed

tests/integration/plugins/test_plugins.py
Mehdi-Bendriss: i think this was missed for a few functions

Mehdi-Bendriss (commented):
Thank you Gulsum, great work! I only have small comments

lib/charms/opensearch/v0/constants_charm.py
gatici: yes,  I replaced it with more actionable one.

tests/integration/plugins/test_plugins.py
gatici: Added docstrings for other functions.

tests/integration/plugins/test_plugins.py
gatici: Done

lib/charms/opensearch/v0/opensearch_keystore.py
gatici: I hit this error while testing:
```
  File "/var/lib/juju/agents/unit-main-0/charm/lib/charms/opensearch/v0/opensearch_keystore.py", line 74, in remove_entries
    self._opensearch.run_bin("keystore", f"remove {key}")
  File "/var/lib/juju/agents/unit-main-0/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 213, in run_bin
    return self._run_cmd(opensearch_command, args, stdin=stdin)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-main-0/charm/venv/lib/python3.12/site-packages/tenacity/__init__.py", line 338, in wrapped_f
    return copy(f, *args, **kw)
           ^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-main-0/charm/venv/lib/python3.12/site-packages/tenacity/__init__.py", line 477, in __call__
    do = self.iter(retry_state=retry_state)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-main-0/charm/venv/lib/python3.12/site-packages/tenacity/__init__.py", line 378, in iter
    result = action(retry_state)
             ^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-main-0/charm/venv/lib/python3.12/site-packages/tenacity/__init__.py", line 420, in exc_check
    raise retry_exc.reraise()
          ^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-main-0/charm/venv/lib/python3.12/site-packages/tenacity/__init__.py", line 187, in reraise
    raise self.last_attempt.result()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/var/lib/juju/agents/unit-main-0/charm/venv/lib/python3.12/site-packages/tenacity/__init__.py", line 480, in __call__
    result = fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^
  File "/var/lib/juju/agents/unit-main-0/charm/lib/charms/opensearch/v0/opensearch_distro.py", line 396, in _run_cmd
    raise OpenSearchCmdError(output.stderr)
charms.opensearch.v0.opensearch_exceptions.OpenSearchCmdError: warning: no-jdk distributions that do not bundle a JDK are deprecated and will be remov
ed in a future release
ERROR: Setting [opensearch.notifications.core.email.smtp-sender-h68805701-gmail-com.username] does not exist in the keystore.

```

In opensearch_distro.py line 396, it calls `raise OpenSearchCmdError(output.stderr)` so first positional argument which is called `cmd` is getting the stderr. That's why err.cmd is used to look for error txt.

But now I fixed it in the Opensearch_distro.py.

lib/charms/opensearch/v0/opensearch_notifications.py
gatici: removed.

lib/charms/opensearch/v0/opensearch_notifications_manager.py
gatici: All the fields are required, none of them optional.

lib/charms/opensearch/v0/opensearch_base_charm.py
gatici: Done

lib/charms/opensearch/v0/opensearch_notifications_manager.py
gatici: TransportSecurity keep the enum values which come from smtp relation, so it should match what smtp integrator sends.

lib/charms/opensearch/v0/opensearch_notifications_events.py
gatici: Done

lib/charms/opensearch/v0/opensearch_smtp.py
gatici: Good catch. First the config should be deleted, the keystore cleanup. If we remove keystore entries first, the smtp account config can still exist in OpenSearch but without credentials. Any send using that account may fail. 
Order is changed.

lib/charms/opensearch/v0/opensearch_notifications_events.py
gatici: I renamed them notifications_manager and notifications_events. I also renamed SmtpEvents to NotifcationEvents.

Mehdi-Bendriss (dismissed):
Thanks Gulsum!

akram09 (dismissed):
Seems good for me. Thank you @gatici

Mehdi-Bendriss (dismissed):

tests/integration/relations/test_oauth.py
```diff
@@ -290,6 +319,13 @@ async def test_setup_large_cluster(ops_test: OpsTest, charm, series, microk8s_mo
         f"{DATA_APP}:opensearch-client", f"{DATA_INTEGRATOR_NAME}:opensearch"
     )
 
+    # Let Juju settle while the cluster forms TLS + security index + peer orchestration
+    await ops_test.model.wait_for_idle(
```
Mehdi-Bendriss: this is redundant, it is what the `wait_until` code does

tests/integration/plugins/test_plugins.py
```diff
@@ -1459,19 +1470,25 @@ async def test_custom_codecs_plugin(ops_test: OpsTest, deploy_type: str) -> None
...
 
-    # compare size of indices, zstd index should be smaller
+    # compare size of indices; zstd should be smaller or equal
     stats = await http_request(ops_test, "GET", f"{base_url}/{zstd},{default}/_stats/store")
     zstd_size = stats["indices"][zstd]["total"]["store"]["size_in_bytes"]
     default_size = stats["indices"][default]["total"]["store"]["size_in_bytes"]
 
     logger.info(f"Index sizes - zstd: {zstd_size} default: {default_size}")
-    assert zstd_size < default_size
+    assert zstd_size > 0 and default_size > 0, "Index store sizes should be positive after bulk"
```
Mehdi-Bendriss:
```suggestion
    assert zstd_size > 0 and default_size > 0, "Index store sizes should be positive after bulk indexing"
```

tests/integration/plugins/test_plugins.py
```diff
@@ -1459,19 +1470,25 @@ async def test_custom_codecs_plugin(ops_test: OpsTest, deploy_type: str) -> None
...
+    assert (
+        zstd_size <= default_size
```
Mehdi-Bendriss, line 1507: I think this should be `<` as per the previous code, right?

tests/integration/relations/test_oauth.py
gatici: The wait_until was missing data_integrator charm. The data_integrator charm added to wait_until and wait_for_idle is removed.

tests/integration/plugins/test_plugins.py
gatici: This is one of the flaky test and we were getting following result in the previous executions.
```
  File "/root/spread_project/tests/integration/plugins/test_plugins.py", line 1474, in test_custom_codecs_plugin
    assert zstd_size < default_size
AssertionError: assert 147474 < 416

  File "/root/spread_project/tests/integration/plugins/test_plugins.py", line 1474, in test_custom_codecs_plugin
    assert zstd_size < default_size
AssertionError: assert 416 < 416
```
Issue (147474 > 416): The compression happens during segment merges, not immediately after indexing. Without force merge, unmerged segments made zstd appear larger. This is fixed by forcing merges before comparing sizes.
Issue (416 == 416): For very small datasets, fixed overhead (metadata, segment headers, minimum segment sizes) can dominate, making sizes equal. This is fixed by using <= instead of < to allow equality.

Mehdi-Bendriss (approved):
Thanks Gulsum!

akram09 (approved):

PR #799, opened 2026-02-09
Bug-fix: removing the hard-coded unit value + acomodating for the mutual exclusivness of units and machines variables

sonyavalo:
The bug is [here](https://github.com/canonical/opensearch-operator/blob/83eea3b52e0555cb387a875b33d7025539288db6/terraform/charm/simple_deployment/main.tf#L66)
The `units` value is hard-coded to 1 in the `self-signed-certificates` juju application definition. 

## Solution:
Changed the `units` value to follow `condition ? true_value : false_value` syntax. If the `machines ` value is not defined `units=var.self-signed-certificates.units`, but if  `machines ` value is defined, `units=null`.
This accommodates for the terrafrom juju [schema ](https://registry.terraform.io/providers/juju/juju/1.1.1/docs/resources/application#:~:text=machines%20(Set%20of%20String)%20Specify%20the%20target%20machines%20for%20the%20application%27s%20units.%20The%20number%20of%20machines%20in%20the%20set%20indicates%20the%20unit%20count%20for%20the%20application.%20Removing%20a%20machine%20from%20the%20set%20will%20remove%20the%20application%27s%20unit%20residing%20on%20it.%20machines%20is%20mutually%20exclusive%20with%20units.) that states that `machines` is mutually exclusive with `units`. 
Additionally made sure that all the `variables.tf` files expect `units` value.

- [ ] Manually
- [ ] Unit tests
- [ ] Integration tests

## Checklist

### DIFF
--- a/terraform/charm/simple_deployment/main.tf
+++ b/terraform/charm/simple_deployment/main.tf
@@ -63,7 +63,7 @@ resource "juju_application" "self-signed-certificates" {
 
   config = var.self-signed-certificates.config
 
-  units       = 1
+  units       = (var.self-signed-certificates.machines == null || length(var.self-signed-certificates.machines) == 0) ? var.self-signed-certificates.units : null
   constraints = var.self-signed-certificates.constraints
   machines    = (var.self-signed-certificates.machines == null || length(var.self-signed-certificates.machines) == 0) ? null : var.self-signed-certificates.machines
 }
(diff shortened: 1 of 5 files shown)


terraform/charm/simple_deployment/main.tf
```diff
@@ -63,7 +63,7 @@ resource "juju_application" "self-signed-certificates" {
 
   config = var.self-signed-certificates.config
 
-  units       = 1
+  units       = (var.self-signed-certificates.machines == null || length(var.self-signed-certificates.machines) == 0) ? var.self-signed-certificates.units : null
```
Mehdi-Bendriss, line 66: with this, we are prioritizing `machines` over `units`, which i believe is the right thing to do

Mehdi-Bendriss (approved):
Thank you Sofya for the contribution!

gatici (approved):
Thank you @sonyavalo!

PR #815, opened 2026-04-06
Fix/temporary file service account gcs

akram09:
This Pull Request fix an issue in opensearch snapshots when related with gcs integrator. The charm was saving the `service_account.json` in filesystem which can cause an issue of leaking credentials. To address the issue we use temporary files to save the credentials temporary while we add it to the opensearch keystore and the tempfile will be deleted.


Mehdi-Bendriss (approved):
Nice, thank you

skourta (approved):
Thank you @akram09

PR #824, opened 2026-06-12
docs: Add tutorial test

izmalk:

We want to test our tutorial.

## Solution:

We add tests using [Spread](https://github.com/canonical/spread) and [STAT](https://github.com/izmalk/stat).

- [x] Manually
- [ ] Unit tests
- [ ] Integration tests

Run the tests with `make test` from the tests/tutorial path.

## Checklist

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -6,3 +6,4 @@ self-hosted-runner:
     - tiobe
     - noble
     - jammy
+    - xlarge
--- a/.github/workflows/tutorial-tests.yaml
+++ b/.github/workflows/tutorial-tests.yaml
@@ -11,7 +11,7 @@ on:
 jobs:
   tutorial-test:
     name: Tutorial end-to-end
-    runs-on: [self-hosted, linux, AMD64, X64, xlarge, noble]
+    runs-on: [self-hosted, linux, amd64, xlarge, noble]
     timeout-minutes: 180
     steps:
       - name: Checkout
--- a/.github/workflows/tutorial-tests.yaml
+++ b/.github/workflows/tutorial-tests.yaml
@@ -1,0 +1,53 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+name: Tutorial tests
+
+on:
+  workflow_dispatch:
+  workflow_call:
+  schedule:
+    - cron: '0 3 1 * *'  # 1st of every month at 03:00 UTC
+
+jobs:
+  tutorial-test:
+    name: Tutorial end-to-end
+    runs-on: [self-hosted, linux, AMD64, X64, xlarge, noble]
+    timeout-minutes: 180
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Check KVM availability
+        run: |
+          if [[ ! -e /dev/kvm ]]; then
+            echo "::error::KVM is not available on this runner. Multipass requires KVM for VM creation."
+            exit 1
+          fi
+
+      - name: Install Multipass
+        run: |
+          if ! command -v multipass &>/dev/null; then
+            sudo snap install multipass
+          fi
+
+      - name: Install Go and Spread
+        run: |
+          if ! command -v go &>/dev/null; then
+            sudo snap install go --classic
+          fi
+          go install github.com/canonical/spread/cmd/spread@latest
+          echo "$HOME/go/bin" >> "$GITHUB_PATH"
+
+      - name: Install tox
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y pipx
+          pipx install tox
+
+      - name: Run tutorial tests
+        run: tox -e tutorial
+
+      - name: Cleanup Multipass VM
+        if: always()
+        run: |
+          multipass delete --purge opensearch-tutorial-vm 2>/dev/null || true
--- a/docs/tutorial/4-integrate-with-a-client-application.md
+++ b/docs/tutorial/4-integrate-with-a-client-application.md
@@ -349,7 +349,7 @@ curl --cacert demo-ca.pem -XGET https://username:password@opensearch_node_ip:920
 ```
 
 <!-- test:assert
-curl --cacert demo-ca.pem -XGET "https://${OS_USERNAME}:${OS_PASSWORD}@${OS_ENDPOINT}/albums/_search?q=Jazz" | grep -q '"value":3'
+curl --cacert demo-ca.pem -XGET "https://${OS_USERNAME}:${OS_PASSWORD}@${OS_ENDPOINT}/albums/_search?q=Jazz" | grep -q '"value".*3'
 -->
 
 This should return a JSON response with all the Jazz albums in the index:
--- a/docs/tutorial/4-integrate-with-a-client-application.md
+++ b/docs/tutorial/4-integrate-with-a-client-application.md
@@ -185,10 +185,13 @@ Before connecting to OpenSearch, it is mandatory that you
 You can access the OpenSearch REST API in any way you prefer. In this tutorial, we will use `curl`.
 Get the IP of an OpenSearch node from the output of `juju status`
 (any of the nodes should work fine), and store the CA certificate in a local file
-(e.g. `demo-ca.pem`). Then, run the following command to connect to the OpenSearch cluster:
+(e.g. `demo-ca.pem`).
+
+Then, replacing `<username>`, `<password>`, and `<opensearch_node_ip>` with the values from
+the `get-credentials` output above, run the following command to connect to the OpenSearch cluster:
 
 ```bash
-curl --cacert demo-ca.pem -XGET https://username:password@opensearch_node_ip:9200/
+curl --cacert demo-ca.pem -XGET https://<username>:<password>@<opensearch_node_ip>:9200/
 ```
 
 <!-- test:assert
@@ -233,7 +236,7 @@ To index some data, run the following command:
 
 ```bash
 curl --cacert demo-ca.pem \
-  -XPOST https://username:password@opensearch_node_ip:9200/albums/_doc/1?refresh=true \
+  -XPOST https://<username>:<password>@<opensearch_node_ip>:9200/albums/_doc/1?refresh=true \
   -d '{"artist": "Vulfpeck", "genre": ["Funk", "Jazz"], "title": "Thrill of the Arts"}' \
   -H 'Content-Type: application/json' -H 'Accept: application/json'
 ```
@@ -267,7 +270,7 @@ Note from the response that our request was successful and the document was inde
 Use the following command to retrieve the previous document:
 
 ```bash
-curl --cacert demo-ca.pem -XGET https://username:password@opensearch_node_ip:9200/albums/_doc/1
+curl --cacert demo-ca.pem -XGET https://<username>:<password>@<opensearch_node_ip>:9200/albums/_doc/1
 ```
 
 <!-- test:assert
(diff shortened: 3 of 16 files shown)


docs/tutorial/4-integrate-with-a-client-application.md
```diff
@@ -85,30 +92,32 @@ Integrate the two applications with:
 juju integrate data-integrator opensearch
 ```
 
+<!-- test:await-idle --timeout 600 -->
+
 Wait for `juju status --relations --watch 1s` to show that the `data-integrator` application is now active:
 
 ```bash
```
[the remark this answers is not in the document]
izmalk:
```suggestion
```text
```

docs/tutorial/4-integrate-with-a-client-application.md
```diff
@@ -364,31 +419,33 @@ Run the following to remove the relation:
 juju remove-relation opensearch data-integrator
 ```
 
+<!-- test:await-idle --timeout 600 --allow-blocked data-integrator -->
+
 if you run `juju status --relations` you will see that the relation has been removed and that the
 `data-integrator` application is now in a `blocked` state.
 
 ```bash
```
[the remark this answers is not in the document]
izmalk:
```suggestion
```text
```

tests/tutorial/extract_commands.py
```diff
@@ -0,0 +1,558 @@
...
+    priority = meta.get("priority", "0")
+    kill_timeout = meta.get("kill-timeout", "30m")
+    summary = heading or script_path
+    return (
+        f'summary: "{summary}"\n'
+        f"priority: {priority}\n"
+        f"kill-timeout: {kill_timeout}\n"
+        f"execute: |\n"
+        f'  bash "$SPREAD_PATH/{script_path}"\n'
+    )
```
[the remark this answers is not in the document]
izmalk: I don't see the need for this change, as the script works correctly. It was tested here and in other documentation sets.

spread.yaml
```diff
@@ -107,10 +107,50 @@ backends:
...
+        --memory "${SPREAD_VM_MEM:-16G}" --name "$instance_name" 24.04
+
+      # Enable root SSH for Spread. Tutorial steps install the required snaps.
+      multipass exec "$instance_name" -- sudo sh -c "
+        set -e
+        echo root:${SPREAD_PASSWORD} | chpasswd
+        printf 'PasswordAuthentication yes\nPermitRootLogin yes\n' \
+          > /etc/ssh/sshd_config.d/10-spread.conf
+        systemctl restart ssh
+      "
```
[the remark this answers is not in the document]
izmalk: Doesn't make much sense for a local deployment.

spread.yaml
```diff
@@ -107,10 +107,50 @@ backends:
       - self-hosted-linux-amd64-noble-xlarge:
           username: ubuntu
 
+  multipass:
+    type: adhoc
+    allocate: |
+      instance_name="opensearch-tutorial-vm"
+      multipass delete --purge "$instance_name" 2>/dev/null || true
```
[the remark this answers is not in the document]
izmalk: I don't see a problem with that, given the provisioning of a GH runner, which is a dedicated VM. No chance of duplication in CI/CD. Local tests are unlikely to be run in parallel: 1. Why? 2. Resources limitation.

spread.yaml
```diff
@@ -107,10 +107,50 @@ backends:
...
 
+  tests/tutorial/:
+    summary: Tutorial end-to-end test
+    systems:
+      - ubuntu-24.04-64
+    variants:
+      - ubuntu24
+    prepare: |
+      export HOME=/root
+      mkdir -p /root/.local/share/juju
```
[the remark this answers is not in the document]
izmalk: We generate the shell scripts and task.yaml files by a script before we run spread. It makes no sense to generate them with spread, because it scans for the files on start.

.github/workflows/tutorial-tests.yaml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  tutorial-test:
+    name: Tutorial end-to-end
+    runs-on: [self-hosted, linux, amd64, xlarge, noble]
+    timeout-minutes: 180
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Check KVM availability
```
akram09, line 20: Why do we need to install Multipass, It seems we are allocating a self-hosted machine. We could install juju inside it directly and run the tests ?

.github/workflows/tutorial-tests.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Install Multipass
+        run: |
+          if ! command -v multipass &>/dev/null; then
+            sudo snap install multipass
+          fi
+
+      - name: Install Go and Spread
+        run: |
+          if ! command -v go &>/dev/null; then
```
akram09, line 35: We can have `sudo snap installl go --classic` without the if condition since it will not block the CI

docs/tutorial/4-integrate-with-a-client-application.md
```diff
@@ -125,10 +133,21 @@ This relation created the user, password, and CA certificate for the `data-integ
 
 To retrieve information such as the username, password, and database. Run the following command:
 
-```shell
+```bash
 juju run data-integrator/leader get-credentials
 ```
 
+<!-- test:set-variables
+command: juju run data-integrator/leader get-credentials
```
akram09, line 141: For my understanding, how does this works ? It directly inject the variables from the ouput ?

docs/tutorial/4-integrate-with-a-client-application.md
```diff
@@ -296,10 +346,14 @@ The `items` field contains the results of each operation in the bulk request.
...
 in our `albums` index, using the following command:
 
-```shell
-curl --cacert demo-ca.pem -XGET https://username:password@opensearch_node_ip:9200/albums/_search?q=Jazz
+```bash
+curl --cacert demo-ca.pem -XGET https://<username>:<password>@<opensearch_node_ip>:9200/albums/_search?q=Jazz
 ```
 
+<!-- test:assert
+curl --cacert demo-ca.pem -XGET "https://${OS_USERNAME}:${OS_PASSWORD}@${OS_ENDPOINT}/albums/_search?q=Jazz" | grep -q '"value".*3'
```
akram09, line 354: Shouldn't we compare the output with the one we are expecting to be more precise ?

docs/tutorial/4-integrate-with-a-client-application.md
```diff
@@ -400,13 +455,13 @@ self-signed-certificates:certificates  opensearch:certificates                tl
 
 Now try again to connect in the same way as the previous section
 
-```shell
-curl --cacert demo-ca.pem -XGET https://username:password@opensearch_node_ip:9200/
+```bash
+curl --cacert demo-ca.pem -XGET https://<username>:<password>@<opensearch_node_ip>:9200/
 ```
```
akram09, line 460: Shouldn't we add an assert block here ?

docs/tutorial/5-manage-passwords.md
```diff
@@ -107,7 +116,7 @@ There are three types of private keys available to be updated on this charm, and
 To change a private key to a random value, run the following command, setting `category`
 equal to your preferred type of private key:
 
-```shell
+```bash
```
akram09, line 119: I don't see tests for this tutorial

tests/tutorial/extract_commands.py
```diff
@@ -0,0 +1,551 @@
...
+    without the ``.md`` extension.
+    """
+    md_files = sorted(input_dir.glob("*.md"))
+    if not md_files:
+        sys.exit(f"Error: no .md files found in {input_dir}")
+
+    processed = 0
+    for md_file in md_files:
+        source = md_file.read_text(encoding="utf-8")
+        if not extract_spread_meta(source):
```
akram09: Shouldn't we instead just look if there is spread meta instead of doing the extracting two times ?

akram09 (commented):
Thank you Vladimir, a couple of comments from me

.github/workflows/tutorial-tests.yaml
izmalk: I just wanted to avoid failing the workflow due to something not being there.
I can't find any official docs confirming the bahavior for the snap install in case it is already installed. So I just prefere the safer option.
Do you want us to drop the check? It doesn't hurt anyone to check.

.github/workflows/tutorial-tests.yaml
izmalk: Good question. I have been running this test locally on my machine, where I prefer to use a Multipass VM for isolation and reproducibility. I guess we don't need it that much in CI/CD. Unless we decide to envelope our tutorial in a Multipass VM for isolation too. @Mehdi-Bendriss, what do you think?

docs/tutorial/4-integrate-with-a-client-application.md
izmalk: Good question. It is an intersting one. It saves values from the output of the command in a variable, that can be used later.
For example: we use `OS_USERNAME` later, which will have the value of the `username` field from the output of this command.

docs/tutorial/4-integrate-with-a-client-application.md
izmalk: But why? We expect it to fail. And it's not that improtant to actual;ly check that it fails as expected.

docs/tutorial/5-manage-passwords.md
izmalk: I'm sorry, can you clarify what do you mean by this tutorial? This page, this code block, or something else?

tests/tutorial/extract_commands.py
izmalk: Good catch. I've tried fixing it here -https://github.com/canonical/opensearch-operator/pull/824/commits/a198e0fa36b547ea3106c9702dfac5509be767ec

.github/workflows/tutorial-tests.yaml
Mehdi-Bendriss, line 20: I would rather we confirm whether KVM comes preinstalled in the self hosted runners and drop this step if that's confirmed.

.github/actionlint.yaml
```diff
@@ -6,3 +6,4 @@ self-hosted-runner:
     - tiobe
     - noble
     - jammy
+    - xlarge
```
Mehdi-Bendriss, line 9: this is for Tiobe / Tics workflow, why do we need an xlarge runner?

Mehdi-Bendriss (commented):
Thanks Vladimir, exciting stuff! I have a couple of small comments

.github/workflows/tutorial-tests.yaml
izmalk: This is for self-hosted runners - they are not immutable, AFAIK. I'm just playing it safe here by checking. The check itself does not take long. How can I reliably check whether a self-hosted runner has it?

.github/actionlint.yaml
izmalk: This linter was triggered by my tutorial test. The tutorial test requires a lot of RAM.

izmalk (commented):
Fixes and replies after review

docs/tutorial/4-integrate-with-a-client-application.md
izmalk: A more detailed comparison will be more succeptible for false positives due to possible format changes. This one is a minimal check but a very effective one to determine that we do have a connection to a working DB.

.github/workflows/tutorial-tests.yaml
Mehdi-Bendriss: @izmalk It makes sense for the tutorial to run on multipass, as this is for people to play around, but for CI not so much

izmalk: If we make the tutorial to run in Mutlipass (which I would prefer), then the test should repeat the path too.

Mehdi-Bendriss (approved):

Mehdi-Bendriss:
to be merged when ci passes.

akram09 (approved):

izmalk:
> to be merged when ci passes.

Please ignore the linkcheck. Most issues are fixed already in the Docs revamp PR.

PR #834, opened 2026-07-24
patch: Remove tag job from release workflows

akram09:
This Pull Request removes the `tag` job from the release workflow since this is used to generate the compatibility version which is not needed in refresh v2. If the release pipeline still fails, we may need to go back to the previous version of the workflow.


skourta (dismissed):

Mehdi-Bendriss (approved):
Thanks Akram

skourta (approved):

PR #857, opened 2026-09-11
patch(DPE-10924): k8s upgrades docs

smethnani:
Adds upgrades on K8s docs

I moved the "recover from rollback" section to VM only as it hasn't been tested on K8s

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,7 @@
 # OpenSearch Operator
 [![Charmhub](https://charmhub.io/opensearch/badge.svg)](https://charmhub.io/opensearch)
 [![Release](https://github.com/canonical/opensearch-operator/actions/workflows/release.yaml/badge.svg)](https://github.com/canonical/opensearch-operator/actions/workflows/release.yaml)
-[![Tests](https://github.com/canonical/opensearch-operator/actions/workflows/ci.yaml/badge.svg?branch=main)](https://github.com/canonical/opensearch-operator/actions/workflows/ci.yaml)
-[![Docs](https://github.com/canonical/opensearch-operator/actions/workflows/sync_docs.yaml/badge.svg)](https://github.com/canonical/opensearch-operator/actions/workflows/sync_docs.yaml)
+[![Tests](https://github.com/canonical/opensearch-operator/actions/workflows/ci.yaml/badge.svg?branch=2/edge)](https://github.com/canonical/opensearch-operator/actions/workflows/ci.yaml)
 
 ## Description
 
--- a/docs/how-to/upgrade.md
+++ b/docs/how-to/upgrade.md
@@ -1,7 +1,7 @@
 ---
 myst:
   html_meta:
-    description: "Upgrade Charmed OpenSearch from one minor version to another, including pre-upgrade checks, in-place upgrades, health verification, and rollback procedures."
+    description: "Upgrade Charmed OpenSearch on machines or Kubernetes from one minor version to another, including pre-upgrade checks, in-place upgrades, health verification, and rollback procedures."
 ---
 
 (how-to-guides-upgrade-index)=
@@ -67,6 +67,11 @@ The first step is to record the revision of the running application,
 as a safety measure for a rollback action.
 To accomplish this, run the `juju status` command and look for the deployed
 Charmed OpenSearch revision in the command output, e.g.:
+`````{tab-set}
+:sync-group: subtrate
+
+````{tab-item} VM
+:sync: vm
 
 ```text
 Model  Controller           Cloud/Region         Version  SLA          Timestamp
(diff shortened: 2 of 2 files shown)


docs/how-to/upgrade.md
```diff
@@ -103,9 +132,26 @@ Optionally, it is recommended to scale the application up by one unit before upg
...
 ```shell
 juju add-unit opensearch
 ```
+````
+
+````{tab-item} K8s
+:sync: k8s
+
+```shell
+juju add-unit opensearch-k8s
```
izmalk: Question: Is it the same `add-unit` in K8s? I thought we have to use `juju scale-application` instead.

docs/how-to/upgrade.md
```diff
@@ -148,10 +226,32 @@ If the charm is running on a revision prior to `185`, the `testing` profile is t
...
+````
+
+````{tab-item} K8s
+:sync: k8s
+
+```{caution}
+Charmed OpenSearch supports performance profiles with different RAM consumption:
+
+* `production`: JVM heap set to 50% of the available RAM, capped at 31 GB
+* `testing`: JVM heap fixed at ~1 GB of RAM
```
izmalk: Suggestion: This part is exactly the same and has specific numbers. Let's move it outside of the tabs, so we won't need to duplicate it.

docs/how-to/upgrade.md
```diff
@@ -228,9 +399,26 @@ To do so, follow the [Perform a minor rollback](how-to-minor-rollback) section b
...
 ```shell
 juju remove-unit opensearch/<highest unit number>
 ```
+````
+
+````{tab-item} K8s
+:sync: k8s
+
+```shell
+juju remove-unit opensearch-k8s --num-units 1
```
izmalk: Question: Same question about `juju scale-application`.

docs/how-to/upgrade.md
```diff
@@ -445,7 +771,7 @@ Follow the steps in this section to restore the cluster to a healthy state.
 For more information, please refer to the upstream
 [OpenSearch documentation about rolling upgrades](https://docs.opensearch.org/latest/migrate-or-upgrade/rolling-upgrade/#preparing-to-upgrade).
 
-### Check Juju status
+**Check Juju status**
```
izmalk: Suggestion: This is a big section below. And there is no alternative for K8s. I feel like we can drop the tabs here, for this section, and just add a warning at the start that K8s don't support this. That way we avoid sacrificing the convenience of proper headings and navigation for just having an almost empty K8s tab.

README.md
```diff
@@ -1,8 +1,7 @@
 # OpenSearch Operator
 [![Charmhub](https://charmhub.io/opensearch/badge.svg)](https://charmhub.io/opensearch)
 [![Release](https://github.com/canonical/opensearch-operator/actions/workflows/release.yaml/badge.svg)](https://github.com/canonical/opensearch-operator/actions/workflows/release.yaml)
-[![Tests](https://github.com/canonical/opensearch-operator/actions/workflows/ci.yaml/badge.svg?branch=main)](https://github.com/canonical/opensearch-operator/actions/workflows/ci.yaml)
-[![Docs](https://github.com/canonical/opensearch-operator/actions/workflows/sync_docs.yaml/badge.svg)](https://github.com/canonical/opensearch-operator/actions/workflows/sync_docs.yaml)
```
izmalk, line 5: praise: Thanks for fixing it.

izmalk (changes requested):
Looks good. I've made a few suggestions and a question about K8s Juju scaling.

docs/how-to/upgrade.md
smethnani: It's what I've been using with k8s but I can see the juju [docs](https://canonical.com/juju/docs/juju-cli/3.6/howto/manage-units/#control-the-number-of-units) do suggest `scale-application`. But they also say `remove-unit <unit>` works for k8s which doesn't (it needs to be just the application name with `--num-units <n>` or juju will complain) :/ I'll switch both to `scale-application`

docs/how-to/upgrade.md
smethnani: Ah I was trying to avoid the revision part in k8s. I think it makes sense to move it to the next section's vm tab as I don't believe the k8s section needs this warning.

Mehdi-Bendriss (dismissed):
Thank you Shams!

docs/how-to/upgrade.md
```diff
@@ -648,7 +974,16 @@ Finally, confirm the cluster is healthy again — the cluster health API should
...
 curl --cacert cert.pem -XGET "https://<unit-ip>:9200/_cluster/health?pretty" -u admin:<password>
 ```
+````
+
+````{tab-item} K8s
+:sync: k8s
+
+This procedure applies to machine deployments. On Kubernetes, restore from backup to a
+new deployment as described in [How to back up and restore](how-to-migrate-a-cluster).
+````
```
akram09: Why do we have to separate them here ?

akram09 (commented):
Thanks Shams!  one comment for me

docs/how-to/upgrade.md
smethnani: Because this hasn't been tested on K8s

akram09: Meaning it should be added in a separate PR ?

smethnani: No, I don't believe we have a revision of k8s to test this on. Or rather if we eventually do have one then yes.

akram09 (dismissed):
Thank you !

izmalk (approved):
LGTM.

akram09 (approved):

Mehdi-Bendriss (approved):

